PR #1105, opened 2016-01-29
Make use of mcg package and update mcg references

tijoytom:
Mcg now available as a package , use it in dotnet-mcg command.
Use System.CommandLine for command parsing.
Pass in AppDep and ILSDK as mcg references.

### DIFF
--- a/src/Microsoft.DotNet.Tools.Compiler.Native/Program.cs
+++ b/src/Microsoft.DotNet.Tools.Compiler.Native/Program.cs
@@ -52,16 +52,7 @@ namespace Microsoft.DotNet.Tools.Compiler.Native
 
                     string interopAssemblyPath = Path.Combine(Path.GetDirectoryName(config.InputManagedAssemblyPath), "Interop");
                     string inputAssemblyName = Path.GetFileNameWithoutExtension(config.InputManagedAssemblyPath);
-                    
-                    // update input managed assembly to be the one re-written by mcg. This's temporary 
-                    // until we avoid re-writing. 
-                    config.InputManagedAssemblyPath = Path.Combine(interopAssemblyPath, inputAssemblyName+".dll");
-
-                    // these addtional references are injected by mcg
-                    config.AddReference(Path.Combine(config.IlcSdkPath, "System.Private.Interop.dll"));
-                    config.AddReference(Path.Combine(config.AppDepSDKPath, "System.Private.Mcg.dll"));
                     config.AddReference(Path.Combine(interopAssemblyPath, inputAssemblyName + ".mcginterop.dll"));
-                    config.AddReference(Path.Combine(config.IlcSdkPath, "System.Private.Mcg.dll"));
                 }
 
                 var nativeCompiler = NativeCompiler.Create(config);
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Tools.Compiler.Native/Program.cs
```diff
@@ -91,6 +82,24 @@ private static int RunMcg(NativeCompileSettings config)
...
+
+                mcgArgs.Add("--r");
+                mcgArgs.Add(refPath);
+            }
+
+            foreach (string refPath in Directory.EnumerateFiles(config.AppDepSDKPath, "*.dll"))
+            {
+                // System.Runtime.Extensions define an internal type called System.Runtime.InteropServices.Marshal which 
+                // conflicts with Marshal from S.P.Interop , we don't need System.Runtime.Extensions anyways,skip it.
+                if (refPath.Contains("System.Runtime.Extensions.dll")) continue;
```
brthor: Is this a Point in Time thing or an expected permanent solution?

src/Microsoft.DotNet.Tools.Mcg/Program.cs
```diff
@@ -40,73 +36,20 @@ private static string[] ParseResponseFile(string rspPath)
...
-                }
-            }
-            var mcgArguments = TransformCommandLine(args);
-            return ExecuteMcg(mcgArguments);
+            return ExecuteMcg(args);
         }
 
         private static int ExecuteMcg(IEnumerable<string> arguments)
         {
             // TODO : Mcg is assumed to be present under the current path.This will change once
```
brthor: Can we remove this todo?

src/Microsoft.DotNet.Tools.Compiler.Native/Program.cs
tijoytom: AppDep SDK is a point in time thing and is expected to go-away. Mcg do not understand packaged now , so it cant consume ILSDK package directly.But it's something which will be fixed once mcg is out in public.

src/Microsoft.DotNet.Tools.Mcg/Program.cs
tijoytom: Yes ,thanks for noticing.

src/Microsoft.DotNet.Tools.Compiler.Native/Program.cs
tijoytom: Oh ,you are specifically asking about if ( contains check.)  ( i still haven't figured how git comments work) . yes , it's a point in time thing , we will rename the Marshal in  System.Runtime.Extensions.dll.

brthor: Alright, makes sense. 

Should we put a TODO/file an issue to remove so we don't forget?

piotrpMSFT: @tijoytom this is a public repo. MCG _is_ out in public ;)

piotrpMSFT: @tijoytom let's please get a tracking issue before merging.

tijoytom: Yes, i have the go-ahead to make mcg binaries public.
Filed  #1111

PR #1108, opened 2016-01-29
Piotrp msft/merge master

piotrpMSFT:


davidfowl:
Why is this in a PR?

piotrpMSFT:
@davidfowl because I was concerned about the MCG merge and didn't want to break master.

PR #1113, opened 2016-01-29
Piotrp msft/issue1112/nu get 559

piotrpMSFT:
@emgarten


emgarten:
The NuGet bump looks good, 559 will bring some bug fixes and perf improvements.

PR #1128, opened 2016-01-29
Remove Xlinker flag for native

schellap:


brthor:
Ubuntu failures are nuget and nothing here touches Ubuntu. 

IMO :shipit:

schellap:
@brthor noted offline that this is related to the quoting issue (with arg forwarding changes). We don't need `-Xlinker`. So removing it.

`"-Xlinker my.lib"`

Should be `"-Xlinker" "my.lib"`

schellap:
OSX passed merging

schellap:
Cc @blackdwarf

blackdwarf:
@schellap this fixes the weird linker errors on OS X?

schellap:
Yes @blackdwarf

PR #1134, opened 2016-01-30
dotnet-resgen should use a file to accept argument values

### LINKED ISSUE #1122: dotnet-resgen should use a file to accept argument values
Ran into this issue building a file with localized content. It seems like there's a character limit to the `ProcessStartInfo.Arguments` value (8192 since we have `UseShellExecute` = true) used to execute the `dotnet-resgen` command as part of `dotnet build`. In my application, the argument length was 13911. The recommended fix for this (https://support.microsoft.com/en-us/kb/830473) is to use a file to pass the arguments around.

https://gist.github.com/pranavkm/1b93142d1756c54f660d has the argument list printed as part of running `dotnet --verbose build`.

pranavkm:
Fixes #1122

### DIFF
--- a/src/Microsoft.DotNet.Tools.Compiler/Program.cs
+++ b/src/Microsoft.DotNet.Tools.Compiler/Program.cs
@@ -327,7 +327,7 @@ namespace Microsoft.DotNet.Tools.Compiler
 
             if (success)
             {
-                success &= GenerateCultureResourceAssemblies(context.ProjectFile, dependencies, outputPath);
+                success &= GenerateCultureResourceAssemblies(context.ProjectFile, dependencies, intermediateOutputPath, outputPath);
             }
 
             bool generateBindingRedirects = false;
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.Tools.Compiler/Program.cs
```diff
@@ -444,11 +444,14 @@ private static bool AddNonCultureResources(Project project, List<string> compile
...
+        private static bool GenerateCultureResourceAssemblies(
+            Project project,
+            List<LibraryExport> dependencies,
+            string intermediateOutputPath,
+            string outputPath)
         {
             var referencePaths = CompilerUtil.GetReferencePathsForCultureResgen(dependencies);
-            var resgenReferenceArgs = referencePaths.Select(referencePath => $"-r \"{referencePath}\"").ToList();
-
+            var resgenReferenceArgs = referencePaths.Select(path => $"-r:{path}").ToList();
```
pranavkm: I looked at the code for Compiler.Csc which uses the response file format and they trim out quotes around the input. Unless I'm missing something, the quotes don't seem to do much here.

davidfowl:
Needs a rebase

Issue #1149, opened 2016-01-31
Intrinsic CLI commands should be in a single project


davidfowl:
Today we're doing something pretty ridiculous, making each of the built in commands and making them separate projects and then publishing them to a single folder which ends up stomping all over each other. Instead of this madness, we should have a single project for the intrinsic commands and separate projects for the commands that aren't `built in`. Here's an example of what that could look like:

```
src/
   dotnet/
      commands/
         dotnet-compile/
         dotnet-compile-csc/
         dotnet-build/
         dotnet-new/
         dotnet-pack/
         dotnet-publish/
         dotnet-projectmodel-server/
         dotnet-resgen/
         dotnet-restore/
         dotnet-run/
         dotnet-test/
         dotnet-repl/
         dotnet-repl-csi/
         dotnet-compile-fsc/
```

Today, everything that ships with the dotnet CLI tools would be in the commands folder. Other commands can already be added as nuget packages so there's no blocker there. If we end up with another global install mechanism then we can revisit then.

/cc @piotrpMSFT @brthor @livarcocc

davidfowl:
Added benefits:
- Simpler install
- Conflicts will be detected during restore and compile
- Faster builds

blackdwarf:
@davidfowl this makes sense, but we should first go through the exercise of figuring out which commands go into the "core package" and which are extensions and then do this bit of work. 

/cc @piotrMSFT @Petermarcu

davidfowl:
Instead of waiting on that, the non core commands can be ripped out when we figure out what those are

anurse:
> the non core commands can be ripped out when we figure out what those are

This. It's been a bit difficult to keep things going when we have a bit of a mix of random new commands. Perhaps we need a separate cli-labs repo where we can work on the non-core commands? IMO, `repl` should be in a separate repo. I'd even suggest F# support be separate **for now** at least. I think F# is awesome and should be fully supported and maintained, but it feels too soon to be piling everything on into one repo. It slows down our development when we have to keep maintaining a large set of other unrelated projects...

davidfowl:
Yes agreed, my only point was that we need a single project.json (aka set of dependencies) per executable. So a single project for built in things instead of pretending like we have separate executables and then stomping dependencies like crazy.

blackdwarf:
@davidfowl but why do the double work? Or is it not double work? 

[...]

PR #1151, opened 2016-01-31
Everything in the same project

davidfowl:
- Build in intrinsic commands as part of `dotnet`.

Starting a discussion about taking the path described in #1149. A few things to note:

- We could remove the special casing of built in command and instead generate stub executables for each built in command.
- With this change you can't remove override built in commands via packages. This wouldn't be too hard to allow but that's how it looks right now.
- I probably need to fix help for built in commands.


Petermarcu:
the proposal is this.. Can we make this PR follow along? These are the high level commands that we think will be in the core and we want to make sure there is a first class experience for everything in the next layer of "Power tools". 

There are some sub commands that are not explicitly listed that these depend on.

dotnet new
dotnet install
dotnet restore
dotnet build
dotnet compile
dotnet test
dotnet debug
dotnet pack
dotnet publish

davidfowl:
> the proposal is this.. Can we make this PR follow along? These are the high level commands that we think will be in the core and we want to make sure there is a first class experience for everything in the next layer of "Power tools".

We should do that in another PR. I don't want to fold this into a single change and restructure.

Petermarcu:
Is it cleaner to just abandon this one and merge the right projects once as opposed to merging and then pulling some back out?

davidfowl:
> Is it cleaner to just abandon this one and merge the right projects once as opposed to merging and then pulling some back out?

Nope, I'm going to merge this thing. We can pull things out in another change.

davidfowl:
Do we know how those commands will ship? Or where they'll be getting built?

PR #1157, opened 2016-02-01
Cleanup command options for Windows CppCompileStep and LinkStep

jkotas:
- Pass command line options correctly to the linker (all options were passed as one giant quoted string and thus they had no effect)
- Remove unnecessary command line options
- Factored out command line options that are shared by debug and release configurations


jkotas:
@schellap PTLA

schellap:
LGTM. wd4477 what does it do? Couldn't find on MSDN. Can be merged, other CI tasks don't use this logic.

jkotas:
> wd4477 what does it do?

It disables warning 4477. We should not actually need it here - I will remove it.

PR #1158, opened 2016-02-01
Add "additionalArguments" compilation option

agocke:
The `compilationOptions` section of the project.json provides support for a lot of commonly used compilation options but it will not, and should not, provide the full set of options available to all possible compilers. However, there should still be a way to pass specific options down to those compilers, ignoring the layers and filtering in between.

The `additionalArguments` section of the `compilationOptions` provides this by allowing for an arbitrary list of strings to be added to the compilation options which will be directly appended to the options eventually given to the language-specific compiler.

/cc @anurse @brthor @davidfowl

### DIFF
--- a/scripts/configuration/testProjects.csv
+++ b/scripts/configuration/testProjects.csv
@@ -3,3 +3,4 @@ StreamForwarderTests
 dotnet-publish.Tests
 dotnet-compile.Tests
 dotnet-build.Tests
+Compiler.Common.Tests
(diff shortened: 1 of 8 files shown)


scripts/test/runtests.sh
```diff
@@ -25,6 +25,7 @@ TestProjects=( \
     dotnet-publish.Tests \
     dotnet-compile.Tests \
     dotnet-build.Tests \
+    Compiler.Common.Test \
```
anurse: Let's make it Test**s** for consistency

anurse:
Looks good. :shipit: after minor comments and builds are resolved.

agocke:
@anurse Looks like Ubuntu queue is having some issues -- is this a known problem?

piotrpMSFT:
@agocke are we good to merge?

agocke:
Yup

PR #1160, opened 2016-02-01
Infer the analyzer languageId from compilerName

agocke:
Simplifies using a different compiler since you only have to set one value if the the compiler is known to dotnet-compile.

/cc @anurse @brthor @srivatsn @jasonmalinowski @davidfowl

### DIFF
--- a/src/dotnet/commands/dotnet-compile/CompilerUtil.cs
+++ b/src/dotnet/commands/dotnet-compile/CompilerUtil.cs
@@ -26,12 +26,29 @@ namespace Microsoft.DotNet.Tools.Compiler
 
             return compilerName;
         }
-        
+
+        private static readonly KeyValuePair<string, string>[] s_compilerNameToLanguageId =
+        {
+            new KeyValuePair<string, string>("csc", "cs"),
+            new KeyValuePair<string, string>("vbc", "vb"),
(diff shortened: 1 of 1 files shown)


agocke:
@jasonmalinowski Was this what you were thinking?

src/dotnet/commands/dotnet-compile/CompilerUtil.cs
```diff
@@ -26,12 +26,29 @@ public static string ResolveCompilerName(ProjectContext context)
...
             return compilerName;
         }
-        
+
+        private static readonly KeyValuePair<string, string>[] s_compilerNameToLanguageId =
+        {
+            new KeyValuePair<string, string>("csc", "cs"),
+            new KeyValuePair<string, string>("vbc", "vb"),
+            new KeyValuePair<string, string>("fsc", "fs")
+        };
```
piotrpMSFT: Does adding a language require a CLI update? It doesn't happen frequently, so might be ok...

agocke: No, this is just the default. You can override it manually if you have a new language.

piotrpMSFT:
:shipit:

Issue #1098, opened 2016-01-28
ProjectModelServer - am seeing an argument exception opening one of my projects


BillHiebert:
dotnet version: 1.0.1.001134

The error is: An item with the same key has already been added. Key: fx/System.Collections
```
 	mscorlib.ni.dll!System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(object key)	Unknown
 	mscorlib.ni.dll!System.Collections.Generic.Dictionary<string, Microsoft.DotNet.ProjectModel.Compilation.LibraryExport>.Insert(string key, Microsoft.DotNet.ProjectModel.Compilation.LibraryExport value, bool add)	Unknown
 	System.Linq.dll!System.Linq.Enumerable.ToDictionary<Microsoft.DotNet.ProjectModel.Compilation.LibraryExport, string>(System.Collections.Generic.IEnumerable<Microsoft.DotNet.ProjectModel.Compilation.LibraryExport> source, System.Func<Microsoft.DotNet.ProjectModel.Compilation.LibraryExport, string> keySelector, System.Collections.Generic.IEqualityComparer<string> comparer)	Unknown
 	dotnet-projectmodel-server.dll!Microsoft.DotNet.ProjectModel.Server.ProjectContextSnapshot.Create(Microsoft.DotNet.ProjectModel.ProjectContext context, string configuration, System.Collections.Generic.IEnumerable<string> currentSearchPaths) Line 35	C#
 	dotnet-projectmodel-server.dll!Microsoft.DotNet.ProjectModel.Server.ProjectSnapshot.Create(string projectDirectory, string configuration, Microsoft.DotNet.ProjectModel.WorkspaceContext workspaceContext, System.Collections.Generic.IReadOnlyList<string> projectSearchPaths) Line 44	C#
 	dotnet-projectmodel-server.dll!Microsoft.DotNet.ProjectModel.Server.ProjectManager.UpdateProject() Line 250	C#
 	dotnet-projectmodel-server.dll!Microsoft.DotNet.ProjectModel.Server.ProjectManager.DoProcessLoop() Line 160	C#
>	dotnet-projectmodel-server.dll!Microsoft.DotNet.ProjectModel.Server.ProjectManager.ProcessLoop() Line 121	C#
 	mscorlib.ni.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)	Unknown
[...]

PR #1169, opened 2016-02-01
Fix projectmodel-server regression

troydai:
1. Address the null reference exception when a project dependency is
broken.
2. Address the duplicate key issues after the logic of redirecting
System pacage to reference assembly was added
3. Include projectmodel-server test in the build script

Issue: https://github.com/dotnet/cli/issues/1098


troydai:
Hi @Sridhar-MS, the CI keep failing to restore packages. Are there networking issues with them?

Sridhar-MS:
@brthor recently had a fix/workaround for this.

brthor:
@troydai adding `--disable-parallel` to all restore calls seems to fix the issue on ubuntul.

If you rebase on rel, the fix is checked in.

troydai:
Moves to https://github.com/dotnet/cli/pull/1173 on `rel/1.0.0`

PR #1170, opened 2016-02-01
Change dropped NuGet.config endpoint to core-rel

brthor:


blackdwarf:
@brthor it would be **great** if you can merge this ASAP.

blackdwarf:
@dotnet-bot test Ubuntu Debug Build please

weshaggard:
Can I ask why this is being switched? There is a pretty good chance we are very soon going to kill the dotnet-core-rel feed.

blackdwarf:
@weshaggard when you say "kill" do you mean "replace with another feed" or something else? We need a solution to make `dotnet new` drop a restorable project.

weshaggard:
Originally we thought we would split our dev and rel feeds because they might conflict but the conflicts still caused lots of trouble with user package caches so we are instead solving the conflicting problem a different way so we plan to just use one feed which would be our dev dotnet-core feed.

brthor:
@weshaggard Is there any expected time when this would switch?

As I understand it this NuGet.config is a temporary artifact anyways. We just need for people to be able to restore the same stdlib in their dotnet new projects that we depend on for the cli.

weshaggard:
As part of our feed switch over we are stopped publishing to dotnet-core-rel and so any packages should come from dotnet-core. That includes RC2 packages we build out of our release branch.

Issue #1096, opened 2016-01-28
dotnet restore NREs when the feed cannot be reached


moozzyk:
Apparently there are some problems with myget today. When I run `dotnet restore` I get this: 

```
PS C:\temp\WebApplication6\src\WebApplication6> dotnet restore
info : Restoring packages for C:\temp\WebApplication6\src\WebApplication6\project.json...
error: A task was canceled.
Restoring Tool 'dotnet-publish-iis' for 'C:\temp\WebApplication6\src\WebApplication6\obj\90926872-e67b-45f7-a4a5-b00d4b991403\bin\project.json' in 'C:\temp\WebApplication6\src\WebApplication6\obj\90926872-e67b-45f7-a4a5-b00d4b991403\bin'
info : Restoring packages for C:\temp\WebApplication6\src\WebApplication6\obj\90926872-e67b-45f7-a4a5-b00d4b991403\bin\project.json...
error: A task was canceled.
Object reference not set to an instance of an object.
PS C:\temp\WebApplication6\src\WebApplication6>
```

PR #1175, opened 2016-02-02
Fix NRE in restore

pakrym:
@davidfowl #1096

### DIFF
--- a/src/dotnet/commands/dotnet-restore/Program.cs
+++ b/src/dotnet/commands/dotnet-restore/Program.cs
@@ -131,11 +131,12 @@ namespace Microsoft.DotNet.Tools.Restore
             {
                 var tempPath = Path.Combine(tempRoot, Guid.NewGuid().ToString(), "bin");
 
-                RestoreToolToPath(tooldep, restoreTask.Arguments, tempPath, quiet);
-
-                CreateDepsInPackageCache(tooldep, tempPath);
-
-                PersistLockFile(tooldep, tempPath, restoreTask.ProjectDirectory);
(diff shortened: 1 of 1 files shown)


src/dotnet/commands/dotnet-restore/Program.cs
```diff
@@ -131,11 +131,12 @@ private static void RestoreTool(LibraryRange tooldep, RestoreTask restoreTask, b
...
                 var tempPath = Path.Combine(tempRoot, Guid.NewGuid().ToString(), "bin");
 
-                RestoreToolToPath(tooldep, restoreTask.Arguments, tempPath, quiet);
-
-                CreateDepsInPackageCache(tooldep, tempPath);
-
-                PersistLockFile(tooldep, tempPath, restoreTask.ProjectDirectory);
+                var restoreSucceded = RestoreToolToPath(tooldep, restoreTask.Arguments, tempPath, quiet);
+                if (restoreSucceded)
+                {
```
piotrpMSFT: How can we get some diagnostics here? This is a silent failure. NRE was nicer since it indicates a problem...

pakrym: Nuget outputs restore errors

piotrpMSFT: when we saw the NREs I hadn't noticed any Nuget errors. I'll keep an eye out for undiagnosable issues...

davidfowl: You should propagate the exit code

Issue #1147, opened 2016-01-30
Compilation errors are hidden when using dotnet run


davidfowl:
This is the cause.

https://github.com/dotnet/cli/blob/master/src/dotnet-run/RunCommand.cs#L105

We'll likely need an `--errors-only` mode so that only errors show up when using `dotnet run`

pakrym:
I think this is fixed.

Issue #1059, opened 2016-01-27
Command Output Forwarding misses some output


piotrpMSFT:
During NuGet push we faced an issue where NuGet was returning '1' despite no error messages. It was later discovered that this was due to a WriteLine from NuGet not being forwarded out through the CLI. I suspect this has to do with stream flushing behavior near process exit. We need to improve StreamForwarderTests to cover these scenarios.

bricelam:
Possibly related to #925

piotrpMSFT:
@brthor, can you take a pass at this once #999 goes in?

brthor:
Looking at this and #925 now

brthor:
Tried to repro this in the following ways:
* A command which throws unhandled exceptions, WriteLining before
* A command which handles exceptions and prints messages, with Writes before and after
* Interleaving StdOut and StdErr in a command
* Nesting a command two processes deep and printing 100,000+ Lines

In all cases I compared output to the output of the process if run directly (`dotnet cmd` vs `dotnet-cmd.exe`). 

I did not see messages being dropped in any cases. I did see significant misordering of the writes, more in line with the issue in #925 

Would recommend closing this out unless we see it repro again.

brthor:
Giving this more thought,

With the issue I found in StreamForwarder, where the Read operation would block until it's buffer was full, it is possible that a message would get dropped if the driver or coreclr crashed before that read stopped blocking (Which is when child process exits). 

Although unlikely, in all of the craziness in that week I wouldn't be very surprised.

brthor:
#925 is tracking the fix currently in PR. 

Let's reopen this if we see it happen again.

PR #1182, opened 2016-02-02
Move error reporter to use stderr

pakrym:
Ran build couple of times, looks good, needs to be fixed in Forwarder anyway.
@davidfowl 
https://github.com/dotnet/cli/issues/1147


piotrpMSFT:
This is is being rolled into #1059

PR #1197, opened 2016-02-03
Invoke intrinsic commands directly instead of creating a process.

AustinWise:
Since issue #1149 intrinsic commands are now invoked directly from the dotnet.exe program rather than creating a new a new dotnet-intrinsic.exe process. Now that all the commands live inside the same assembly, intrinsic commands that invoke other commands could potentially do so without creating a new process. As creating a process, loading the CLR, resolving dependencies, and JIT-ing code all take a bit of time, this change could speed things up a bit.

I believe the principal change in behavior this change causes is for the DOTNET_CLI_CONTEXT_ environmental variables to be inherit by intrinsic commands created by other intrinsic commands. Currently Microsoft.DotNet.Cli.Program.ProcessArgs() seems to overwrite this environmental variable every time based on command line flags.

There are probably nicer ways of doing this, but I wanted to try the simplest thing possible to see if it would work and if y'all are interested in doing something like this.


davidfowl:
@AustinWise can you send this PR to rel/1.0.0

AustinWise:
There does not appear to be a way to edit which branch I'm targeting, so I'm going to close and open a new one.

PR #1206, opened 2016-02-03
Removing www.myget.org in our tests.

eerhardt:
We have a bunch of NuGet.Config files in our tests that were needed before because the tests were run out of a temp directory.  Now the tests are run out of the artifacts folder, which is a subdirectory of our repo root, so the root NuGet.Config will get picked up.

I also updated all the dotnet-core feeds from www.myget.org to dotnet.myget.org.


eerhardt:
cc @anurse

PR #1207, opened 2016-02-03
Port native fixes

schellap:
@brthor porting fixes from master so we can use other stable command versions from our repo.

### DIFF
--- a/src/dotnet/project.json
+++ b/src/dotnet/project.json
@@ -28,7 +28,7 @@
         "Microsoft.DotNet.ProjectModel": "1.0.0-*",
         "Microsoft.DotNet.Compiler.Common": "1.0.0-*",
         "Microsoft.DotNet.Cli.Utils": "1.0.0-*",
-        "Microsoft.DotNet.ILCompiler.SDK": "1.0.4-prerelease-00006",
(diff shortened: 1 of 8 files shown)


src/dotnet/project.json
```diff
@@ -28,7 +28,7 @@
         "Microsoft.DotNet.ProjectModel": "1.0.0-*",
         "Microsoft.DotNet.Compiler.Common": "1.0.0-*",
         "Microsoft.DotNet.Cli.Utils": "1.0.0-*",
-        "Microsoft.DotNet.ILCompiler.SDK": "1.0.4-prerelease-00006",
+        "Microsoft.DotNet.ILCompiler.SDK": "1.0.5-prerelease-00002",
```
brthor: ping @eerhardt 
Some of these deps are changing.

brthor:
LGTM why closed?

src/dotnet/project.json
eerhardt: Thanks for the heads up.  AFAIK, the 1.0.5 version is already in our dependency feed, since I was working in master.

FYI - I'm pretty close to merging my change to remove www.myget.org from our NuGet.Config file.

PR #1203, opened 2016-02-03
Updating NuGet.config to no longer point to myget.org.

eerhardt:
We are using our own cli-deps feed in an enterprise myget account in order to ensure we depend on stable nuget repositories.

@piotrpMSFT @anurse

PR #1209, opened 2016-02-03
Updating the Compiler and Interactive toolsets to v1.2.0-beta1-20160202-02

tannergooding:
**Contained in this Pull Request**
* Updates `Microsoft.Net.Compilers.netcore` to v1.2.0-beta1-20160202-02
* Updates `Microsoft.Net.CSharp.Interactive.netcore` to v1.2.0-beta1-20160202-02
* Adds the `csi` and `Microsoft.CodeAnalysis.CSharp.Scripting` to the list of assemblies to crossgen
* Adds a direct dependency to `Microsoft.DiaSymReader.Native`, v1.3.3
* Removes the `/ReadyToRun` switch from the crossgen arguments for Windows crossgen

### DIFF
--- a/scripts/crossgen/crossgen_roslyn.cmd
+++ b/scripts/crossgen/crossgen_roslyn.cmd
@@ -29,16 +29,24 @@ if exist mscorlib.ni.dll (
 
 set READYTORUN=
 
+echo Crossgenning System.Collections.Immutable
 crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% System.Collections.Immutable.dll >nul 2>nul
 if not %errorlevel% EQU 0 goto fail
 
+echo Crossgenning System.Reflection.Metadata
 crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% System.Reflection.Metadata.dll >nul 2>nul
 if not %errorlevel% EQU 0 goto fail
 
+echo Crossgenning Microsoft.CodeAnalysis
 crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% Microsoft.CodeAnalysis.dll >nul 2>nul
 if not %errorlevel% EQU 0 goto fail
 
+echo Crossgenning Microsoft.CodeAnalysis.CSharp
 crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% Microsoft.CodeAnalysis.CSharp.dll >nul 2>nul
+if not %errorlevel% EQU 0 goto fail
+
+echo Crossgenning Microsoft.CodeAnalysis.CSharp.Scripting
+crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% Microsoft.CodeAnalysis.CSharp.Scripting.dll >nul 2>nul
 if not %errorlevel% EQU 0 goto fail
 
 echo Crossgenning Microsoft.CodeAnalysis.VisualBasic
@@ -47,6 +55,10 @@ if not %errorlevel% EQU 0 goto fail
 
 echo Crossgenning csc
 crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% csc.dll >nul 2>nul
+if not %errorlevel% EQU 0 goto fail
+
+echo Crossgenning csi
+crossgen /nologo %READYTORUN% /Platform_Assemblies_Paths %BIN_DIR% csi.dll >nul 2>nul
 if not %errorlevel% EQU 0 goto fail
 
 echo Crossgenning vbc
--- a/scripts/crossgen/crossgen_roslyn.sh
+++ b/scripts/crossgen/crossgen_roslyn.sh
@@ -54,19 +54,33 @@ chmod +x crossgen
 
 ./crossgen -nologo $READYTORUN -platform_assemblies_paths $BIN_DIR mscorlib.dll
 
+info "Crossgenning System.Collections.Immutable"
 ./crossgen -nologo $READYTORUN -platform_assemblies_paths $BIN_DIR System.Collections.Immutable.dll
 
+info "Crossgenning System.Reflection.Metadata"
 ./crossgen -nologo $READYTORUN -platform_assemblies_paths $BIN_DIR System.Reflection.Metadata.dll
 
+info "Crossgenning Microsoft.CodeAnalysis"
 ./crossgen -nologo $READYTORUN -platform_assemblies_paths $BIN_DIR Microsoft.CodeAnalysis.dll
 
+info "Crossgenning Microsoft.CodeAnalysis.CSharp"
 ./crossgen -nologo $READYTORUN -platform_assemblies_paths $BIN_DIR Microsoft.CodeAnalysis.CSharp.dll
 
+info "Crossgenning Microsoft.CodeAnalysis.CSharp.Scripting"
+./crossgen -nologo $READYTORUN -platform_assemblies_paths $BIN_DIR Microsoft.CodeAnalysis.CSharp.Scripting.dll
+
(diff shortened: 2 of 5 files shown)


tannergooding:
This will likely require the dotnet.myget.org feed to be updated (see #1203).

NuGet.Config
```diff
@@ -13,6 +13,7 @@
     <add key="corert" value="https://www.myget.org/F/dotnet/api/v3/index.json" />
     <add key="dotnet-buildtools" value="https://www.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <add key="fsharp-daily" value="https://www.myget.org/F/fsharp-daily/api/v3/index.json" />
+	<add key="dotnet-core" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
```
davidfowl: Fix the formatting

davidfowl:
@tannergooding Once we verify that crossgen still works then :shipit:

NuGet.Config
tannergooding: Fixed.

tannergooding:
@davidfowl, cross-gen is currently broken due to a known bug when running cross-gen/native image generation on System.Reflection.Metadata (see Internal Bug 185474). 

Cross-Gen will complete succesfully on Windows (but hangs indefinitely on other platforms). However, on Windows, attempting to execute code in the repl (csi) will cause it to exit with an AccessViolationException.

The temporary workaround would be to prevent System.Reflection.Metadata from going through cross-gen until the internal bug is either fixed or has a workaround.

tannergooding:
FYI. @tmat

tannergooding:
@davidfowl, could you please loop in the necessary people to determine if disabling cross-gen for System.Reflection.Metadata is a reasonable workaround for the time being. There may (of course) be some perf impact, but this change brings in several fixes (especially for dotnet repl).

davidfowl:
We're not going to be disabling cross-gen. The performance is just too bad when it's not working. We'll need to fix that bug before getting this change in.

jaredpar:
CC @tmat who's looking into a work around for System.Reflection.Metadata.

jaredpar:
@tannergooding @JohnChen0 my original understanding is that Bug 185474 was an ngen only issue.  If this is truly a ngen and crossgen issue then why are we tracking it internally instead of externally?  Tracking it internally makes it harder to cross reference the issue on Github (as I'd like to do here).

tannergooding:
@jaredpar, the issue has the same repro as the internal bug. Just run `crossgen` on `System.Reflection.Metadata.dll` and it will hang indefinitely (/ReadyToRun seems to allow crossgen to complete, but you get all kinds of AccessViolationExceptions during runtime instead). As such, I see no reason to not move the bug into the open.

tannergooding:
@davidfowl, @jaredpar, @tmat, @agocke  The PR is now updated with a new version of System.Reflection.Metadata that resolves the ngen/crossgen issue. However, this new version still has an issue when run through crossgen with the `/ReadyToRun` switch (resulting in ExecutionEngineExceptions at runtime). Removing the '/ReadyToRun' switch (which is only used on windows) allows everything to work as expected.

What is the `/ReadyToRun` switch and does it make a large impact on the perf?

jaredpar:
ReadToRun is essentially a version tolerant ngen switch.  Version tolerance isn't necessary for our consumption here, I'm unsure why we are using it.  Maybe I'm missing a scenario.

agocke:
@jaredpar Nope, just cruft. I wasn't sure if it was necessary when I first added crossgen. I've later been convinced it's not necessary.

agocke:
Although I would still like if someone from the CLR could confirm that :)

davidfowl:
I'm not sure why we're using R2R at all. It's slower and we don't care about servicing csc outside of the CLI itself do we? (Maybe the roslyn team does :smile: )

jaredpar:
We don't care about servicing in this scenario.

I feel like we're all staring at this switch wondering why it ever got added in the first place.  Let's stop staring and just delete it :smile:

tannergooding:
@jaredpar, @agocke, @davidfowl I have updated the PR to remove the `ReadyToRun` switches from the Windows crossgen script (since it sounds like we never wanted/needed them in the first place, and they are blocking this PR).

@ericeil, Only one new package: `System.Reflection.Metadata, v1.2.0-rc3-23803` (currently exists in the `dotnet-core` myget feed). This just needs to be added to the new cli-deps feed, and then I can remove the change to nuget.config and retest.

davidfowl:
@tannergooding It's not just that dependency, it's that dependency and its dependency closure :smile: . Does it bump everything else in the graph to rc3 packages?

tannergooding:
@davidfowl, Comparing the project.lock.json files from before and after the shows the following new dependencies:

# MyGet Dependencies
* System.Reflection.Metadata: 1.2.0-rc3-23721 (indirect dependency of the `Microsoft.Net.Compilers.netcore` package)
* System.Reflection.Metadata: 1.2.0-rc3-23803 (direct dependency of the `src/dotnet/project.json`)

# NuGet Dependencies
* System.Linq: 4.0.0 (indirect dependency of the `System.Reflection.Metadata` package)
* Microsoft.DiaSymReader.Native: 1.3.3 (direct dependency of the `src/dotnet/project.json`)
* Microsoft.Net.Compilers.netcore: 1.2.0-beta1-20160202-02 (direct dependency of the `src/dotnet/project.json`)
* Microsoft.Net.CSharp.Interactive.netcore: 1.2.0-beta1-20160202-02 (direct dependency of the `src/dotnet/project.json`)
* Microsoft.CodeAnalysis.CSharp: 1.2.0-beta1-20160202-02 (direct dependency of the `src/Microsoft.DotNet.Compiler.Common/project.json`)
* Microsoft.CodeAnalysis.CSharp.Workspaces: 1.2.0-beta1-20160202-02 (direct dependency of the `src/Microsoft.DotNet.ProjectModel.Workspaces/project.json`)
* Microsoft.CodeAnalysis.Workspaces.Common: 1.2.0-beta1-20160202-02 (indirect dependency of the `Microsoft.CodeAnalysis.CSharp.Workspaces` package)
* Microsoft.CodeAnalysis.CSharp.Scripting: 1.2.0-beta1-20160202-02 (indirect dependency of the `Microsoft.Net.CSharp.Interactive.netcore` package)
* Microsoft.CodeAnalysis.Scripting: 1.2.0-beta1-20160202-02 (indirect dependency of the `Microsoft.Net.CSharp.Interactive.netcore` package)
* Microsoft.CodeAnalysis.Scripting.Common: 1.2.0-beta1-20160202-02 (indirect dependency of the `Microsoft.Net.CSharp.Interactive.netcore` package)
* Microsoft.CodeAnalysis.VisualBasic: 1.2.0-beta1-20160202-02 (indirect dependency of the `Microsoft.Net.Compilers.netcore` package)
* Microsoft.CodeAnalysis.Common: 1.2.0-beta1-20160202-02 (indirect dependency of the `Microsoft.Net.Compilers.netcore` package)

tannergooding:
test CentOS7.1 Debug Build please
test CentOS7.1 Release Build please
test Ubuntu Debug Build please
test Ubuntu Release Build please

tannergooding:
test CentOS7.1 Debug Build please
test Ubuntu Debug Build please
test Ubuntu Release Build please

tannergooding:
test this please

tannergooding:
test this please

tannergooding:
@davidfowl, build is succeeding for everything except for Ubuntu/CentOS (which are also failing in the commit builds).

I don't believe there is anything left on my side blocking this from going in (all temporary workarounds are gone and all dependencies are now on MyGet).

FYI. @ManishJayaswal. This is still pending

tannergooding:
test OSX Release Build please
test Windows_NT Debug Build please
test Windows_NT Release Build please

tannergooding:
FYI. @ManishJayaswal, @davidfowl This has been merged. All tests were passing.

PR #867, opened 2016-01-15
Fix diffs in generated nuspec with nuget.core

ericstj:
There was a typo in file list that used Source
value for target attribute.

The ordering of metadata was different than
nuget.core and nuget.core never added
DevelopmentDependency if it was set to
false.

There are still some differences between a
nuspec generated through Manifest.Save
in Nuget.Core and this, but all of those
make it more consistent with the nuspec
that is put in the actual package so I'm
leaving them.

The primary difference is that NuGet.Core
used string types to represent the TFM in
groupable items so it would just pass through
whatever you gave the API to the nuspec.
The downside of this is that you might have
an invalid string and you wouldn't know that
until you packed (and cracked open the nuspec).
The new implementation parses the TFM
and writes a normalized version to the nuspec.
This is more wysiwyg so I'm leaving it.

Contained in branches: master#gitext://gotobranch/master
Contained in no tag

PR #1211, opened 2016-02-03
Fix diffs in generated nuspec with nuget.core

ericstj:
There was a typo in file list that used Source
value for target attribute.

The ordering of metadata was different than
nuget.core and nuget.core never added
DevelopmentDependency if it was set to
false.

There are still some differences between a
nuspec generated through Manifest.Save
in Nuget.Core and this, but all of those
make it more consistent with the nuspec
that is put in the actual package so I'm
leaving them.

The primary difference is that NuGet.Core
used string types to represent the TFM in
groupable items so it would just pass through
whatever you gave the API to the nuspec.
The downside of this is that you might have
an invalid string and you wouldn't know that
until you packed (and cracked open the nuspec).
The new implementation parses the TFM
and writes a normalized version to the nuspec.
This is more wysiwyg so I'm leaving it.

replacement PR for https://github.com/dotnet/cli/pull/867
/cc @piotrpMSFT

### DIFF
--- a/src/dotnet/commands/dotnet-pack/NuGet/ManifestContentFiles.cs
+++ b/src/dotnet/commands/dotnet-pack/NuGet/ManifestContentFiles.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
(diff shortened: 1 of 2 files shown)


ericstj:
@dotnet-bot please test Ubuntu Debug Build

davidfowl:
LGTM

PR #1212, opened 2016-02-03
Fix color printing and --verbose

AustinWise:
I noticed that the escape codes are being printed instead of colors on Windows. There were a couple problems:

* When running intrinsic commands, the AnsiPassThrough was being enabled. This caused the escape codes to never be processed for intrinsic commands.

* Reporter.Write did not transform escape codes into colors. The Command uses this method to forward some output.

Calling ShouldPassAnsiCodesThrough() for the side effect of reading the environmental variable seems a little lame. Perhaps instead Command.CreateDotNet should always set this variable? I'm not sure if Microsoft.DotNet.Cli.Utils is meant for consumption by externals tools as a means of invoking dotnet.exe, so I did not do it there.

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/Command.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/Command.cs
@@ -168,7 +168,7 @@ namespace Microsoft.DotNet.Cli.Utils
                 if (to == null)
                 {
                     _stdOut.ForwardTo(writeLine: Reporter.Output.WriteLine);
-                    _process.StartInfo.Environment[CommandContext.Variables.AnsiPassThru] = ansiPassThrough.ToString();
+                    EnvironmentVariable(CommandContext.Variables.AnsiPassThru, ansiPassThrough.ToString());
                 }
                 else
                 {
@@ -186,7 +186,7 @@ namespace Microsoft.DotNet.Cli.Utils
                 if (to == null)
                 {
                     _stdErr.ForwardTo(writeLine: Reporter.Error.WriteLine);
-                    _process.StartInfo.Environment[CommandContext.Variables.AnsiPassThru] = ansiPassThrough.ToString();
+                    EnvironmentVariable(CommandContext.Variables.AnsiPassThru, ansiPassThrough.ToString());
                 }
                 else
                 {
--- a/src/Microsoft.DotNet.Cli.Utils/Command.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/Command.cs
@@ -160,7 +160,7 @@ namespace Microsoft.DotNet.Cli.Utils
             return this;
         }
 
-        public ICommand ForwardStdOut(TextWriter to = null, bool onlyIfVerbose = false)
+        public ICommand ForwardStdOut(TextWriter to = null, bool onlyIfVerbose = false, bool ansiPassThrough = true)
         {
             ThrowIfRunning();
             if (!onlyIfVerbose || CommandContext.IsVerbose())
@@ -168,6 +168,7 @@ namespace Microsoft.DotNet.Cli.Utils
                 if (to == null)
                 {
                     _stdOut.ForwardTo(writeLine: Reporter.Output.WriteLine);
+                    _process.StartInfo.Environment[CommandContext.Variables.AnsiPassThru] = ansiPassThrough.ToString();
                 }
                 else
                 {
@@ -177,7 +178,7 @@ namespace Microsoft.DotNet.Cli.Utils
             return this;
         }
 
-        public ICommand ForwardStdErr(TextWriter to = null, bool onlyIfVerbose = false)
+        public ICommand ForwardStdErr(TextWriter to = null, bool onlyIfVerbose = false, bool ansiPassThrough = true)
         {
             ThrowIfRunning();
             if (!onlyIfVerbose || CommandContext.IsVerbose())
@@ -185,6 +186,7 @@ namespace Microsoft.DotNet.Cli.Utils
                 if (to == null)
                 {
                     _stdErr.ForwardTo(writeLine: Reporter.Error.WriteLine);
+                    _process.StartInfo.Environment[CommandContext.Variables.AnsiPassThru] = ansiPassThrough.ToString();
                 }
                 else
                 {
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -74,7 +74,7 @@ Common Commands:
         {
             // CommandLineApplication is a bit restrictive, so we parse things ourselves here. Individual apps should use CLA.
 
-            var verbose = false;
+            bool? verbose = null;
             var success = true;
             var command = string.Empty;
             var lastArg = 0;
@@ -114,6 +114,11 @@ Common Commands:
 
             var appArgs = (lastArg + 1) >= args.Length ? Enumerable.Empty<string>() : args.Skip(lastArg + 1).ToArray();
 
+            if (verbose.HasValue)
+            {
+                Environment.SetEnvironmentVariable(CommandContext.Variables.Verbose, verbose.ToString());
+            }
+
             if (string.IsNullOrEmpty(command) || command.Equals("help", StringComparison.OrdinalIgnoreCase))
             {
                 return RunHelpCommand(appArgs);
@@ -140,15 +145,10 @@ Common Commands:
             Func<string[], int> builtIn;
             if (builtIns.TryGetValue(command, out builtIn))
             {
-                // mimic the env variable
-                Environment.SetEnvironmentVariable(CommandContext.Variables.Verbose, verbose.ToString());
-                Environment.SetEnvironmentVariable(CommandContext.Variables.AnsiPassThru, bool.TrueString);
                 return builtIn(appArgs.ToArray());
             }
 
             return Command.Create("dotnet-" + command, appArgs, FrameworkConstants.CommonFrameworks.DnxCore50)
-                .EnvironmentVariable(CommandContext.Variables.Verbose, verbose.ToString())
-                .EnvironmentVariable(CommandContext.Variables.AnsiPassThru, bool.TrueString)
                 .ForwardStdErr()
                 .ForwardStdOut()
                 .Execute()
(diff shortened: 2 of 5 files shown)


src/dotnet/Program.cs
```diff
@@ -140,6 +140,8 @@ private static int ProcessArgs(string[] args)
             {
                 // mimic the env variable
                 Environment.SetEnvironmentVariable(CommandContext.Variables.Verbose, verbose.ToString());
+                // Use the current value of AnsiPassThru but make sure child processes have it set to true.
+                CommandContext.ShouldPassAnsiCodesThrough();
```
livarcocc: Is this a query or a command? If a query, should you be testing the result here? If a command, can we call it PassAnsiCodesThrough?

brthor: :+1: This just returns a value, why put this here?

AustinWise: The intent to cause CommandContext to read the current value of the environmental variable before changing it on the following line. The intention is for the current process to use the value of the DOTNET_CLI_CONTEXT_ANSI_PASS_THRU as it is when the process starts but for all the children to have it set to true. Otherwise dotnet.exe will print out a bunch of escape codes on Windows because AnsiPassThrough will always be true.

Since calling that getter method for it's side effect is obtuse, perhaps I should add a new function to CommandContext that reads the environmental variable. Something like "InitialAnsiPassThroughFromEnvironment()". What do you think?

brthor: I get the intent here but I don't get the side effect you're going after. 

You're forcing the variable to get read, but what does that affect?

AustinWise:
I changed the way the environmental variable is read to be explicit, instead of calling a getter method for the side effect of lazy initialization.
I also expanded the scope of this pull request a little bit by making sure the verbose environmental variable is inherited by all child processes. Currently only the first process is getting it and it will be overwritten in any child processes.

src/dotnet/Program.cs
```diff
@@ -113,6 +113,16 @@ private static int ProcessArgs(string[] args)
 
             var appArgs = (lastArg + 1) >= args.Length ? Enumerable.Empty<string>() : args.Skip(lastArg + 1).ToArray();
 
+            if (verbose.HasValue)
+            {
+                Environment.SetEnvironmentVariable(CommandContext.Variables.Verbose, verbose.ToString());
+            }
+            //This process should inherit AnsiPassThrough or use the default of false.
+            CommandContext.InitializeAnsiPassThroughFromEnvironment();
```
brthor: How does this affect the logical flow of the program?

It reads the environment variable but how does that affect subsequent code which runs? I may be missing something here.

brthor: Is this capturing the value of the environment variable before it gets changed in the next bit?

AustinWise: Yes, CommandContext.InitializeAnsiPassThroughFromEnvironment() reads the environmental variable (or uses the default value), while the following line sets the environmental variable true, which will effect any spawned processes.

As some background, all color formatting of output is done using [ASNI escape codes](https://github.com/dotnet/cli/blob/1376228e4ceade0e41e5723442672abdc17ea0e8/src/Microsoft.DotNet.Cli.Utils/AnsiColorExtensions.cs). However these are not supported by Windows so [Ansi Console](https://github.com/dotnet/cli/blob/a7f438f85d5a68c8f4c61bb422c39bbf3c1e1822/src/Microsoft.DotNet.Cli.Utils/AnsiConsole.cs#L66) parses them and changes Console.ForegroundColor appropriately.

The changes to Console.ForegroundColor have to be in the process that was directly invoked by the user, as the output redirection used when spawning processes does not capture the color changes. So the Report class [checks](https://github.com/dotnet/cli/blob/a7f438f85d5a68c8f4c61bb422c39bbf3c1e1822/src/Microsoft.DotNet.Cli.Utils/Reporter.cs#L35) the environmental variable DOTNET_CLI_CONTEXT_ANSI_PASS_THRU, and skips converting the ANSI codes.

Before the this pull request, the DOTNET_CLI_CONTEXT_ANSI_PASS_THRU was getting always set for intrinsic commands, so the ANSI escape codes were being written directly to the Windows console. This pull request makes the process use the current value of the environmental variable, which will default to false for the original invokcation of dotnet.exe. However we want any child process we spawn to pass through ANSI escape codes, so we need to set the environmental variable to true.

brthor: I have an incoming PR which may overlap a bit but it shouldn't be too bad. :smile: 

I get what you are trying to do now, but the way of doing this feels hacky. 
Reading through this bit it wouldn't be clear to me that we are saving the value so that it's not affected by the subsequent set.

With your other change this should be entirely unnecessary for internal commands now no?
So why not only set this environment variable when calling external commands, leaving it in the command.create below but removing it altogether for internal commands.

AustinWise:
For reference, this is what it looks like before this change:
Before
And after:
After
Note the funny looking "←[33m" text in the before picture.

src/dotnet/Program.cs
AustinWise: Yeah, it does not seem a bit hacky.

I think it would not be too hard to make all external external commands. Since all the commands use [Command.CreateDotnet](https://github.com/dotnet/cli/blob/65c026afbf2f7dab3ce9fd1ef480ae6233d5fb54/src/Microsoft.DotNet.Cli.Utils/Command.cs#L43), I could change that to apply the ANSI forwarding.

AustinWise:
I push a new commit that just applies the environmental variable to all children created with the Command class. I think this is the right place to set the variable as the Command does the output forwarding that depends on this variable being set in children.

src/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -28,6 +28,7 @@ private Command(CommandSpec commandSpec)
                 RedirectStandardError = true,
                 RedirectStandardOutput = true
             };
+            psi.Environment[CommandContext.Variables.AnsiPassThru] = bool.TrueString;
```
brthor: I think there are situations where we might want to use command without being forced to set this value. Testing is the first thing that comes to mind, but also this is intended to be a general purpose Command Class for use by CLI.

I prefer the more explicit method of declaring exactly what environment variables you want set with each command.

AustinWise: I pushed a revision where I only set the environmental variable if output is being redirected to the Reporter, which knows how to handle the ANSI escape codes.

This way if someone is using the Command class to forward output, colors will just work. And if the command class is being used to capture output, the consumer of the output will not unexpectedly see escape codes unless they explicitly ask for them.

I think this is preferable to having to remember to set the environmental variable each time you want to redirect output.

src/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -149,6 +149,7 @@ public Command ForwardStdOut(TextWriter to = null, bool onlyIfVerbose = false)
                 if (to == null)
                 {
                     _stdOut.ForwardTo(write: Reporter.Output.Write, writeLine: Reporter.Output.WriteLine);
+                    _process.StartInfo.Environment[CommandContext.Variables.AnsiPassThru] = bool.TrueString;
```
brthor: Alright this seems like a reasonable default, but I think there are testing scenarios where we won't want to have this set.

Can we create a Boolean option which defaults to true for these two methods, and set the value of this environment variable to that Boolean?

That way we will have a method of turning it off if we need to.

AustinWise: Oh, that's a good idea, I'll add that in.

brthor:
This LGTM.

We are having some issues with tests on the CI right now so we'll hold off on merging for the moment. 

I'd also like to see how this interacts with the new streamforwarder changes in #1240. I don't expect any issues but there is potential.

AustinWise:
If #1240 get's merged first I'll rebase ontop of it and try it out. It does not look like it would cause any problem.
Thanks for reviewing my changes!

AustinWise:
I rebased onto the tip of 1.0.0 to fix the merge conflicts.

AustinWise:
I rebased and it seems like #1364 has fixed the build problems.
I tested out my changes the colors are still fixed.

brthor:
I think we should merge this today.

@anurse want to take a quick look?

anurse:
LGTM too. ANSI pass-through should be able to go away once we change Command.cs to only redirect stdout/stderr when we actually want to capture the output. It's only there because Windows uses APIs to manage Console Colors rather than ANSI codes :)

AustinWise:
Since ANSI pass through is temporary, when I rebased I did not add the new parameter to ICommand.

As an aside, during testing I discovered that Windows 10 v1511 added support for these ANSI codes.

brthor:
@dotnet-bot Windows_NT Debug Build please

AustinWise:
@brthor Odd, I don't understand why ProcessStartInfo.Enironment stopped working on just Windows. Do you have any ideas?
```
'ProcessStartInfo' does not contain a definition for 'Environment'
```

src/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -149,6 +149,7 @@ public ICommand ForwardStdOut(TextWriter to = null, bool onlyIfVerbose = false)
                 if (to == null)
                 {
                     _stdOut.ForwardTo(writeLine: Reporter.Output.WriteLine);
+                    _process.StartInfo.Environment[CommandContext.Variables.AnsiPassThru] = bool.TrueString;
```
brthor: Why change this back?

brthor:
@AustinWise Not sure, seems weird only on Windows. Will need to investigate

AustinWise:
@dotnet-bot Windows_NT Debug Build please

AustinWise:
@brthor I fixed the compile error. As part of doing that I put the ansiPassThrough back on Command and added it to ICommand.

piotrpMSFT:
Thanks @AustinWise!

AustinWise:
@piotrpMSFT Thanks for @brthor reviewing my code. I'm looking forward to seeing pretty colors in my terminal again!

PR #1217, opened 2016-02-04
Update build readme

cdmihai:
Document the --force-incremental-unsafe flag

### DIFF
--- a/src/dotnet/commands/dotnet-build/README.md
+++ b/src/dotnet/commands/dotnet-build/README.md
@@ -30,3 +30,6 @@ In addition Compile's parameters, Build adds the following flag:
 
 --build-profile
 Prints out the incremental safety checks that users need to address in order for incremental compilation to be automatically turned on.
+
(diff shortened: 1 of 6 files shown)


src/dotnet/commands/dotnet-build/README.md
```diff
@@ -30,3 +30,6 @@ In addition Compile's parameters, Build adds the following flag:
 
 --build-profile
 Prints out the incremental safety checks that users need to address in order for incremental compilation to be automatically turned on.
+
+--force-incremental-unsafe
```
davidfowl: Can we rename this thing to `--force-rebuild`?

cdmihai: How about `--no-incremental`? It is closer the meaning of the flag.

With `--force-rebuild` users might assume the flag does more things than just turning incremental off (e.g. deleting lock or other spurious files and other artifacts).

PR #1221, opened 2016-02-04
[WIP]: Crossgen publish

davidfowl:
Added ability to crossgen after publish. Today we're using core clr from the runtime folder so the crossgen we'll use in the built in one.

/cc @jkotas @gkhanna79 @piotrpMSFT

### DIFF
--- a/src/dotnet/commands/dotnet-publish/PublishCommand.cs
+++ b/src/dotnet/commands/dotnet-publish/PublishCommand.cs
@@ -157,19 +157,15 @@ namespace Microsoft.DotNet.Tools.Publish
             {
                 Reporter.Output.WriteLine($"Generating native images for {context.ProjectFile.Name.Bold()}...");
 
-                var loopResult = Parallel.ForEach(runtimeAssemblies, (path, state) =>
+                var results = runtimeAssemblies.AsParallel().Select(path =>
                 {
                     var crossgenResult = Command.Create("crossgen", new[] { "-platform_assemblies_paths", outputPath, path }, FrameworkConstants.CommonFrameworks.DnxCore50)
                         .WorkingDirectory(outputPath)
                         .Execute();
-
-                    if (crossgenResult.ExitCode != 0)
-                    {
-                        state.Stop();
-                    }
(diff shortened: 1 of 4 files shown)


src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -142,6 +147,23 @@ private static bool PublishProjectContext(ProjectContext context, string outputP
                 }
             }
 
+            if (Crossgen)
+            {
+                Reporter.Output.WriteLine($"Generating native images for {context.ProjectFile.Name.Bold()}...");
+                
+                foreach (var path in runtimeAssemblies)
```
jkotas: This would benefit from running one crossgen per processor in parallel.

src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -142,6 +147,23 @@ private static bool PublishProjectContext(ProjectContext context, string outputP
...
+                {
+                    var crossgenResult = Command.Create("crossgen", new[] { "-platform_assemblies_paths", outputPath, path }, FrameworkConstants.CommonFrameworks.DnxCore50)
+                        .WorkingDirectory(outputPath)
```
jkotas: Does this need to be `-readytorun`? (Depends how you expect servicing to work.)

gkhanna79:
CC @JohnChen0 

I concur with @jkotas that we should consider starting with R2R and change course only if there is a good reason to do so.

davidfowl:
@dotnet-bot test this please

davidfowl:
> I concur with @jkotas that we should consider starting with R2R and change course only if there is a good reason to do so.

There needs to be a flag to go non R2R. I plan to use this on the CLI itself where dependencies don't change at all. R2R is a bit slower in some cases so I don't always want it on.

davidfowl:
@dotnet-bot test this please

davidfowl:
@dotnet-bot build OSX Debug please

davidfowl:
@dotnet-bot test OSX Debug please

PR #1225, opened 2016-02-04
Fixed dictionary key conflict on binding redirects generation

### LINKED ISSUE #865: Error on dotnet run
Guys,

Not sure how to proceed.

Thank you!

A bit more context:

```
"Log4Net/2.0.5": {
        "type": "package",
        "compile": {
        },
        "runtime": {
        }
      },
        "type": "package",
        "dependencies": {
        "log4net": "2.0.3"
        },
        "compile": {
        },
        "runtime": {
        }
      }
```

UPDATE2:

marono:

Fix #865

### DIFF
--- a/src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
@@ -109,7 +109,7 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         private static AssemblyRedirect[] CollectRedirects(IEnumerable<LibraryExport> dependencies)
         {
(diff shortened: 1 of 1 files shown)


davidfowl:
This needs a test

marono:
Is that something I've messed up? When building it locally it seems to succeed...

shahid-pk:
@marono all work is done in the rel/1.0.0 branch may be master was broken before your patch.

marono:
Do you need me to rebase it, or will you do it ?

davidfowl:
@maririos 2 things:
- Send the pull request to rel/1.0.0
- Write a test that verifies the problem is fixed.

marono:
Ok, will do, thanks!

PR #1226, opened 2016-02-04
Handle Windows directory separators in PopulateLegacyPortableDependencies

bricelam:
@davidfowl


davidfowl:
I forget we normalize the slashes when reading the lock file. Might be better to just use `Path.DirectorySeparatorChar`

bricelam:
Updated.

Issue #1091, opened 2016-01-28
Support installing a particular version of the CLI


victorhurdugaci:
It would be nice if the install scripts under `scripts\obtain` would support installing a particular version of the CLI, not only the latest. It would help us when we find a blocking bug and we need to stick to a particular version of the CLI for a while

piotrpMSFT:
:+1:

pranavkm:
Fixed in https://github.com/dotnet/cli/commit/ed1177a664106ee312003983474b55f0e3bc627f

PR #1228, opened 2016-02-04
sync install script params with PowerShell a bit

anurse:
Small change to install.sh to allow channel to be passed as an arg, like it is in install.ps1

/cc @victorhurdugaci @piotrpMSFT @brthor @Sridhar-MS

### DIFF
--- a/scripts/obtain/install.sh
+++ b/scripts/obtain/install.sh
@@ -54,15 +54,6 @@ exec "$MY_TARGET" "$@"
 EOF
 )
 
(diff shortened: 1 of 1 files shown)


victorhurdugaci:
While you're at this, don't you want to add an extra parameter for https://github.com/dotnet/cli/issues/1091 ? :)

anurse:
I could, but that would mean more work and more testing :). It doesn't make it any harder to add support for #1091. I just have other stuff to get back to :(

davidfowl:
:shipit:

PR #1232, opened 2016-02-04
Add a auto-generated header to the assembly info file that is created on build

### LINKED ISSUE #1177: AssemblyInfoFileGenerator should generate an auto generated header
For tools and analyzers the assembly info file should contain a header that states that it is auto generated.

pdelvo:
I also renamed the `Generate` method to `GenerateCSharp` to match the FSharp version and the (future) VB version of it.

Fixes #1177


davidfowl:
@pdelvo Can you move this PR to the rel/1.0.0 branch?

PR #1234, opened 2016-02-05
compile native.standalone

piotrpMSFT:
#1213 

This covers stage 1, making dotnet-compile-native stand-alone.

/cc @brthor @schellap @davidfowl

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1234)


src/dotnet/commands/dotnet-compile/Program.cs
```diff
@@ -47,89 +47,17 @@ private static bool OnExecute(List<ProjectContext> contexts, CompilerCommandApp
...
             {
-                nativeArgs.Add("--arch");
-                nativeArgs.Add(args.ArchValue);
+                
+                var result = Command.CreateDotNet("compile-native", new string[] {})
+                    .ForwardStdErr()
+                    .ForwardStdOut()
+                    .Execute();
             }
+<<<<<<< ours
```
brthor: check this.

src/dotnet/commands/dotnet-build/Program.cs
```diff
@@ -33,7 +33,7 @@ public static int Run(string[] args)
             }
         }
 
-        private static bool OnExecute(List<ProjectContext> contexts, CompilerCommandApp args)
+        private static bool OnExecute(List<ProjectContext> contexts, NativeCompilerCommandApp args)
```
cdmihai: Indeed, that looks so horrible. I'll just clone the code in the --no-dependency change

src/dotnet/commands/dotnet-compile-native/NativeCompilerCommandApp.cs
```diff
@@ -0,0 +1,147 @@
...
+            baseClassOptions = new Dictionary<string, CommandOption>();
+
+            AddCompileParameters();
+        }
+
+        private void AddCompileParameters()
+        {
+            _app.HelpOption("-h|--help");
+
+            _outputOption = _app.Option("-o|--output <OUTPUT_DIR>", "Directory in which to place outputs", CommandOptionType.SingleValue);
```
tijoytom: Does this mean native compilation will do an IL compilation ( ie dotnet compile) first and then use the managed output?I don't see the managed assembly being passed in to ilc , unless someone is changing ilc to take in a project.json

schellap: The native tool is invoked now. See CompileCommand.

tijoytom: I see the temporary workaround, what is the long term command. "dotnet compile-native  <project.json> | managed dll"? or something like "dotnet compile native" which essentially pipe the commands. I think the "dotnet compile native" pipelining has value since dotnet native can work with any managed compiler ..example "dotnet compile-fsc native".

schellap: @blackdwarf, @piotrpMSFT for user experience, the current --native is better as the user simply passes native option to get a native exe rather than a two step process to get the native exe.

schellap: @piotrpMSFT, why are we moving ilc alone out of the repo to corert? I thought you mentioned csc fsc etc, moving too in the future? @gkhanna79

tijoytom: @blackdwarf, @piotrpMSFT @schellap  For user experience "dotnet compile --native" and "dotnet compile native" is not very much different. Also if i want to conditionally enable subcommands to native , i will have to add another option to CompileCommand ,  for example  i want to conditionally enable mcg  "dotnet compile --native  --enablemcg". I am not sure if there is another neat way to to pass sub-commands to native rather than  polluting CompileCommand.

schellap: @tijoytom, `dotnet compile native` and `dotnet compile-native` are different, the second one means the user explicitly invokes it. To answer your original question, @piotrpMSFT correct me if I am wrong, the final orchestrator will be `build -n` which will work like today except that it will invoke `compile-native` in the end? Are you planning to include that as well?

src/dotnet/commands/dotnet-build/Program.cs
piotrpMSFT: sweet!

src/dotnet/commands/dotnet-compile-native/NativeCompilerCommandApp.cs
piotrpMSFT: @schellap, that's right. They are different. Native is a post-IL compile step. `dotnet-compile`s job is strictly to generate IL and the existing native integration has been a rather strange leftover from preliminary ideas. 

dotnet-build is the orchestrator, so if any orchestration of compile/compile-native occurs, that should occur in build.

dotnet-build should not be explicitly aware of native in the rel branch. That will need to get brought into the rel branch that stabilizes the first CLI release to include a RTM native. We should experiment with this in the master branch for the time being.

@tijoytom now that dotnet-compile-native is a stand-alone command it should have its own rules for managing its own options/parameters.

src/dotnet/commands/dotnet-compile-native/CompileNativeCommand.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            // Write Response File
+            var rsp = Path.Combine(nativeIntermediateOutputPath, $"dotnet-compile-native.{context.ProjectFile.Name}.rsp");
+            File.WriteAllLines(rsp, nativeArgs);
+
+            // TODO Add -r assembly.dll for all Nuget References
+            //     Need CoreRT Framework published to nuget
+
+            // Do Native Compilation
+            var exitCode = CompileNativeCommandInternal.Run(new string[] { "--rsp", $"{rsp}" });
```
brthor: I think you could just do `.Run(nativeArgs)` rather than writing the response file here. Now that this isn't a separate process not sure this is necessary.

piotrpMSFT: @brthor I think there is a lot of opportunity for cleanup in another PR

schellap:
@piotrpMSFT LGTM. Please resolve and squash!

src/dotnet/commands/dotnet-compile/CompileCommand.cs
```diff
@@ -47,109 +47,17 @@ private static bool OnExecute(List<ProjectContext> contexts, CompilerCommandApp
...
-            }
-
-            // Architecture
-            if (args.ArchValue != null)
+            
+            if (args.IsNativeValue && success)
             {
-                nativeArgs.Add("--arch");
-                nativeArgs.Add(args.ArchValue);
+                var result = Command.CreateDotNet("compile-native", new string[] {})
```
schellap: Didn't pass in the args?

schellap:
@piotrpMSFT, @tijoytom and I are going to edit native to enable a couple of scenarios that customers who are trying our have been asking about (basically wire dependencies from project.json and in turn use response files for ILC and MCG.) Can you hold off till then? I can merge for you once we are done.

piotrpMSFT:
Ok

schellap:
@piotrpMSFT once you are able to resolve the merge conflicts, I can take a look. Let me know if you need help with this.

Cc @tijoytom

scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -13,15 +13,17 @@ namespace Microsoft.DotNet.Cli.Build
 {
     public class TestTargets
     {
-        public static readonly string[] TestPackageProjects = new[]
+        public static readonly dynamic[] PackageProjects = new[]
```
brthor: Why dynamic vs some object?

PR #1235, opened 2016-02-05
Print RID

natemcmaster:
`dotnet --rid` prints the default RID.

Why? For general use with other scripting tools/languages. e.g. save users the hassle of figuring out the magic combitation of `dotnet --version`, `grep`, and `awk` to get RID when writing x-plat friendly scripts.

### DIFF
--- a/Documentation/manpages/dotnet.1
+++ b/Documentation/manpages/dotnet.1
@@ -7,7 +7,8 @@
 dotnet \-\- general driver for running the command\-line commands
 .SH SYNOPSIS
 .PP
-dotnet [\-\-version] [\-\-help] [\-\-verbose] < command > [< args >]
+dotnet [\-\-version] [\-\-rid] [\-\-help] [\-\-verbose] < command > [<
+args >]
 .SH DESCRIPTION
 .PP
 dotnet is a generic driver for the CLI toolchain.
@@ -33,6 +34,14 @@ Enable\ verbose\ output.
 .nf
 \f[C]
 Print\ out\ the\ version\ of\ the\ CLI\ tooling
+\f[]
+.fi
+.PP
+\f[C]\-\-rid\f[]
+.IP
+.nf
+\f[C]
+Print\ out\ the\ default\ runtime\ identifier\ of\ the\ CLI\ tooling
 \f[]
 .fi
 .PP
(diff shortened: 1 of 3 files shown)


davidfowl:
This doesn't deserve to be its own switch. It's already printed as part of the version. How are you using this today?

natemcmaster:
>  It's already printed as part of the version.

I'm aware of that. It can be extracted by finding the substring of version, but why make every script perform the substring to get the rid? Adding an option to get rid directly is generally useful.

My use case: labeling artifacts produced on CI agents. `dotnet publish -o artifacts/myApp/${RID}`. I don't want to depend on the exact spacing and output of `--version` since who knows when that will change.

Currently:
```xml
  <Target Name="_GetRid">
    <Exec Command="dotnet.exe --version" ConsoleToMSBuild="true">
        <Output TaskParameter="ConsoleOutput" PropertyName="DotNetVersion" />
    </Exec>
    <PropertyGroup>
        <Rid>$(DotNetVersion.Substring($([MSBuild]::Add(13, $(DotNetVersion.IndexOf('Runtime Id'))))))</Rid>
    </PropertyGroup>
  </Target>
```

davidfowl:
@natemcmaster doesn't that feel super special casey?

natemcmaster:
@davidfowl it could still be useful in other scenarios. But as long as the format of version string doesn't change too often, making devs bite the `awk` and `sed` bullet is a probably a fair tradeoff for (your presumed intention of) avoiding top-level switches except as necessary.

factormystic:
Would it be possible to make `rid` a param of `--version`? Then that'd open the possibility of having porcelain access to other version properties.

natemcmaster:
Won't need this anymore when some new tooling comes along. Not worth adding as a top level switch

sejongoh:
I was looking for a way to automatically get rid in a test script to download a right package and got it by:
 
```
../Tools/dotnetcli/bin/dotnet --version | grep 'Runtime Id:' | sed 's/^ *Runtime Id: *//g'
```
 
Although I am not sure how many other scripts need to extract rid, something like ‘dotnet –rid’ looks useful at least to me since I don’t need to worry about the output format of ‘dotnet –version’.

schellap:
I just found myself in need of the `dotnet --rid` or `dotnet rid` to pipe the machine `rid` into the corehost at build time. It would be nice to just get it out of dotnet directly rather than do jingbang.

+1 for this change.

PR #1243, opened 2016-02-05
Remove DOTNET_HOME from the CoreCLR probe path

schellap:


davidfowl:
```
Test Failed: test\PackagedCommands\Consumers\AppWithDirectAndToolDependency\dotnet hello
             printed Could not resolve coreclr path
The following tests failed.
E2E.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\E2E-testResults.xml'
StreamForwarderTests.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\StreamForwarderTests-testResults.xml'
dotnet-publish.Tests.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\dotnet-publish.Tests-testResults.xml'
dotnet-compile.Tests.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\dotnet-compile.Tests-testResults.xml'
dotnet-build.Tests.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\dotnet-build.Tests-testResults.xml'
Compiler.Common.Tests.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\Compiler.Common.Tests-testResults.xml'
package-command-test.ps1.dll failed. Logs in 'D:\j\workspace\dotnet_cli\release_windows_nt_prtest\artifacts\tests\package-command-test.ps1-testResults.xml'
Command Failed: '& D:\j\workspace\dotnet_cli\release_windows_nt_prtest\scripts\test\run-tests.ps1'
Command Failed: '& D:\j\workspace\dotnet_cli\release_windows_nt_prtest\scripts\test\test.ps1'
```

Seems like it doesn't work :smile:

schellap:
Haven't tested yet :)

src/corehost/cli/deps_resolver.cpp
```diff
@@ -415,6 +415,45 @@ void deps_resolver_t::get_local_assemblies(const pal::string_t& dir)
...
+    const pal::string_t& package_dir,
+    const pal::string_t& package_cache_dir)
+{
+    pal::string_t candidate;
+    if (!package_cache_dir.empty())
+    {
+        for (const deps_entry_t& entry : m_deps_entries)
+        {
+            if (entry.asset_type == _X("native") &&
+                entry.asset_name == LIBCORECLR_NAME &&
```
davidfowl: There's no extension on the asset name  it's `libcoreclr` on OSX for example:

```
...,"runtime","mscorlib","runtimes/osx.10.10-x64/lib/dotnet/mscorlib.dll"
...,"native","System.Globalization.Native","runtimes/osx.10.10-x64/native/System.Globalization.Native.dylib"
...,"native","libcoreclr","runtimes/osx.10.10-x64/native/libcoreclr.dylib"
...,"native","libdbgshim","runtimes/osx.10.10-x64/native/libdbgshim.dylib"
...,"native","libmscordaccore","runtimes/osx.10.10-x64/native/libmscordaccore.dylib"
...,"native","libmscordbi","runtimes/osx.10.10-x64/native/libmscordbi.dylib"
...,"native","libryujit","runtimes/osx.10.10-x64/native/libryujit.dylib"
...,"native","libsos","runtimes/osx.10.10-x64/native/libsos.dylib"
```

schellap:
The problem is CoreCLR.dll wants to find mscorlib in its own dir. See deps -- mscorlib is inside `lib` and not inside `native`. When I copied over mscorlib.ni.dll with coreclr.dll it works fine.

davidfowl:
@schellap Then we're going to have to copy those 2 app local when in package mode. As part of publish.

schellap:
@davidfowl for now I am just removing the dotnet_home from the probe path. APP_DIR will win. Publish can copy over or publish the coreclr package alone.

The right long term fix is to fetch mscorlib from TPA -- if thats bad for perf, use a property to explicitly specify MSCORLIB_PATH. It is not clear to me (guardrails) why they can't live together in the nupkg. @gkhanna79, who can help exposing a property or load mscorlib from TPA?

https://github.com/dotnet/coreclr/blob/43b39a73cbf832ec13ec29ed356cb75834e7a8d7/src/binder/assemblybinder.cpp#L700

gkhanna79:
I have published a change where GH built CoreCLR nuget packages will have mscorlib.ni.dll next to CoreCLR. Once I consolidate these with TFS builds, published packages would have the same property.

piotrpMSFT:
@dotnet-bot test this please

schellap:
@gkhanna79 could you please let us know when the TFS integration and the packages will be pushed? I will update the NETStandard.Library references to use the new version and this change should be able to go in.

gkhanna79:
@schellap Will do. Right now, I am trying to get CLI for CentOS working to restore build tools. Once this is fixed, I intend to merge with TFS side and you should be able to consume this.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -32,12 +32,16 @@ enum StatusCode
...
+    append_path(&svc_clr, _X("runtime"));
+    append_path(&svc_clr, _X("coreclr"));
+
     const pal::string_t* dirs[] = {
-        &args.dotnet_runtime_servicing, // DOTNET_RUNTIME_SERVICING
-        &args.app_dir,                  // APP LOCAL
-        &args.dotnet_home               // DOTNET_HOME
+        &svc_clr,                       // DOTNET_RUNTIME_SERVICING
+        &deps_clr_dir,                  // DEPS BASED CLR
+        &args.app_dir                   // APP LOCAL
```
gkhanna79: @schellap Why are changing the order of probing? wont this deprioritize app-local lookup? 

CC @davidfowl

davidfowl:
@schellap to unblock this we need to copy mscorlib side by side coreclr in the clr package

schellap:
@gkhanna79 @davidfowl copying over on windows due to UAC with symlinking. On unix we symlink.

@shhsu, once this change goes in please give it a shot, as we load coreclr slightly differently now.

CLR probe order is similar to TPA etc.

davidfowl:
Oh don't copy it at runtime do it at build time.

schellap:
How would it work if it was published on Windows for OSX rid and ran on OSX?

schellap:
Oh restore will take care of the cache.

davidfowl:
@schellap doing it on restore sounds much better but it would be such and uber hack...

davidfowl:
@schellap Does this logic work? Restore is pretty gross since it's so specific to this particular package and bug. I'm thinking 3 options:

1. Do it on restore. This matches what will happen with the packages once @gkhanna79 gets his change in but it's too hacky for my liking.
2. Do it at runtime - bad and platform specific, and hard to get right
3. Do it at compile time. Copy mscorlib to the coreclr folder - This is the best solution since the problem only reproduces on build not publish.

I'm going to have to vote for 3. Now this logic won't fix the packages  cache, we'll need to wait on @gkhanna79 's fix for that one.

schellap:
I think this is a hack and super temporary, probably will last for a week. Why not go ahead with the option 2 I have in the PR? It works on OSX, Ubuntu and Windows in my testing. Copies once per package version on a machine and does exists check after that.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -415,6 +454,109 @@ void deps_resolver_t::get_local_assemblies(const pal::string_t& dir)
 }
 
 // -----------------------------------------------------------------------------
+// Patch mscorlib from package relative dir if not sxs coreclr.
+//
+// TEMPORARY WORKAROUND for NuGet packages that don't contain sxs mscorlib.
```
gkhanna79: We should file an issue to remove this once the package comes with mscorlib.ni.dll

src/corehost/common/pal.unix.cpp
```diff
@@ -20,6 +21,17 @@
 #define symlinkEntrypointExecutable "/proc/curproc/exe"
 #endif
 
+bool pal::ensure_file(const pal::string_t& src, const pal::string_t& dest)
```
gkhanna79: If this is present only for mscorlib patching, lets have a TODO to remove this as well. And given that it is temporary, I would suggest we keep the behaviour consistent and copy on non-Windows as well (eventually, it will be a non-symlinked mscorlib.ni.dll).

gkhanna79:
> Why not go ahead with the option 2 I have in the PR? It works on OSX, Ubuntu and Windows in my testing. 

One advantage of going with (3) is that we dont need to add any runtime code and thus, when my package changes lightup, we wont have to do any undo in the CLI implementation.

schellap:
@gkhanna79, We don't have to do anything once your changes go in, as file exists would succeed and patch is a no-op then. I am inclined to merge as is than do more work on this already awful hack which is throw away in 1 week. Cc @piotrpMSFT.

gkhanna79:
You will need to still file an issue to remove this code once the actual changes are in.

anurse:
We really shouldn't do the symlinking/copying at runtime. It seems that it messed up the package cache on one of the OSX agents and caused some other builds to fail. I had to ssh in and remove the symlinked mscorlib from `.nuget/packages` to unblock the build.

We could probably do it on build, and should definitely do it via a copy on all platforms (since I think it was the symlink itself that messed things up in the cache; I'm not sure all the NuGet code is able to cope with symlinks).

schellap:
@dotnet-bot test this please

schellap:
@dotnet-bot test Windows_NT Release Build please

anurse:
> 10:18:24 warn : Dependency specified was System.Reflection.Metadata (>= 1.2.0-rc2-23811) but ended up with System.Reflection.Metadata 1.3.0-beta-23811.
> 10:18:24 warn : Dependency specified was System.Reflection.Metadata (>= 1.2.0-rc2-23811) but ended up with System.Reflection.Metadata 1.3.0-beta-23811.
> 10:18:24 warn : Dependency specified was System.Reflection.Metadata (>= 1.2.0-rc2-23811) but ended up with System.Reflection.Metadata 1.3.0-beta-23811.

Looks like System.Reflection.Metadata `1.2.0-rc2-23811` is not available on `cli-deps` so we're bumping to `1.3.0-beta-23811` which **is** on our feed /cc @eerhardt . I'm not sure that's causing the failure though, since the actual error is a failure to initialize the CoreCLR, but that can be caused by messed up assemblies :).

eerhardt:
I've pushed 1.2.0-rc2-23811 of System.Reflection.Metadata to our cli-deps feed.

schellap:
@davidfowl want to have the honor of merging this ;)

piotrpMSFT:
Fantastic!

PR #1256, opened 2016-02-05
Fix the package cache on CI on Unix.

eerhardt:
On the CI machines, the $HOME variable is redirected to the artifacts folder, which is under the repo root. The package cache is using $HOME. Thus between runs, the cache is removed.

To fix this, change the package cache on CI runs to be outside of the repo.

@davidfowl @mmitche

### DIFF
--- a/scripts/common/_nuget.sh
+++ b/scripts/common/_nuget.sh
@@ -4,7 +4,15 @@
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 #
 
-export NUGET_PACKAGES=~/.nuget/packages
+if [ ! -z "$CI_BUILD" ]; then
+	# On CI, $HOME is redirected under the repo, which gets deleted after every build.
+	# So make $NUGET_PACKAGES outside of the repo.
+	NUGET_PACKAGES=$REPOROOT/../.nuget/packages
+else
+	NUGET_PACKAGES=~/.nuget/packages
+fi
+
+export NUGET_PACKAGES
 export DOTNET_PACKAGES=$NUGET_PACKAGES
 export DNX_PACKAGES=$NUGET_PACKAGES
 


eerhardt:
@dotnet-bot test this please

mmitche:
@eerhardt Please stand by, the server is being restarted.

mmitche:
@dotnet-bot test this please

eerhardt:
CentOS is broke right now because we don't have a corresponding runtime.centos.7.1-x64.microsoft.netcore.runtime.coreclr package for our current .Net Core - version 1.0.1-rc2.23805.

@ericstj @dagood - is CentOS package publishing broke right now?

ericstj:
@ellismg removed it and replaced it with RHEL package.  RHEL should be applicable to centos if you're using the latest rid graph (which is part of the same build in Microsoft.NETCore.Platforms).

eerhardt:
@ellismg and @ericstj - why did we remove it?  Why not just add rhel and keep both for a while?

ericstj:
I mentioned that to @ellismg but I think he preferred to have a single build.  How is it breaking?  The centos RIDs import RHEL so it should just work.

ellismg:
We did not unpublish existing packages and the rid graph updates happened in the same version where we stopped producing CentOS specific packages.  Maybe there is a mismatch of versions somewhere?  Are you picking up the right version of the platforms package?

eerhardt:
dotnet/cli broke when moving to the latest .Net Core build - rc2-23805. We have some RIDs hard-coded in our scripts.  I cc'd you guys on my PR that should fix it.

eerhardt:
@dotnet-bot test CentOS7.1 Debug Build

eerhardt:
@dotnet-bot test CentOS7.1 Release Build

eerhardt:
@dotnet-bot test Ubuntu Release Build

mmitche:
What kind of troubles are you seeing?

Sent from my Windows Phone

eerhardt:
There was an I/O error during the build. This could be caused by a hung process on the target machine, preventing it from cleaning up the workspace, or other issues, such as low disk space.
Indication 3

mmitche:
I don't think it's low disk space or anything like that.  I also saw this in a different cli PR on two different OS's.

mmitche:
@dotnet-bot test CentOS7.1 Release Build please

eerhardt:
@dotnet-bot test this please

Issue #453, opened 2015-12-09
dotnet-compile-native does not work on centos


Sridhar-MS:
There is no corert packages for centos 7 yet.

/cc: @schellap @brthor

gkhanna79:
@Sridhar-MS @piotrpMSFT I dont see CentOs at https://github.com/dotnet/cli. Is CentOs support for CLI targeted for RC2?

Sridhar-MS:
@gkhanna79 CentOS builds are currently enabled and we are targeting for beta(still working through some bugs). We have pushed native compilation on CentOS to RC2.

gkhanna79:
We are not targeting CoreRT for CentOS yet. It will likely be a post-RTM effort.

piotrpMSFT:
Closing this issue. CLI enables compile-native wherever it's available.

PR #1262, opened 2016-02-06
Fix CentOS on the latest .Net Core.

eerhardt:
The latest .Net Core renamed the RID of the 'centos' packages to 'rhel'.  Migrating our scripts to use this name.

@ericstj @ellismg @piotrpMSFT @anurse

### DIFF
--- a/scripts/crossgen/crossgen_roslyn.sh
+++ b/scripts/crossgen/crossgen_roslyn.sh
@@ -16,7 +16,7 @@ if [ "$OSNAME" == "osx" ]; then
 elif [ "$OSNAME" == "ubuntu" ]; then
     RID=ubuntu.14.04-x64
 elif [ "$OSNAME" == "centos" ]; then
-    RID=centos.7.1-x64
+    RID=rhel.7-x64
 else
     echo "Unknown OS: $OSNAME" 1>&2
     exit 1


ellismg:
You shouldn't need to do this and it is pointing to a problem in the system somewhere.  The centos.7 rids should be pulling in the rhel.7 assets.

eerhardt:
> You shouldn't need to do this and it is pointing to a problem in the system somewhere.

@anurse @davidfowl - care to comment? How else are we supposed to get the RID in our scripts?

scripts/common/_rid.sh
```diff
@@ -23,7 +23,7 @@ if [ -z "$RID" ]; then
             export RID=ubuntu.14.04-x64
         elif [ "$(cat /etc/*-release | grep -cim1 centos)" -eq 1 ]; then
             export OSNAME=centos
-            export RID=centos.7-x64
+            export RID=rhel.7-x64
```
anurse: This one shouldn't need to be changed, it's given to NuGet and NuGet knows what to do.

scripts/build/build_appdeps.sh
```diff
@@ -22,7 +22,7 @@ if [ "$OSNAME" == "osx" ]; then
 elif [ "$OSNAME" == "ubuntu" ]; then
     RID=ubuntu.14.04-x64
 elif [ "$OSNAME" == "centos" ]; then
-    RID=centos.7.1-x64
+    RID=rhel.7.1-x64
```
anurse: I don't think we need this unless the CoreRT toolchain package has changed it's ID.

anurse:
@ellismg Actually, it's not about the RID, it's just about the Package ID. Since the package ID now has `rhel` as the RID, and we need to find crossgen inside it, we have to change our search path. The RID inheritance is working fine AFAIK

scripts/build/build_appdeps.sh
eerhardt: Looks like there is no CoreRT toolchain for centos or rhel.  @schellap - is this a known TODO?

eerhardt:
OK - thanks @anurse and @ellismg.  I've updated just the one crossgen script, which should unblock our CI.

scripts/build/build_appdeps.sh
schellap: Yes. We don't have CentOS support tracked by https://github.com/dotnet/cli/issues/453

piotrpMSFT:
One? We should need to update ps1 and sh...

eerhardt:
> One? We should need to update ps1 and sh...

We don't need to update ps1 since that doesn't run on CentOS.

eerhardt:
@dotnet-bot test CentOS7.1 Debug Build

ellismg:
Sorry for the trouble this caused, BTW.

eerhardt:
> Sorry for the trouble this caused, BTW.

No worries. It was pretty easy to fix, once I understood what the issue was. I thought we were missing packages at first.

PR #1271, opened 2016-02-06
Disable crossgen, and provide centralized ability to do so.

piotrpMSFT:
CrossGen'd images are still causing CLI commands to fail. This change disables crossgen and provides a single flag for controlling crossgen.


davidfowl:
We can't ship this way. We need to keep a crossgened CLI working.

piotrpMSFT:
Agreed. I'm working with Jan and Stephen to get to the bottom of this. For now let's unblock CI

PR #1281, opened 2016-02-08
* Don't throw if Assembly.GetEntryAssembly is null

pranavkm:
* Add a CompilationOptions.Default for convenient use when DependencyContext.Default is null.


pranavkm:
cc @pakrym

pakrym:
@dotnet-bot run tests please

src/Microsoft.Extensions.DependencyModel/CompilationOptions.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Linq;
```
pakrym: ?

pakrym:
:shipit:

Issue #1251, opened 2016-02-05
Dependency names collusion cause DependencyModel failure during build


troydai:
Build following project:
```
{
    "version": "1.0.0-*",
    "compilationOptions": {
        "emitEntryPoint": true,
        "preserveCompilationContext": true
    },

    "dependencies": {
        "System.IO": "4.1.0-*"
    },

    "frameworks": {
        "dnx451": { 
            "frameworkAssemblies": {
                "System.IO": ""
            }
        }
    }
}
```

Build Error:
```
Can not add property System.IO to Newtonsoft.Json.Linq.JObject. Property with the same name already exists on object.
```

PR #1284, opened 2016-02-08
Fix dependency context bug and add load overload

pakrym:
https://github.com/dotnet/cli/issues/1251

@davidfowl @troydai @anurse

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
+++ b/src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
@@ -68,8 +68,12 @@ namespace Microsoft.Extensions.DependencyModel
 
             var libraryAssets = runtime ? export.RuntimeAssemblies : export.CompilationAssemblies;
 
-            foreach (var libraryDependency in export.Library.Dependencies)
+            foreach (var libraryDependenciesGroup in export.Library.Dependencies.GroupBy(d => d.Name))
             {
+                LibraryRange libraryDependency = libraryDependenciesGroup
+                    .OrderByDescending(d => d.Target == LibraryType.ReferenceAssembly)
+                    .First();
+
                 Dependency dependency;
                 if (dependencyLookup.TryGetValue(libraryDependency.Name, out dependency))
                 {
--- a/src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
+++ b/src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
@@ -38,7 +38,12 @@ namespace Microsoft.Extensions.DependencyModel
         private static DependencyContext LoadDefault()
         {
             var entryAssembly = Assembly.GetEntryAssembly();
-            var stream = entryAssembly?.GetManifestResourceStream(entryAssembly.GetName().Name + DepsResourceSufix);
+            return Load(entryAssembly);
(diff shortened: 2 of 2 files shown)


src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
```diff
@@ -38,7 +38,12 @@ public DependencyContext(string target, string runtime, CompilationOptions compi
         private static DependencyContext LoadDefault()
         {
             var entryAssembly = (Assembly)typeof(Assembly).GetTypeInfo().GetDeclaredMethod("GetEntryAssembly").Invoke(null, null);
-            var stream = entryAssembly.GetManifestResourceStream(entryAssembly.GetName().Name + DepsResourceSufix);
+            return Load(entryAssembly);
+        }
+
+        public static DependencyContext Load(Assembly assembly)
+        {
+            var stream = assembly.GetManifestResourceStream(assembly.GetName().Name + DepsResourceSufix);
```
troydai: Just curious, why do you split the method here?

troydai:
:shipit: 
/cc @davidfowl could you take a look? this blocks OmniSharp build.

troydai:
@pakrym I'd like to have this fix checked in as soon as possible. However, it occurs to me that there is no test case associated. Can we add test coverage afterwards?

src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
```diff
@@ -68,8 +68,12 @@ private static CompilationOptions GetCompilationOptions(CommonCompilerOptions co
 
             var libraryAssets = runtime ? export.RuntimeAssemblies : export.CompilationAssemblies;
 
-            foreach (var libraryDependency in export.Library.Dependencies)
+            foreach (var libraryDependenciesGroup in export.Library.Dependencies.GroupBy(d => d.Name))
             {
+                LibraryRange libraryDependency = libraryDependenciesGroup
+                    .OrderByDescending(d => d.Target == LibraryType.ReferenceAssembly)
```
troydai: Why not `.First(d => d.Target == LibraryType.ReferenceAssembly)`? is there any possibility that two reference assemblies coexist?

pakrym: There may be multiple dependencies with the same name and we want to prefer `ReferenceAssembly` but if it doesn't exist take whats left

src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
pakrym: If you want to load dependency context for non entry point assembly, for example `xunit` is entry assembly and you wand to get dependency context for test assembly.

src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
troydai: ok. order could be expensive. but i guess its ok, since there shouldn't be too many items in one group.

troydai:
:shipit:

fearthecowboy:
When will this make it to the 'latest' build? I'm blocked on this, and really need new bits :)

PR #1293, opened 2016-02-08
Use csc-compile command from mcg.

tijoytom:
Fix mcg command , recent changes broke it.Also make use of csc-compile command to compile mcg generated code.I have been waiting for stand-alone native command change to merge to integrate this, but not having mcg is blocking(https://github.com/dotnet/corert/issues/828)


tijoytom:
@piotrpMSFT @brthor I have been waiting for peter's change , but this is blocking simple test in CoreRT and we need to get this merged.Probably i will have some re-work to do once peter's change comes in , but those should be straight forward.

src/dotnet/commands/dotnet-compile-native/Program.cs
```diff
@@ -2,11 +2,13 @@
 using System.IO;
 using Microsoft.DotNet.Cli.Utils;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.DotNet.Tools.Compiler.Native
 {
     public class CompileNativeCommand
     {
+        private const string McgInteropAssemblySuffix = ".mcginterop.dll";
```
jkotas: This needs to be `.McgInterop` . It will not work on Unix otherwise because of Unix is case sensitive.

PR #1317, opened 2016-02-09
Move the debian package_tool into our \tools directory.

eerhardt:
The package_tool isn't really part of our product, so moving it to the tools directory.

@brthor


brthor:
Unfortunately we won't know if this is actually working because of the Ubuntu issues, but the change looks good to me.

IMO :shipit:

eerhardt:
@brthor - Do you think we should wait with this until the Ubuntu build is fixed?

brthor:
Is this move blocking you? 

If it's not blocking I think it would make sense, but if it is we're not producing Ubuntu binaries at the moment anyways so it's probably okay.

eerhardt:
> Is this move blocking you?

No, I'm just going through cleaning up our assets. I'll wait a few days. Hopefully Ubuntu is unblocked soon.

brthor:
@dotnet-bot test this please

brthor:
@eerhardt 
looks like you may need to do some chmodding.. did you move things around on windows?
If possible it'd be best to do this on a unix system to retain the file permissions.

/opt/code/tools/DebianPackageTool/package_tool: Permission denied

@anurse Why would this be coming up green?

eerhardt:
> looks like you may need to do some chmodding.. did you move things around on windows?
If possible it'd be best to do this on a unix system to retain the file permissions.

Yes that is my problem, I used Windows. Does git preserve file permissions? If so, I didn't know that.

piotrpMSFT:
@eerhardt git does indeed!

brthor:
git generally does when modifying files, but I'm not sure how it handles when files are being moved

anurse:
@brthor likely because something isn't returning the correct exit code. Are you able to confirm that `package.sh` is returning a non-zero exit code?

brthor:
hmmm appears that it is:
opt/code/scripts/package/package.sh exited with 0

Will check it out

brthor:
Looks like set -e got taken out a while back, will add it back today.

eerhardt:
@brthor - I will add it to this PR when I fix the file permission issue.

brthor:
@eerhardt think I opened the PR same time as you commented
https://github.com/dotnet/cli/pull/1454

Will be good to get it in ASAP anyways

eerhardt:
I changed the file permissions on package_tool to 755 like it previously was and rebased.

I'll manually check the Ubuntu builds to make sure the package is created.

brthor:
Taking a look at the diff, there are a number of files that had their permissions changed.

It doesn't look like these are causing breaks, but for the following files the change is not desirable because they are used when devving on the package tool itself:

`.../debian/package_tool/setup/build_setup.sh → tools/DebianPackageTool/setup/build_setup.sh  100755 → 100644 `

`...g/debian/package_tool/setup/test_setup.sh → tools/DebianPackageTool/setup/test_setup.sh  100755 → 100644 `

`...path_relative_to_package_root/test_exe.sh → ...path_relative_to_package_root/test_exe.sh  100755 → 100644`

`...ackage_layout/package_root/test_called.sh → ...ackage_layout/package_root/test_called.sh  100755 → 100644`

`packaging/debian/package_tool/test.sh → tools/DebianPackageTool/test.sh  100755 → 100644`

eerhardt:
:(  I'll fix those as well.  Thanks for the catch.

brthor:
If only the file permissions were x-plat :smile:

eerhardt:
OK - I fixed all the file permissions.
Since #1454 is merged now, we should get failures in Ubuntu if this move broke anything.

brthor:
:shipit:

Issue #1250, opened 2016-02-05
Refactor test layout


Sridhar-MS:

Sridhar-MS:
The layout of test resources for CLI will be moved to this layout:

```
/%REPOROOT%
|--- /src
        |---/TestFramework
|--- /assets
        |--- /TestProjects
        |--- /TestPackages
|--- /test
        |--- /{commandName}.Tests
        |--- /{libraryName}.Tests
        |--- /Installer
                 |--- /Deb
                 |--- /MSI         
        |--- /EndToEnd
```

blackdwarf:
@Sridhar-MS @eerhardt is this work done? Can this issue be closed?

eerhardt:
The following tests still need to be renamed/moved as appropriate:

- ArgumentForwardingTests
- ArgumentsReflector
- ScriptExecutorTests

But I think that's the only thing left.

davidfowl:
There's also this `https://github.com/dotnet/cli/tree/rel/1.0.0/testapp` that needs to be deleted /cc @troydai

troydai:
It can be moved, but there are projects under `testapp` will fail the `dotnet restore` intentionally. Make sure it doesn't break build script.

PR #1328, opened 2016-02-09
Refactoring test directories.

eerhardt:
- Rename Compiler.Common.Tests to Microsoft.DotNet.Compiler.Common.Tests.
- Create the Microsoft.DotNet.Cli.Utils.Tests and move StreamForwarderTests to it.
    - ArgumentForwardingTests will be moved here in the future
- Remove LoadContextTest, since it is old and no longer relevant.

Partial #1250 

@livarcocc @anurse


piotrpMSFT:
:shipit:

PR #1332, opened 2016-02-09
NuGet 607

piotrpMSFT:
@Sridhar-MS @emgarten

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/project.json
+++ b/src/Microsoft.DotNet.ProjectModel/project.json
@@ -7,7 +7,7 @@
     "dependencies": {
     "System.Reflection.Metadata": "1.2.0-rc2-23811",
 
-        "NuGet.Packaging": "3.4.0-beta-583",
+        "NuGet.Packaging": "3.4.0-beta-618",
         "Microsoft.Extensions.FileSystemGlobbing": "1.0.0-rc2-15996",
         "Microsoft.Extensions.JsonParser.Sources": {
             "type": "build",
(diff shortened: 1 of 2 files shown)


emgarten:
:+1:

piotrpMSFT:
@dotnet-bot test this please

emgarten:
@piotrpMSFT  The failure here looks like an issue with the project model. NuGet 607 does not flow build only dependencies to parent projects, publish should not either.

davidfowl:
Why is this a build dependency... https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/project.json#L27 wtf

emgarten:
Looking at this further I see that NuGet is leaving build type dependencies in as dependency entries on the parent target entries. This ends up looking the same as an unresolved package to the project model.

I have a fix here: https://github.com/NuGet/NuGet.Client/pull/269

piotrpMSFT:
@davidfowl just changed the dependency

emgarten:
The dependency fix isn't in NuGet 612 yet, I'll provide the number once this finishes building.

piotrpMSFT:
Thanks @emgarten. I just updated my PR to drop the latent "build" dependencies as well.

emgarten:
@piotrpMSFT NuGet 613 contains the fix for removing build type dependencies.

SajayAntony:
This pretty much blocks our cross plat pert runs. At this point .

piotrpMSFT:
rebasing now.

piotrpMSFT:
we're seeing a bunch of version hoisting... I'm investigating if it repros on 583...

davidfowl:
I'll take over this change. I have another branch that I'm working on

Issue #255, opened 2015-11-19
dotnet compile --native should pass app dependencies to ILC.exe


brthor:
This functionality was taken out because the CoreRT framework libraries are currently only included in the AppDepSDK Nuget package. 

We need to think about where the framework libraries will eventually come from, and how we deal with third party app dependencies.

/cc @gkhanna79 @schellap @livarcocc

gkhanna79:
@brthor - To confirm, are you saying that "-r" is only passed for System.Private.Corelib.dll and AppDepSDK?

schellap:
@gkhanna79, we need more libraries in AppDepSDK like EntityFramework. If there are third party libraries, how is ILC supposed to work? Will we do -r ThirdPartyApp.dll or our goal is to have a native image for each IL binary?

gkhanna79:
Longterm, acquisition of 3rd party libraries would be part of restoring the app which would declare dependency on such libraries. At that point in time, dotnet compile native should be passing -r for all libraries constituting the app to ilc.

schellap:
--ilcargs "-r reference1.dll -r reference2.dll" can be passed in such a case as of now.

davidfowl:
this includes packaging for aot things. Today projects only target dnxcore50 but the aot things are netcore50 and this needs to be sorted

brthor:
@gkhanna yes currently only appdepsdk/*.dll and System.Private.CoreLib are passed as references

gkhanna79:
@davidfowl - I agree. Who is driving that? @richlander may have some context on who maybe driving the transition to netcore50.

piotrpMSFT:
@brthor can you start linking issues for the pre-requisites of this issue?

brthor:
@piotrpMSFT Child issues have been created. 

#402 
#404 
#405 
#406 

@schellap @gkhanna79 
Please review these issues to ensure we are still on the same page. Feel free to offer any corrections.

brthor:
@schellap You have been working on this. Want to take this issue?

blackdwarf:
@schellap do we want to move this over to dotnet/corert?

gkhanna79:
@nattress Can you please create an issue to reference this and close this one?

piotrpMSFT:
@nattress ping

nattress:
See https://github.com/dotnet/corert/issues/1584

Issue #402, opened 2015-12-03
Standardize on a TFM for aot and non-aot Framework Libraries


brthor:
Currently aot libraries exist under the TfM `netcore50`, whilst non-aot libraries exist under the TfM `dnxcore50`

The aot libraries are needed to compile apps with CoreRT. Currently we include a hardcoded list of aot libraries that are compiled into every app when doing native compilation. This is blocking scenarios beyond Hello World.

To move forward with proper framework dependency compilation for `dotnet compile --native` we should consolidate these TfMs. This will reduce the complexity of acquiring these aot libraries, and improve the user experience in doing so.

This will unblock #255 

/cc @piotrpMSFT @weshaggard @gkhanna79

weshaggard:
We need a TFM for for dotnet-cli/netstandadard or what ever we end up with and then we can use that in combination with the aot RID to uniquely select assets for the --native option.

brthor:
Thanks @weshaggard 

Do you know the status of this TFM transition or who I could follow up with?

Petermarcu:
@ericstj who is working with the NuGet team to get support added for the netstandard tfm's. Once NuGet supports them we can move a bunch of the rest of the system to it. For now, I think we can use the dotnet?? tfm's plus the aot RID to get these assets.

ericstj:
Nuget work is being tracked in https://github.com/NuGet/Home/issues/1691. @davidfowl enumerated the work here: https://github.com/dotnet/corefx/issues/4367

Until that is done you can use "dotnetX.Y".  We'll need to do a mop up when the name change happens, but I expect NuGet to support both for some time to aide in transition.

blackdwarf:
Since the TFMs have changed significantly as we approached RC2 and seeing as how CoreRT bits are not shipped with the CLI anymore, I'm closing this. Reopen if still needed.

PR #1363, opened 2016-02-11
Include project.json dependencies and use response files

schellap:
WIP because I am still debugging why I am not able to pass: `--ilcarg --verbose` or `--ilcarg -r:c:\foo.dll` to the ILC.

@gkhanna79 @brthor PTAL.

* Depends on new ILC (corresponding changes, mainly using `System.CommandLine` will be sent for PR shortly)
* Instead of a single gigantic `ilcargs`, we will be able to chain multiple single `--ilcarg` and add them to the response file.

Cc @tijoytom @piotrpMSFT 

Spent a bunch of time debugging why EndToEnd failed but apparently because `DOTNET_HOME` was set to some old CLI install.

### DIFF
--- a/src/dotnet/commands/dotnet-build/CompileContext.cs
+++ b/src/dotnet/commands/dotnet-build/CompileContext.cs
@@ -339,10 +339,10 @@ namespace Microsoft.DotNet.Tools.Build
                 args.Add(_args.ArchValue);
             }
 
-            if (!string.IsNullOrWhiteSpace(_args.IlcArgsValue))
+            foreach (var ilcArg in _args.IlcArgsValue)
             {
-                args.Add("--ilcargs");
-                args.Add(_args.IlcArgsValue);
+                args.Add("--ilcarg");
+                args.Add(ilcArg);
             }
 
             if (!string.IsNullOrWhiteSpace(_args.IlcPathValue))
--- a/src/dotnet/commands/dotnet-compile-native/ILCompilerInvoker.cs
+++ b/src/dotnet/commands/dotnet-compile-native/ILCompilerInvoker.cs
@@ -1,13 +1,15 @@
 using System.Collections.Generic;
 using System.IO;
+using System.Text;
 using Microsoft.DotNet.Cli.Utils;
 
 namespace Microsoft.DotNet.Tools.Compiler.Native
 {
     public class ILCompilerInvoker
     {
-        private readonly string ExecutableName = "corerun" + Constants.ExeSuffix;
-        private readonly string ILCompiler = "ilc.exe";
+        private static readonly string HostExeName = "corerun" + Constants.ExeSuffix;
+        private static readonly string ILCompiler = "ilc.exe";
+
         private IEnumerable<string> Args;
         private NativeCompileSettings config;
 
@@ -27,14 +29,6 @@ namespace Microsoft.DotNet.Tools.Compiler.Native
         {
             var argsList = new List<string>();
 
-            var managedPath = Path.Combine(config.IlcPath, ILCompiler);
-            if (!File.Exists(managedPath))
-            {
-                throw new FileNotFoundException("Unable to find ILCompiler at " + managedPath);
-            }
-
-            argsList.Add($"{managedPath}");
-            
             // Input File 
             var inputFilePath = config.InputManagedAssemblyPath;
             argsList.Add($"{inputFilePath}");
@@ -43,32 +37,29 @@ namespace Microsoft.DotNet.Tools.Compiler.Native
             var coreLibsPath = Path.Combine(config.IlcSdkPath, "sdk");
             foreach (var reference in Directory.EnumerateFiles(coreLibsPath, "*.dll"))
             {
-                argsList.Add($"-r");
-                argsList.Add($"{reference}");
+                argsList.Add($"-r:{reference}");
             }
             
             // AppDep References
             foreach (var reference in config.ReferencePaths)
             {
-                argsList.Add($"-r");
-                argsList.Add($"{reference}");
+                argsList.Add($"-r:{reference}");
             }
             
             // Set Output DetermineOutFile
             var outFile = DetermineOutputFile(config);
-            argsList.Add($"-out");
-            argsList.Add($"{outFile}");
+            argsList.Add($"-o:{outFile}");
             
             // Add Mode Flag TODO
             if (config.NativeMode == NativeIntermediateMode.cpp)
             {
-                argsList.Add("-cpp");
+                argsList.Add("--cpp");
             }
             
             // Custom Ilc Args support
-            if (! string.IsNullOrEmpty(config.IlcArgs))
+            foreach (var ilcArg in config.IlcArgs)
             {
-                argsList.Add(config.IlcArgs);
+                argsList.Add(ilcArg);
             }
                         
             Args = argsList;
@@ -76,9 +67,20 @@ namespace Microsoft.DotNet.Tools.Compiler.Native
 
         public int Invoke()
         {
-            var executablePath = Path.Combine(config.IlcPath, ExecutableName);
-            
-            var result = Command.Create(executablePath, Args)
+            // Check if ILCompiler is present
+            var ilcExePath = Path.Combine(config.IlcPath, ILCompiler);
+            if (!File.Exists(ilcExePath))
+            {
+                throw new FileNotFoundException("Unable to find ILCompiler at " + ilcExePath);
+            }
+
+            // Write the response file
+            var intermediateDirectory = config.IntermediateDirectory;
+            var rsp = Path.Combine(intermediateDirectory, "dotnet-compile-native-ilc.rsp");
+            File.WriteAllLines(rsp, Args, Encoding.UTF8);
+
+            var hostPath = Path.Combine(config.IlcPath, HostExeName);
+            var result = Command.Create(hostPath, new string[] { ilcExePath, "@" + $"{rsp}" })
                 .ForwardStdErr()
                 .ForwardStdOut()
                 .Execute();
--- a/src/dotnet/commands/dotnet-compile-native/IntermediateCompilation/Linux/LinuxCppCompileStep.cs
+++ b/src/dotnet/commands/dotnet-compile-native/IntermediateCompilation/Linux/LinuxCppCompileStep.cs
@@ -25,7 +25,7 @@ namespace Microsoft.DotNet.Tools.Compiler.Native
 
         private readonly string[] _appdeplibs = 
             {
-                "libSystem.Native.a"
+                "System.Native.a"
             };
 
         public LinuxCppCompileStep(NativeCompileSettings config)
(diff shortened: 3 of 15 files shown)


NuGet.Config
```diff
@@ -4,6 +4,7 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="dotnet-corert" value="https://dotnet.myget.org/F/dotnet-corert/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
```
schellap: @piotrpMSFT added new feed, opinions? or you want me to push into your feeds?

src/dotnet/commands/dotnet-compile/Program.cs
```diff
@@ -71,16 +71,35 @@ private static bool OnExecute(List<ProjectContext> contexts, CompilerCommandApp
 
             var managedOutput = outputPaths.CompilationFiles.Assembly;
 
+            // Create the library exporter
+            var exporter = context.CreateExporter(args.ConfigValue);
```
davidfowl: How does this even work? Don't you need to pass in the runtime assemblies to the AOT compiler? This will get you the wrong set of assemblies. You need the project context for the AOT runtime ID, which isn't in the projects by default. Tons of work to do here.

schellap: Yes. We need to in the long term. But we wanted to have simple cases work without blocking on the tons of work. We got rid of System.Console from AppDep and it now comes from the project file. System.IO.FileSystem comes from the project file. When we have the AOT rid, we'll look for it -- tracked by #402. By design and what you say is future work. The AOT libraries will override just like AppDep overrides in the PR.

https://github.com/dotnet/cli/issues/402

davidfowl: Don't understand how this works. Don't you need to at least use the correct runtime id? Do the core clr implementations and CoreRT implementations line up?

schellap: Do you mean the rid? It comes from `compile` option for both cross and native compilation. If unspecified, defaults to platform rid. I see error handling missing here for other specified rids unknown to platform ILC.

Otherwise, coreclr and corert target dotnet core and should be compatible for the most part except few assemblies which we will acquire through the aot rid or put in the appdep which gets the first preference.

src/dotnet/commands/dotnet-build/CompileContext.cs
```diff
@@ -328,10 +328,10 @@ private bool InvokeCompileOnRootProject()
                 args.Add(_args.ArchValue);
             }
 
-            if (!string.IsNullOrWhiteSpace(_args.IlcArgsValue))
+            foreach (var ilcArg in _args.IlcArgsValue)
             {
-                args.Add("--ilcargs");
-                args.Add(_args.IlcArgsValue);
+                args.Add("--ilcarg");
```
gkhanna79: Why do we need to make this split?

src/dotnet/commands/dotnet-compile-native/ILCompilerInvoker.cs
```diff
@@ -43,32 +37,29 @@ private void InitializeArgs(NativeCompileSettings config)
...
             var outFile = DetermineOutputFile(config);
-            argsList.Add($"-out");
-            argsList.Add($"{outFile}");
+            argsList.Add($"-o:{outFile}");
             
             // Add Mode Flag TODO
             if (config.NativeMode == NativeIntermediateMode.cpp)
             {
-                argsList.Add("-cpp");
+                argsList.Add("--cpp");
```
gkhanna79: This seems to be out of sync with the other args being added with a single "-". I believe we want "--" to be exposed at a public level (e.g. where user interacts with dotnet command) but for ILC, "-cpp" is fine.

src/dotnet/commands/dotnet-compile-native/IntermediateCompilation/Linux/LinuxCppCompileStep.cs
```diff
@@ -25,7 +25,7 @@ public class LinuxCppCompileStep : IPlatformNativeStep
 
         private readonly string[] _appdeplibs = 
             {
-                "libSystem.Native.a"
+                "System.Native.a"
```
gkhanna79: Has the library name changed?

brthor:
This fixes #255

NuGet.Config
brthor: We don't want to take any new feeds in our repo right now. 

Should work with @eerhardt to get these dependencies into our deps feed for the time being.

src/dotnet/commands/dotnet-compile-native/ILCompilerInvoker.cs
```diff
@@ -43,32 +37,29 @@ private void InitializeArgs(NativeCompileSettings config)
             var coreLibsPath = Path.Combine(config.IlcSdkPath, "sdk");
             foreach (var reference in Directory.EnumerateFiles(coreLibsPath, "*.dll"))
             {
-                argsList.Add($"-r");
-                argsList.Add($"{reference}");
+                argsList.Add($"-r:{reference}");
```
brthor: Want to point out this will cause this arg to be quoted and escaped as a single arg. 

So for a space with a path, it will end up like this: `"-r:/some path/with/a/space"

Is this intentional?

brthor: Ah I see that you are dropping a response file. You may need to manually quote the piece after `-r:`
since this won't go through the argument escaper.

We do this in CSC. Recommend to write some test cases around spaces with paths here.

src/dotnet/commands/dotnet-compile-native/NativeCompileSettings.cs
```diff
@@ -168,8 +168,11 @@ private NativeCompileSettings()
             BuildType = DefaultBuiltType;
             NativeMode = DefaultNativeModel;
             AppDepSDKPath = Path.Combine(AppContext.BaseDirectory, "appdepsdk");            
-
-            _referencePaths = new List<string>(Directory.EnumerateFiles(AppDepSDKPath, "*.dll"));
+            _referencePaths = new Dictionary<string, string>();
+            foreach (var file in Directory.EnumerateFiles(AppDepSDKPath, "*.dll"))
+            {
+                _referencePaths.Add(Path.GetFileName(file), file);
```
gkhanna79: Is this being done to address a symlink issue? If not, then why do we need this since EnumerateFiles is expected to return an absolute path?

src/dotnet/commands/dotnet-compile/Program.cs
```diff
@@ -71,16 +71,35 @@ private static bool OnExecute(List<ProjectContext> contexts, CompilerCommandApp
...
+
+            // Gather exports for the project
+            var dependencies = exporter.GetDependencies().ToList();
+
+            var references = new List<string>();
+            foreach (var dependency in dependencies)
```
tijoytom: Can you please move all the code get runtime references to a separate function so I can re-use it?

NuGet.Config
eerhardt: The 1.0.6 versions are now in our feed.  You shouldn't need this change.

src/dotnet/commands/dotnet-build/CompileContext.cs
schellap: We write response files now and in the file, you have to have arguments in this format:

```
-option1:value1
-option2:value2
```

With a opaque string like `--ilcargs`, we would end up with in the response file.
```
-r:c:\foo.dll -r:c:\bar.dll -r:c:\baz.dll
```

unless you write a parser that resolves these into per ilc arg in the above format and separates them with a newline.

With `--ilcarg` when you chain each arg, you end up with:

```
-r:c:\foo.dll
-r:c:\bar.dll
-r:c:\baz.dll
```

in the response file. @terrajobst any other suggestions, we pass in `--ilcargs` as an opaque string through a response file to ILC from `dotnet compile-native`

src/dotnet/commands/dotnet-compile-native/ILCompilerInvoker.cs
schellap: We should start using `System.CommandLine` syntax rules for ILC. I know this means we have to update all docs...

src/dotnet/commands/dotnet-compile-native/IntermediateCompilation/Linux/LinuxCppCompileStep.cs
schellap: Yes. It has changed to System.Native.a. We will soon consume that as well from the dependencies (through NETStandard.Library) I don't think it is wired up that way right now.

Cc @tarekgh

src/dotnet/commands/dotnet-compile-native/NativeCompileSettings.cs
schellap: This is a de-duplicator for AppDep assemblies vs. added references. AppDep needs to get higher priority.

schellap:
CI failing due to: https://github.com/dotnet/corefx/issues/6092

src/dotnet/commands/dotnet-compile/NativeCompiler.cs
```diff
@@ -22,16 +23,37 @@ public override bool Compile(ProjectContext context, CompilerCommandApp args)
 
             var managedOutput = outputPaths.CompilationFiles.Assembly;
 
+            // Create the library exporter
+            var exporter = context.CreateExporter(args.ConfigValue);
```
davidfowl: AFAIK this still doesn't work unless I'm missing something fundamental about the CoreRT implementations.

schellap: @davidfowl, what do you think would work eventually in your mental model?

schellap: Cc @anurse

schellap: The missing piece is the look up of `aot-win7-x64` (for example) assemblies and wiring it through to ILC skipping the ones that are in this list too.

davidfowl: This just wont work by default until we figure out how to restore for the RID.

schellap: @davidfowl, yeah the user would put this in the project.json runtimes, if native is needed. See my comment earlier, we don't want to block on this work for now, thats why we have the AppDep package.

But this is a good time to get this conversation started. See #402.

davidfowl: So a few questions:
- What happens when the wrong implementations get passed into the compiler? Are you looking specifically for an aot-* RID?
- How are you getting that RID at compile time? Are you defaulting to the current aot-{RID}?

schellap:
The full fledged aot lookup needs more discussion and work across nuget, corefx and cli. The change here is just a few lines in getting the runtime assemblies to the ILC, but it unblocks several tests and MCG for corert. So I am merging... @davidfowl expressed concern over why just `compile` is taking rid specific assemblies. This should be fixed separately.

PR #1375, opened 2016-02-11
Bindings redirects fix

### LINKED ISSUE #865: Error on dotnet run

marono:

Fix #865


marono:
How do I prevent my test from running on non-Windows platforms? Added WindowsOnlyFactAttribute, but doesn't seem to do the trick.

piotrpMSFT:
@marono this is blocked on both CI and CLA.

marono:
@piotrpMSFT how do I prevent my tests from running on OSX/Linux? Because binding redirects are only generated for 'Desktop' runtimes, all my test projects are targeted for net461.

Not sure why the Debug build on windows fails, I'll investigate it as soon as the CI server comes back up.

marono:
@piotrpMSFT, windows debug build failed because of a timeout. Don't think that's related to my changes, could you trigger it again please?

src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
```diff
@@ -109,26 +109,24 @@ private static XElement GetOrAddElement(XContainer parent, XName elementName)
 
         private static AssemblyRedirect[] CollectRedirects(IEnumerable<LibraryExport> dependencies)
         {
-            var allRuntimeAssemblies = dependencies.SelectMany(d => d.RuntimeAssemblies).Select(GetAssemblyInfo).ToArray();
-            var assemblyLookup = allRuntimeAssemblies.ToDictionary(r => r.Identity.ToLookupKey());
+            var dependenciesList = dependencies.ToList();
+            var allRuntimeAssemblies = dependenciesList.SelectMany(d => d.RuntimeAssemblies).Select(GetAssemblyInfo).ToArray();
+            var assemblyLookup = GetListOfUniqueRefs(dependenciesList).Select(GetAssemblyInfo).ToDictionary(r => r.Identity.ToLookupKey());
```
pakrym: You are calling GetAssemblyInfo on some assemblies twice, I'd prefer not doing this.

marono:
Now I don't understand it, yesterday Windows Release was green, but not anymore. What's changed?

marono:
Guys,

This is ready for merging, I've just signed my CLA, sorry for the delay, it's the first time my employer does this so they had to check the agreement.

I've also amended the code to cache assembly info during scanning.

For some reason, builds on non-Windows machines are not triggered, maybe you know why that is?

marono:
I've been asked to resubmit the PR as there was an issue with my CLA form, I'll close this one for now.

PR #1382, opened 2016-02-12
Use ApplicationBasePath instead of entry assembly to locate reference assemblies.

pranavkm:


pranavkm:
cc @pakrym \ @davidfowl

davidfowl:
:shipit:

GuardRex:
Side ?: Can I use ...
```
PlatformServices.Default.Application.ApplicationBasePath
```
... with ...
```
.UseApplicationBasePath()
```
... to get the .exe path when launching app under IIS from `wwwroot`?

pranavkm:
@GuardRex, that should work. Are you seeing otherwise?

davidfowl:
You don't need to set the application base to the exe location. That's the default

GuardRex:
Haven't tried it yet. Will give it a shot when I get back to the office. Thx.

GuardRex:
Oh... didn't see last msg. Will try removing path I supply manually and see if it works now (didn't work a few weeks back).

moozzyk:
@GuardRex - https://github.com/aspnet/IISIntegration/blob/dev/src/dotnet-publish-iis/WebConfigTransform.cs#L85:

```
<configuration>
  <system.webServer>
    <handlers>
      <add name="httpPlatformHandler" path="*" verb="*" modules="httpPlatformHandler" resourceType="Unspecified" />
    </handlers>
    <httpPlatform processPath="..\HelloMvc.exe" stdoutLogEnabled="false" stdoutLogFile="..\logs\stdout.log" startupTimeLimit="3600">
      <environmentVariables>
        <environmentVariable name="ASPNET_APPLICATIONBASE" value="." />
      </environmentVariables>
    </httpPlatform>
  </system.webServer>
</configuration>
```

GuardRex:
No thanks! That's exactly the crap I'm trying to avoid. The app should know where it lives IMHO.

moozzyk:
@GuardRex - this is showing exactly that - the `.` is resolved relative to the folder the application lives in.

GuardRex:
@pranavkm @davidfowl @moozzyk 
```c#

var host = new WebHostBuilder()
    .UseServer("Microsoft.AspNetCore.Server.Kestrel")
    .UseApplicationBasePath(PlatformServices.Default.Application.ApplicationBasePath)
```
... works perfectly everywhere. thx. :beers:

I recommend you update https://github.com/aspnet/cli-samples/blob/master/HelloMvc/Program.cs to use this approach, as `Directory.GetCurrentDirectory()` will resolve to the `wwwroot` folder under IIS and :boom:. As for an env var in `web.config` ... well ... that's just plain disgusting.

moozzyk:
@GuardRex - I thought we have already [covered it](https://github.com/aspnet/IISIntegration/pull/80#issuecomment-179524246). Setting appBase to current directory allows loading views during development without having to copy the Views folder since the exe goes to bin\Debug\dnxcore50 and Views stay in the project root, no?

GuardRex:
@moozzyk Yeah ... rough day here. Bad news all day, and I'm getting jerked around by GoDaddy (yuck!) mail servers right now. But can't the ...
```
.UseApplicationBasePath(PlatformServices.Default.Application.ApplicationBasePath)
```
... solve both problems?

Locally, it's the right path. Under IIS, it's the right path. It seems like it's the ultimate-insanely-cool-super-perfect-it-works-crazy-good-everywhere solution. Or ... what is the :monkey: not getting? I must be missing something ... about to open my fourth :beer:. :smile:

moozzyk:
@GuardRex what will be the appbase if you run `dotnet run` from the project folder?

GuardRex:
@moozzyk I don't plan on doing that. Are you saying that's the problem with doing it the `PlatformServices`-way?

moozzyk:
If you don't plan on running your app with `dotnet run` (or potentially with VS - I am not sure at the moment how VS is going to run the app) then it may be fine. I actually think you could just do `.UseApplicationBasePath(".")` since the `"."` will be resolved related to .exe. I don't think however we should change the samples. 
(As a side note - if you use the publish-iis tool and run the app with IIS/httpPlatformHandler it does not really matter what you set in the code since the tool will update your web.config to set the env variable which wins)

GuardRex:
@moozzyk  I see. Good to know. I like the "." idea. That's cool. However, I'm going to go with `PlatformServices` right now ... sounds good ... feels good. After four cheap :beer:'s, I'm feel'in that right now. As for the publish-iis tool, that's just fluff. I don't see value there.

> I don't think however we should change the samples. 

Fine. Your samples. Just wait until people put that on an IIS server. They won't be too happy with the outcome ... I wasn't.

GuardRex:
@moozzyk Forgive me for yesterday ... rough day at the office. The CEO **and** the company President stomping on my last nerve.

I just ran a test on `dotnet run` with ...
```
.UseApplicationBasePath(PlatformServices.Default.Application.ApplicationBasePath)
```
... and it works ... of course, as you pointed out with the caveat of Views (and other assets) location relative to the executable in https://github.com/aspnet/IISIntegration/pull/80#issuecomment-179524246. It only works because I'm publishing and running the published output. I hear ya on that. I'm good now. Thx again for your help.

davidfowl:
You don't need to do that because it's the default.https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L195

PR #1392, opened 2016-02-12
Make Portable PDB the default on Windows

### LINKED ISSUE #704: CLI should produce portable PDBs on Windows
Currently the .NET CLI invokes Roslyn on non-Windows platforms so that portable PDBs are produced, but on Windows, standard PDBs are still used. After discussing this with the Roslyn team, they would like the default to be that portable PDBs are produced everywhere, but there is an option to produce Windows PDBs on Windows.

Note that the support for portable PDBs in VS isn't in yet, so I would recommend that this bug be fixed **after** the beta release, so that there is time for VS to come out with a release that supports this.

jaredpar:
VS 2015 Update 2 CTP has been released which has support for debugging portable PDBs.  Now is a good time to make the transition so a) we can root out any remaining issues with portable PDBs and b) have larger dogfooding of the VS debugging experience here.

closes #704

### DIFF
--- a/src/Microsoft.DotNet.Compiler.Common/CommonCompilerOptionsExtensions.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/CommonCompilerOptionsExtensions.cs
@@ -32,6 +32,8 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         internal static readonly OptionTemplate s_publicSignTemplate = new OptionTemplate("public-sign");
 
+        internal static readonly OptionTemplate s_debugTypeTemplate = new OptionTemplate("debug-type");
+
         internal static readonly OptionTemplate s_emitEntryPointTemplate = new OptionTemplate("emit-entry-point");
 
         internal static readonly OptionTemplate s_generateXmlDocumentation = new OptionTemplate("generate-xml-documentation");
@@ -44,6 +46,7 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             IReadOnlyList<string> suppressWarnings = null;
             string languageVersion = null;
             string platform = null;
+            string debugType = null;
             bool? allowUnsafe = null;
             bool? warningsAsErrors = null;
             bool? optimize = null;
@@ -61,6 +64,8 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             syntax.DefineOptionList(s_suppressWarningTemplate.LongName, ref suppressWarnings, "Suppresses the specified warning");
 
             syntax.DefineOptionList(s_additionalArgumentsTemplate.LongName, ref additionalArguments, "Pass the additional argument directly to the compiler");
+
+            syntax.DefineOption(s_debugTypeTemplate, ref debugType, "The type of PDB to emit: portable or full");
 
             syntax.DefineOption(s_languageVersionTemplate.LongName, ref languageVersion,
                     "The version of the language used to compile");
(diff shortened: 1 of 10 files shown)


davidfowl:
I think we need a flag to control this as part of this change.

jaredpar:
@davidfowl this can be controlled on an individual project level by just adding an explicit setting.  Are you thinking about a more global control?  If so what's the general pattern for dotnet to implement such an override?

Petermarcu:
Is there logic somewhere else to keep it using Windows pdb's for portable targets and .NET Framework targets? I think we also talked about warning when we use a portable pdb for a .NET Framework or portable target on Linux/Mac.

piotrpMSFT:
How do we test this? We need a way to determine if a pdb is portable.

gregg-miskelly:
@piotrpMSFT I have a trivial command line app I can send you. You can just check the first four bytes.

piotrpMSFT:
Thanks @gregg-miskelly!

Petermarcu:
@jaredpar , Making sure we saw this. The design we discussed was:

|Target|OS|Result|
|------|---|------|
|Full .NET Or Portable Library|Windows|Windows Pdb|
|Full .NET Or Portable Library|Non-Windows|Portable Pdb with warning|
|.NET Standard App or Library|Everywhere|Portable Pdb|

davidfowl:
> @davidfowl this can be controlled on an individual project level by just adding an explicit setting. Are you thinking about a more global control? If so what's the general pattern for dotnet to implement such an override?

A compilation option.

jaredpar:
@davidfowl such a compilation option already exists.  I'm unsure what you want here.

davidfowl:
Ah I didn't see it, how do I set it in project.json

jaredpar:
@Petermarcu I don't understand why we want to issue a warning for portable PDB anywhere.  This is the path forward for PDBs in .Net.  A warning is only going to serve to push users away from portable PDBs vs. getting them to embrace them.  I think we should change the design to remove the warning.  

CC @tmat

jaredpar:
@davidfowl ah okay.  I assumed it was already specified in the project.json entry.  I'll check and add it in if missing.

tmat:
@jaredpar Re warning: it's only for "Full .NET Or Portable Library". As a library (package) author you won't be able to deliver the same debugging/tooling experience for your customers on Full .NET FX. Hence the warning informing you of that. We need to set the expectations right. It would be a mistake imo to push Portable PDB to hard atm, since people would be disappointed with debugging experience vs. on Windows (relative to Windows PDBs).

davidfowl:
@dotnet-bot test Windows Build please

davidfowl:
@dotnet-bot test Windows_NT Build please

davidfowl:
@dotnet-bot test Windows_NT Debug Build please

davidfowl:
@dotnet-bot test Windows_NT Release Build please

piotrpMSFT:
@jaredpar can you please include tests using the tool @gregg-miskelly offered to build as part of this PR? The matrix is going to be easy to break if not tested.

jaredpar:
Latest commits add a new entry to project.json that allows tweaking between full and portable PDBs:

``` json
    "compilationOptions": {
        "debugType": "portable"
    },
```

The value `"full"` can be used to specify Windows PDBs.  I'm awaiting @gregg-miskelly tool to verify that it's actually a portable PDB.  But visually looking at the bytes, and the difference between the two settings, it's producing the correct PDB for the setting.  

Will squash commits post review before committing.

PR #1406, opened 2016-02-15
Fixed the build

davidfowl:
- csc kept crashing on stage 2 and it turned out to be a version mismatch with System.Reflection.Metadata.
- Silenced crossgen
- Made omnisharp see all of the projects

### DIFF
--- a/src/dotnet/project.json
+++ b/src/dotnet/project.json
@@ -54,7 +54,7 @@
         "Microsoft.NETCore.ConsoleHost": "1.0.0-rc2-23811",
         "Microsoft.NETCore.TestHost": "1.0.0-rc2-23811",
         "NETStandard.Library": "1.0.0-rc2-23811",
-        "System.Reflection.Metadata": "1.2.0-rc3-23811",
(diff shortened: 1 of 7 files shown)


davidfowl:
@dotnet-bot test this please

src/dotnet/project.json
```diff
@@ -54,7 +54,7 @@
         "Microsoft.NETCore.ConsoleHost": "1.0.0-rc2-23811",
         "Microsoft.NETCore.TestHost": "1.0.0-rc2-23811",
         "NETStandard.Library": "1.0.0-rc2-23811",
-        "System.Reflection.Metadata": "1.2.0-rc3-23811",
+        "System.Reflection.Metadata": "1.3.0-beta-23811",
```
Petermarcu: Why is dotnet on a different version of S.R.M than project model?

davidfowl: It doesn't need to be. You can use project model stand alone so it doesn't matter. This was legit breaking Roslyn in dotnet, that's why I updated it

PR #1411, opened 2016-02-16
Updated to a newer version of NuGet

davidfowl:
- Update the newest nuget
- Filter out build dependencies from the project resolver based on hte lock file
- Fix more erroneous build dependencies

/cc @emgarten @piotrpMSFT


emgarten:
Looks good to me :shipit:

PR #1416, opened 2016-02-16
Improved VSO build logging during VS Environment Variable collection.

piotrpMSFT:


scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -228,13 +230,20 @@ public static BuildTargetResult ValidateDependencies(BuildTargetContext c)
                     File.Delete(temp);
                 }
             }
+            
             result.EnsureSuccessful();
+            
             var vars = new Dictionary<string, string>();
             foreach (var line in result.StdOut.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries))
             {
+                c.Info($"Adding variable '{line}'");
```
davidfowl: verbose?

davidfowl: set will print out ALL of the environment variables set on the machine.

PR #1417, opened 2016-02-16
This change mimics the nuget lock file generation logic

davidfowl:
We should eventually move all of these intrinsics to NuGet and just use them as is.

/cc @emgarten


emgarten:
:shipit:

Issue #1422, opened 2016-02-16
Error: No HOME path available


piotrpMSFT:
When attempting to upload build artifacts to Azure during VSO builds we see `Error: No HOME path available` on Windows. This is because the Azure x-plat CLI is not meant to be used within PowerShell. PowerShell provides first-class Azure CmdLets which are recommended. 

Given that our scripts are currently built around the CLI I'm going to try to work around this issue by setting $env:HOME on Windows CI builds. If this proves to be a rats nest, however, we should consider moving to PS CmdLets.

blackdwarf:
@piotrpMSFT is this solved? Should we close?

PR #1424, opened 2016-02-16
Fix publish scripts

piotrpMSFT:
Addreses issue #1422 

I'm working on a Mac, so no PS to test with...


davidfowl:
Should convert this to C#

piotrpMSFT:
agreed. for the time being, I want to get builds working.

PR #1428, opened 2016-02-16
Use runtime context only where we require it

pakrym:
@davidfowl

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
@@ -83,7 +83,7 @@ namespace Microsoft.DotNet.ProjectModel
                         .WithRuntimeIdentifiers(runtimeIdentifiers)
                         .Build();
         }
-
+        
         /// <summary>
         /// Creates a project context for each framework located in the project at <paramref name="projectPath"/>
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -85,6 +85,18 @@ public static ProjectContext Create(string projectPath, NuGetFramework framework
         }
 
         /// <summary>
+        /// Creates a project context based on existing context but using runtime target
+        /// </summary>
+        /// <param name="context"></param>
+        /// <param name="runtimeIdentifiers"></param>
+        /// <returns></returns>
+
+        public static ProjectContext CreateRuntimeContext(ProjectContext context, IEnumerable<string> runtimeIdentifiers)
```
davidfowl: This should be an instance method.

src/dotnet/commands/dotnet-build/CompileContext.cs
```diff
@@ -310,6 +316,12 @@ private bool InvokeCompileOnRootProject()
             args.Add("--configuration");
             args.Add(_args.ConfigValue);
 
+            if (!string.IsNullOrWhiteSpace(_args.RuntimeValue))
```
davidfowl: Remove this and break the script :smile:

src/dotnet/commands/dotnet-build/CompileContext.cs
```diff
@@ -290,6 +290,12 @@ private bool InvokeCompileOnDependency(ProjectDescription projectDependency)
             args.Add(_args.ConfigValue);
             args.Add(projectDependency.Project.ProjectDirectory);
 
+            if (!string.IsNullOrWhiteSpace(_args.RuntimeValue))
```
davidfowl: Remove it.

pakrym:
@dotnet-bot test this please

PR #1441, opened 2016-02-16
forward packaged commands test outputs to console

anurse:
This should help diagnose failures in these tests.

/cc @pakrym @davidfowl @piotrpMSFT

### DIFF
--- a/scripts/dotnet-cli-build/TestTargets.cs
+++ b/scripts/dotnet-cli-build/TestTargets.cs
@@ -163,7 +163,12 @@ namespace Microsoft.DotNet.Cli.Build
             // Test the apps
             foreach (var dir in Directory.EnumerateDirectories(consumers))
             {
-                var result = dotnet.Exec("hello").WorkingDirectory(dir).CaptureStdOut().CaptureStdErr().Execute();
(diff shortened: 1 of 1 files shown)


scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -163,7 +163,12 @@ public static BuildTargetResult RunPackageCommandTests(BuildTargetContext c)
             // Test the apps
             foreach (var dir in Directory.EnumerateDirectories(consumers))
             {
-                var result = dotnet.Exec("hello").WorkingDirectory(dir).CaptureStdOut().CaptureStdErr().Execute();
+                var result = dotnet.Exec("hello").WorkingDirectory(dir)
+                    .ForwardStdErr()
+                    .ForwardStdOut()
+                    .CaptureStdOut()
```
pakrym: Why capture and forward?

anurse: We need to check the output further down. So we want to do both here.

pakrym:
Didn't notice checking the output :shipit: after green build.

piotrpMSFT:
@anurse merging this as it'll help unblock CI.

PR #1443, opened 2016-02-16
fix version number padding

anurse:
Version numbers weren't padded properly.

Someone with the MyGet admin creds is going to have to purge the one or two versions that got published with the broken version strings.

/cc @piotrpMSFT @brthor @davidfowl


piotrpMSFT:
:shipit:

PR #1444, opened 2016-02-16
Remove dotnet-restore RID override

### LINKED ISSUE #1352: Remove RID detection override
In order to work around https://github.com/NuGet/Home/issues/1941, I added this override: https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-restore/Program.cs#L41 to our dotnet-restore wrapper. We need to remove that now, as it a) no longer necessary and b) causing issues when a `project.json` defines a `runtimes` section.

/cc @piotrpMSFT - I didn't have cycles to get this done right now but @dagood needed it for porting some repos. It should just be a matter of removing that entire section from the Program.cs, but it would require a little testing and verification to make sure restore still works for projects that don't specify `runtimes`

piotrpMSFT:
Fixes #1352

### DIFF
--- a/src/dotnet/commands/dotnet-restore/Program.cs
+++ b/src/dotnet/commands/dotnet-restore/Program.cs
@@ -7,7 +7,6 @@ using System.IO;
 using System.Linq;
 using System.Text;
 using Microsoft.DotNet.Cli.Utils;
-using Microsoft.DotNet.Cli.Compiler.Common;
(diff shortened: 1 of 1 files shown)


piotrpMSFT:
@anurse, can you sign off on this one?

anurse:
:shipit: (retroactively ;))

PR #1451, opened 2016-02-17
fix up casing of PS1 environment variables

piotrpMSFT:
The recent build system update seems to have renamed $Configuration to $CONFIGURATION but not all use cases were updated.


anurse:
Out of curiosity, did something break? I think Environment Variable names are case-insensitive. I do agree that using consistent casing is good so I'm happy with the change, just curious.

PR #1457, opened 2016-02-17
Revert System.Reflection.Metadata version to rc2

troydai:


troydai:
@piotrpMSFT @davidfowl verified locally. This change fixed the incorrect introduced 1.3.0.0 version of System.Reflection.Metadata dependency.

piotrpMSFT:
:shipit:

PR #1458, opened 2016-02-17
Fixes #1412.

### LINKED ISSUE #1412: Auto define for PCL profiles should be the profile number
Probably something like `PORTABLE_PROFILE{number}`.

``` C#

#if PORTABLE_PROFILE7
   // Something
#endif
```

scottdorman:
This adds support for automatically defining a compiler symbol for PCL
profiles.

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/ProjectReader.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectReader.cs
@@ -577,14 +577,16 @@ namespace Microsoft.DotNet.ProjectModel
 
         private static string MakeDefaultTargetFrameworkDefine(NuGetFramework targetFramework)
         {
-            var shortName = targetFramework.GetTwoDigitShortFolderName();
+            var candidateName = targetFramework.GetTwoDigitShortFolderName().ToUpperInvariant();
 
-            if (targetFramework.IsPCL)
+            if (targetFramework.IsPCL && string.IsNullOrWhiteSpace(targetFramework.Profile))
             {
                 return null;
             }
-
-            var candidateName = shortName.ToUpperInvariant();
+            else
+            {
+                candidateName = $"{FrameworkConstants.FrameworkIdentifiers.Portable.TrimStart('.')}_{targetFramework.Profile}".ToUpperInvariant();
+            }
 
             // Replace '-', '.', and '+' in the candidate name with '_' because TFMs with profiles use those (like "net40-client")
             // and we want them representable as defines (i.e. "NET40_CLIENT")


src/Microsoft.DotNet.ProjectModel/ProjectReader.cs
```diff
@@ -577,14 +577,16 @@ private static CommonCompilerOptions GetCompilationOptions(JsonObject rawObject,
...
-            if (targetFramework.IsPCL)
+            if (targetFramework.IsPCL && string.IsNullOrWhiteSpace(targetFramework.Profile))
             {
                 return null;
             }
-
-            var candidateName = shortName.ToUpperInvariant();
+            else
+            {
+                candidateName = $"{FrameworkConstants.FrameworkIdentifiers.Portable.TrimStart('.')}_{targetFramework.Profile}".ToUpperInvariant();
```
davidfowl: I don't see how this is right. Doesn't this happen for all targets? I also think we should test this.

scottdorman:
I see the issue and have it fixed locally right now. In testing, I'm seeing the following behavior though (from `NuGetFramework.Parse` and from the `GetTwoDigitShortFolderName()` extension method):

 NuGetFramework.Parse | GetTwoDigitShortFolderName | Expected?
------------ | -------------  | -------------
"dotnet5.2" | "dotnet5.2" | Y
"dotnet" | "dotnet5.0" | Y
"net40" | "net40" | Y
".NETPortable,V4.0,Profile=Profile328" | "portable-net40+sl5+win8+wp8+wpa81" | N - The TFM isn't valid
".NETPortable,Version=4.0,Profile=Profile328" | "portable40-net40+sl5+win8+wp8+wpa81" | N - The short folder name is wrong.
".NETPortable,Version=4.5,Profile=Profile328" | "portable45-net40+sl5+win8+wp8+wpa81" | N - The TFM isn't valid.
".NETPortable,Version=4.5,Profile=Profile259" | "portable45-net45+win8+wp8+wpa81" | N - The short folder name result is wrong.

The problem I'm running in to is that even when the TFM is wrong, the `NuGetFramework.Parse` method still works and sets `IsPCL` to `true` and `Profile` to the profile specified.

I'm going to look and see if there's a way of determining if the TFM is actually a valid PCL profile or not.

scottdorman:
After some more digging, it looks like the `Version` portion of the PCL profile string is effectively being ignored. I get the same results above if I just call `GetShortFolderName()` directly because it calls `DefaultFrameworkNameProvider.Instance.TryGetPortableFrameworks()`. That method uses only the numeric portion of the profile string to look for the key (so for Profile328, it looks for a key of 328), which it finds and returns the list of frameworks.

PR #1168, opened 2016-02-01
Update developer-guide with the correct paths

eerhardt:

PR #1460, opened 2016-02-17
Update developer-guide with the correct paths

eerhardt:
Moving #1168 to change branches from master to rel/1.0.0.

@davidfowl @piotrpMSFT


davidfowl:
:shipit:

PR #1463, opened 2016-02-18
Fix release builds in docker.

Sridhar-MS:
Move the args processing to run-build.sh. Earlier all the args were processed in build.sh and not passed on properly to a docker container.

cc @brthor @anurse @piotrpMSFT

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,48 +24,32 @@ if [ -z "$HOME" ]; then
     mkdir -p $HOME
 fi
 
+args=( "$@" )
+
 while [[ $# > 0 ]]; do
     lowerI="$(echo $1 | awk '{print tolower($0)}')"
     case $lowerI in
-        -c|--configuration)
-            export CONFIGURATION=$2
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -24,48 +24,32 @@ if [ -z "$HOME" ]; then
...
-            ;;
-        --nopackage)
-            export DOTNET_BUILD_SKIP_PACKAGING=1
-            ;;
         --docker)
             export BUILD_IN_DOCKER=1
             export DOCKER_IMAGENAME=$2
+            # remove docker args
+            args=( "${args[@]/$1}" )
+            args=( "${args[@]/$2}" )
```
brthor: this is cool, didn't know you could do this :smile:

Sridhar-MS: A caveat- It just removes the string. There is still an empty element in the array.

brthor:
Assuming Builds Pass, LGTM

PR #1471, opened 2016-02-18
Fixing versions

davidfowl:
- Added --version-suffix to build and publish
- Support reading DOTNET_* version variables everywhere versions can be read
- Show the commit sha in dotnet --version
- Added tests that check the assembly output version
- Set DOTNET_BUILD_VERSION when producing the CLI app itself so that it has the version information stamped in for help.

/cc @piotrpMSFT

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -3,8 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.Test;
+using Microsoft.Extensions.PlatformAbstractions;
+using NuGet.Frameworks;
 using Microsoft.DotNet.ProjectModel.Server;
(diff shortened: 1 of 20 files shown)


davidfowl:
@dotnet-bot test this please

src/dotnet/Program.cs
```diff
@@ -134,6 +135,12 @@ private static void PrintVersionInfo()
         {
             HelpCommand.PrintVersionHeader();
 
+            var commitSha = GetCommitSha();
+            Reporter.Output.WriteLine();
+            Reporter.Output.WriteLine("Product Information:");
+            Reporter.Output.WriteLine($" Version:     {HelpCommand.ProductVersion}");
+            Reporter.Output.WriteLine($" Commit Sha:  {commitSha.Substring(0, 10)}");
```
akoeplinger: this will cause a NullReferenceException when commitSha is null (e.g. when the file doesn't exist).

shahid-pk: yup i think it does get NRE when commitSha is null filled #1498

PR #1477, opened 2016-02-18
fix dotnet-compile.deps.json space in the path issue.

brthor:
Resources are a special case in quoting for CSC in response files.

This is the quick fix. Going to work on some abstractions so this stops happening.

cc @pakrym


test/dotnet-compile.Tests/CompilerTests.cs
```diff
@@ -88,6 +88,24 @@ public void LibraryWithAnalyzer()
             Assert.Contains("CA1018", result.StdErr);
         }
 
+        [Fact]
+        public void CompilingAppWithPreserveCompilationContextWithSpaceInThePathShouldSucceed()
```
davidfowl: Why aren't you using the new test framework?

PR #1484, opened 2016-02-18
Enable EndToEnd tests against the debian package

### LINKED ISSUE #381: Uncomment E2E Tests for Dotnet Debian Package
The debian package installation should be tested with End to end tests. 

This functionality is commented out because of https://github.com/dotnet/coreclr/issues/2215 

This issue prevents `dotnet restore` from running properly inside a docker container, which is where we do our package testing.

Before closing, the E2E test should be uncommented in `scripts/package/package-debian.sh`

brthor:
Fixes #381 

Enable end to end test to run against the binaries installed by the debian package.

### DIFF
--- a/scripts/package/package-debian.sh
+++ b/scripts/package/package-debian.sh
@@ -97,28 +97,43 @@ create_debian_package(){
 test_debian_package(){
     header "Testing debian package"
     
+    install_bats
+    run_package_integrity_tests
+
+    install_debian_package
+    run_e2e_test
+    remove_debian_package
+}
+
+install_bats() {
+    rm -rf $TEST_STAGE_DIR
+    git clone https://github.com/sstephenson/bats.git $TEST_STAGE_DIR
+}
+
+install_debian_package() {
+    sudo dpkg -i $DEBIAN_FILE
+}
+
+remove_debian_package() {
+    sudo dpkg -r $DOTNET_DEB_PACKAGE_NAME
+}
+
+run_package_integrity_tests() {
     # Set LAST_VERSION_URL to enable upgrade tests
     export LAST_VERSION_URL="https://dotnetcli.blob.core.windows.net/dotnet/$CHANNEL/Installers/Latest/dotnet-ubuntu-x64.latest.deb"
 
-    rm -rf $TEST_STAGE_DIR
-    git clone https://github.com/sstephenson/bats.git $TEST_STAGE_DIR
-    
     $TEST_STAGE_DIR/bin/bats $PACKAGE_OUTPUT_DIR/test_package.bats
-
-    # E2E Testing of package surface area
-    # Disabled: https://github.com/dotnet/cli/issues/381
-    # run_e2e_test
 }
 
 run_e2e_test(){
-    set +e    
-    sudo dpkg -i $DEBIAN_FILE
-    $REPOROOT/scripts/test/e2e-test.sh
-    result=$?
-    sudo dpkg -r dotnet
-    set -e
+    local dotnet_path="/usr/bin/dotnet"
+
+    header "Running EndToEnd Tests against debian package using ${dotnet_path}"
     
-    return result
+    # Won't affect outer functions
+    cd $REPOROOT/test/EndToEnd
+    $dotnet_path build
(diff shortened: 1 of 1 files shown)


scripts/package/package-debian.sh
```diff
@@ -97,28 +97,43 @@ create_debian_package(){
 test_debian_package(){
     header "Testing debian package"
     
-    # Set LAST_VERSION_URL to enable upgrade tests
-    export LAST_VERSION_URL="https://dotnetcli.blob.core.windows.net/dotnet/$CHANNEL/Installers/Latest/dotnet-ubuntu-x64.latest.deb"
+    install_bats
```
piotrpMSFT: Now that we have some nice support in our test framework for validating on-disk content, can we move these tests into xunit C# tests?

scripts/package/package-debian.sh
```diff
@@ -97,28 +97,43 @@ create_debian_package(){
...
+    run_package_integrity_tests
 
+    install_debian_package
+    run_e2e_test
```
piotrpMSFT: Do you see an advantage to running E2E twice? Should it just run out of the Native Installers?

scripts/package/package-debian.sh
```diff
@@ -97,28 +97,43 @@ create_debian_package(){
...
-    set -e
+    local dotnet_path="/usr/bin/dotnet"
+
+    header "Running EndToEnd Tests against debian package using ${dotnet_path}"
     
-    return result
+    # Won't affect outer functions
+    cd $REPOROOT/test/EndToEnd
+    $dotnet_path build
+    $dotnet_path test -xml $TEST_STAGE_DIR/debian-endtoend-testResults.xml
```
piotrpMSFT: Once we move to C#, would it make sense to unify our "Native Installer Test Suite" such that we had the same tests but with different implementations for each platform?

scripts/package/package-debian.sh
brthor: I think it will help with debugging if the issue is in the installer or in the source code.

Is the gain in perf worth the potential loss in time?

scripts/package/package-debian.sh
brthor: As far as the package_tool goes, these tests are dropped by the tool itself next to the deb package, and are not at all specific to our package. For that reason I think leaving that functionality makes sense for the package_tool, if it ever had additional consumers.

We could have a separate implementation in C# which would be independent to our repo. I think this makes sense and will bring consistency.

scripts/package/package-debian.sh
brthor: Sure, makes sense.

This will help reduce logic duplication for things like running E2E

scripts/package/package-debian.sh
piotrpMSFT: why do you prefer an external dependency on bats over an external dependency on dotnet-test?

scripts/package/package-debian.sh
piotrpMSFT: Let's keep an eye on it. That's a very expensive test, no?

piotrpMSFT:
:shipit:

scripts/package/package-debian.sh
brthor: Less so since it only does restore once now, but more expensive than any of the other tests (except builder tests :smile: )

scripts/package/package-debian.sh
brthor: That's a fair question and I hadn't thought of it in this frame.

I'm not really in favor of bats, it doesn't give us much and is only acquirable via git. 
Bringing dotnet test in the picture has more complexity but I think having it as the test runner, with a graceful fallback would be better than bats. 

What I want to avoid is package_tool having a dependency inside the repo itself (so it can remain conceptually as an independent tool). So far that independence hasn't slowed us down and I think it has value over the long term.

If it had a dependency on the dotnet test binary (via apt-get or whatever kind of install) I think that makes sense. The test would be more useful anyways since it will integrate better with build systems.

So an invocation of package_tool could be `package_tool -t` where -t specifies to compile and drop a test assembly next to the deb package. What I'm not sure of is whether the package_tool should auto run this assembly (via `dotnet test`) or if that should be an external action.

What do you think?

PR #1487, opened 2016-02-19
Fixes in corehost for building x86 arch.

Sridhar-MS:
- Set the calling convention to '__stdcall' explicitly else the default is '_cdecl' on windows which mismatches with coreclr hosting APIs.
- Use the newly built 'corehost' in stage1 instead of one from stage0.

cc @schellap

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -195,6 +195,10 @@ namespace Microsoft.DotNet.Cli.Build
                 }
             }
 
+            // dotnet.exe is from stage0. But we must be using the newly built corehost in stage1
+            File.Delete(Path.Combine(binDir, $"dotnet{Constants.ExeSuffix}"));
+            File.Copy(Path.Combine(binDir, $"corehost{Constants.ExeSuffix}"), Path.Combine(binDir, $"dotnet{Constants.ExeSuffix}"));
+
             // Crossgen Roslyn
             var result = Crossgen(c, binDir);
             if (!result.Success)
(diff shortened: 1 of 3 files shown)


src/corehost/cli/coreclr.cpp
```diff
@@ -9,7 +9,7 @@
 static pal::dll_t g_coreclr = nullptr;
 
 // Prototype of the coreclr_initialize function from coreclr.dll
-typedef pal::hresult_t(*coreclr_initialize_fn)(
+typedef pal::hresult_t(__stdcall *coreclr_initialize_fn)(
```
schellap: Compiles on unix?

piotrpMSFT:
@dotnet-bot test OSX Debug Build please.

piotrpMSFT:
once @schellap's comment is handled, :shipit:

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -195,6 +195,10 @@ private static BuildTargetResult CompileStage(BuildTargetContext c, DotNetCli do
                 }
             }
 
+            // dotnet.exe is from stage0. But we must be using the newly built corehost in stage1
```
piotrpMSFT: This is bootstrapping code, I assume...

Sridhar-MS: Well stage1 dotnet.exe is not the locally built corehost. Instead it is copied from stage0 (due to `dotnet publish`). So any fix in corehost src code cannot be bootstrapped in our builds easily. So I am planning to leave it as is.

src/corehost/cli/coreclr.cpp
Sridhar-MS: Fixed

schellap:
LGTM. Thanks.

PR #1499, opened 2016-02-19
Running Windows installer tests on Docker.

eerhardt:
Cleaning up the Installer tests
- rename project to conform to the rest of the tests
- convert to .xproj
- clean up unused usings


eerhardt:
cc @Sridhar-MS

piotrpMSFT:
:shipit:

test/Installer/testmsi.ps1
```diff
@@ -48,22 +61,26 @@ try {
...
+        CopyTestXUnitRunner $testBin
+
+        Write-Host "Running installer tests in Windows Container"
+        
+        $MsiFileName = [System.IO.Path]::GetFileName($inputMsi)
+        docker run `
+            -v "$testBin\:D:" `
+            -e "CLI_MSI=D:\$MsiFileName" `
+            windowsservercore `
+            D:\xunit.console.exe D:\$testName.dll | Out-Host
```
Sridhar-MS: I am assuming this hard doing is ok since it's happening inside docker.

Sridhar-MS:
:shipit:

test/Installer/testmsi.ps1
piotrpMSFT: Good catch... What happens when we want to run this outside docker? E.g. Repro on win7?

eerhardt: When you want to run it outside of docker, you can run it from $testBin.  This hard-coding in the docker command is fine since all the hard-coding happens on this one command in the 3 arguments.

```
-v "$testBin\:D:" `
```
That says "map the $testBin directory to the D: drive on the docker container". (Using drive letters, instead of full paths improves the experience inside the docker container. For example, inside the docker container you can't run 'git' commands if you use a path other than a drive letter, since using a path utilizes symlinks. But if you use a drive letter, 'git' commands work inside the container. I have an email to the Windows Container team on this issue.)

```
-e "CLI_MSI=D:\$MsiFileName" `
```
That sets the environment variable to tell the tests where to pick up the .msi.

```
D:\xunit.console.exe D:\$testName.dll | Out-Host
```
That says run the xunit.console.exe on the test assembly, which are both in the $testBin, which has been mapped to the D drive.


If/when we want to run this outside of docker, you wouldn't use this docker command, instead you'd just need to say 
```
$env:CLI_MSI=$inputMsi
$testBin\xunit.console.exe $testBin\$testName.dll
```

piotrpMSFT:
:shipit:

Issue #1497, opened 2016-02-19
Display better error message when runtime target is not found in project lock


pakrym:
See https://github.com/dotnet/cli/issues/1492

blackdwarf:
@pakrym will close this since it seems it is fixed. If its not, please reopen.

Issue #1414, opened 2016-02-16
If a single framework is specific in project.json use it for dotnet run


davidfowl:
I currently get this error:

```
System.InvalidOperationException: Couldn't find default target with framework: DNXCore,.NETStandardApp
   at Microsoft.DotNet.Tools.Run.RunCommand.CalculateDefaultsForNonAssigned()
   at Microsoft.DotNet.Tools.Run.RunCommand.RunExecutable()
   at Microsoft.DotNet.Tools.Run.RunCommand.Start()
   at Microsoft.DotNet.Tools.Run.RunCommand.Run(String[] args)
```

/cc @pakrym

pakrym:
Run restore, there is no runtime target in project.lock

davidfowl:
This used to work and it broke. Running `dotnet run -f net451` works.

pakrym:
Will fix.

blackdwarf:
@pakrym @davidfowl is this fixed? Should we close?

PR #1501, opened 2016-02-19
Fix run behaviour when only single target exists

pakrym:
Plays good with https://github.com/dotnet/cli/issues/1497 which shows nice error message if selected target is not restored for current runtime.

@davidfowl


davidfowl:
Where is the test bro?

pakrym:
https://github.com/dotnet/cli/issues/1414

piotrpMSFT:
:) we still need a test in dotnet-run.Tests to make sure we don't regress!

PR #1513, opened 2016-02-21
Fix net20 compatibility

### LINKED ISSUE #1480: dotnet compile fails for net20 target, because of TargetFrameworkAttribute in AssemblyInfo.cs
How to reproduce:
 * dotnet new
 * change configure project.json to target .net 2.0: _"frameworks": { "net20": { } }_
 * dotnet restore
 * dotnet build

Then the following error pops up:
> C:\Program Files\dotnet\bin\dotnet-compile-csc.exe "@D:\dev\coretestx\obj\Debug\net20\dotnet-compile.rsp" returned Exit Code 1
> D:\dev\coretestx\obj\Debug\net20\dotnet-compile.assemblyinfo.cs(4,37): error CS0234: The type or namespace name 'TargetFrameworkAttributeAttribute' does not exist in the namespace 'System.Runtime.Versioning' (are you missing an assembly reference?)
D:\dev\coretestx\obj\Debug\net20\dotnet-compile.assemblyinfo.cs(4,37): error CS0234: The type or namespace name 'TargetFrameworkAttribute' does not exist in the namespace 'System.Runtime.Versioning' (are you missing an assembly reference?)

> Compilation failed.
    0 Warning(s)
    2 Error(s)

When looking into _\obj\Debug\net20\dotnet-compile.assemblyinfo.cs_, obviously there's a _TargetFrameworkAttribute_ which is only available .net 4.0+.

discostu105:
Fixes #1480

Added TargetFrameworkVersion to AssemblyInfoOptions in order to let AssemblyInfoFileGenerator decide if attributes are compatible with framework version. E.g. TargetFrameworkAttribute is not .NET 2.0 compatible.

### DIFF
--- a/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
@@ -10,6 +10,7 @@ using Microsoft.CodeAnalysis.CSharp;
 using System.IO;
 using System.Runtime.Versioning;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using NuGet.Frameworks;
 
 namespace Microsoft.DotNet.Cli.Compiler.Common
 {
@@ -59,7 +60,10 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         private static Dictionary<Type, string> GetProjectAttributes(AssemblyInfoOptions metadata)
         {
-            var attributes = new Dictionary<Type, string>() {
+            NuGetFramework targetFramework = NuGetFramework.Parse(metadata.TargetFramework);
+
+            var attributes = new Dictionary<Type, string>()
+            {
                 [typeof(AssemblyTitleAttribute)] = EscapeCharacters(metadata.Title),
                 [typeof(AssemblyDescriptionAttribute)] = EscapeCharacters(metadata.Description),
                 [typeof(AssemblyCopyrightAttribute)] = EscapeCharacters(metadata.Copyright),
@@ -71,7 +75,8 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             };
             
             // only .net 4.0+ compatible
-            if (metadata.TargetFrameworkVersion == null || metadata.TargetFrameworkVersion >= new Version(4, 0)) {
+            if (targetFramework.Version >= new Version(4, 0))
+            {
                 attributes[typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework);
             };
             return attributes;
--- a/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
@@ -60,8 +60,6 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         private static Dictionary<Type, string> GetProjectAttributes(AssemblyInfoOptions metadata)
         {
-            NuGetFramework targetFramework = NuGetFramework.Parse(metadata.TargetFramework);
-
             var attributes = new Dictionary<Type, string>()
             {
                 [typeof(AssemblyTitleAttribute)] = EscapeCharacters(metadata.Title),
@@ -74,10 +72,10 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
                 [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage)
             };
             
-            // only .net 4.0+ compatible
-            if (targetFramework.Version >= new Version(4, 0))
+            NuGetFramework targetFramework = string.IsNullOrEmpty(metadata.TargetFramework) ? null : NuGetFramework.Parse(metadata.TargetFramework);
+            if (targetFramework != null && !(targetFramework.IsDesktop() && targetFramework.Version < new Version(4, 0)))
             {
-                attributes[typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework);
+                attributes[typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework); // TargetFrameworkAttribute only exists since .NET 4.0
             };
             return attributes;
         }
--- a/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
@@ -71,13 +71,31 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
                 [typeof(AssemblyCultureAttribute)] = EscapeCharacters(metadata.Culture),
                 [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage)
             };
-            
-            NuGetFramework targetFramework = string.IsNullOrEmpty(metadata.TargetFramework) ? null : NuGetFramework.Parse(metadata.TargetFramework);
-            if (targetFramework != null && !(targetFramework.IsDesktop() && targetFramework.Version < new Version(4, 0)))
+
+            if (IsAllowV4Attributes(metadata))
             {
                 attributes[typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework); // TargetFrameworkAttribute only exists since .NET 4.0
             };
+
             return attributes;
+        }
+
+        private static bool IsAllowV4Attributes(AssemblyInfoOptions metadata)
+        {
+            if (string.IsNullOrEmpty(metadata.TargetFramework))
+            {
+                // target framework is unknown. to be on the safe side, return false.
+                return false;
+            }
+
+            NuGetFramework targetFramework = NuGetFramework.Parse(metadata.TargetFramework);
+            if (!targetFramework.IsDesktop())
+            {
+                // assuming .NET Core, which should support .NET 4.0 attributes
+                return true;
+            }
+
+            return targetFramework.Version >= new Version(4, 0);
         }
 
         private static bool IsSameAttribute(Type attributeType, AttributeSyntax attributeSyntax)
--- a/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
@@ -74,7 +74,8 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
             if (SupportsTargetFrameworkAttribute(metadata))
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
```diff
@@ -68,9 +68,15 @@ public static string GenerateFSharp(AssemblyInfoOptions metadata)
                 [typeof(AssemblyVersionAttribute)] = EscapeCharacters(metadata.AssemblyVersion?.ToString()),
                 [typeof(AssemblyInformationalVersionAttribute)] = EscapeCharacters(metadata.InformationalVersion),
                 [typeof(AssemblyCultureAttribute)] = EscapeCharacters(metadata.Culture),
-                [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage),
-                [typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework)
+                [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage)
             };
+            
+            // only .net 4.0+ compatible
+            if (metadata.TargetFrameworkVersion == null || metadata.TargetFrameworkVersion >= new Version(4, 0))
```
davidfowl: This isn't right. You need to parse the target framework as a NuGetFramework. The version isn't enough to tell. You need to make sure it's `IsDesktop()`  as well.

This also needs a test

discostu105: You're right. I just found that there is a NuGetFramework.Parse method, so I'll use that and dump the TargetFrameworkVersion again.

Concerning tests. I'm happy to add one, but I'll need some guidance. Should I add a new project in TestAssets\TestProjects which targets multiple frameworks (incl. net20), or is there already something to reuse? Then add a new test class in 'dotnet-build.Tests' like 'IncrementalTestsMultiTargetFrameworks'?

davidfowl:
For the test, copy this one as a template:

https://github.com/dotnet/cli/blob/rel/1.0.0/test/dotnet-build.Tests/BuildOutputTests.cs#L116

You'll need to make a new test project under test assets that has a `net20` framework target. We basically want to verify a few cases:

- There's no attribute in the `net20` case
- There's an attribute with the long target framework name in all other cases.

src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
```diff
@@ -68,9 +69,15 @@ public static string GenerateFSharp(AssemblyInfoOptions metadata)
                 [typeof(AssemblyVersionAttribute)] = EscapeCharacters(metadata.AssemblyVersion?.ToString()),
                 [typeof(AssemblyInformationalVersionAttribute)] = EscapeCharacters(metadata.InformationalVersion),
                 [typeof(AssemblyCultureAttribute)] = EscapeCharacters(metadata.Culture),
-                [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage),
-                [typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework)
+                [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage)
             };
+            
+            NuGetFramework targetFramework = string.IsNullOrEmpty(metadata.TargetFramework) ? null : NuGetFramework.Parse(metadata.TargetFramework);
```
davidfowl: When is `metadata.TargetFramework` null?

src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
```diff
@@ -68,9 +69,15 @@ public static string GenerateFSharp(AssemblyInfoOptions metadata)
...
+            if (targetFramework != null && !(targetFramework.IsDesktop() && targetFramework.Version < new Version(4, 0)))
```
davidfowl: This condition is hard to grok.

src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
discostu105: The unit test "TestProjectWithCultureSpecificResource" failed, so I added the null check. Is this unintended?
```
Project TestProjectWithCultureSpecificResource (DNXCore,Version=v5.0) will be compiled because some of its inputs were newer than its oldest output.
Compiling TestProjectWithCultureSpecificResource for DNXCore,Version=v5.0
System.ArgumentNullException: Value cannot be null.
Parameter name: folderName
   at NuGet.Frameworks.NuGetFramework.Parse(String folderName, IFrameworkNameProvider mappings)
   at Microsoft.DotNet.Cli.Compiler.Common.AssemblyInfoFileGenerator.IsAllowV4Attributes(AssemblyInfoOptions metadata) in C:\dev\dotnet-cli\src\Microsoft.DotNet.Compiler.Common\AssemblyInfoFileGenerator.cs:line 91
   at Microsoft.DotNet.Cli.Compiler.Common.AssemblyInfoFileGenerator.GetProjectAttributes(AssemblyInfoOptions metadata) in C:\dev\dotnet-cli\src\Microsoft.DotNet.Compiler.Common\AssemblyInfoFileGenerator.cs:line 75
   at Microsoft.DotNet.Cli.Compiler.Common.AssemblyInfoFileGenerator.Generate(AssemblyInfoOptions metadata, IEnumerable`1 sourceFiles) in C:\dev\dotnet-cli\src\Microsoft.DotNet.Compiler.Common\AssemblyInfoFileGenerator.cs:line 21
   at Microsoft.DotNet.Tools.Resgen.ResourceAssemblyGenerator.Generate(ResourceSource[] sourceFiles, Stream outputStream, AssemblyInfoOptions metadata, String assemblyName, String[] references) in C:\dev\dotnet-cli\src\dotnet\commands\dotnet-resgen\ResourceAssemblyGenerator.cs:line 59
   at Microsoft.DotNet.Tools.Resgen.ResgenCommand.Execute() in C:\dev\dotnet-cli\src\dotnet\commands\dotnet-resgen\ResgenCommand.cs:line 36
   at Microsoft.DotNet.Tools.Resgen.ResgenCommand.Run(String[] args) in C:\dev\dotnet-cli\src\dotnet\commands\dotnet-resgen\Program.cs:line 63
```

src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
discostu105: You're right. Let me make one more suggestion (it's your call of course :))

```
        private static bool IsAllowV4Attributes(AssemblyInfoOptions metadata)
        {
            if (string.IsNullOrEmpty(metadata.TargetFramework))
            {
                // target framework is unknown. to be on the safe side, return false.
                return false;
            }

            NuGetFramework targetFramework = NuGetFramework.Parse(metadata.TargetFramework);
            if (!targetFramework.IsDesktop())
            {
                // assuming .NET Core, which should support .NET 4.0 attributes
                return true;
            }

            return targetFramework.Version >= new Version(4, 0);
        }
```
usage
```
            if (IsAllowV4Attributes(metadata))
            {
                attributes[typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework); // TargetFrameworkAttribute only exists since .NET 4.0
            };
```

davidfowl: Like it but you should fix the name, `SupportsTargetFrameworkAttribute`

discostu105:
sorry, i kind of messed up the git history on this. i'm working on it.

src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
```diff
@@ -68,9 +69,33 @@ public static string GenerateFSharp(AssemblyInfoOptions metadata)
...
                 [typeof(AssemblyInformationalVersionAttribute)] = EscapeCharacters(metadata.InformationalVersion),
                 [typeof(AssemblyCultureAttribute)] = EscapeCharacters(metadata.Culture),
-                [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage),
-                [typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework)
+                [typeof(NeutralResourcesLanguageAttribute)] = EscapeCharacters(metadata.NeutralLanguage)
             };
+
+            if (SupportsTargetFrameworkAttribute(metadata))
+            {
+                attributes[typeof(TargetFrameworkAttribute)] = EscapeCharacters(metadata.TargetFramework); // TargetFrameworkAttribute only exists since .NET 4.0
```
davidfowl: nit: move the comment above this line. Putting at the end makes the line super long.

src/Microsoft.DotNet.Compiler.Common/AssemblyInfoFileGenerator.cs
```diff
@@ -68,9 +69,33 @@ public static string GenerateFSharp(AssemblyInfoOptions metadata)
...
+
+        private static bool SupportsTargetFrameworkAttribute(AssemblyInfoOptions metadata)
+        {
+            if (string.IsNullOrEmpty(metadata.TargetFramework))
+            {
+                // target framework is unknown. to be on the safe side, return false.
+                return false;
+            }
+
+            NuGetFramework targetFramework = NuGetFramework.Parse(metadata.TargetFramework);
```
davidfowl: var

discostu105:
Not sure why tests cannot be executed. Jenkins says "Hung processs on target machine".

piotrpMSFT:
What is the experience on non-Windows? Do we now show a reasonable error?

davidfowl:
> What is the experience on non-Windows? Do we now show a reasonable error?

We should errors when you don't have the targeting pack installed for the specified framework. https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs#L238

discostu105:
> We should errors when you don't have the targeting pack installed for the specified framework. https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs#L238

I've changed the test to assert Should().Fail() instead of not running the test. We could as well assert on some output on stdout/stderr such as _Framework not installed: {TargetFramework.DotNetFrameworkName}_, but IMHO then this test would cover more than what it's supposed to cover.

Maybe there should be a separate test, which covers all supported frameworks and asserts on messages such as _Framework not installed_?

discostu105:
By the way. The build for net20 on *nix platforms printed errors like this on stderr:
```
error CS0246: The type or namespace name 'System' could not be found (are you missing a using directive or an assembly reference?)
error CS0518: Predefined type 'System.String' is not defined or imported
error CS0518: Predefined type 'System.Object' is not defined or imported
error CS0518: Predefined type 'System.Void' is not defined or imported
```

Nothing like `Framework not installed` in the logs.

davidfowl:
@discostu105 need bettar error handling! :smile: Not related to this change.

davidfowl:
LGTM

PR #1554, opened 2016-02-24
Fix program files probe and HOME dir

schellap:
Updating as per corehost spec.


src/corehost/common/pal.windows.cpp
```diff
@@ -27,7 +26,19 @@ bool pal::find_coreclr(pal::string_t* recv)
         }
     }
 
-    // TODO: Try somewhere in Program Files, see https://github.com/dotnet/cli/issues/249
```
schellap: @shhsu just implemented this `TODO`

schellap:
@gkhanna79 PTAL.

Cc @anurse, @piotrpMSFT.

Probe `ProgramFiles` for Windows as a last resort. Fallback on `getpwuid` if `$HOME` is absent as recommended by `getpwuid` man pages.

schellap:
In my testing NuGet *fails* to restore properly into `~/.nuget` as well if `HOME` is not set.

src/corehost/common/pal.windows.cpp
```diff
@@ -27,7 +26,19 @@ bool pal::find_coreclr(pal::string_t* recv)
...
+
+    // Try %ProgramFiles%. Note this works for both x86 and x64/wow64 as per:
```
gkhanna79: Do we need this comment given that you are looking up ProgramFiles env var?

src/corehost/common/pal.unix.cpp
```diff
@@ -83,12 +86,22 @@ bool pal::is_path_rooted(const pal::string_t& path)
...
+    }
+    if (dir.empty())
     {
         return false;
     }
-    append_path(&*recv, _X(".nuget"));
-    append_path(&*recv, _X("packages"));
+    append_path(&dir, _X(".nuget"));
+    append_path(&dir, _X("packages"));
+    recv->assign(dir);
```
gkhanna79: With this change in place, HOME is no longer required to be defined. CI machines do not have it defined and that broke CoreCLR builds when using CLI to perform restore. See https://github.com/dotnet/coreclr/commit/f025f9c678b7194d3b125d0cfa4f8c8e9d27beab#diff-c02be70df3e835d946f675bd0bedc234 for details.

You should file an issue to remove the CI specific change from CoreCLR once CoreCLR uses the updated CLI.

Otherwise, change LGTM

src/corehost/common/pal.windows.cpp
schellap: Yeah. This comment helps how it works for x86/x64/wow64. Not sure about nano server? Last ditch, if unset, fail.

src/corehost/common/pal.unix.cpp
schellap: Apparently NuGet is not resilient to HOME not being present. So I believe we need it if they do `dotnet restore`

Cc @yishaigalatzer

anurse: Can you file an issue on NuGet/Home for that? If they need to P/Invoke something for this fallback that should be reasonable as long as it's simple. Ideally there should be a CoreFX API for getting the home directory, but whatever works :).

schellap:
@dotnet-bot test this please

PR #1583, opened 2016-02-25
Links to dev docs in root readme

AlexKeySmith:
As a newbie to building cli from source it was easy to miss the dev documentation and jump straight into running build.cmd not realizing I needed cmake etc to be installed.


blackdwarf:
:shipit:

Issue #1602, opened 2016-02-26
[DotNet.ProjectModel.Server] Error: An item with the same key has already been added. Key: Microsoft.CSharp


abpiskunov:
When I have a dotnet project (no sdk/version in global.json) and following project.json, I see error:
"An item with the same key has already been added. Key: Microsoft.CSharp". Error is confusing since project.json has only one dependency with such name. 

```
{
  "version": "1.0.0-*",
  "description": "KProjectB Class Library",

  "frameworks": {
    "dnx451": { 
      "dependencies": {
        "Microsoft.CSharp": "4.0.1-*",
        "System.Collections": "4.0.11-*",
        "System.Linq": "4.0.1-*",
        "System.Runtime": "4.0.21-*",
        "System.Threading": "4.0.11-*"
      }
    }
  }
}
```

abpiskunov:
Related issue:
If immediately after I saw error above, I remove "Microsoft.CSharp" dependency from project.json (keep other packages). I see following error (warning about dependencies are expected, but error is not)


Severity	Code	Description	Project	File	Line	Suppression State
**Error		Value cannot be null.**
**Parameter name: o	KProjectLibraryNonSrc		1	**
Warning	NU1007	Dependency specified was System.Collections >= 4.0.11-* but ended up with System.Collections 4.0.0.0.	KProjectLibraryNonSrc	D:\temp\TestSolution\References\ProjectReferences\nonsrc\KProjectLibraryNonSrc\project.json	8	
Warning	NU1007	Dependency specified was System.Linq >= 4.0.1-* but ended up with System.Linq 4.0.0.0.	KProjectLibraryNonSrc	D:\temp\TestSolution\References\ProjectReferences\nonsrc\KProjectLibraryNonSrc\project.json	9	
Warning	NU1007	Dependency specified was System.Runtime >= 4.0.21-* but ended up with System.Runtime 4.0.10.0.	KProjectLibraryNonSrc	D:\temp\TestSolution\References\ProjectReferences\nonsrc\KProjectLibraryNonSrc\project.json	10	
Warning	NU1007	Dependency specified was System.Threading >= 4.0.11-* but ended up with System.Threading 4.0.0.0.	KProjectLibraryNonSrc	D:\temp\TestSolution\References\ProjectReferences\nonsrc\KProjectLibraryNonSrc\project.json	11

troydai:
The warnings is not related in this case. If you `dotnet build` following project on command line you got a similar warning.

```
{
[...]

PR #1608, opened 2016-02-26
Ensure Microsoft.CSharp reference is not duplicated

troydai:
Address: https://github.com/dotnet/cli/issues/1602
With tests.

/cc @davidfowl


davidfowl:
:shipit:

PR #1630, opened 2016-03-01
Do not download a pinned version of cli if it is already installed.

### LINKED ISSUE #1628: Obtain script always downloads cli on Mac when the version is pinned
If the $VERSION of cli is not specified the script gets the (hash for the latest version)[https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/obtain/install.sh#L178] from the server which in majority of cases does not match what's already on the drive. As a result it will re-download cli again even if this is the same version that is already installed.

```
cat /Users/###/.dotnet/share/dotnet/cli/.version
6aeed1f52d73ffa8bb01229e4638fe3d8250b9e7
1.0.0.001540
```

1540 was on the box and because 1573 is the latest on the server the scripts downloads 1540 again:

```
dotnet_install: Preparing to install .NET Tools from 'beta' channel to '/Users/###/.dotnet'
dotnet_install: Latest Version: 1.0.0.001573
dotnet_install: Local Version: 1.0.0.001540
dotnet_install: Downloading dotnet-osx-x64.1.0.0.001540.tar.gz from https://dotnetcli.blob.core.win
```

moozzyk:
Fixes #1628

### DIFF
--- a/scripts/obtain/install.sh
+++ b/scripts/obtain/install.sh
@@ -174,21 +174,25 @@ install_dotnet()
 
     if [ "$RELINK" = "0" ]; then
         if [ "$FORCE" = "0" ]; then
-            # Check if we need to bother
-            local remoteData="$(curl -s https://dotnetcli.blob.core.windows.net/dotnet/$CHANNEL/dnvm/latest.$os.version)"
-            [ $? != 0 ] && say_err "Unable to determine latest version." && return 1
-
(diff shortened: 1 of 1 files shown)


piotrpMSFT:
@krwq I'm ok with merging this after CR, but please be aware that the scripts are currently being entirely rewritten per #1482

krwq:
Looks good to me!

moozzyk:
Thanks!

PR #1636, opened 2016-03-01
Update runtime config spec

anurse:
/cc @gkhanna79 @schellap @davidfowl @DamianEdwards @piotrpMSFT @pakrym


Documentation/specs/runtime-configuration-file.md
```diff
@@ -25,6 +25,10 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files.
     "runtimeConfig": {
         "gcServer": true,
         "gcConcurrent": false
+    },
+    "targetFramework": {
```
davidfowl: No we aren't doing this. We're going to use the package. There's no point duplicating that information. On the edge we'll put redist: true or something

anurse: That's conflicting with what I heard yesterday. Is that a recent change? I'm pretty sure I explicitly asked if it was in the editable ".config.json" file and heard that it was :)

anurse:
@dotnet-bot test this please

Documentation/specs/runtime-configuration-file.md
schellap: @anurse +1

gkhanna79: @davidfowl Where do you expect to find the SharedFramework name and version if not in runtime.config.json? Also, I hope we are not setting serverGC to true by default, right?

Documentation/specs/runtime-configuration-file.md
```diff
@@ -128,9 +123,13 @@ This section is copied verbatim from an identical section in the input `project.
 
 These settings are read by `corehost` to determine how to initialize the runtime. All versions of `corehost` **must ignore** settings in this section that they do not understand (thus allowing new settings to be added in later versions).
 
+### `targetFramework` Section (`.config.json`)
+
+This section indicates the name and version of the shared framework in use. Where this data comes from is TBD right now, as is the name of this section. If not present, default values of `NetCoreApp` for `name` and `1.0.0` for `version` should be assumed (REVIEW: Is this the right behavior? Should we error out if it isn't present?).
```
gkhanna79: @anurse Isn't publish operation writing the Framework name and version? What is TBD where?

gkhanna79: I dont believe we want to start with any defaults and fail to make errors obvious. This would be in line with app specifying a version and host not finding it and failing.

Documentation/specs/runtime-configuration-file.md
anurse: Chatted with @davidfowl - He misunderstood this, we are doing this as discussed in this document.

Documentation/specs/runtime-configuration-file.md
anurse: Sure, no defaults is fine with me. "TBD" here refers to where publish gets the data. Publish will put it here, but it needs a way to know what to put here :). We have ideas, but they are out of the scope of this document. I'll clarify that.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -10,21 +10,25 @@ There are two runtime configuration files for a particular application. Given a
...
-### [appname].config.json
+### [appname].runtimeconfig.json
 ```json
 {
-    "runtimeConfig": {
+    "runtimeOptions": {
         "gcServer": true,
         "gcConcurrent": false
+    },
+    "targetRuntime": {
```
piotrpMSFT: @pmarcu, want to make sure that targetRuntime is the right property name here.

gkhanna79: Shouldnt this be "targetFramework"?

anurse: We went back and forth in the meeting and really wanted to avoid the connotations that `targetFramework` bring up. I'm certainly open to ideas.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -282,7 +276,54 @@ Some of the sections in the `.deps.json` file contain data used for runtime comp
 
 An application can be deployed in a "portable" deployment model. In this case, the RID-specific assets of packages are published within a folder structure that preserves the RID metadata. However, `corehost` does not use this folder structure, rather it reads data from the `.deps.json` file. Also, during deployment, the `.exe` file (`corehost` renamed) is not deployed.
 
-In a `portable` target, the package entries may have an additional `subtargets` section detailing RID-specific assets. The `corehost` application should use this data, along with the current RID and the fallback data defined in the `runtimes` section to select one **and only one** "subtarget" out of each package individually. The most specific subtarget should always be selected. In practice, this means selecting the first RID shown on the appropriate line in `runtimes`. For example, given a package containing the following subtargets:
+In the portable deployment model, the `*.runtimeConfig.json` file will contain the `targetRuntime` section:
+
+```json
+{
+    "targetRuntime": {
+        "name": "NETCore.App",
```
schellap: Should either be runtimeTarget or targetRuntime (or framework) uniform.

Documentation/specs/runtime-configuration-file.md
piotrpMSFT: Many TFMs will use the same runtime. That's why this string seems funny. targetPlatform?

Documentation/specs/runtime-configuration-file.md
```diff
@@ -282,7 +276,54 @@ Some of the sections in the `.deps.json` file contain data used for runtime comp
...
+
+This data is used to locate the shared framework folder. The exact mechanics of which version are selected are defined elsewhere, but in general, it locates the shared runtime in the `shared` folder located beside it by using the relative path `shared/[targetRuntime.name]/[targetRuntime.version]`. Once it has applied any version roll-forward logic and come to a final path to the shared framework, it locates the `[targetRuntime.name].deps.json` file within that folder and loads it **first**.
+
+Next, the deps file from the application is loaded and merged into this deps file (this is conceptual, the host implementation doesn't necessary have to directly merge the data ;)). Data from the app-local deps file trumps data from the shared framework.
+
+The shared framework's deps file will also contain a `runtimes` section defining the fallback logic for all RIDs known to that shared framework. For example, a shared framework deps file installed into a Ubuntu machine may look something like the following:
+
+```json
+{
+    "targets": {
```
schellap: Missing runtimeTarget.

anurse: Yeah. This also raises an issue during the merging where the `runtimeTarget` from the shared framework deps file should be merged with the `runtimeTarget` from the app-local deps file. It's less important to have a `runtimeTarget` in the shared framework deps file because it will only have one target, but for consistency we should have it.

Documentation/specs/runtime-configuration-file.md
weshaggard: How will publish recognize these references and know it needs to create this entry? Perhaps that might be a clue as to how we should name it.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -128,9 +123,13 @@ This section is copied verbatim from an identical section in the input `project.
 
 These settings are read by `corehost` to determine how to initialize the runtime. All versions of `corehost` **must ignore** settings in this section that they do not understand (thus allowing new settings to be added in later versions).
 
+### `targetRuntime` Section (`.runtimeconfig.json`)
+
+This section indicates the name and version of the shared runtime in use. The `dotnet publish` command will place this data here (where it gets that data from is undecided but out of the scope of this document). If not present, the host will produce an error.
+
```
weshaggard: Why would the host produce an error if this isn't present? I would expect this to not be present in a standalone application for example.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -10,21 +10,25 @@ There are two runtime configuration files for a particular application. Given a
 
 * `MyApp.dll` - The managed assembly for `MyApp`, including an ECMA-compliant entry point token.
 * `MyApp.exe` - A copy of the `corehost.exe` executable.
-* `MyApp.config.json` - An **optional** configuration file containing runtime configuration settings.
+* `MyApp.runtimeconfig.json` - An **optional** configuration file containing runtime configuration settings.
```
weshaggard: Is there a reason we need a separate runtime config from the deps file? Can they just be a single file?

Documentation/specs/runtime-configuration-file.md
weshaggard: I see below that the trigger isn't clear. However I did notice that the deps file as a "runtimeTarget" isn't that representing the same information?

anurse: No, `runtimeTarget` is expressing which entry in the `targets` section `corehost` should use, since we don't want `corehost` to deal with parsing Target Names.

Documentation/specs/runtime-configuration-file.md
anurse: Good point, I'll update.

Documentation/specs/runtime-configuration-file.md
anurse: Because the `.runtimeconfig.json` file is designed to be user-editable (much like `[assemblyname].exe.config` in full framework) whereas `.deps.json` is not. I'll clarify that.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -128,9 +122,13 @@ This section is copied verbatim from an identical section in the input `project.
 
 These settings are read by `corehost` to determine how to initialize the runtime. All versions of `corehost` **must ignore** settings in this section that they do not understand (thus allowing new settings to be added in later versions).
 
+### `targetRuntime` Section (`.runtimeconfig.json`)
+
+This section indicates the name and version of the shared runtime in use. The `dotnet publish` command will place this data here (where it gets that data from is undecided but out of the scope of this document). If not present, the host will assume the app is published as a standalone app. The presence of this section indicates if the application is standalone or not.
```
anurse: FYI @schellap We no longer need the `portable` flag in `runtimeTarget`. The presence of this section in the runtime configuration file is sufficient to identify a portable application. Since the `.runtimeconfig.json` file is optional, if the file is not present, we will assume the app is standalone.

Documentation/specs/runtime-configuration-file.md
anurse: Though that confusion seems likely to be a recurring issue. Since one property is called `targetRuntime` and the other `runtimeTarget` :). Perhaps we need to avoid `target` in this property name.

`sharedFramework`? `frameworkReference`? Basically we're trying to identify the name and version of the Redist in use (without using the term "Redist" because it's not a great term :))

pakrym:
```
Only dependencies with a type value of package will have asset lists (compile, runtime, resources, native). Dependencies which are satisfied by projects will have all of their assets copied to the output directory, so they will not be listed in this file.
``` 
We need `project` type and their dependency info to do Controller resolution and similar things. Also `ReferenceAssemblies` for compilation.

Documentation/specs/runtime-configuration-file.md
schellap: Are you sure, there will still be a single target name always if you remove this section?

schellap: Also parsing 2 files to get a sense of the app and one user editable doesn't sound like deps is self sufficient...

Documentation/specs/runtime-configuration-file.md
muratg: Yeah, this will be confusing if we have `targetRuntime` in one file and `runtimeTarget` in another.

Would just `runtime` work?

anurse:
@dotnet-bot test Ubuntu Release Build please

Documentation/specs/runtime-configuration-file.md
schellap: Stick it in `runtimeOptions: FX: { name: {} version: {} }`

Documentation/specs/runtime-configuration-file.md
```diff
@@ -292,61 +336,42 @@ In a `portable` target, the package entries may have an additional `subtargets`
                 "compile": {
                     "ref/netstandard1.5/System.Data.SqlClient.dll": {}
                 },
-                "subtargets": {
-                    "runtime": {
-                        "runtimes/unix/lib/netstandard1.5/System.Data.SqlClient.dll": { "rid": "unix" },
-                        "runtimes/win7-x64/lib/netstandard1.5/System.Data.SqlClient.dll": { "rid": "win7-x64" },
-                        "runtimes/win7-x86/lib/netstandard1.5/System.Data.SqlClient.dll": { "rid": "win7-x86" }
+                "runtimeTargets": {
```
anurse: @schellap In order to unify this format with how NuGet is producing the data (/cc @emgarten), we've changed the layout of this section just slightly. Basically, we've flattened out the child nodes of `subtargets` and renamed it to `runtimeTargets`. The behavior is the same, and I've updated the resolution algorithm below to correspond to these changes.

Documentation/specs/runtime-configuration-file.md
gkhanna79: @schellap +1

anurse: Sure, makes sense.

Documentation/specs/runtime-configuration-file.md
anurse: No, we'd still have `runtimeTarget` to identify the target, it just doesn't need the `portable` flag. Instead, the host just looks at `runtimeOptions.framework` to determine if it needs to merge in the additional deps file provided by the shared framework. Since we decided we want that data to be user-editable (to allow a consumer to adjust the framework target), we put it in the runtimeconfig file.

The deps file is self-sufficient when in standalone mode because you don't need to merge the shared framework deps.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -282,7 +275,58 @@ Some of the sections in the `.deps.json` file contain data used for runtime comp
 
 An application can be deployed in a "portable" deployment model. In this case, the RID-specific assets of packages are published within a folder structure that preserves the RID metadata. However, `corehost` does not use this folder structure, rather it reads data from the `.deps.json` file. Also, during deployment, the `.exe` file (`corehost` renamed) is not deployed.
 
-In a `portable` target, the package entries may have an additional `subtargets` section detailing RID-specific assets. The `corehost` application should use this data, along with the current RID and the fallback data defined in the `runtimes` section to select one **and only one** "subtarget" out of each package individually. The most specific subtarget should always be selected. In practice, this means selecting the first RID shown on the appropriate line in `runtimes`. For example, given a package containing the following subtargets:
+In the portable deployment model, the `*.runtimeConfig.json` file will contain the `targetRuntime` section:
+
+```json
+{
+    "targetRuntime": {
```
richlander: I'm curious about the 'targetRuntime' term. I also see 'runtimeTarget(s)' below.  Are those supposed to be the same or intentionally different.

I was also wondering how about the difference between our use of 'runtime' and 'framework' in various places. I'm guessing that we're using runtime here since it is more RID-oriented and we'll use 'framework' when it is more surface area oriented?

anurse: Yes, we moved the setting to `runtimeOptions.framework` because of exactly that confusion. I'm updating everything now, but the example at the very top is up-to-date.

PR #1638, opened 2016-03-01
Update to latest NuGet and add manual RID inference to preserve current behavior

anurse:
The newest NuGet disabled RID inference (as we requested). Unfortunately, we don't actually have support for Portable App publishing, so we still need RID inference :). So this introduces a temporary new switch (`--infer-runtime`) to re-enable RID inference. This is used in dotnet cli and will be used by ASP.NET Core to ensure we can keep advancing to new NuGet versions without breaking our builds.

/cc @pakrym @piotrpMSFT @brthor @Sridhar-MS @davidfowl @muratg

### DIFF
--- a/src/dotnet/project.json
+++ b/src/dotnet/project.json
@@ -24,7 +24,7 @@
         "Microsoft.CodeAnalysis.CSharp":  "1.3.0-beta1-20160225-02",
         "Microsoft.DiaSymReader.Native": "1.3.3",
 
-        "NuGet.CommandLine.XPlat": "3.4.0-beta-632",
+        "NuGet.CommandLine.XPlat": "3.4.0-beta-680",
         "System.CommandLine": "0.1.0-e160119-1",
 
         "Microsoft.DotNet.ProjectModel": "1.0.0-*",
--- a/src/dotnet/commands/dotnet-restore/Program.cs
+++ b/src/dotnet/commands/dotnet-restore/Program.cs
@@ -36,6 +36,17 @@ namespace Microsoft.DotNet.Tools.Restore
             var quiet = args.Any(s => s.Equals("--quiet", StringComparison.OrdinalIgnoreCase));
             args = args.Where(s => !s.Equals("--quiet", StringComparison.OrdinalIgnoreCase)).ToArray();
 
+            // NuGet disabled RID inference, but we need it back for now, while we implement the rest of the portable app stuff.
+            // Re-enabling it like this won't mess with our ability to support the other scenarios, it just means we're restoring more than we need for now.
+            // This **will** mean we are ignoring the RIDs provided in the project.json, if any, though.
+            if(!args.Contains("--runtime", StringComparer.OrdinalIgnoreCase))
+            {
+                args = Enumerable.Concat(
+                        args,
+                        PlatformServices.Default.Runtime.GetOverrideRestoreRuntimeIdentifiers().SelectMany(s => new [] { "--runtime", s })
+                    ).ToArray();
+            }
+
             app.OnExecute(() =>
             {
                 try


scripts/dotnet-cli-build/Utils/DotNetCli.cs
```diff
@@ -42,5 +51,19 @@ private static string GetStage0Path()
                 return Path.Combine(Directory.GetCurrentDirectory(), ".dotnet_stage0", PlatformServices.Default.Runtime.OperatingSystemPlatform.ToString(), "share", "dotnet", "cli", "bin");
             }
         }
+
+        // Mega-hack because we can't take a reference on a project within the build script
+        internal static IEnumerable<string> GetOverrideRestoreRuntimeIdentifiers(IRuntimeEnvironment env)
```
anurse: This will be removed in #1639 very shortly after merging this PR (once it hits stage 0)

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/RestoreCommand.cs
```diff
@@ -15,7 +17,9 @@ public RestoreCommand()
 
         public override CommandResult Execute(string args="")
         {
-            args = $"restore {args}";
+            // Temporary until portable app publishing is working
+            var ridArgs = PlatformServices.Default.Runtime.GetOverrideRestoreRuntimeIdentifiers().SelectMany(s => new[] { "--runtime", s });
```
anurse: This will change to `--infer-runtime` once this PR hits stage 0

muratg:
@victorhurdugaci I think we'll need to get this in KoreBuild as well. 

cc @BrennanConroy

anurse:
@dotnet-bot test OSX Debug Build please
@dotnet-bot test OSX Release Build please

anurse:
@victorhurdugaci suggested that since we'd have to change SO much to compensate for a new command line switch, we should just turn RID inference back on for dotnet-restore. I've done that via reviving a hack we had a while ago :). We'll remove it once we have the portable app model working.

anurse:

anurse:
@dotnet-bot test OSX Debug Build please

davidfowl:
Be problem with turning it on by default is guard rails for things like netstandard. Did nuget also remove the hack for "compile time" tfms?

emgarten:
> Did nuget also remove the hack for "compile time" tfms?

that was removed, whatever runtimes are in the project.json file will be used now, regardless of tfm

anurse:
Well, we either need to turn it on by default, add `runtimes` to every project that needs to be published, restore the compile-time TFM hack, or just wait until we have full portable app support before updating NuGet (though unfortunately we need to take a new version of NuGet to get the restore changes necessary to handle portable apps...)

anurse:
I suppose one option would be to just revert the change in NuGet that removed RID inference for now. Thoughts on that @davidfowl @emgarten ? It's still the change we want, but maybe we just did things in the wrong order :).

anurse:
I think we're just going to see if we can walk back the RID inference changes. It's the easiest solution to get new NuGet versions even though we're not quite ready to do portable publishing.

src/dotnet/project.json
```diff
@@ -24,7 +24,7 @@
         "Microsoft.CodeAnalysis.CSharp":  "1.3.0-beta1-20160225-02",
         "Microsoft.DiaSymReader.Native": "1.3.3",
 
-        "NuGet.CommandLine.XPlat": "3.4.0-beta-632",
+        "NuGet.CommandLine.XPlat": "3.4.0-beta-680",
```
eerhardt: There is a reference to NuGet here as well: https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/project.json#L9

anurse: :+1: thanks!

anurse:
No longer relevant. Will open a new one to update to latest NuGet

PR #1645, opened 2016-03-01
Move to CoreFX build rc2-23901

eerhardt:
@piotrpMSFT @anurse @livarcocc

### DIFF
--- a/TestAssets/FSharpTestProjects/CompileFail/project.json
+++ b/TestAssets/FSharpTestProjects/CompileFail/project.json
@@ -9,12 +9,12 @@
     ],
     "dependencies": {
         "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-151221",
-        "NETStandard.Library": "1.0.0-rc2-23811"
+        "NETStandard.Library": "1.0.0-rc2-23901"
     },
 
     "frameworks": {
-        "dnxcore50": {
-            "imports": "portable-net45+win8"
+        "netstandardapp1.5": {
+            "imports": "dnxcore50"
         }
     }
 }
--- a/TestAssets/FSharpTestProjects/TestApp/project.json
+++ b/TestAssets/FSharpTestProjects/TestApp/project.json
@@ -9,13 +9,13 @@
     ],
     "dependencies": {
         "TestLibrary": "1.0.0-*",
-        "NETStandard.Library": "1.0.0-rc2-23811",
-        "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-151221",
+        "NETStandard.Library": "1.0.0-rc2-23901",
+        "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-151221"
     },
 
     "frameworks": {
-        "dnxcore50": {
-            "imports": "portable-net45+win8"
+        "netstandardapp1.5": {
+            "imports": "dnxcore50"
         }
     }
 }
--- a/TestAssets/FSharpTestProjects/TestAppWithArgs/project.json
+++ b/TestAssets/FSharpTestProjects/TestAppWithArgs/project.json
@@ -9,12 +9,12 @@
     ],
     "dependencies": {
         "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-151221",
-        "NETStandard.Library": "1.0.0-rc2-23811"
+        "NETStandard.Library": "1.0.0-rc2-23901"
     },
 
     "frameworks": {
-        "dnxcore50": {
-            "imports": "portable-net45+win8"
+        "netstandardapp1.5": {
+            "imports": "dnxcore50"
         }
     }
 }
--- a/TestAssets/FSharpTestProjects/TestLibrary/project.json
+++ b/TestAssets/FSharpTestProjects/TestLibrary/project.json
@@ -2,7 +2,7 @@
     "version": "1.0.0-*",
     "dependencies": {
         "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-151221",
-        "NETStandard.Library": "1.0.0-rc2-23811"
+        "NETStandard.Library": "1.0.0-rc2-23901"
     },
     "compilerName": "fsc",
     "compileFiles": [
@@ -10,8 +10,8 @@
         "Helper.fs"
     ],
     "frameworks": {
-        "dnxcore50": {
-            "imports": "portable-net45+win8"
+        "netstandardapp1.5": {
+            "imports": "dnxcore50"
         }
     }
 }
--- a/TestAssets/ProjectModelServer/DthTestProjects/src/BrokenProjectPathSample/project.json
+++ b/TestAssets/ProjectModelServer/DthTestProjects/src/BrokenProjectPathSample/project.json
@@ -1,9 +1,11 @@
 {
     "dependencies": {
-        "NETStandard.Library": "1.0.0-rc2-23811",
+        "NETStandard.Library": "1.0.0-rc2-23901",
         "EmptyLibrary": "1.0.0-*"
     },
     "frameworks": {
-        "dnxcore50": { }
+        "netstandardapp1.5": {
+            "imports": "dnxcore50"
+        }
     }
 }
--- a/TestAssets/ProjectModelServer/DthTestProjects/src/EmptyConsoleApp/project.json
+++ b/TestAssets/ProjectModelServer/DthTestProjects/src/EmptyConsoleApp/project.json
@@ -1,9 +1,10 @@
 {
   "dependencies": { },
   "frameworks": {
-    "dnxcore50": {
+    "netstandardapp1.5": {
+      "imports": "dnxcore50",
       "dependencies": {
-        "NETStandard.Library": "1.0.0-rc2-23811"
+        "NETStandard.Library": "1.0.0-rc2-23901"
       }
     },
     "dnx451": { }
--- a/TestAssets/ProjectModelServer/DthTestProjects/src/EmptyLibrary/project.json
+++ b/TestAssets/ProjectModelServer/DthTestProjects/src/EmptyLibrary/project.json
@@ -2,9 +2,10 @@
   "version": "1.0.0-*",
   "dependencies": { },
   "frameworks": {
-    "dnxcore50": {
-      "dependencies":{
-        "NETStandard.Library": "1.0.0-rc2-23811"
+    "netstandard1.3": {
+      "imports": "dnxcore50",
+      "dependencies": {
+        "NETStandard.Library": "1.0.0-rc2-23901"
       }
     }
   }
--- a/TestAssets/ProjectModelServer/DthTestProjects/src/FailReleaseProject/project.json
+++ b/TestAssets/ProjectModelServer/DthTestProjects/src/FailReleaseProject/project.json
@@ -1,8 +1,9 @@
 {
   "frameworks": {
-    "dnxcore50": {
+    "netstandardapp1.5": {
+      "imports": "dnxcore50",
       "dependencies": {
-        "NETStandard.Library": "1.0.0-rc2-23811"
+        "NETStandard.Library": "1.0.0-rc2-23901"
       }
     }
   },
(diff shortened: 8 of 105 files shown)


piotrpMSFT:
@ericstj this update should require new tfm, no?

ericstj:
@piotrpMSFT yes.  You need to start using `netstandard` instead of `dotnet`.  If you still have packages in the graph that use `dotnet` you can add an import statement for that.

anurse:
Yeah, that'd be whats causing all the `System.IO.FileSystem 4.0.1-rc2-23901 is not compatible with .NETPlatform,Version=v5.4.` errors.

We need to basically change every occurrence of `dotnetX.Y` in the `frameworks` section to `netstandard(X-4).(Y-1)`, and add `"imports": [ "dotnetX.Y" ]"`. See https://github.com/aspnet/Mvc/pull/4170/files for some examples.

piotrpMSFT:
We need to take this opportunity to also purge dnxcore50.

anurse:
Yeah, we should be able to simply replace `dnxcore50` with `netstandardapp1.5` now. We'll need one more change once we finalize the name for `netstandardapp`.

ericstj:
> simply replace dnxcore50 with netstandardapp1.5 now
RC2 build doesn't yet have all the removals of DNXCore50 from the corefx packages so you may still need the import.  RC3-23901 doesn't need the import.  I believe that change should FI into KRel today.

eerhardt:
So I've changed the project.json in `Microsoft.DotNet.InternalAbstractions` to look like this:

```
{
  "description": "Abstractions for making code that uses file system and environment testable.",
  "version": "1.0.0-*",
  "repository": {
    "type": "git",
    "url": "git://github.com/dotnet/cli"
  },
  "compilationOptions": {
    "warningsAsErrors": true,
    "keyFile": "../../tools/Key.snk"
  },
  "dependencies": {
    ~"Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-16537"~
    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20019"
  },
  "frameworks": {
    "net451": { },
    ~"dotnet5.4": {~
      ~"dependencies": {~
        ~"System.Runtime": "4.0.21-rc2-23811"~
      ~}~
    ~},~
    "netstandardapp1.5": {
      "imports": [ "dotnet5.4" ],
      "dependencies": {
        "NETStandard.Library": "1.0.0-rc2-23901"
      }
    }
  },
  "scripts": {
  }
}
```

(Note: lines that start and end with `~` are the old lines and being removed.)

And I get the following output during restore:

```
warn : Detected package downgrade: System.Reflection from 4.1.0-rc2-23901 to 4.1.0-rc2-23819
warn :  Microsoft.DotNet.InternalAbstractions (>= 1.0.0) -> Microsoft.Extensions.PlatformAbstractions (>= 1.0.0-rc2-20019) -> System.Resources.ResourceManager (>= 4.0.1-rc2-23819) -> System.Reflection (>= 4.1.0-rc2-23901)

error: System.Resources.ResourceManager 4.0.1-rc2-23901 provides a compile-time reference assembly for System.Resources.ResourceManager on .NETStandardApp,Version=v1.5, but there is no run-time assembly compatible with win7-x64.
error: System.Reflection 4.1.0-rc2-23901 provides a compile-time reference assembly for System.Reflection on .NETStandardApp,Version=v1.5, but there is no run-time assembly compatible with win7-x64.
```

eerhardt:
But then I change it to:

```
  "frameworks": {
    "net451": { },
    "netstandardapp1.5": {
      "imports": [ "dotnet5.4", "dnxcore50" ],
      "dependencies": {
        "NETStandard.Library": "1.0.0-rc2-23901"
      }
    }
```

And I only get the warning output:

```
warn : Detected package downgrade: System.Reflection from 4.1.0-rc2-23901 to 4.1.0-rc2-23819
warn :  Microsoft.DotNet.InternalAbstractions (>= 1.0.0) -> Microsoft.Extensions.PlatformAbstractions (>= 1.0.0-rc2-20019) -> System.Resources.ResourceManager (>= 4.0.1-rc2-23819) -> System.Reflection (>= 4.1.0-rc2-23901)
warn :  Microsoft.DotNet.InternalAbstractions (>= 1.0.0) -> Microsoft.Extensions.PlatformAbstractions (>= 1.0.0-rc2-20019) -> System.Reflection (>= 4.1.0-rc2-23819)
```

No errors.

Can I just ignore that warning?

ericstj:
You don't need to import dotnet and dnxcore.  DNXCore supports dotnet.  You can ignore the warning for now.  That seems like it may be a nuget issue, but we'll be eliminating the entries in runtime.json that are causing this.

eerhardt:
The next place I'm running into an issue is in our test projects.  Here is our EndToEnd project.json:

```
{
  "version": "1.0.0-*",
  "compilationOptions": {
    "emitEntryPoint": true
  },

  "dependencies": {
    "NETStandard.Library": "1.0.0-rc2-23901",

    "Microsoft.DotNet.ProjectModel": { "target": "project" },
    "Microsoft.DotNet.Cli.Utils": { "target": "project" },
    "Microsoft.DotNet.Tools.Tests.Utilities": { "target": "project" },

    "xunit": "2.1.0",
    "xunit.netcore.extensions": "1.0.0-prerelease-*",
    "dotnet-test-xunit": "1.0.0-dev-48273-16"
  },

  "frameworks": {
    "netstandardapp1.5": {
      "imports": [
        "dnxcore50",
        "portable-net451+win8"
      ]
    }
  },

  "testRunner": "xunit"
}
```

And when running `dotnet restore` on this, I'm getting the following error output:

```
Errors in C:\cli\test\EndToEnd\project.json
    System.IO.Compression 4.1.0-rc2-23901 provides a compile-time reference assembly for System.IO.Compression on .NETStandardApp,Version=v1.5, but there is no run-time assembly compatible with win7-x64.
    Some packages are not compatible with .NETStandardApp,Version=v1.5 (win7-x64).
    System.IO.Compression 4.1.0-rc2-23901 provides a compile-time reference assembly for System.IO.Compression on .NETStandardApp,Version=v1.5, but there is no run-time assembly compatible with win7-x86.
    Some packages are not compatible with .NETStandardApp,Version=v1.5 (win7-x86)
```


A crazy thing is, if I change `netstandardapp1.5` to `netstandard1.3`, the restore works, but then build fails because this has a project-to-project reference to Microsoft.DotNet.ProjectModel which uses `netstandardapp1.5`.

ericstj:
This is happening because something in the compile graph is bringing in the old version of System.IO.Compression that contained a runtime.json.  That System.IO.Compression's runtime.json has more specific RIDs and overrides the version from our lineup.  You can workaround it by either a) upgrading the project that uses System.IO.Compression to use the latest or b) upgrading some parent project to a if you don't happen to build the a project in your repo.

eerhardt:
@stephentoub @ericeil  - do you guys know any reason for the test failures on *nix?

> System.EntryPointNotFoundException : Unable to find an entry point named 'GetHostName' in DLL 'System.Native'.
+++++++++++++++++++
STACK TRACE:
at Interop.Sys.GetHostName(Byte* name, Int32 nameLength) at Interop.Sys.GetHostName() at System.Net.Dns.GetHostName() at System.Net.Sockets.Socket.InitializeSockets() at System.Net.Sockets.Socket..ctor(AddressFamily addressFamily, SocketType socketType, ProtocolType protocolType) at Microsoft.DotNet.ProjectModel.Server.Tests.DthTestServer.FindFreePort() at Microsoft.DotNet.ProjectModel.Server.Tests.DthTestServer..ctor(ILoggerFactory loggerFactory) at Microsoft.DotNet.ProjectModel.Server.Tests.DthTests.DthStartup_ProtocolNegotiation(Int32 requestVersion, Int32 expectVersion)

Have there been changes in this area in the last week?

stephentoub:
That method was renamed to SystemNative_GetHostName at the beginning of January... I think by you :smile:
https://github.com/dotnet/corefx/blame/18652a64e02047b46a83a354a7f99cfecf5c9063/src/Native/System.Native/pal_networking.cpp#L905

It seems like you've got an old System.Net.Sockets.dll trying to use a newer System.Native.so.

eerhardt:
That's ood, looking in my .deps file, I see I am getting System.Net.Sockets out of the 4.1.0-rc2-23901 package:

> "Package","runtime.unix.System.Net.Sockets","4.1.0-rc2-23901","sha512-lYYNIKVScW6OUHJX/wgxVaW+jO5UDETTlJo6JcGDuc8I/KJWnC7NiMgN8gycW377DCYMxi2JgSYELEbj80k3qw==","runtime","System.Net.Sockets","runtimes/unix/lib/netstandard1.3/System.Net.Sockets.dll"

It doesn't appear to be an old System.Net.Sockets.dll

ericstj:
@ellismg maybe a missing change in rel?

ellismg:
> @ellismg maybe a missing change in rel?

I don't think so.  It looks like the call is coming from `System.Net.NameResolution`.  I looked at `runtime.unix.System.Net.NameResolution.4.0.0-rc2-23901` and it had the correct import:

```
.method private hidebysig static pinvokeimpl("System.Native" as "SystemNative_GetHostName" lasterr winapi) 
        int32  GetHostName(uint8* name,
                           int32 nameLength) cil managed preservesig
{
}
```

Does the lock file show you are pulling in that version of the runtime package for System.Net.NameResolution?

eerhardt:
Well, that was the problem.  Thanks for the tip, @ellismg.

Looking in my deps file, I see:

> "Package","runtime.linux.System.Net.NameResolution","4.0.0-beta-23516","sha512-yLGF09g7bW5cu8EsX94iZZ+i53mriQ9aSQ/w+fedDmDiip0CJp8uJoUpmjO5mf5i4N83m6G+IzAKdgoeUH3zxA==","runtime","System.Net.NameResolution","runtimes/unix/lib/dnxcore50/System.Net.NameResolution.dll"

eerhardt:
@ericstj - Is this issue is in the same family as the DataContractSerialization issue I was seeing earlier?  I'm guessing not because this repros on rc2-23903 as well.

I can repro this issue in a stand-alone app on ubuntu:

```C#
using System;
using System.Net.Sockets;

namespace ConsoleApplication
{
    public class Program
    {
        {
            var s = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
            Console.WriteLine(s);
        }
    }
}
```

with project.json:

```JSON
{
    "version": "1.0.0-*",
    "compilationOptions": {
        "emitEntryPoint": true
    },
    "dependencies": {
        "NETStandard.Library": "1.0.0-rc2-23903",
        "Microsoft.Net.Compilers.netcore": "1.3.0-beta1-20160225-02"
    },
    "frameworks": {
        "netstandardapp1.5": {
            "imports": [ "dnxcore50", "portable-net45+win8" ]
        }
    }
} 
```

When I `dotnet restore` and `dotnet run` that app:

```
Unhandled Exception: System.EntryPointNotFoundException: Unable to find an entry point named 'GetHostName' in DLL 'System.Native'.
   at Interop.Sys.GetHostName(Byte* name, Int32 nameLength)
   at Interop.Sys.GetHostName()
   at System.Net.Dns.GetHostName()
   at System.Net.Sockets.Socket.InitializeSockets()
   at System.Net.Sockets.Socket..ctor(AddressFamily addressFamily, SocketType socketType, ProtocolType protocolType)
   at ConsoleApplication.Program.Main(String[] args)
```

ericstj:
Can you share the lock file?

ericstj:
Yeah, similar problem to IO.Compression: https://github.com/dotnet/corefx/issues/6603

In this case Microsoft.Net.Compilers.netcore is bringing in "System.Net.NameResolution": "4.0.0-beta-23504" to the compile graph.  This is the highest version in the compile graph.  This contains RID mapping for win\osx\linux.  The latest lineup only uses win\unix for this package.  Since linux is more specific that wins and the old implementation is brought in.  Since the new shim is not compatible with the old implementation this fails at runtime.

The workaround here is the same as IO.Compression.  Ensure the latest version of the package is referenced in the compile graph.

ericstj:
I see the same problem for System.Security.Cryptography.Algorithms in this project.

eerhardt:
Alright, I added a temporary workaround to the test - explicitly referencing the latest version. This time it should be green.

@anurse @piotrpMSFT @livarcocc - Can you guys take a look again at my changes now that I have the TFM changes in?

eerhardt:
> I see the same problem for System.Security.Cryptography.Algorithms in this project.

This isn't an issue for this test because it doesn't use Crypto at runtime. So I've opted not to add a "workaround" for it.

Issue #1656, opened 2016-03-02
Commands are bleeding bold


bricelam:
The compile command (and possibly others) are leaving my console foreground color bolded after exiting.

![image](https://cloud.githubusercontent.com/assets/1226749/13471355/13224b8e-e065-11e5-8980-c52cec66b820.png)

Eilon:
That is a **bold** accusation.

*mic drop*

Peter-Schneider:
I think the problem is in this line: https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/AnsiConsole.cs#L41 - if _boldRecursion > 0 then the color should be light.

brthor:
the opposite thing is happening here. When your original foreground color is already bolded, it is set to light because `boldRecursion < 1`

@Peter-Schneider that is where the issue is occurring but I don't believe that is the correct fix. We should skip the SetColor call when going back to the OriginalForegroundColor altogether.

Peter-Schneider:
Thanks @brthor for your feedback! If you take a look at line https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-compile/ManagedCompiler.cs#L44 you can see that the color is set to Yellow. In the current implementation the value for _boldRecursion is not initialized - therefore the color is highlighted in this line  https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/AnsiConsole.cs#L41 and also when setting back to the OriginalForegroundColor the color is incorrectly highlighted. 

Bold should mean high-intensity color, where as not bold should use the dark colors - therefore my suggestion to light the color. 
So in the above sample the first yellow string should be DarkYellow as only Yellow() is used and not .Bold().Yellow(). What do you think?

brthor:
Asking around it seems folks are of the opinion that the bold should be the lighter color. 

It is a separate issue though, and we shouldn't mix that change with the solution to bleeding the colors after the command exits.

Can you open up a new issue to discuss that?

PR #1660, opened 2016-03-02
Fix for issue #1656

Peter-Schneider:


brthor:
If bold recursion is > 0, we are inside a bold tag and the color should be dark, why change it this way?

brthor:
@Peter-Schneider this does not fix the issue of bold bleeding like in #1656 but I still think it's a valuable fix.

Please open an issue for the bold colors being darker and let's get this merged :smile:

Peter-Schneider:
@brthor I've submitted issue #1686 :wink:

brthor:
:shipit:

piotrpMSFT:
Thanks for the contribution, @Peter-Schneider

Peter-Schneider:
:+1: Thanks for merging, @piotrpMSFT - I think issue #1686 can be closed then...

PR #1667, opened 2016-03-02
Add support for building dotnet CLI on RHEL.

Sridhar-MS:


davidfowl:
Looks pretty simple

test/EndToEnd/EndToEndTest.cs
```diff
@@ -244,7 +244,8 @@ private bool IsCentOS()
 
                 if(File.Exists(OSIDFILE))
```
anurse: Could this not use `PlatformServices.Default.Runtime.OperatingSystem`? That property value is derived from the same file.

anurse:
Minor comment, and needs rebasing, other than those: :shipit:

Sridhar-MS:
@dotnet-bot test OSX Debug Build please.

test/EndToEnd/EndToEndTest.cs
Sridhar-MS: Fixed.

PR #1678, opened 2016-03-03
Fix host case comparison

### LINKED ISSUE #1600: Could not load file or assembly 'dotnet.EXE.dll'
I downloaded net cli binaries and executed dotnet restore within folder which contains asp.net core project. Some packages were restored and several errors about incompativlity were displayed.
Then I tried execute dotnet new within empty directory and got error
`Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'dotnet.EXE.dll' or one of its dependencies.`
I tried to uninstall and reinstall (and also repair) but error still occurs.

version 1.0.0.001540

schellap:
Fixes #1600

### DIFF
--- a/src/corehost/common/pal.h
+++ b/src/corehost/common/pal.h
@@ -95,6 +95,9 @@ namespace pal
 
     inline int strcmp(const char_t* str1, const char_t* str2) { return ::wcscmp(str1, str2); }
     inline int strcasecmp(const char_t* str1, const char_t* str2) { return ::_wcsicmp(str1, str2); }
+    inline int strncmp(const char_t* str1, const char_t* str2, int len) { return ::wcsncmp(str1, str2, len); }
+    inline int strncasecmp(const char_t* str1, const char_t* str2, int len) { return ::_wcsnicmp(str1, str2, len); }
+
     inline size_t strlen(const char_t* str) { return ::wcslen(str); }
     inline void err_vprintf(const char_t* format, va_list vl) { ::vfwprintf(stderr, format, vl); ::fputws(_X("\r\n"), stderr); }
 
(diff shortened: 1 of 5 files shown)


davidfowl:
Add a test?

schellap:
Filed an issue for RTM to get back to it.

schellap:
@piotrpMSFT or @gkhanna79 PTAL.

davidfowl:
Why not add a test now?

schellap:
Right now, I don't want to spend time learning how cli test infrastructure works as there are other higher pri bugs/work items on my plate. If an accompanying test helps completeness, I will close this PR and come back to it later, but having a tracking bug does the same...

src/corehost/common/pal.h
```diff
@@ -95,6 +95,9 @@ namespace pal
 
     inline int strcmp(const char_t* str1, const char_t* str2) { return ::wcscmp(str1, str2); }
     inline int strcasecmp(const char_t* str1, const char_t* str2) { return ::_wcsicmp(str1, str2); }
+    inline int strncmp(const char_t* str1, const char_t* str2, int len) { return ::wcsncmp(str1, str2, len); }
+    inline int strncasecmp(const char_t* str1, const char_t* str2, int len) { return ::_wcsnicmp(str1, str2, len); }
```
gkhanna79: We are doing case-insensitive comparison here but not in the !Win32 case?

gkhanna79:
LGTM other than one comment.

src/corehost/common/pal.h
schellap: `strcasecmp` is a funny name but used by unix, Ruby as `casecmp` and Python as `casefold`

schellap:
@davidfowl, merging... but I will definitely get to the issue #1682

PR #1701, opened 2016-03-04
Added InstallDir option for install.ps1.

hal-ler:
This change brings the "-d" functionality of install.sh into install.ps1. It allows installing dotnetcli from a build script without having to change the environment. This is useful when developing platform independent build scripts using Cake for example. The existing functionality is unaffected, the new option is only used if set explicitly.

### DIFF
--- a/scripts/obtain/install.ps1
+++ b/scripts/obtain/install.ps1
@@ -6,7 +6,8 @@
 param(
    [string]$Channel="dev",
    [string]$version="Latest",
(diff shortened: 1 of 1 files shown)


troydai:
LGTM 
/cc @piotrpMSFT @Sridhar-MS @pranavkm

hal-ler:
Closing it since the functionality was added to the script. Thank you guys!

PR #1703, opened 2016-03-04
deps.json pass 1

pakrym:
Does not have yet:
1. Support for resource assembly exports
2. Reading subtargets from lock.json and passing them to DependencyContext (Reader/Writesr support it)
3. Discovering host deps and merging deps files
4. Reading rid graph from lock.json (Reader/Writesr support it)

@davidfowl @anurse


src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
```diff
@@ -15,13 +15,46 @@ public class DependencyContext
...
         private static readonly Lazy<DependencyContext> _defaultContext = new Lazy<DependencyContext>(LoadDefault);
 
-        public DependencyContext(string target, string runtime, CompilationOptions compilationOptions, CompilationLibrary[] compileLibraries, RuntimeLibrary[] runtimeLibraries)
+        public DependencyContext(string target,
+            string runtime,
+            bool isPortable,
+            CompilationOptions compilationOptions,
+            CompilationLibrary[] compileLibraries,
+            RuntimeLibrary[] runtimeLibraries,
+            IReadOnlyList<KeyValuePair<string, string[]>> runtimeGraph)
```
anurse: I generally prefer taking in `IEnumerable` in these situations for the most flexibility. Personal preference mostly though, so if there are good reasons to leave it as `IReadOnlyList` that's OK too.

src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
```diff
@@ -15,13 +15,46 @@ public class DependencyContext
 
         private static readonly Lazy<DependencyContext> _defaultContext = new Lazy<DependencyContext>(LoadDefault);
 
-        public DependencyContext(string target, string runtime, CompilationOptions compilationOptions, CompilationLibrary[] compileLibraries, RuntimeLibrary[] runtimeLibraries)
+        public DependencyContext(string target,
+            string runtime,
+            bool isPortable,
+            CompilationOptions compilationOptions,
+            CompilationLibrary[] compileLibraries,
+            RuntimeLibrary[] runtimeLibraries,
```
anurse: `IEnumerable`?

src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
pakrym: Yeah, looks super inconsistent, will change everything to IEnumerable

src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -28,23 +28,92 @@ public DependencyContext Read(Stream stream)
 
         private DependencyContext Read(JObject root)
         {
+            string runtime = string.Empty;
+            string target = string.Empty;
```
anurse: `var`

src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -28,23 +28,92 @@ public DependencyContext Read(Stream stream)
...
-            var compileTargetProperty = targetsObject.First(target => !IsRuntimeTarget(target.Key));
+            JObject runtimeTarget = null;
+            JObject compileTarget = null;
+            if (targetsObject != null)
+            {
+                var compileTargetProperty = targetsObject.FirstOrDefault(t => !IsRuntimeTarget(t.Key));
+                compileTarget = (JObject) compileTargetProperty.Value;
+                target = compileTargetProperty.Key;
+
+                if (!string.IsNullOrEmpty(runtimeTargetInfo.Name))
```
anurse: I wonder if maybe we should put the compilation target name in as well just for completeness... that way nobody is parsing target names. Not sure if that might be overkill though.

src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -28,23 +28,92 @@ public DependencyContext Read(Stream stream)
...
+                {
+                    runtimeTarget = (JObject) targetsObject.FirstOrDefault(t => t.Key == runtimeTargetInfo.Name).Value;
```
anurse: Does `targetsObject[runtimeTargetInfo.Name]` not work?

src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -28,23 +28,92 @@ public DependencyContext Read(Stream stream)
...
+        private RuntimeTargetInfo ReadRuntimeTargetInfo(JObject root)
+        {
+
+            var runtimeTarget = (JObject)root[DependencyContextStrings.RuntimeTargetPropertyName];
+            if (runtimeTarget != null)
+            {
+                return new RuntimeTargetInfo()
+                {
+                    Name = runtimeTarget[DependencyContextStrings.RuntimeTargetNamePropertyName]?.Value<string>(),
+                    Portable = runtimeTarget[DependencyContextStrings.PortablePropertyName]?.Value<bool>() == true
```
anurse: This is now just a simple string property, see https://github.com/dotnet/cli/blob/anurse/update-config-spec/Documentation/specs/runtime-configuration-file.md#appnameruntimeconfigjson

That changed recently :). The host will use the presence of the `runtimeOptions.framework` section in the `.runtimeconfig.json` file to determine if the target is portable.

src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -28,23 +28,92 @@ public DependencyContext Read(Stream stream)
...
+        private IEnumerable<KeyValuePair<string, string[]>> ReadRuntimeGraph(JObject runtimes)
+        {
+            if (runtimes == null)
+            {
+                yield break;
+            }
+
+            var targets = runtimes.Children();
+            var runtime = (JProperty)targets.Single();
+            foreach (var pair in (JObject)runtime.Value)
```
anurse: `.ToDictionary`? Since a dictionary is `IEnumerable<KeyValuePair<K, V>>` you can just return that.

src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
```diff
@@ -75,11 +102,22 @@ private JObject WriteCompilationOptions(CompilationOptions compilationOptions)
             {
                 o[DependencyContextStrings.EmitEntryPointPropertyName] = compilationOptions.EmitEntryPoint;
             }
+            if (compilationOptions.GenerateXmlDocumentation != null)
```
anurse: These seem ripe for refactoring into a helper :). `AddProperty(jobj, name, value)`

src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
```diff
@@ -127,12 +199,75 @@ private JObject WriteTargetLibrary(Library library)
                 );
         }
 
-        private JObject WriteAssemblies(IReadOnlyList<string> assemblies)
+        private JObject WritePortableTargetLibrary(RuntimeLibrary runtimeLibrary, CompilationLibrary compilationLibrary)
```
anurse: I think it would be good to go back and look at refactoring `RuntimeLibrary` and `CompilationLibrary` later. It seems like there's quite a bit of duplication in behavior. Not something we need to deal with now though.

src/Microsoft.Extensions.DependencyModel/RuntimeLibrary.cs
```diff
@@ -8,12 +8,23 @@ namespace Microsoft.Extensions.DependencyModel
...
+            string hash,
+            RuntimeAssembly[] assemblies,
+            RuntimeTarget[] subTargets,
+            Dependency[] dependencies,
+            bool serviceable)
             : base(libraryType, packageName, version, hash, dependencies, serviceable)
         {
-            Assemblies = assemblies.Select(path => new RuntimeAssembly(path)).ToArray();
+            Assemblies = assemblies;
+            SubTargets = subTargets;
```
anurse: We call it `RuntimeTargets` elsewhere (to match the new name in lock.json), I assume this just needs to be updated to match.

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextBuilderTests.cs
```diff
@@ -0,0 +1,301 @@
...
+            context.CompilationOptions.WarningsAsErrors.Should().Be(true);
+
+            context.CompilationOptions.Defines.Should().BeEquivalentTo(new[] { "Define", "D" });
+            context.CompilationOptions.KeyFile.Should().Be("Key.snk");
+            context.CompilationOptions.LanguageVersion.Should().Be("C#8");
+            context.CompilationOptions.Platform.Should().Be("Platform");
+        }
+
+
+        private LibraryExport Export(
```
anurse: `private` stuff goes at the bottom

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextBuilderTests.cs
```diff
@@ -0,0 +1,301 @@
...
+                Enumerable.Empty<LibraryRange>(),
+                _defaultFramework,
+                true,
+                true);
+        }
+
+        [Fact]
+        public void FillsRuntimeAndTarget()
+        {
+            var context = Build(target: new NuGetFramework("SomeFramework",new Version(1,2)), runtime: "win8-32");
```
anurse: nit: I know it's just test data, but let's use a real RID: `win8-x86`

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs
```diff
@@ -0,0 +1,356 @@
...
+                .Subject.Should().BeOfType<JObject>().Subject;
+
+            rids.Should().HaveProperty("win7-x64")
+                .Subject.Should().BeOfType<JArray>()
+                .Which.Values<string>().ShouldBeEquivalentTo(new[] { "win6", "win5" });
+
+            rids.Should().HaveProperty("win8-x64")
+                .Subject.Should().BeOfType<JArray>()
+                .Which.Values<string>().ShouldBeEquivalentTo(new[] { "win7-x64" });
+        }
```
anurse: I think I'd like one or two tests that assert the actual JSON emitted. Since it's a little brittle, I'm fine with just one, but I'd like to see at least one test that asserts the actual outputted JSON string, for end-to-end verification.

pakrym: How would it help?

src/Microsoft.Extensions.DependencyModel/DependencyContext.cs
```diff
@@ -15,13 +15,46 @@ public class DependencyContext
...
+            if (runtimeLibraries == null)
+            {
+                throw new ArgumentNullException(nameof(runtimeLibraries));
+            }
+            if (runtimeGraph == null)
+            {
+                throw new ArgumentNullException(nameof(runtimeGraph));
+            }
+
             Target = target;
```
anurse: I know this isn't new, but this should be `TargetFramework`. I feel like `Target` is too generic.

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs
anurse: Right now, we're asserting that the generated JObject is structured correctly, which is, I agree, more effective in general. I think that having one end-to-end, `DependencyContext` goes in, string comes out, test would just make it easier to look at the tests and verify that the JSON is what we expect it to be.

It's easier (at least to me) to read the expected JSON string and verify it than to look at the JObjects/JArrays/JWhatevers :).

pakrym: Would you be ok with me doing it in pass 2? Don't want to fix one test a million times.

anurse: Yeah, that's fine with me

anurse:

PR #1704, opened 2016-03-04
Cleanup equality / hash code implementations

jaredpar:
A number of types in CLI had incorrectly implemented equality.  Typically by using object.GetHashCode for the hash value (reference based hashing) when value equality was desired.  The code today wasn't excercising the hashing semantics of the types hence we hadn't noticed.  Decided to clean it up anyways + add some tests.

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/AnalyzerOptions.cs
+++ b/src/Microsoft.DotNet.ProjectModel/AnalyzerOptions.cs
@@ -1,9 +1,10 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
 
 namespace Microsoft.DotNet.ProjectModel
 {
-    public class AnalyzerOptions
+    public sealed class AnalyzerOptions : IEquatable<AnalyzerOptions>
     {
         /// <summary>
         /// The identifier indicating the project language as defined by NuGet.
@@ -15,7 +16,7 @@ namespace Microsoft.DotNet.ProjectModel
 
         public static bool operator ==(AnalyzerOptions left, AnalyzerOptions right)
         {
-            return left.LanguageId == right.LanguageId;
+            return object.Equals(left, right);
         }
 
         public static bool operator !=(AnalyzerOptions left, AnalyzerOptions right)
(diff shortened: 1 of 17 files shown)


jaredpar:
CC @agocke

src/Shared/Hash.cs
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
```
agocke: Different license header. Did you steal this from Roslyn? ;)

jaredpar: Steal is such a negative word ... I pasted it from Roslyn.  :smile:

src/Microsoft.DotNet.ProjectModel/AnalyzerOptions.cs
```diff
@@ -15,7 +16,7 @@ public class AnalyzerOptions
 
         public static bool operator ==(AnalyzerOptions left, AnalyzerOptions right)
         {
-            return left.LanguageId == right.LanguageId;
```
agocke: Why reverse the call order from Equals->== to ==->Equals? Just convention?

jaredpar: It's just convention that is hard coded into my brain. I got burned going `Equals` to `==` so many times due to not being careful with `==` usage that I stopped doing it.  Don't have a hard preference here though.

agocke: Well, I'll nit pick and say that I think your way produces an unnecessary vcall, but I don't really care either way :)

agocke:
:+1: Aside from license

src/Shared/Hash.cs
anurse: We have other code using this combiner: https://github.com/aspnet/Common/tree/dev/src/Microsoft.Extensions.HashCodeCombiner.Sources

I don't mind which we use, but we should consolidate on one :). Doesn't have to happen in the PR, but might be worth tracking for clean-up later.

eerhardt:
@jaredpar - are you still interested in this PR?  If not, I'll close it out.

eerhardt:
@jaredpar - I'm going to close this out at EOD if I don't hear anything.

PR #1708, opened 2016-03-04
Update dotnet-build to produce portable layout

anurse:
dotnet-build will produce a deps file for portable builds, and will now
create "runnable" outputs for RID-less targets

the outputs won't actually be runnable today because we need corehost
changes and to generate a deps.json file for corehost to use.

Note: This **temporarily** adds a `--portable` switch to `dotnet build` in order to emulate a RID-inference-less NuGet when we want to test that.

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -45,6 +45,13 @@ namespace Microsoft.DotNet.Cli.Build
             "Microsoft.Extensions.Testing.Abstractions"
         };
 
+        // Updates the stage 2 with recent changes.
+        [Target(nameof(PrepareTargets.Init), nameof(CompileStage2))]
+        public static BuildTargetResult UpdateBuild(BuildTargetContext c)
+        {
+            return c.Success();
+        }
+
         [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(CompileStage1), nameof(CompileStage2))]
         public static BuildTargetResult Compile(BuildTargetContext c)
         {
--- a/src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
@@ -83,7 +83,7 @@ namespace Microsoft.DotNet.ProjectModel
                         .WithRuntimeIdentifiers(runtimeIdentifiers)
                         .Build();
         }
-        
+
         public static ProjectContextBuilder CreateBuilder(string projectPath, NuGetFramework framework)
         {
             if (projectPath.EndsWith(Project.FileName))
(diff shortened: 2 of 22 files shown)


src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs
```diff
@@ -130,29 +141,35 @@ public int Execute(OnExecute execute, string[] args)
...
                 {
-                    contexts = _frameworkOption.Values
-                        .Select(f =>
-                        {
-                            return ProjectContext.CreateBuilder(ProjectPathValue, NuGetFramework.Parse(f))
-                                                 .WithReaderSettings(settings)
-                                                 .Build();
-                        });
+                    // Project is missing 'frameworks' section
+                    Reporter.Error.WriteLine("Project does not have any frameworks listed in the 'frameworks' section.");
```
pakrym: Won't we show this error when project is not restored?

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -143,15 +140,20 @@ public static IEnumerable<ProjectContext> CreateContextForEachTarget(string proj
 
         public ProjectContext CreateRuntimeContext(IEnumerable<string> runtimeIdentifiers)
         {
-            var context = Create(ProjectFile.ProjectFilePath, TargetFramework, runtimeIdentifiers);
-            if (context.RuntimeIdentifier == null)
+            // Check if there are any runtime targets (i.e. are we portable)
```
pakrym: What if user targeted `netstandard` instead of `netstandardapp` will we display useful message?

src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs
anurse: Indeed. I'll remove it. I think I was just overdoing the error checks :)

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
anurse: If they do target `netstandard`, we'll publish just as we do for any other project. With portable apps/redists the TFM is no longer a pivot that defines a runnable application.

anurse:
it's green! yaaay

anurse:
Last call :). I'm rebasing and then merging when it goes green.

anurse:
\*sigh\* now both OSX Builds are failing with the NuGet ZIP corruption error. I'm going to merge without those once the others pass, the OSX VSO builds are unaffected.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -45,6 +45,13 @@ public class CompileTargets
             "Microsoft.Extensions.Testing.Abstractions"
         };
 
+        // Updates the stage 2 with recent changes.
+        [Target(nameof(PrepareTargets.Init), nameof(CompileStage2))]
+        public static BuildTargetResult UpdateBuild(BuildTargetContext c)
```
brthor: NICE :+1:

anurse: INORITE? So handy :)

anurse: It even skips stage1, which puts your artifacts folder in a weird state, but that's just fine for this :)

PR #1724, opened 2016-03-05
compile-fsc fix and features

enricosada:
Some fixes to build github.com/Microsoft/visualfsharp with .NET CLI

Support project.json compilationOptions:

- nowarn
- warningsAsErrors
- keyFile
- delaySign

Use env var to configure compile-fsc behaviour:

- DOTNET_FSC_PATH the fsc path. Default the bundled fsc.exe
- DOTNET_COMPILEFSC_USE_RESPONSE_FILE if '1' then pass a response file
  to fsc instead of all arguments. Default pass all arguments
- DOTNET_FSC_EXEC configure how to run the fsc. Values:
   - RUN run fsc passing fsc args
   - COREHOST run corehost passing fsc and fsc args. The default

Use the same order of fsproj msbuild task for fsc arguments to make it easier
compare fsproj build and .net cli build

Fix --resource with path and name

Enable --debug (-g) if pdb

The generated assembly info file must be in the last - 1 position
in source files list.

1. the generated assembly info file must be in the last possibile position to
   override the attributes
2. The last file is the source file with main

### DIFF
--- a/src/dotnet/commands/dotnet-compile-fsc/Program.cs
+++ b/src/dotnet/commands/dotnet-compile-fsc/Program.cs
@@ -252,7 +252,7 @@ namespace Microsoft.DotNet.Tools.Compiler.Fsc
         private static Command RunFsc(List<string> fscArgs)
         {
             var fscExe = Environment.GetEnvironmentVariable("DOTNET_FSC_PATH")
-                      ?? Path.Combine(AppContext.BaseDirectory, "fsc" + Constants.ExeSuffix);
+                      ?? Path.Combine(AppContext.BaseDirectory, "fsc.exe");
 
             var exec = Environment.GetEnvironmentVariable("DOTNET_FSC_EXEC")?.ToUpper() ?? "COREHOST";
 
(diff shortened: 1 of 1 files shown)


src/dotnet/commands/dotnet-compile-fsc/Program.cs
```diff
@@ -80,18 +80,13 @@ public static int Run(string[] args)
...
-            var allArgs = new List<string>(translated);
-            allArgs.AddRange(GetDefaultOptions());
-
-            // Generate assembly info
-            var assemblyInfo = Path.Combine(tempOutDir, $"dotnet-compile.assemblyinfo.fs");
-            File.WriteAllText(assemblyInfo, AssemblyInfoFileGenerator.GenerateFSharp(assemblyInfoOptions));
-            allArgs.Add($"{assemblyInfo}");
 
+            // TODO less hacky
             bool targetNetCore = commonOptions.Defines.Contains("DNXCORE50");
```
abrodersen: Can `netstandard` be included in this check until there is a better solution?

enricosada: added it, thx!

enricosada:
closing in favor of #1898

PR #1731, opened 2016-03-07
Bindings redirects fix

### LINKED ISSUE #865: Error on dotnet run

marono:
Added tests and assembly info caching as requested in the previous PR,
which I had to close due to CLA.

Fix #865

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1731)


davidfowl:
Why are there so many commits?

marono:
@davidfowl took me a while to figure out how to get tests to run only on windows machines.
Binding redirects are only generated for desktop runtimes which won't build on *nix.

The number of changed files is quite small, all in all.

cwe1ss:
Should this also fix #1860 ?

marono:
@cwe1ss Yes it does. While I haven't tested your particular scenario, the test projects at $/TestAssets/TestProjects/TestBindingRedirectGeneration2 reproduce a similar situation.

marono:
This seems to have been fixed as part of a different PR. Although I'm not completely convinced the process is deterministic in generating the version to redirect to, I don't have any test scenario to prove it.

PR #1737, opened 2016-03-07
Move CLI Packaging Scripts

ellismg:
We'll be producing other packages shortly (for the shared framework) so
move all the CLI specific package stuff into it's own folder under
packaging.


brthor:
`packaging/osx/scripts/postinstall → packaging/cli/osx/scripts/postinstall  100755 → 100644`

Intentional permissions change?

Thinking this may break the PKG

ellismg:
Not intentional, will update index.

ellismg:
This will interact poorly with https://github.com/dotnet/cli/pull/1739, so I will just do this once that has landed.

PR #1738, opened 2016-03-07
Add ability to specify `outputPath` when creating a `ProjectContext`s `AssemblyLoadContext`.

NTaylorMullen:
- Prior to this change `outputPath` would always default to `null` which didn't work well with `VisualStudio` tooling. The tooling enables the user to specify a project's `outputPath` but more importantly defaults the project's `outputPath` to the root directory (instead of the project directory).


anurse:
:shipit: once the builds finish.

Issue #1562, opened 2016-02-24
Produce deps.json file


anurse:
We've outgrown the `deps` file :). This has been coming for a while, but I wanted to codify the full design to make sure we know what we're doing. I chatted with @muratg and it sounds like we're going to put @pakrym on this.

Rather than a `.deps` text file, we should generate a `.deps.json` file containing all necessary metadata from the lock file. As much as possible, we want this file to look like we took the `project.lock.json` file, and simply removed unnecessary data, so the file should have a nearly identical format (with some fields simply removed).

**In the near-term, we will generate BOTH files since `corehost` has not yet been updated to support the file***

See the full spec here for details as to the format: https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/specs/runtime-configuration-file.md (**IMPORTANT**: I have updated this document in https://github.com/dotnet/cli/pull/1564, so use that version for reference right now)

This file will also replace the embedded `.deps.json` data that was used for runtime compilation. We will still only place compilation data into the external `.deps.json` file when `preserveCompilationContext` is true (feel free to change the name to something better ;P)

/cc @piotrpMSFT @pakrym @muratg @davidfowl

piotrpMSFT:
This workitem will break unit testing. `dotnet-test` invokes the test runner by passing the deps file to CoreHost using --depsfile. We need to update --depsfile to reflect the new format before/in concert with checking in this change.

/cc @livarcocc @brthor @gkhanna79 @schellap

piotrpMSFT:
Just chatted with @pakrym. When this is checked in we will be producing both new-style and old-style deps files. I will open issues to track the --depsfile requirement along with tracking the removal of the old-style deps file.

PR #1741, opened 2016-03-08
deps.json pass 2

pakrym:
Add resource assemblies and runtime targets to dependency context builder.


schellap:
@pakrym a quick question, does this fix issue #1562, I am at a point in host implementation to read this file and do some testing...

anurse:
It looks like this phase doesn't yet write the "runtime targets" to the deps file. Am I correct in that observation? Is that the next phase?

pakrym:
@schellap @anurse yes, this doesn't write file yet

anurse:
:shipit:

brthor:
0 / 11 passing :frowning:

pakrym:
Everything was green, squash messed up ci.

anurse:
With the CI times we're seeing and given the amount of traffic, it's really hard to wait for the CI after rebasing since the change often needs to be rebased again when that finishes.

Issue #1728, opened 2016-03-06
corehost does not support cross-build


mylibero:
I'm trying to use CLI on ARM.
After a series of struggles, I could run cli on ARM. 
It crashes for errors from coreclr, though. (this is another issue that we are working on)

Soon, I will be sending pull-requests about my work.
One of issues is that corehost does not support cross-build.
Since CLI does not build on ARM yet without an ARM CLI version, I had to cross-build on x64.
Alternatively, I could do native-build corehost on ARM and copy those back to x64, 
but I found it easier if corehost supports cross-build.

I changed 'src/corehost/build.sh' to take an architecture option. (x64 by default)
Any other idea? suggestion?

mylibero:
corehost is too simple program to have a proper cross-build script.
I'm thinking of taking a cross compiler as an argument along with 'cross' option, instead of the architecture option.

mylibero:
@Sridhar-MS This issue may be closed now. https://github.com/dotnet/cli/pull/1747 has been merged.

shootdaj:
@mylibero Can you tell me how I can use the dotnet CLI to build/publish a .NET Core RC2 application to Raspberry Pi with Windows IoT? I read that you have to actually build the CLI on the ARM platform itself for the output to be compatible with ARM. It seems that you were able to cross-build it on the x64 platform. How were you able to do that? And is the output dotnet.exe runnable on a Raspberry Pi? I'm a little confused. Bottom line is, I want to be able to run an ASP.NET Core RC2 application on Raspberry Pi running Windows 10 IoT. Thank you.

mylibero:
@shootdaj dotnet.exe is just a rename of corehost. Note that corehost is now moved to core-setup.
Compiling corehost for ARM is possible now. Other files except dotnet.exe are platform independent.
So, you can run dotnet-cli on ARM, but I doubt if it would build your project for ARM yet.
Alternatively, if your project does not have native code, you can just build it on x64 and run it on RPi.

PR #1747, opened 2016-03-08
Support cross-build in corehost

mylibero:
Add ability to specify `cross` and `C++ Cross Compiler` when building corehost for ARM.

Signed-off-by: Dongyun Jin <dongyun.jin@samsung.com>


mylibero:
This PR is based on this issue https://github.com/dotnet/cli/issues/1728

Sridhar-MS:
cc @schellap @gkhanna79

schellap:
Please look at `scripts\dotnet-cli-build\CompileTargets.cs` and search for build.sh to wire it in through the CLI build process. You might have a bootstrapping problem here to obtain a stage0 CLI.

@piotrpMSFT, do you build a stage0 CLI by hand first?

mylibero:
You're right. Definitely, there is a bootstrapping problem.
Here's our plan.

1) Modify corehost to be built for ARM. **[This PR]**
2) Include the corehost binary for ARM in the release package (the one used as a stage0 CLI).
     I'm suggesting `bin/corehost.arm`.
3) Add the ARM architecture option to `dotnet publish`, `dotnet build` (currently x64 only), and `dotnet compile-native`. 
4) Add a cross compile option in the CLI build scripts.
    Here, we should modify `scripts\dotnet-cli-build\CompileTargets.cs` as you mentioned, and other scripts as well.

Sridhar-MS:
@schellap Yes. We do build stage0 manually just for once.

@mylibero The steps you mentioned are mostly correct. A few suggestion
- you can add a `--xcompile arm` option to the build scripts.
- In build scripts we 3 stages
   - stage0 download from previous successful builds.
   - stage1 built using stage0.
   - stage2 built using stage1. After a successful test pass stage2 is uploaded (to be consumed as stage0 for the next build).
  
   But for cross-compile you can only do stage0 (an x64 version of CLI) and build stage1 (the arm version of CLI) using stage0.
- You need not add any new option to `dotnet publish` and `dotnet build`. You must be using runtime   identifiers like `ubuntu.14.04-arm` to publish, build for cross-plat.

manu-silicon:
I am also interested in having dotnet-cli support for arm/arm64 on Linux. How can we help in achieving this goal as we do not have control over the restore phase to get stage0?

mylibero:
@Sridhar-MS Agreed. To add the runtime identifier 'ubuntu.14.04-arm', what should we do? 
So far, I figured out that these files should be modified
`scripts/dotnet-cli-build/CompileTargets.cs`
`src/Microsoft.DotNet.Cli.Utils/Constants.cs`
Is there any other component that I should modify?
Also, it seems that a json lock file describes runtime identifiers, but I don't know which one does.
I'll investigate on this, but if you can give us a guide, it would be a big help.

@manu-silicon Good to hear that :) 
How about adding `--xcompile arm` to the build scripts. 
You should modify `./build.sh` and all sub scripts from it.
BTW, it will take some time for us to make cli work on ARM. 
First of all, CoreCLR does not work on ARM properly yet.
A good news is that we have other guys working on CoreCLR for ARM.

myungjoo:
@Sridhar-MS The projects.lock.json file downloaded with Nuget locks us out from any other architectures than the currently used. Therefore, unless we manually create projects.lock.json (almost impossible) for whateverdistro-arm, it is impossible to do any sort of cross-building via dotnet CLI.

Therefore, we need a mechanism somehow to tell dotnet CLI to get projects.lock.json not based on the current platform/architecture, but based on the given "runtime identifier". However, because normally, it is natural to assume that the target platform/arch is the same as the host platform/arch, without any additional options, dotnet CLI may work as it is, which in turn, requires us to add another option to build/publish to notify dotnet CLI that the target arch/platform is going to be different from the host. 

Note that that's how coreclr build mechanism is doing (there is "cross" option in build to notify the build system not to assume that that the target arch is going to be same as the host arch).

mylibero:
@myungjoo I modified NuGet.Client and generated project.lock.json with arm runtime identifier.
I'm sending a pull request to NuGet.

mylibero:
@Sridhar-MS I sent a pull request to NuGet.Client: https://github.com/NuGet/NuGet.Client/pull/398
I checked that CLI can recognize the arm runtime identifier from this patch.
Should I work on this PR and branch? 
Since building corehost for ARM is a separate implementation from the build option and the runtime identifier thing, can you ship this PR and let me send another PR?

@manu-silicon Since I don't hear from you, I will start working on the `--xcompile arm` option.

manu-silicon:
@mylibero Sorry, got my hands full with arm code generation issue on CoreCLR.

schellap:
@mylibero, lot of changes went into the host this week, please rebase against those.

mylibero:
@schellap done. ready for review.

src/corehost/build.sh
```diff
@@ -55,5 +77,9 @@ __cmake_defines="${__cmake_defines} ${__define}"
 
 
 echo "Building Corehost from $DIR to $(pwd)"
-cmake "$DIR" -G "Unix Makefiles" $__cmake_defines -DCLI_CMAKE_RUNTIME_ID:STRING=$__runtime_id
+if [ $__CrossBuild == 1 ]; then
```
schellap: Looks like `if` and `else` are inverted.

src/corehost/build.sh
```diff
@@ -4,6 +4,19 @@
...
+    echo ""
+    echo "Options:"
+    echo "  --arch <Architecture>             Target Architecture (amd64, x86, arm)"
+    echo "  --rid <Runtime Identifier>        Target Runtime Identifier"
+    echo "  --xcompiler <Cross C++ Compiler>  Cross Compiler when the target is arm"
+    echo "                                    e.g.) /usr/bin/arm-linux-gnueabi-g++-4.7"
+
+    exit 1
+}
+
```
schellap: Thanks!

schellap:
Lgtm otherwise

src/corehost/build.sh
mylibero: oops! Fixed now.

PR #1750, opened 2016-03-08
Fix dependency context builder regression

pakrym:
Fix regression in dependency builder when having multiple dependencies with same name.
@anurse


anurse:
:shipit: when builds succeed

pranavkm:
Any idea where the `IDictionary<>.Add` call is? Could we change that to indexer assignment so it doesn't throw?

pakrym:
I wasn't dictionary add call it's `Can not add property System.Collections to Newtonsoft.Json.Linq.JObject. Property with the same name already exists on object.` thrown at https://github.com/dotnet/cli/blob/pakrym/deps-duplication-fix/src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs#L246

troydai:
:shipit:

pranavkm:
:shipit:

davidfowl:
When does this happen?

pakrym:
Won't merge per @davidfowl request, we'll need to fix root cause

PR #1763, opened 2016-03-08
fix bad merge that lost removal of net45 from portable app test

anurse:
not really using this to review, just to run the CI builds :)


anurse:
OSX builds failing as expected because of NuGet issue (which is fixed and rolling through the build process! yay!)

PR #1766, opened 2016-03-09
Generate actual deps.json file

pakrym:
@anurse


anurse:
Can you update dotnet-build.Tests/BuildPortableTests as well? It should just be like BuildOutputTests, add the output file (I didn't use the constants, so if you feel like a little clean-up you could change it to use the constants too ;))

anurse:
:shipit: after updating BuildPortableTests and ensuring we get (mostly) green CI

anurse:

anurse:
:shipit:

PR #1768, opened 2016-03-09
Allow to pin a specific runtime version

schellap:
@anurse, @piotrpMSFT, @gkhanna79, @leecow, @richlander

Provide a way to pin the runtime version.

### DIFF
--- a/Documentation/specs/runtime-configuration-file.md
+++ b/Documentation/specs/runtime-configuration-file.md
@@ -29,7 +29,8 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files. 
         "gcConcurrent": false,
         "framework": {
             "name": "Microsoft.DotNetCore",
-            "version": "1.0.1"
+            "version": "1.0.1",
+            "rollForward": false,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


anurse:
Looks correct to me. /cc @DamianEdwards

davidfowl:
Not sure about this. A cleaner way to handle it would be to support version ranges. 

Min version:

```
"framework": {
    "name": "Microsoft.DotNetCore",
    "version": "1.0.1"
}
```

Exact version:

```
"framework": {
    "name": "Microsoft.DotNetCore",
    "version": "[1.0.1]"
}
```

Future (not for v1.1)

```
"framework": {
    "name": "Microsoft.DotNetCore",
    "version": "1.0.*"
}
```

Wildcard syntax.

anurse:
That seems OK to me, as long as we keep the version range syntax clean and simple (the NuGet syntax is pretty simple: https://docs.nuget.org/create/versioning).

Documentation/specs/runtime-configuration-file.md
```diff
@@ -118,7 +119,10 @@ This section is copied verbatim from an identical section in the input `project.
...
 * `gcConcurrent` - Boolean indicating if background garbage collection should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).
-* `framework` - Indicates the name and version of the shared framework to use when activating the application. The presence of this section indicates that the application is a portable app designed to use a shared redistributable framework.
+* `framework` - Indicates the `name`, `version` and other properties of the shared framework to use when activating the application. The presence of this section indicates that the application is a portable app designed to use a shared redistributable framework.
+  * `autoRollForward` - When `false`, the runtime version is strictly obeyed by the host. When `autoRollForward` is not specified, or specified as `true`, then the runtime from the highest version that differs only in the `SemVer` patch field will be picked up.
```
gkhanna79: "runtime" - this should be "framework".

Documentation/specs/runtime-configuration-file.md
```diff
@@ -118,7 +119,10 @@ This section is copied verbatim from an identical section in the input `project.
...
+    * For example, if `version=1.0.1` and `autoRollForward` is `true`, the host would load `coreclr.dll` from `1.0.{max}` but will not load from `1.1.0` even if present. When `autoRollForward` is `false`, `coreclr.dll` will be loaded from `1.0.1` strictly.
```
gkhanna79: There should be no reference to Coreclr.dll. Instead, refer to the "shared framework".

Documentation/specs/runtime-configuration-file.md
```diff
@@ -118,7 +119,10 @@ This section is copied verbatim from an identical section in the input `project.
 
 * `gcServer` - Boolean indicating if the server GC should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/ms229357.aspx) setting)
 * `gcConcurrent` - Boolean indicating if background garbage collection should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).
-* `framework` - Indicates the name and version of the shared framework to use when activating the application. The presence of this section indicates that the application is a portable app designed to use a shared redistributable framework.
+* `framework` - Indicates the `name`, `version` and other properties of the shared framework to use when activating the application. The presence of this section indicates that the application is a portable app designed to use a shared redistributable framework.
```
GuardRex: Suggestion: Don't imply the presence of the section for standalone apps. You might want to add to the end ...
```
This section will not be present for standalone applications that do not rely upon a shared framework. 
```
Nit: Lists have (for the most part) adopted the comma prior to the conjunction format:
```
the `name`, `version`, and other properties
```

Documentation/specs/runtime-configuration-file.md
GuardRex: Comma splice ...
```
... is not specified, or specified as ...
```
to
```
... is not specified or specified as ...
```
Remove "then," as it is implied that what follows must be true if the conditions are met ...
```
... as `true`, then the runtime ...
```
to
```
... as `true`, the runtime ...
```

schellap:
@davidfowl, that definitely sounds cleaner and I like it.

@gkhanna79 and I discussed and decided that we do not want to pay the cost of building this now. 

However, in ask mode, if allowed, we'll make the necessary changes.

anurse:
We should error out. The `rollforward` property is expected to be a JSON boolean, so that would be a JSON parsing error.

PR #1772, opened 2016-03-09
take NuGet with OSX fix

anurse:
Not 100% sure it has the fix, since I can't access the NuGet CI, but the build will help us find out :)

/cc @piotrpMSFT @brthor @Sridhar-MS @davidfowl @pakrym @joelverhagen


anurse:
Sounds like this does not yet have the fix. Will keep the PR up-to-date as the NuGet build rolls out

anurse:
FYI, this is not actually going to fix in our CI right now because the stage0 has the old build. Assuming all other builds pass, I'm going to merge this and check if Jenkins is working with the new stage 0.

anurse:
Windows 2016 is so SLOW. I'm confident this won't affect it given the other passing builds. Gonna merge it.

PR #1773, opened 2016-03-09
Add version field for future-proofing

anurse:
We did this in the lock file and it was extremely useful. It seemed like a useful small addition to make to ensure we can make significant changes and ensure older hosts will produce useful errors.

If people think this looks like a good idea, I'll file actionable bugs for the two main work items: Add this to the generated deps file; `corehost` to read it and error if the value is unsupported.

/cc @gkhanna79 @schellap @davidfowl @DamianEdwards @piotrpMSFT @Petermarcu


anurse:
This arose because I was thinking about some format simplification stuff for post-v1 (or just, y'know, never :)) and it would change the format enough to want some kind of version toggle. We could just wait until we actually change the format to introduce this field, but it seems like a good thing to have in v1 anyway.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -123,6 +123,12 @@ This section is copied verbatim from an identical section in the input `project.
 
 These settings are read by `corehost` to determine how to initialize the runtime. All versions of `corehost` **must ignore** settings in this section that they do not understand (thus allowing new settings to be added in later versions).
 
+### `version` Property (`.deps.json`)
+
+This is a simple integer version indicating the version of the deps file format in use. It begins at `2` because the deps file format closely matches the format for `project.lock.json` which is at version `2`. In general, hosts should fail to load applications that use deps files with "unsupported" versions. It's up to the host to determine it's support policy. There is no guarantee of compatibility between versions of the deps file, each version could be completely different and require quirking to load properly, so no assumption should be made about the version field. The goal of this field is to allow us to make major changes without causing undefined behavior in existing hosts (instead providing a clearer "This application is not compatible with this host" error).
```
gkhanna79: If the format of the deps file changes, then it mandates an update to the host that reads it (both the outer host that reads the FX name/version and the impl host that performs assembly lookup). Without this invariant, we dont expect to push a deps.json update at all. Given that, do you still feel there is a scenario where an older host would be endup reading a new deps.json?

anurse: The scenario I'm envisioning is publishing using tools that know about deps version 3, then taking that application and moving it to another machine that only has the older muxer (which only knows about deps version 2). It would be good if the user were to get a clear error message like "This application requires a newer version of the .NET Tools blah blah blah" rather than potentially undefined behavior.

Indeed, updating the deps file format mandates a change in the host, but how does the user know that the update is required? The idea here is to make that clear to the user. Similarly, it is likely that if a deps file change is made, we will want newer hosts to understand older formats for compatibility (or at least provide clear error messaging) and they might wish to change behavior based on this version field.

gkhanna79: The deps.json is processed by dotnethostimpl.dll, which is part of the SharedFramework redist. Given that, is it even possible for an application to target a version of the SharedFramework (which is already public) and yet carry a deps.json that dotnethostimpl.dll (part of the public SharedFramework that the app claims to target) does not understand?

anurse: That's a fair point, since the version of the shared framework comes from `.runtimeconfig.json`.

PR #1782, opened 2016-03-09
Add support for building dotnet/cli on Debian.

jeremymeng:
Currently `/etc/os-release` still needs to be patched to have `VERSION_ID="8.2"` because corefx and coreclr are using different Debian RID's.

/cc @Sridhar-MS


Sridhar-MS:
@jeremymeng Can you do another rebase?

scripts/Microsoft.DotNet.Cli.Build.Framework/CurrentPlatform.cs
```diff
@@ -65,11 +65,20 @@ public static bool IsUnix
...
+                return string.Equals(osname, "debian", StringComparison.OrdinalIgnoreCase);
+            }
+        }
+
         public static bool IsLinux
         {
             get
             {
-                return IsUbuntu || IsCentOS || IsRHEL;
+                return IsUbuntu || IsCentOS || IsRHEL || IsDebian;
```
Sridhar-MS: You will need to add Debian to `IsPlatform` method too.

Sridhar-MS:
LGTM module one comment.

PR #1784, opened 2016-03-09
add support for publishing portable application layout

anurse:
With this change, `dotnet-publish` produces a portable application layout, where runtime-specific assets are laid out in subdirectories so that `corehost` can read them as appropriate. This ties in with #1766 to basically complete the portable apps stuff for `dotnet restore` and `dotnet publish`. Once this PR and #1766 are in, we should (strong emphasis on *should*) have everything `corehost` needs checked in. Certainly the bulky stuff should be in and we can start iterating the last few issues down.

/cc @piotrpMSFT @davidfowl @pakrym @schellap


test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/PublishCommand.cs
```diff
@@ -88,27 +85,13 @@ public string GetOutputExecutable()
...
-
-        private string GetConfigOption()
-        {
-            return string.IsNullOrEmpty(_config) ? "" : $"-c {_output}";
-        }
+        private string GetPortableOption() => _forcePortable ? "--portable" : "";
+        private string GetFrameworkOption() => string.IsNullOrEmpty(_framework) ? "" : $"-f {_framework}";
+        private string GetRuntimeOption() => string.IsNullOrEmpty(_runtime) ? "" : $"-r {_runtime}";
+        private string GetOutputOption() => string.IsNullOrEmpty(_output) ? "" : $"-o \"{_output}\"";
+        private string GetConfigOption() => string.IsNullOrEmpty(_config) ? "" : $"-c {_output}";
```
pakrym: Make them properties?

anurse: :+1:

test/dotnet-publish.Tests/PublishPortableTests.cs
```diff
@@ -0,0 +1,47 @@
...
+            Tuple.Create("debian-x64", "libuv.so"),
+            Tuple.Create("rhel-x64", "libuv.so"),
+            Tuple.Create("osx", "libuv.dylib"),
+            Tuple.Create("win7-arm", "libuv.dll"),
+            Tuple.Create("win7-x86", "libuv.dll"),
+            Tuple.Create("win7-x64", "libuv.dll")
+        };
+
+        [Fact]
+        public void PortableAppWithRuntimeTargetsIsPublishedCorrectly()
```
pakrym: I'd like to have build portable test, which checks that libuv is copied for current runtime.

anurse: You mean in the standalone case?

pakrym:
:shipit:

anurse:
There are some issues here. I'm going to just open a new one after some redesign

Issue #1762, opened 2016-03-08
Writing to project.lock.json fails sometimes when creating a web project


barrytang:
## Steps to reproduce
Create a web project

## Expected  behavior
No error in the Error list

## Environment data
`dotnet --version` output:

barrytang:
![image](https://cloud.githubusercontent.com/assets/1648865/13617843/9380824e-e535-11e5-8f78-861ac96b4e83.png)

BillHiebert:
I am seeing builds fail with this same issue, and this is affecting a lot of our tests..

piotrpMSFT:
@pakrym, I just spoke with @BillHiebert and verified that VS itself is not likely to be looking at project.lock.json. There is suspicion that ProjectModelServer might be keeping the file in use. Any thoughts?

troydai:
[`WorkspaceContext`](https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs) is where we cached the `ProjectContext` and `LockFile`. `WorkspaceContext` itself doesn't cache a file handler by itself. So I'm looking at `LockFile` and `ProjectContext` now.

barrytang:
+@Eilon

@troydai, any updates? This is blocking us from running tests reliably...

troydai:
Sorry I was sidetracked this morning. I'll run a simulation now.

troydai:
I doubt it is caused by project model server. In this [test](https://github.com/dotnet/cli/blob/troy/1668/test/dotnet-projectmodel-server.Tests/DthTests.cs#L392) the lockfile is updated while server is still running and `WorkspaceContext` is still catching all the data. The test pass on both Windows and Mac OSX.

troydai:
Todd and I were trying to repo this just now. We created a project, wait for the restore and delete the lock file. No error. The lock file is deleted successful. That means dotnet-projectmodel-server doesn't occupy the lock file. 

More like it is related to the VS started `dotnet restore` commands. Could there be parallel `restore` happen?

troydai:
@barrytang

troydai:
Update: Todd manage to create a repo and save the Process Monitor logs. I'm investigating.

troydai:
```
[...]

PR #1789, opened 2016-03-10
Retry when lock file is occupied by other process

troydai:
Address #1762

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
@@ -17,7 +17,7 @@ namespace Microsoft.DotNet.ProjectModel.Graph
     {
         public static LockFile Read(string lockFilePath)
         {
-            using (var stream = new FileStream(lockFilePath, FileMode.Open, FileAccess.Read, FileShare.Read))
+            using (var stream = ResilientFileStreamOpener.OpenFile(lockFilePath))
             {
                 try
                 {
@@ -34,7 +34,7 @@ namespace Microsoft.DotNet.ProjectModel.Graph
             }
         }
 
-        internal static LockFile Read(string lockFilePath, Stream stream)
+        public static LockFile Read(string lockFilePath, Stream stream)
         {
             try
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
```diff
@@ -17,7 +18,17 @@ public static class LockFileReader
     {
         public static LockFile Read(string lockFilePath)
```
davidfowl: There should be no retry by default. Only the project model server should be doing the retry.

troydai: This is the case here. Only `WorkspaceContext` call the overload.

troydai:
@dotnet-bot test Windows_NT Debug Build please

src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
pakrym: I'd prefer not having retry code in ProjectModel at all, this could be done by helper method easily.

davidfowl: Yah, it's pretty weird to put it in the core really.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -205,8 +206,23 @@ private FileModelEntry<LockFile> AddLockFileEntry(string projectDirectory, FileM
                 else
                 {
                     currentEntry.FilePath = Path.Combine(projectDirectory, LockFile.FileName);
-                    currentEntry.Model = LockFileReader.Read(currentEntry.FilePath);
-                    currentEntry.UpdateLastWriteTimeUtc();
+
+                    using (var fs = ResilientFileStreamOpener.OpenFile(currentEntry.FilePath, retry: 2))
```
pakrym: Would 2 retries be enough? `dotnet restore` can run much longer then a second

troydai: The rety window is not for the entire restore, is for the during of actually writing the file.

PR #1792, opened 2016-03-10
Shared FX and Portable Support for Unified Host

schellap:
This change is primarily for folks who feel adventurous to take an early look.

- Most of the code is in place.
- Some untested code could have bad or obvious bugs.
- I've not done much testing at this point.
- `--depsfile` and `--additionalprobingpath` need to be wired into the logic

@gkhanna79


src/corehost/cli/args.cpp
```diff
@@ -105,13 +107,10 @@ bool parse_arguments(const int argc, const pal::char_t* argv[], arguments_t& arg
         args.deps_path.append(app_base);
         args.deps_path.push_back(DIR_SEPARATOR);
         args.deps_path.append(app_name, 0, app_name.find_last_of(_X(".")));
-        args.deps_path.append(_X(".deps"));
+        args.deps_path.append(_X(".deps.json"));
     }
 
-    pal::getenv(_X("NUGET_PACKAGES"), &args.nuget_packages);
     pal::getenv(_X("DOTNET_PACKAGES_CACHE"), &args.dotnet_packages_cache);
     pal::getenv(_X("DOTNET_SERVICING"), &args.dotnet_servicing);
```
gkhanna79: We should ad  TODO here to fix this for servicing location to be computed at runtime.

src/corehost/cli/deps_entry.cpp
```diff
@@ -0,0 +1,138 @@
...
+//    base - The base directory to look for the relative path of this entry
+//    str  - If the method returns true, contains the file path for this deps
+//           entry relative to the "base" directory
+//
+// Returns:
+//    If the file exists in the path relative to the "base" directory.
+//
+bool deps_entry_t::to_full_path(const pal::string_t& base, pal::string_t* str) const
+{
+    pal::string_t& candidate = *str;
```
gkhanna79: Check if base is empty or not and bail out accordingly.

src/corehost/cli/deps_format.cpp
```diff
@@ -0,0 +1,259 @@
...
+}
+
+bool deps_json_t::load_portable(const json_value& json, const pal::string_t& target_name, const rid_fallback_graph_t& rid_fallback_graph)
+{
+    portable_assets_t portable_assets;
+
+    for (const auto& package : json.at(_X("targets")).at(target_name).as_object())
+    {
+        const auto& targets = package.second.as_object();
+        auto iter = targets.find(_X("subtargets"));
```
gkhanna79: What is subtargets for? I dont see it in the spec (https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/specs/runtime-configuration-file.md).

gkhanna79: Is this for PHAT packages?

src/corehost/cli/deps_format.cpp
```diff
@@ -0,0 +1,259 @@
...
+            });
+            if (iter == fallback_rids.end() || (*iter).empty())
+            {
+                trace::error(_X("Did not find a matching fallback rid for package %s for the host rid %s"), package.first.c_str(), host_rid.c_str());
+                return false;
+            }
+            matched_rid = *iter;
+        }
+        assert(!matched_rid.empty());
+        for (auto iter = package.second.begin(); iter != package.second.end(); ++iter)
```
gkhanna79: Is this just required to cleanup the entries in the package representation once the applicable fallback RID has been found?

src/corehost/cli/deps_format.cpp
```diff
@@ -0,0 +1,259 @@
...
+                entry.library_name = library.first.substr(0, pos);
+                entry.library_version = library.first.substr(pos + 1);
+                entry.library_type = _X("package");
+                entry.library_hash = hash;
+                entry.asset_name = asset_name;
+                entry.asset_type = s_known_asset_types[i];
+                entry.relative_path = rel_path;
+                entry.is_serviceable = serviceable;
+
+                // TODO: Deps file does not follow spec. It uses '\\', should use '/'
```
gkhanna79: Please file an issue for this. CC @anurse

src/corehost/cli/deps_resolver.cpp
```diff
@@ -440,43 +153,89 @@ pal::string_t deps_resolver_t::resolve_coreclr_dir(
...
+            }
+        }
+
+        return pal::string_t();
+    };
 
-    // App dir.
-    trace::verbose(_X("Probing for CoreCLR in app directory=[%s]"), app_dir.c_str());
-    if (coreclr_exists_in_dir(app_dir))
+    trace::info(_X("--- Starting CoreCLR Proble from app deps.json"));
```
gkhanna79: Probe or Proble?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -516,19 +275,35 @@ void deps_resolver_t::resolve_tpa_list(
         const pal::string_t& clr_dir,
         pal::string_t* output)
 {
+    std::vector<deps_entry_t> empty(0);
+
+    pal::string_t ni_package_cache_dir;
+    if (package_cache_dir.empty())
```
gkhanna79: !empty

src/corehost/cli/deps_resolver.cpp
```diff
@@ -539,45 +314,63 @@ void deps_resolver_t::resolve_tpa_list(
         {
             add_tpa_asset(entry.asset_name, candidate, &items, output);
         }
-        // Is this entry present in the secondary package cache?
+        // Is an NI image for this entry present in the secondary package cache?
+        else if (entry.to_hash_matched_path(ni_package_cache_dir, &candidate))
```
gkhanna79: Shouldnt be matching hash. @anurse @davidfowl  will hash file be present for the NI image?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -539,45 +314,63 @@ void deps_resolver_t::resolve_tpa_list(
...
+            add_tpa_asset(entry.asset_name, m_sxs_assemblies.find(entry.asset_name)->second, &items, output);
+        }
+        // The app is portable so the asset should be picked up from relative subpath.
+        else if (is_portable && deps->try_ni(entry).to_full_path(app_dir, &candidate))
+        {
+            add_tpa_asset(entry.asset_name, candidate, &items, output);
         }
         // Is this entry present in the package restore dir?
-        else if (entry.to_full_path(package_dir, &candidate))
+        else if (deps->try_ni(entry).to_full_path(package_dir, &candidate))
```
gkhanna79: Check for package_Dir being empty

src/corehost/cli/deps_resolver.cpp
anurse: The hash would be for the whole nupkg, so I think it would be present. The hash is really just a unique string that's expected to be identical between the deps file and the cache. A SHA-1 hash of the nupkg is the most convenient way to achieve that, but it's OK if the hash is used to match assets **derived** from the package (i.e. NIs) as well as assets contained directly in the package.

schellap:
@dotnet-bot test this please

src/corehost/cli/deps_resolver.h
```diff
@@ -65,9 +60,13 @@ class deps_resolver_t
 
 private:
 
-    bool load();
-
-    bool parse_deps_file(const arguments_t& args);
+    static pal::string_t get_fx_deps(const pal::string_t& fx_dir, const pal::string_t& fx_name)
+    {
+        pal::string_t fx_deps = fx_dir;
+        pal::string_t fx_deps_name = pal::to_lower(fx_name) + _X(".deps.json");
```
gkhanna79: I think Ubuntu is case-sensitive, so if we lower the FX_name and use it later to lookup the file, it may fail. Can you please confirm this is not an issue?

src/corehost/cli/dll/CMakeLists.txt
```diff
@@ -32,6 +40,8 @@ else()
     list(APPEND SOURCES ../../common/pal.unix.cpp)
 endif()
 
+add_definitions(-D_NO_ASYNCRTIMP)
+add_definitions(-D_NO_PPLXIMP)
 add_definitions(-DCOREHOST_MAKE_DLL=1)
 
 add_library(hostpolicy SHARED ${SOURCES})
```
gkhanna79: Given that the host will be called dotnet.exe, we should rename this to be "dotnethostpolicy.dll" and "hostfxr.dll" to "dotnethostfxr.dll"

src/corehost/cli/fxr/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    list(APPEND SOURCES ../../common/pal.windows.cpp)
+else()
+    list(APPEND SOURCES ../../common/pal.unix.cpp)
+endif()
+
+add_definitions(-D_NO_ASYNCRTIMP)
+add_definitions(-D_NO_PPLXIMP)
+add_definitions(-DCOREHOST_MAKE_DLL=1)
+
+add_library(hostfxr SHARED ${SOURCES})
```
gkhanna79: See earlier comment for this.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+        if (!pal::realpath(&app_path))
+        {
+            return StatusCode::LibHostExecModeFailure;
+        }
+
+        runtime_config_t config(get_runtime_config_json(app_path));
+        if (config.get_portable())
+        {
+            pal::string_t fx_dir = resolve_fx_dir(own_dir, &config, app_path);
```
gkhanna79: resolve_fx_dir can return empty fx_dir. We should add a check to bail out for the error case.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            // Transform dotnet exec [--additionalprobingpath path] [--depsfile file] dll [args] -> dotnet dll [args]
+
+            std::vector<const pal::char_t*> new_argv(argc - cur_i + 1); // +1 for dotnet
+            memcpy(new_argv.data() + 1, argv + cur_i, (argc - cur_i) * sizeof(pal::char_t*));
+            new_argv[0] = argv[0];
+
+            pal::string_t deps_file = opts.count(_X("--depsfile")) ? opts[_X("--depsfile")] : _X("");
+            pal::string_t probe_path = opts.count(_X("--additionalprobingpath")) ? opts[_X("--additionalprobingpath")] : _X("");
```
gkhanna79: @brthor Are you planning to pass a hint that muxer can use for non-published apps to continue to support "dotnet <path to app.dll>"?

src/corehost/common/utils.cpp
```diff
@@ -87,3 +114,59 @@ void replace_char(pal::string_t* path, pal::char_t match, pal::char_t repl)
...
+#error "Unknown target"
+#endif
+}
+
+pal::string_t get_own_rid()
+{
+#if defined(COREHOST_RID)
+    return COREHOST_RID;
+#else
+    return _X("win10-x64");
```
gkhanna79: We should not have the "#else" case and always have the host RID. how are we getting COREHOST_RID?

src/corehost/corehost.cpp
```diff
@@ -2,123 +2,156 @@
...
-//    main_fn      - Contains the entrypoint "corehost_main" when returns success.
-//
-// Returns:
-//    Non-zero exit code on failure. "main_fn" contains "corehost_main"
-//    entrypoint on success.
-//
-StatusCode load_host_lib(const pal::string_t& lib_dir, pal::dll_t* h_host, corehost_main_fn* main_fn)
+#ifdef COREHOST_PACKAGE_SERVICING
+    pal::string_t svc_dir;
+    if (!pal::getenv(_X("DOTNET_SERVICING"), &svc_dir))
```
gkhanna79: We should have a Windows specific check here to get the path to PFx86 to initialize servicing folder and override it with the env var if specified.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -70,15 +70,20 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
             var configuration = c.BuildContext.Get<string>("Configuration");
 
             // Run the build
+            string version = DotNetCli.Stage0.Exec("", "--version").CaptureStdOut().Execute().StdOut;
+            string rid = Array.Find<string>(version.Split(Environment.NewLine.ToCharArray()), (e) => e.Contains("Runtime Id:")).Replace("Runtime Id:", "").Trim();
```
anurse: `PlatformServices.Default.Runtime.GetRuntimeIdentifier` should get you this without having to shell out

src/corehost/common/pal.h
```diff
@@ -15,10 +15,12 @@
...
 #include <algorithm>
+#include <cassert>
 
 #if defined(_WIN32)
 
-#include <Windows.h>
+#define NOMINMAX
+#include <windows.h>
 
 #define HOST_EXE_NAME L"corehost.exe"
```
anurse: this will need to change to `dotnet.exe`

schellap:

PR #1795, opened 2016-03-10
Add installer scripts for the shared framework/host, incorporate into build

mellinoe:
Most of the work here is by @ellismg , I am submitting this as i'm going to be keeping this moving along while he's on vacation for a few days.

With this change, we produce many more installers:

* MSI installer for the shared framework and host (x64 and x86)
* deb installer for the shared framework and host (x64)
* pkg installer for the shared framework and host (x64)

There is some shared build logic (in SharedFrameworkTargets.cs) that is responsible for producing the binary layout that all of the installers consume. Essentially, this just performs a "dotnet publish" on a project.json that contains the standard library/host that we want to package (see src/sharedframework/framework/project.json). After that step, platform-specific installer targets take place which handle creating the appropriate bundle for each system.

There is logic for running cross-gen on the shared framework layout before it is packaged up, but this currently is only active on Windows. There is a comment in SharedFrameworkTargets regarding this. The basic problem is we need to update to a newer version of the framework packages in order to pull in mscorlib.ni.dll, but currently doing so is tied with upgrading the rest of the CLI's dependencies. It's possible that we could detach the two and have the shared framework find a matching crossgen version some other way.

@ellismg , @piotrpMSFT , @schellap , @gkhanna79

### DIFF
--- a/packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
+++ b/packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
@@ -1,0 +1,35 @@
+{
+    "maintainer_name":"Microsoft",
+    "maintainer_email": "dotnetcore@microsoft.com",
+
+    "package_name": "%SHARED_FRAMEWORK_DEBIAN_PACKAGE_NAME%",
+    "install_root": "/usr/share/dotnet",
+
+    "short_description": ".NET Core Shared Framework %SHARED_FRAMEWORK_NUGET_NAME% %SHARED_FRAMEWORK_NUGET_VERSION%",
+    "long_description": ".NET Core is a cross-platform implementation of .NET Framework, a modern, modular platform\n for building diverse kinds of applications, from command-line applications to microservices and \n modern websites.\n This package contains a runtime and framework which can be used by .NET Core applications.",
+    "homepage": "https://dotnet.github.io/core",
+
+    "release":{
+        "package_version":"1.0.0.0",
+        "package_revision":"1",
+        "urgency" : "low",
+        "changelog_message" : "Inital shared framework."
+    },
+
+    "control": {
+        "priority":"standard",
+        "section":"libs",
+        "architecture":"amd64"
+    },
+
+    "copyright": "2015 Microsoft",
+    "license": {
+        "type": "MIT",
+        "full_text": "Copyright (c) 2015 Microsoft\nPermission is hereby granted, free of charge, to any person obtaining a copy\nof this software and associated documentation files (the \"Software\"), to deal\nin the Software without restriction, including without limitation the rights\nto use, copy, modify, merge, publish, distribute, sublicense, and/or sell\ncopies of the Software, and to permit persons to whom the Software is\nfurnished to do so, subject to the following conditions:\nThe above copyright notice and this permission notice shall be included in all\ncopies or substantial portions of the Software.\nTHE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\nIMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\nFITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE\nAUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER\nLIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,\nOUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE\nSOFTWARE."
+    },
+
+    "debian_dependencies":{
+        "libssl-dev" : {},
+        "libcurl3" : {}
+    }
+}
--- a/packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
+++ b/packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
@@ -29,7 +29,7 @@
     },
 
     "debian_dependencies":{
-        "libssl-dev" : {},
+        "libssl1.0.0" : {},
         "libcurl3" : {}
     }
 }
--- a/scripts/dotnet-cli-build/SharedFrameworkTargets.cs
+++ b/scripts/dotnet-cli-build/SharedFrameworkTargets.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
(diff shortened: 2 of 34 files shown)


mellinoe:
Seems like CI is dead, not sure why though. Looks like restore failed.

scripts/dotnet-cli-build/SharedFrameworkTargets.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            // Rename the .deps file
+            File.Move(Path.Combine(SharedFrameworkNameAndVersionRoot, "framework.deps"), Path.Combine(SharedFrameworkNameAndVersionRoot, $"{SharedFrameworkName}.deps"));
+
+            // corehost will be renamed to dotnet at some point and then this can be removed.
+            File.Move(Path.Combine(SharedFrameworkNameAndVersionRoot, $"{CoreHostBaseName}{Constants.ExeSuffix}"), Path.Combine(SharedFrameworkNameAndVersionRoot, $"dotnet{Constants.ExeSuffix}"));
+
+            // hostpolicy will be renamed to dotnet at some point and then this can be removed.
+            File.Move(Path.Combine(SharedFrameworkNameAndVersionRoot, $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}"), Path.Combine(SharedFrameworkNameAndVersionRoot, $"{Constants.DynamicLibPrefix}dotnethostimpl{Constants.DynamicLibSuffix}"));
+
```
schellap: @mellinoe, there is one more `dll` for the host in addition to `hostpolicy.dll`. It is called `dotnethostfxr.dll`

gkhanna79: @schellap I presume, for v1.0, this is the one next to the outer muxer, right?

mellinoe: Has this `dotnethostxfr.dll` been added to the NuGet package yet? If so, which version has it?

mellinoe:
Any additional comments about this? This is ready to go in now, even if other work is still in-flight. If additional assemblies are added to the NuGet packages this depends on, they will automatically get picked up by the publishing step and be included in all of the installers. Once this goes in, we can also easily split up the work to normalize the text and branding elements in the installers (once we have concrete guidelines for those).

scripts/package/package-sharedhost-debian.sh
```diff
@@ -0,0 +1,108 @@
...
+
+create_debian_package(){
+    header "Packing .deb"
+
+    mkdir -p "$PACKAGE_OUTPUT_DIR"
+
+    "$PACKAGING_TOOL_DIR/package_tool" -i "$PACKAGE_LAYOUT_DIR" -o "$PACKAGE_OUTPUT_DIR" -v "$SHARED_HOST_DEBIAN_VERSION"
+}
+
+execute_build
```
brthor: Any reason why we don't run the debian package tests against this and shared-framework debian package? 
Similar to how it's done in package-debian.sh

packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
```diff
@@ -0,0 +1,35 @@
...
+        "type": "MIT",
+        "full_text": "Copyright (c) 2015 Microsoft\nPermission is hereby granted, free of charge, to any person obtaining a copy\nof this software and associated documentation files (the \"Software\"), to deal\nin the Software without restriction, including without limitation the rights\nto use, copy, modify, merge, publish, distribute, sublicense, and/or sell\ncopies of the Software, and to permit persons to whom the Software is\nfurnished to do so, subject to the following conditions:\nThe above copyright notice and this permission notice shall be included in all\ncopies or substantial portions of the Software.\nTHE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\nIMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\nFITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE\nAUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER\nLIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,\nOUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE\nSOFTWARE."
+    },
+
+    "debian_dependencies":{
+        "libssl-dev" : {},
```
brthor: This can be replaced with libssl1.0.0

packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
```diff
@@ -0,0 +1,35 @@
...
+        "package_version":"1.0.0.0",
+        "package_revision":"1",
+        "urgency" : "low",
+        "changelog_message" : "Inital shared framework."
+    },
+
+    "control": {
+        "priority":"standard",
+        "section":"libs",
+        "architecture":"amd64"
```
brthor: We should probably change the cli ones from "any" to "amd64" as well.

Could be in a separate change.

scripts/dotnet-cli-build/SharedFrameworkTargets.cs
schellap: Sounds good. The hostfxr or hostmxr is still in PR as of now. The names will change as you note.

packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
mellinoe: Fixed

scripts/package/package-sharedhost-debian.sh
mellinoe: I added a similar test function to both. They currently don't test that you can actually do anything with the shared framework (since you can't yet), but we can add that as the other work is integrated. So this does the steps that install, uninstall, reinstall, etc. the debian package to ensure that it is valid.

packaging/sharedframework/debian/dotnet-sharedframework-debian_config.json
mellinoe: I'll leave that for a separate change as well. The CLI one might be different anyways once it stops shipping the framework/runtime with it. It will probably be architecture-agnostic at that point.

brthor: good point, hadn't considered we'll be IL only at that point.

gkhanna79:
> There is logic for running cross-gen on the shared framework layout before it is packaged up, but this currently is only active on Windows.

@mellinoe Please make this logic opt-in for now. This will ensure we can build and test R2R version of SharedFramework on-demand but it should not happen by default. As we approach RTM, we will determine when to make this default.

PR #1796, opened 2016-03-10
Fix issue preventing MVC gate to take a new dotnet cli

pakrym:
`context.RuntimeIdentifier` was always null there because that's compilation context.
@anurse


anurse:
:shipit:

Issue #1793, opened 2016-03-10
Deps.json should always use `/` for paths


anurse:
In the `project.lock.json` format, we set a requirement that all paths **always** use `/` for delimiting directories, regardless of the platform. It looks like the `.deps.json` file is not doing that. We should make these consistent and have `.deps.json` ensure paths are flipped to `/` as necessary.

The host should always assume paths are going to be delimited with `/`, which implies that it will have to flip the directory separators on Windows.

/cc @pakrym @schellap

Issue #1791, opened 2016-03-10
Runtime section should have RID fallback graph without target subnode


schellap:
`dotnet.deps.json` contains a section called:

```
  "runtimes": {
    ".NETStandardApp,Version=v1.5": {}
  }
```

https://github.com/dotnet/cli/blob/8e78083913dd1614e3183a9c2cf9a3bba3717a6b/Documentation/specs/runtime-configuration-file.md

`RID` fallback also contains a section called `runtimes` with a different object layout, can we use a different name for the RID fallback graph other than "runtimes", for example `runtime-fallback` or `rid-graph`?

```
"runtime-fallback": {
        "ubuntu.15.04-x64": [ "ubuntu.14.10-x64", "ubuntu.14.04-x64", "debian.8-x64", "linux-x64", "linux", "unix", "any", "base" ],
        "ubuntu.14.10-x64": [ "ubuntu.14.04-x64", "debian.8-x64", "linux-x64", "linux", "unix", "any", "base" ],
        "ubuntu.14.04-x64": [ "debian.8-x64", "linux-x64", "linux", "unix", "any", "base" ]
    }
```

@anurse @pakrym @mellinoe @ellismg

pakrym:
Thats a bug, the section you see is fallback graph section, at some point it had target subnode (https://github.com/dotnet/cli/commit/dd6888c02b852dc4bdba9fce4b8fbe74348b724b#diff-b65c14694db4c6fab803eacceb1b5bedR105) and I missed that it changed later. Will fix.

PR #1801, opened 2016-03-10
Fix issue with paths in deps.json not being normalized to have /

pakrym:
Also moves rid graph section to root level instead of being inside 	target subsection

Fixes https://github.com/dotnet/cli/issues/1793 and https://github.com/dotnet/cli/issues/1791

@anurse


anurse:
:shipit: when builds green

PR #1817, opened 2016-03-11
WIP: Installation scenarios spec

krwq:
This includes mostly but not only file naming conventions

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1817)

### DIFF
--- a/Documentation/specs/installation-scenarios.md
+++ b/Documentation/specs/installation-scenarios.md
@@ -1,0 +1,248 @@
+# Notes
+This section describes placeholders used inside this spec.
+
+| Placeholder | Description |
+| ---: | :--- |
+| `<Channel>` | `(future|preview|production)`. [See more info](#channels) |
+| `<Version>` | 4-part-number version |
+| `<OSName>`  | `(win|ubuntu|rhel|osx|debian)` - code for OS name |
+| `<LowestSupportedOSVersion>` | Lowest supported OS Version |
+| `<Architecture>` | Processor architecture related to binaries produced |
+| `<Extension>` | File extension. This will be described in more details later for each OS separately. |
+| `<OSID>` | Abbreviation for: `<OSName><LowestSupportedOSVersion>.<Architecture>`. [See more info](#osid) |
+| `<VersionPointer>` | `(latest|lkg)` |
+| `<ExecutableExtension>` | Executable extension including dot specific to OS (can be empty string) |
+| `<CommitHash>` | Commit hash related to state of repository from where build with specific `<Version>` was build |
+| `<DebianPackageName>` | Name of the debian package. [See more info](#debian-feed-relation) |
+
+# Build Output
+Each official, successful build should create and upload packages to location described by following URLs:
+```
+https://dotnetcli.blob.core.windows.net/dotnet/<Channel>/<Version>/dotnet.<OSID>.<Version>.<Extension>
+
+Currently:
+https://dotnetcli.blob.core.windows.net/dotnet/<Channel>/Binaries/<Version>/dotnet-sharedframework-<OSName>-<Architecture>.<Version>.zip
+https://dotnetcli.blob.core.windows.net/dotnet/<Channel>/Binaries/<Version>/dotnet-host-<OSName>-<Architecture>.<Version>.zip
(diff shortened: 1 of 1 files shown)


Documentation/specs/installation-scenarios.md
```diff
@@ -0,0 +1,243 @@
...
+
+# OSID
+
+OSID represents abbreviation for:
+```
+<OSName><LowestSupportedOSVersion>.<Architecture>
+```
+This naming scheme gives us flexibility to easily create new binaries when OS makes a breaking change without creating confusing names.
+
+In example, we currently put `api-ms-*.dll` files in our binaries. Those files are not needed on Windows 8 and higher. When using name `win7.x64` we can easily decide to get rid of `api-ms-*.dll` in the newest packages and simply call new version `win8.x64` which would mean that from Windows 8 forward those are recommended binaries (there is currently no issue with those files and this should be only treated as an example).
```
mellinoe: Those files won't be needed regardless, won't they be part of the shared framework?

Documentation/specs/installation-scenarios.md
```diff
@@ -0,0 +1,243 @@
...
+For Windows we are using PowerShell script (install-dotnet.ps1).
+For any other OS we are using bash script (install-dotnet.sh)
+
+WIP: Exact script action description.
+
+### Script arguments description
+
+| PowerShell script | Bash script | Default | Description |
+| --- | --- | --- | --- |
+| -Channel | --channel | production | Which [channel](#channels) to install from. Possible values: `future`, `preview`, `production` |
```
mellinoe: We don't use powershell in corefx, but I've found that having different syntax for different scripting languages tends to be confusing, without any real benefit. I know some of the build scripts in here already use this style, but since these installer scripts would be more "public-facing" (i.e. anyone would use them, not just devs in the repo), we might want to consider making them uniform across platforms.

Documentation/specs/installation-scenarios.md
```diff
@@ -0,0 +1,243 @@
...
+
+| `<Extension>` | Description |
+| --- | :--- |
+| exe | Installer bundle. It should be used by end customers |
+| zip | Packed binaries. Used by [installation script](#installation-scripts) |
+| symbols.zip | Packed binaries with included symbols. See [symbol packages](#symbol-packages) |
+
+### Including dotnet cli installer as part of your bundle
+
+In order to install dotnet cli with other installer you need an MSI package. To get MSI, download exe bundle and extract it using `dark` tool which is part of [WiX Toolset](http://wixtoolset.org):
```
mellinoe: maybe "in order to *bundle* the dotnet cli with another installer..."

Documentation/specs/installation-scenarios.md
krwq: I don't have much control on the powershell syntax but I can definietely make powershell switches work in shell

Documentation/specs/installation-scenarios.md
krwq: I haven't wrote down shared framework stuff in here yet

Documentation/specs/installation-scenarios.md
brthor: I agree with @mellinoe  We should choose a format to standardize on. Personally prefer --all-lowercase and -a (where a single dash has only a single character)

piotrpMSFT:
@dotnet-bot test this please.

Documentation/specs/installation-scenarios.md
krwq: ok, we had a discussion, conclusions:
- bash should have switches in linux style
- powershell switches should work everywhere
- --prefix should become --install-dir, PREFIX env should be ignored (PREFIX points to a hive)

Documentation/specs/installation-scenarios.md
Sridhar-MS: https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-sharedframework-win-x64.latest.zip
https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-host-win-x64.latest.zip
https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-win-x64.latest.zip

blackdwarf:
@krwq can we merge this?

PR #1819, opened 2016-03-11
Fix incorrect reference assembly path

troydai:
Yeah, I know. Tested with OmniSharp build this time.


anurse:
:shipit:

troydai:
It's helpful for external project to begin leverage the auto probing feature. I'd like to have this checked in as soon as possible. So I will not wait for Windoes_2016 build here.

PR #1824, opened 2016-03-12
Trigger automated DockerHub builds

naamunds:
This change triggers automated DockerHub builds for successful
official Ubuntu builds of rel.1.0.0. The DockerHub repo for this is
https://hub.docker.com/r/microsoft/dotnet-preview/.

The variables DOCKER_HUB_REPO and DOCKER_HUB_TRIGGER_TOKEN are set in
the "DotNet-CLI-CI (Ubuntu) [rel.1.0.0]" VSTS build definition.

This change also allows the Environment TargetConditionAttribute to be
used to require that an environment variable is set, in addition to the
existing functionality of being able to be used to require that an
environment variable is equal to one of a set of specified values.

cc: @MichaelSimons, @dleeapho, @Sridhar-MS, @brthor

### DIFF
--- a/scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
+++ b/scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.DotNet.Cli.Build.Framework
 {
@@ -13,7 +13,11 @@ namespace Microsoft.DotNet.Cli.Build.Framework
         {
             if (string.IsNullOrEmpty(envVar))
             {
-                throw new ArgumentNullException("envVar");
+                throw new ArgumentNullException(nameof(envVar));
+            }
+            if (expectedVals == null)
+            {
+                throw new ArgumentNullException(nameof(expectedVals));
             }
 
             _envVar = envVar;
@@ -24,15 +28,14 @@ namespace Microsoft.DotNet.Cli.Build.Framework
         {
             var actualVal = Environment.GetEnvironmentVariable(_envVar);
 
-            foreach (var expectedVal in _expectedVals)
+            if (_expectedVals.Any())
             {
-                if (string.Equals(actualVal, expectedVal, StringComparison.Ordinal))
-                {
-                    return true;
-                }
+                return _expectedVals.Any(ev => string.Equals(actualVal, ev, StringComparison.Ordinal));
             }
-
-            return false;
+            else
+            {
+                return !string.IsNullOrEmpty(actualVal);
+            }
         }
     }
 }
--- a/scripts/dotnet-cli-build/PublishTargets.cs
+++ b/scripts/dotnet-cli-build/PublishTargets.cs
@@ -1,12 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Runtime.InteropServices;
+using System.Net.Http;
+using System.Text;
 using Microsoft.DotNet.Cli.Build.Framework;
-using Microsoft.Extensions.PlatformAbstractions;
-using Microsoft.WindowsAzure;
 using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Auth;
 using Microsoft.WindowsAzure.Storage.Blob;
 
 using static Microsoft.DotNet.Cli.Build.Framework.BuildHelpers;
@@ -36,7 +33,8 @@ namespace Microsoft.DotNet.Cli.Build
 
         [Target(nameof(PrepareTargets.Init),
         nameof(PublishTargets.InitPublish),
-        nameof(PublishTargets.PublishArtifacts))]
+        nameof(PublishTargets.PublishArtifacts),
+        nameof(PublishTargets.TriggerDockerHubBuilds))]
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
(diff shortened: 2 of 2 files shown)


scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
```diff
@@ -24,15 +23,22 @@ public override bool EvaluateCondition()
         {
             var actualVal = Environment.GetEnvironmentVariable(_envVar);
 
-            foreach (var expectedVal in _expectedVals)
+            if (_expectedVals.Length == 0)
             {
-                if (string.Equals(actualVal, expectedVal, StringComparison.Ordinal))
+                return !string.IsNullOrEmpty(actualVal);
+            }
+            else
```
MichaelSimons: Consider removing the else as it isn't necessary and adds unnecessary nesting.

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
```diff
@@ -24,15 +23,22 @@ public override bool EvaluateCondition()
         {
             var actualVal = Environment.GetEnvironmentVariable(_envVar);
 
-            foreach (var expectedVal in _expectedVals)
+            if (_expectedVals.Length == 0)
```
MichaelSimons: There doesn't look to be anything that prevents _expectedVals from being set to null.  While it may not be used this way in practice today, you may consider checking for null.

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -118,6 +116,41 @@ public static BuildTargetResult PublishDebFileToDebianRepo(BuildTargetContext c)
...
+        [Environment("DOCKER_HUB_REPO")]
+        [Environment("DOCKER_HUB_TRIGGER_TOKEN")]
+        public static BuildTargetResult TriggerDockerHubBuilds(BuildTargetContext c)
+        {
+            string dockerHubRepo = Environment.GetEnvironmentVariable("DOCKER_HUB_REPO");
+            string dockerHubTriggerToken = Environment.GetEnvironmentVariable("DOCKER_HUB_TRIGGER_TOKEN");
+
+            Uri baseDockerHubUri = new Uri("https://registry.hub.docker.com/u/");
+            Uri dockerHubTriggerUri;
+            if (!Uri.TryCreate(baseDockerHubUri, string.Format("{0}/trigger/{1}/", dockerHubRepo, dockerHubTriggerToken), out dockerHubTriggerUri))
```
MichaelSimons: Consider using the $ interpolated string syntax instead of string.Format to be consistent with the rest of the file.  IMO the $ syntax is more readable.

Sridhar-MS: +1 C# FTW

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
Sridhar-MS: You can throw an `ArgumentNullException` here - https://github.com/dotnet/cli/pull/1824/files#diff-ec153fc2353f931636f6e56f3b335449R11

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
Sridhar-MS: +1

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -118,6 +116,41 @@ public static BuildTargetResult PublishDebFileToDebianRepo(BuildTargetContext c)
             return c.Success();
         }
 
+        [Target]
+        [Environment("DOCKER_HUB_REPO")]
```
Sridhar-MS: You must restrict this target to run only on on Ubuntu or whichever platform for which you are trying to build the docker image.

Sridhar-MS: What are the platforms for which the docker image is built?

Sridhar-MS: I just read your PR message. If you are targeting only ubuntu, please add this attribute - https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/PublishTargets.cs#L106

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -36,7 +33,8 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
 
         [Target(nameof(PrepareTargets.Init),
         nameof(PublishTargets.InitPublish),
-        nameof(PublishTargets.PublishArtifacts))]
+        nameof(PublishTargets.PublishArtifacts),
+        nameof(TriggerDockerHubBuilds))]
```
Sridhar-MS: Can you add `PublishTargets.TriggerDockerHubBuilds` just to maintain uniformity?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -118,6 +116,41 @@ public static BuildTargetResult PublishDebFileToDebianRepo(BuildTargetContext c)
...
+            c.Info("Triggering automated DockerHub builds for " + dockerHubRepo);
+            using (HttpClient client = new HttpClient())
+            {
+                StringContent requestContent = new StringContent("{\"build\": true}", Encoding.UTF8, "application/json");
+                try
+                {
+                    HttpResponseMessage response = client.PostAsync(dockerHubTriggerUri, requestContent).Result;
+                    if (!response.IsSuccessStatusCode)
+                    {
+                        return c.Failed(string.Format("HTTP request to {0} was unsuccessful. HTTP {1}: {2}", dockerHubTriggerUri.ToString(), response.StatusCode, response.ReasonPhrase));
```
eerhardt: Consider writing the response's Content here as well, as that will give you better error messages when things go wrong.

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
naamunds: I considered that, but I think using an `if/else` instead of just an `if` makes the logic clearer. To me, at least, removing the `else` would make the code read like the `_expectedVals.Length == 0` case is an edge case or error condition, whereas I want it to convey that these are just two different options for the use of the method.

Also because the `else` block is small, I'd prefer to keep it as is, because I think the logical clarity outweighs the nesting in this instance.

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
naamunds: I've added logic to throw an ArgumentNullException in the constructor if it's null.

scripts/dotnet-cli-build/PublishTargets.cs
naamunds: Good suggestion; I've updated it to use that syntax.

scripts/dotnet-cli-build/PublishTargets.cs
naamunds: We're controlling which CLI builds trigger DockerHub builds through the use of VSTS build variables.  I think it's better to control it in one place rather than in two places if possible.  If we add a CentOS Docker image in addition to our Ubuntu image, for example, I think it would be better to be able to just add the build variables to the CentOS build definition rather than also needing to add a target condition allowing it to run on CentOS.

Not having the Ubuntu condition also makes it easier to test locally.

scripts/dotnet-cli-build/PublishTargets.cs
naamunds: Done.

scripts/dotnet-cli-build/PublishTargets.cs
naamunds: Good idea.

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
naamunds: I talked to Michael about this offline, and he suggested using LINQ to make the code more readable. I do still prefer keeping the "else" because that seems more logical to me for this case. If using this method without specifying arguments for expected values were truly an edge case, then I would absolutely agree with not having an `else`, but I don't see it as such.

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
```diff
@@ -15,6 +15,10 @@ public EnvironmentAttribute(string envVar, params string[] expectedVals)
             {
                 throw new ArgumentNullException("envVar");
             }
+            if (expectedVals == null)
+            {
+                throw new ArgumentNullException("expectedVals");
```
MichaelSimons: The existing pattern was to use a hard coded string, but I would recommend updating the code to use the nameof operator.  Using the nameof operator in this case helps ensure the the correct argument names are used and helps keep the names in sync when refactoring.

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -118,6 +116,45 @@ public static BuildTargetResult PublishDebFileToDebianRepo(BuildTargetContext c)
...
+            c.Info("Triggering automated DockerHub builds for " + dockerHubRepo);
+            using (HttpClient client = new HttpClient())
+            {
+                StringContent requestContent = new StringContent("{\"build\": true}", Encoding.UTF8, "application/json");
+                try
+                {
+                    HttpResponseMessage response = client.PostAsync(dockerHubTriggerUri, requestContent).Result;
+                    if (!response.IsSuccessStatusCode)
+                    {
+                        var sb = new StringBuilder();
```
MichaelSimons: <nit> Be consistent in your coding style.  Why use `var` here but not when declaring `baseDockerHubUri`, `requestContent`, or `client`?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -118,6 +116,45 @@ public static BuildTargetResult PublishDebFileToDebianRepo(BuildTargetContext c)
...
+            string dockerHubTriggerToken = Environment.GetEnvironmentVariable("DOCKER_HUB_TRIGGER_TOKEN");
+
+            Uri baseDockerHubUri = new Uri("https://registry.hub.docker.com/u/");
+            Uri dockerHubTriggerUri;
+            if (!Uri.TryCreate(baseDockerHubUri, $"{dockerHubRepo}/trigger/{dockerHubTriggerToken}/", out dockerHubTriggerUri))
+            {
+                return c.Failed("Invalid DOCKER_HUB_REPO and/or DOCKER_HUB_TRIGGER_TOKEN");
+            }
+
+            c.Info("Triggering automated DockerHub builds for " + dockerHubRepo);
```
MichaelSimons: nit - This is another place to consider using the $ interpolated string syntax.  This would make the code more consistent.

scripts/Microsoft.DotNet.Cli.Build.Framework/TargetConditions/EnvironmentAttribute.cs
naamunds: Good idea. Fixed.

scripts/dotnet-cli-build/PublishTargets.cs
naamunds: Good point. I changed it to explicitly be `StringBuilder`.

scripts/dotnet-cli-build/PublishTargets.cs
naamunds: Fixed.

MichaelSimons:
:shipit:

naamunds:
Waiting for a second sign-off before merging.

Sridhar-MS:
:shipit:

PR #1838, opened 2016-03-14
Bump NuGet to 763

piotrpMSFT:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1838)


anurse:
We're way past 763 :). I think this is no longer relevant, eh?

PR #1841, opened 2016-03-14
Add native stuff to deps.json

### LINKED ISSUE #1828: deps.json file does not contain `native` assets
These should be present in runtime-specific targets, primarily for the standalone app case.

I've almost got a fix ready but wanted to merge the stuff I had so far and track this so we don't forget.

/cc @schellap @pakrym

anurse:
fixes #1828 

/cc @schellap @pakrym

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
+++ b/src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
@@ -147,7 +147,7 @@ namespace Microsoft.Extensions.DependencyModel
                 return;
             }
             libraryObject.Add(new JProperty(DependencyContextStrings.CompileTimeAssembliesKey,
-                 WriteAssemblies(compilationAssemblies))
+                 WriteAssetList(compilationAssemblies))
              );
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Extensions.DependencyModel/DependencyContextWriter.cs
```diff
@@ -197,6 +197,7 @@ private JObject WriteTargetLibrary(Library library)
                 AddDependencies(libraryObject, runtimeLibrary.Dependencies);
                 AddRuntimeAssemblies(libraryObject, runtimeLibrary.Assemblies);
                 AddResourceAssemblies(libraryObject, runtimeLibrary.ResourceAssemblies);
+                libraryObject.Add(DependencyContextStrings.NativeLibrariesKey, WriteAssetList(runtimeLibrary.NativeLibraries));
```
pakrym: Do a method with `if (libraries.Any())` check

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs
```diff
@@ -151,6 +151,7 @@ public void WritesRuntimeLibrariesToRuntimeTarget()
                                         "1.2.3",
                                         "HASH",
                                         new [] { RuntimeAssembly.Create("Banana.dll")},
+                                        new [] { "native.so" },
```
pakrym: Put it into subdir with wrong `\` to check that it is normalized

pakrym:
:shipit:

PR #1842, opened 2016-03-14
Use Microsoft.NETCore.App for sharedframework publish, pull local corehost for sharedhost

mellinoe:
Contains a few updates:

* Use Microsoft.NETCore.App instead of NETStandard.Library. This is the correct nuget package to be using to publish the shared framework.
* Use locally-copied corehost instead of pulling it from nuget for the shared host installers/zips.

Misc
* Throw an exception if a BuildContext property can't be retrieved @Sridhar-MS 
* Throw an exception if we can't parse the nuget version from the project.json file (means something is mismatched), instead of returning null.


scripts/dotnet-cli-build/SharedFrameworkTargets.cs
```diff
@@ -111,11 +116,15 @@ private static string GetVersionFromProjectJson(string pathToProjectJson)
                 }
             }
 
-            return null;
+            throw new InvalidOperationException("Unable to match the version name from " + pathToProjectJson);
         }
 
-        private static void CrossGenAllManagedAssemblies(string pathToAssemblies)
+        [Target]
+        [Environment("CROSSGEN_SHAREDFRAMEWORK")]
```
Sridhar-MS: Make sure that is run only when set to "true" or 1. https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/PublishTargets.cs#L40

anurse:
:shipit:

Make sure you're rebased, I just merged #1828 which should fix the issue @schellap was encountering where standalone apps weren't getting native assets in their deps files.

scripts/dotnet-cli-build/SharedFrameworkTargets.cs
```diff
@@ -33,7 +39,22 @@ public static BuildTargetResult PublishSharedFramework(BuildTargetContext c)
...
+            if (PlatformServices.Default.Runtime.OperatingSystemPlatform == Platform.Windows)
+            {
+                publishRuntime = $"win7-{PlatformServices.Default.Runtime.RuntimeArchitecture}";
+            }
+            else
+            {
+                publishRuntime = PlatformServices.Default.Runtime.GetRuntimeIdentifier();
+            }
+
+            DotNetCli.Stage0.Publish(
```
Sridhar-MS: Why use stage0 instead of stage2??

anurse: Yeah, this should use stage2

mellinoe: I think the original idea was to stay independent of the rest of the build (so that you could just run the package targets), but since we are pulling corehost from the build as well that seems unimportant now. I switched it.

PR #1849, opened 2016-03-14
Update files copied into shared framework staging directory

mellinoe:
Once https://github.com/dotnet/cli/pull/1792 goes in, we need to copy an extra file (hostfxr.dll) into the shared framework staging directory. We also need an additional file which is already there (hostpolicy.dll), but wasn't being copied.

NOTE: This will fail until #1792 is merged.


schellap:
@dotnet-bot test this please

schellap:
@mellinoe @anurse @Sridhar-MS @piotrpMSFT if anyone sees this green pls. merge... zzzzz

Issue #1688, opened 2016-03-03
Opt-In generation of RID fallback graph data in dotnet publish


anurse:
We need to support generating the RID fallback graph specified here: https://github.com/dotnet/core-proposals/blob/master/DotNetCli-RestoreAndPublish.md#runtime-graph-data within `dotnet publish` (since the shared framework will, itself, be `dotnet publish`ed).

1. We will add a switch to `dotnet publish`: `--include-rid-graph <rid1> --include-rid-graph <rid2> ...`.
2. When generating a deps file, we will load up all the `runtime.json` files specified in the dependency graph (available in the lock file) into NuGet's runtime graph object model
3. For each of the RIDs specified in the `--include-rid-graph` option, we will expand them to their full list and write the data to the deps file.

I'm assigning @pakrym since it relates to the deps file, but we can load balance over as things progress. This is a somewhat lower-pri issue because if we have to, we could just hand-write the file in v1. /cc @ellismg

blackdwarf:
@anurse ugh, those switches are wordy.

anurse:
They are only ever used by people building shared frameworks. I almost **want** them to be wordy to scare people off :)

blackdwarf:
@anurse ah, ok, that makes me feel better. :)

anurse:
FYI, for @pakrym : NuGet will not be generating the RID graph anymore. We'll have to do that ourselves. The NuGet APIs we're already using in ProjectModel should make that fairly easy. DependencyContext probably doesn't **have** to have the logic to do this, since it's only done during special invocations of `dotnet publish`. So as long as DependencyContext can carry the data, we can use NuGet APIs to generate it.

pakrym:
https://github.com/dotnet/cli/pull/1895

PR #1866, opened 2016-03-15
add manual merge of RID fallback graph for Windows

anurse:
Soon, we will have #1688, but until we do, we need to have a RID fallback graph in the shared framework deps file in order to boot Kestrel. This does that via a simple JSON fragment that is merged into the framework deps file after build. I expect the full solution to be available in the next day or so, but I want to get a build that can actually produce a fully working package :).

/cc @mellinoe @davidfowl @pakrym


pakrym:
:shipit:

anurse:
> 0 of 8 checks passed

Just because of the rebase. I had successes before then.

PR #1869, opened 2016-03-15
[DO NOT MERGE] Fix issues in corehost that prevented hellomvc from working

pakrym:
1. Rid graph travers goes to infinite loop because `iter++` changes internal state of iterator but returns old one
2. We need to merge rid and rid-less assets instead of selecting only one.

Code is horrible.

/cc @schellap @anurse

### DIFF
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -139,9 +139,14 @@ bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, c
         assert(!matched_rid.empty());
         for (auto iter = package.second.begin(); iter != package.second.end(); /* */)
         {
-            iter = (iter->first != matched_rid)
-                 ? package.second.erase(iter)
-                 : iter++;
+           if (iter->first != matched_rid)
+           {
+                   iter = package.second.erase(iter);
+           }
+           else
+           {
+                   iter++;
+           }
         }
     }
     return true;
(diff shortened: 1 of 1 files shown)


src/corehost/cli/deps_format.cpp
```diff
@@ -225,9 +230,21 @@ bool deps_json_t::load_portable(const json_value& json, const pal::string_t& tar
         return rid_assets.count(package) || non_rid_assets.count(package);
     };
     auto get_relpaths = [&rid_assets, &non_rid_assets](const pal::string_t& package, int type_index) -> const std::vector<pal::string_t>& {
-        return (rid_assets.count(package))
-            ? rid_assets[package].begin()->second[type_index]
-            : non_rid_assets[package][type_index];
+       // HACK:      v - memory leak
```
pakrym: :warning:

schellap:
It just needs to be ++iter. But this is fine.

src/corehost/cli/deps_format.cpp
schellap: Dont merge

pakrym: Wasn't going to, it's more to show issues and solutions to you.

schellap: The merge shoulf happpen already. Will TAL.

anurse:
These are super-hacks that @pakrym and I bashed out to wrap up a few small things. This PR is illustrative, not an actual fix :).

schellap:
Thanks for the testing it out, guys!

PR #1873, opened 2016-03-15
Add hostfxr.dll to windows host msi

mellinoe:
hostfxr.dll needs to be added to the host installer. Just a one line change here, verified locally.


Sridhar-MS:
:shipit:

Issue #1797, opened 2016-03-10
Enable specifying output assembly name in compiler options


moozzyk:
Currently the name of the assembly produced by dotnet build is taken from `Project.Name`. `Project.Name` is also used as a name of the package. As a result the assembly name in a package is always the same as the package name. For tools we don't want packages to be called `dotnet-` and hence need a way to provide the name of the output assembly. The idea is to add an option to `compilerOptions` and use it if it is specified.

Related https://github.com/aspnet/IISIntegration/issues/95

moozzyk:
Fixed with ca74a9b17756a284667e56dcc10959c6397b1624

PR #1877, opened 2016-03-15
Enable specifying output assembly name in compiler options

moozzyk:
Addresses #1797

### DIFF
--- a/src/Microsoft.DotNet.Compiler.Common/CommonCompilerOptionsExtensions.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/CommonCompilerOptionsExtensions.cs
@@ -40,6 +40,8 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         internal static readonly OptionTemplate s_additionalArgumentsTemplate = new OptionTemplate("additional-argument");
 
+        internal static readonly OptionTemplate s_outputNameTemplate = new OptionTemplate("output-name");
+
         public static CommonCompilerOptions Parse(ArgumentSyntax syntax)
         {
             IReadOnlyList<string> defines = null;
@@ -55,6 +57,7 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             bool? publicSign = null;
             bool? emitEntryPoint = null;
             bool? generateXmlDocumentation = null;
+            string outputName = null;
             IReadOnlyList<string> additionalArguments = null;
 
             Func<string, bool?> nullableBoolConverter = v => bool.Parse(v);
@@ -97,6 +100,8 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             syntax.DefineOption(s_generateXmlDocumentation.LongName, ref generateXmlDocumentation,
                     nullableBoolConverter, "Generate XML documentation file");
 
+            syntax.DefineOption(s_outputNameTemplate.LongName, ref outputName, "Output assembly name");
+
             return new CommonCompilerOptions
             {
                 Defines = defines,
@@ -112,6 +117,7 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
                 DebugType = debugType,
                 EmitEntryPoint = emitEntryPoint,
                 GenerateXmlDocumentation = generateXmlDocumentation,
+                OutputName = outputName,
                 AdditionalArguments = additionalArguments
             };
         }
@@ -131,6 +137,7 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             var publicSign = options.PublicSign;
             var emitEntryPoint = options.EmitEntryPoint;
             var generateXmlDocumentation = options.GenerateXmlDocumentation;
+            var outputName = options.OutputName;
             var additionalArguments = options.AdditionalArguments;
 
             var args = new List<string>();
@@ -203,6 +210,11 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             if (generateXmlDocumentation != null)
             {
                 args.Add(s_generateXmlDocumentation.ToLongArg(generateXmlDocumentation));
+            }
+
+            if (outputName != null)
+            {
+                args.Add(s_outputNameTemplate.ToLongArg(outputName));
             }
 
             return args;
--- a/src/Microsoft.DotNet.ProjectModel/RuntimeOutputFiles.cs
+++ b/src/Microsoft.DotNet.ProjectModel/RuntimeOutputFiles.cs
@@ -37,12 +37,10 @@ namespace Microsoft.DotNet.ProjectModel
                     // The executable is a DLL in this case
                     extension = FileNameSuffixes.DotNet.DynamicLib;
                 }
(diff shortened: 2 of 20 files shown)


moozzyk:
Recreating https://github.com/dotnet/cli/pull/1807 to use new CI

TestAssets/TestProjects/LibraryWithOuputAssemblyName/project.json
```diff
@@ -0,0 +1,13 @@
+{
```
brthor: nit: typo in the directory here

moozzyk: will fix

test/dotnet-compile.Tests/CompilerTests.cs
```diff
@@ -167,6 +167,24 @@ public void DepsDependencyContextIsValidOnBuild()
...
+                    .WithLockFiles();
+
+            var root = testInstance.TestRoot;
+            var outputDir = Path.Combine(root, "bin");
+            var testProject = ProjectUtils.GetProjectJson(root, "LibraryWithOutputAssemblyName");
+            var buildCommand = new BuildCommand(testProject, output: outputDir, framework: DefaultFramework);
+            var result = buildCommand.ExecuteWithCapturedOutput();
+            result.Should().Pass();
+
+            new DirectoryInfo(outputDir).Should().HaveFiles(new [] { "MyLibrary.dll" });
```
pakrym: Can you check that misc files like *.deps also have correct names for executable?

moozzyk: OK - I assume that I will have to use my package as a dependency to check that. Correct? Any other files/scenarios you would recommend verifying?

pakrym: Just make it an app not a library

src/Microsoft.DotNet.Compiler.Common/CommonCompilerOptionsExtensions.cs
```diff
@@ -205,6 +212,11 @@ public static IEnumerable<string> SerializeToArgs(this CommonCompilerOptions opt
                 args.Add(s_generateXmlDocumentation.ToLongArg(generateXmlDocumentation));
             }
 
+            if (outputName != null)
```
troydai: Could it be empty?

troydai:
:shipit:

test/dotnet-compile.Tests/CompilerTests.cs
moozzyk: @pakrym I tried it and I can see that deps now has: 
```
        "runtime": {
          "MyApp.dll": {}
        }
```
Is this what needs to be checked, right?

(on a separate note when creating an app the dll is named as per the requested output name but the .exe itself has the package name - I think we want .exe to have the same name as specified in the outputName right @davidfowl ? if so, I will investigate this separately...)

moozzyk: Actually I also see this in the deps file: 
  "libraries": {
    "LibraryWithOutputAssemblyName/1.0.0": {
      "type": "project",
      "serviceable": false,
      "sha512": ""
    },

and also the same dll on the disk so am not sure if this is right...

pakrym: > (on a separate note when creating an app the dll is named as per the requested output name but the .exe itself has the package name - I think we want .exe to have the same name as specified in the outputName right @davidfowl ? if so, I will investigate this separately...)

That's what I meant, .exe, .deps.json should be named as your assembly name.

moozzyk: OK. I will look into it but I would prefer to do this separately. Tools are not executable so merging this change would unblock further work on tools while I am working on executables.

src/Microsoft.DotNet.Compiler.Common/CommonCompilerOptionsExtensions.cs
moozzyk: Empty values are being checked elsewhere. The error message is not great as it does relate exactly to the name from project.json (`option --output-name requires a value`) - but this is consistent with other options (I think I checked the xmlDoc option)

test/dotnet-compile.Tests/CompilerTests.cs
davidfowl: We should really fix that though, I rather not have anything like this be broken...

moozzyk: In that case I will update the PR with another commit that fixes things for exes

moozzyk:
:up::date:
- added support for outputName in apps
- updated publish to write the correct .exe name

pakrym:
Can we abstract `compilationOptions.OutputName ?? context.ProjectFile.Name` somewhere? It has 4 usages in this PR

moozzyk:
I thought about this but am not sure if it is right to mix compilation options with ProjectContext...

pakrym:
Maybe we can set it to CompilationOptions

pakrym:
Accidental close, sorry

moozzyk:
Folded `ProjectFile.Name ?? compilerOptions.OutputName` to `compilerOptions.OutputName`.

src/Microsoft.DotNet.ProjectModel/RuntimeOutputFiles.cs
```diff
@@ -37,8 +37,12 @@ public string Executable
                     // The executable is a DLL in this case
                     extension = FileNameSuffixes.DotNet.DynamicLib;
                 }
+                
+                var compilationOptions = Project.GetCompilerOptions(Framework, Configuration);
 
-                return Path.Combine(BasePath, Project.Name + extension);
+                return Path.Combine(
+                    BasePath,
+                    (compilationOptions.OutputName ?? Project.Name) + extension);
```
pakrym: What about this usage?

moozzyk: Fixing. Must have had different pattern and VS Code did not find it...

troydai:
Still :shipit:

moozzyk:
Some CI failures are not related to this change

CentOs:
the tool /mnt/resource/j/workspace/dotnet_cli/rel_1.0.0/debug_centos7.1_x64_prtest/.dotnet_stage0/Linux/share/dotnet/cli/bin/dotnet cannot be found

Windows:
20:41:35 dotnet : The term 'dotnet' is not recognized as the name of a cmdlet, 
20:41:35 function, script file, or operable program. Check the spelling of the name, or 
20:41:35 if a path was included, verify that the path is correct and try again.

moozzyk:
Rebased on the latest changes but Ubuntu CI still fails: 

10:29:51 Restoring Build Script projects...
10:29:51 /opt/code/scripts/run-build.sh: line 105: dotnet: command not found

pakrym:
:shipit:

PR #1878, opened 2016-03-15
Update iterator properly and merge rid/non-rid assets

schellap:
@gkhanna79 PTAL, minor clean up.

Cc @anurse @pakrym


gkhanna79:
LGTM

src/corehost/cli/deps_format.cpp
```diff
@@ -139,9 +139,9 @@ bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, c
         assert(!matched_rid.empty());
         for (auto iter = package.second.begin(); iter != package.second.end(); /* */)
         {
-            iter = (iter->first != matched_rid)
-                 ? package.second.erase(iter)
-                 : iter++;
+            (iter->first != matched_rid)
```
pakrym: I'd prefer `if` here because we are not using return value of `?:` for anything.

PR #1879, opened 2016-03-15
Add a directory-delete helper.

mellinoe:
This seems to be needed on Windows to delete temporary staging files correctly.

@Sridhar-MS


Sridhar-MS:
:shipit:

PR #1880, opened 2016-03-15
Fix the VSO signing builds

Sridhar-MS:
As a part of Package init create all the SharedFx, SharedHost and CLI SDK
layouts. This way all other package targets can take a dependency only on
InitPackage.

cc: @mellinoe


Sridhar-MS:
:shipit:

mellinoe:
:shipit:

PR #1883, opened 2016-03-15
Fix a couple build issues on OSX/non-Windows

mellinoe:
We were downloading the wrong 'latest version' of the CLI, and targets passed into --targets were not split correctly, resulting in some errors like "target Prepare Compile didn't exist".

@Sridhar-MS @brthor


Sridhar-MS:
:shipit:

brthor:
:shipit:

PR #1885, opened 2016-03-15
Layout the CLI SDK in the "1.0.0" instead of "1.0.0-beta-001793"

Sridhar-MS:
dotnet host muxer does not support pre-release CLI SDK when no global.json
file is present. So fixing the CLI layout to be in production version
layout. THIS IS A TEMP HACK.

cc: @mellinoe @Petermarcu


Sridhar-MS:
Created issue #1886 to track this.

Petermarcu:
Do it

PR #1889, opened 2016-03-15
Enable crossgen for shared framework, provide opt-out

mellinoe:
crossgen fails on `Microsoft.CodeAnalysis.dll` for a few reasons, so that assembly is currently skipped. This step is now on by default, and there is an opt-out variable which can be set to skip crossgen for all assemblies in the shared framework.


mellinoe:
I'll put up a new PR with this when I've updated it. We need to update the shared framework version once a new version is available that has a change in Roslyn.

PR #1892, opened 2016-03-16
Create combined zip/tar for framework-host-sdk and framework-host

mellinoe:
This creates two new zip files produced by the build. The names for these are messy, we should come up with something better.

* Combined zip/tar containing the shared framework, host, and SDK
* Combined zip/tar containing the shared framework and host

The second commit adds some steps that publish these to Azure.

@piotrpMSFT @Sridhar-MS

### DIFF
--- a/scripts/dotnet-cli-build/PackageTargets.cs
+++ b/scripts/dotnet-cli-build/PackageTargets.cs
@@ -14,7 +14,9 @@ namespace Microsoft.DotNet.Cli.Build
     {
         [Target(nameof(PackageTargets.CopyCLISDKLayout),
         nameof(SharedFrameworkTargets.PublishSharedHost),
-        nameof(SharedFrameworkTargets.PublishSharedFramework))]
+        nameof(SharedFrameworkTargets.PublishSharedFramework),
+        nameof(PackageTargets.CopyCombinedFrameworkSDKHostLayout),
+        nameof(PackageTargets.CopyCombinedFrameworkHostLayout))]
         public static BuildTargetResult InitPackage(BuildTargetContext c)
         {
             Directory.CreateDirectory(Dirs.Packages);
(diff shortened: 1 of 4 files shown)


scripts/dotnet-cli-build/PackageTargets.cs
```diff
@@ -90,6 +92,39 @@ public static BuildTargetResult CopyCLISDKLayout(BuildTargetContext c)
...
 
+        [Target]
+        public static BuildTargetResult CopyCombinedFrameworkSDKHostLayout(BuildTargetContext c)
+        {
+            var combinedRoot = Path.Combine(Dirs.Output, "obj", "combined-framework-sdk-host");
+
+            string sdkPublishRoot = c.BuildContext.Get<string>("CLISDKRoot");
+            Utils.CopyDirectoryRecursively(sdkPublishRoot, combinedRoot);
+
+            string sharedFrameworkPublishRoot = c.BuildContext.Get<string>("SharedFrameworkPublishRoot");
```
Sridhar-MS: Why dont you just call `CopyCombinedFrameworkHostLayout` from here?

Sridhar-MS: Or even add it as dependent target.

Sridhar-MS: They are getting copied to different dirs. Ignore my previous comments.

anurse:
Can we update the README.md links as well to point at these new combined ZIPs? Also the sooner we get this in the better so we can begin testing all the other scenarios :)

anurse:
You can ignore the Ubuntu failures. If you're ready to go, let's get this checked in. Ignore my comment about the README, I'll do another PR to update the README and the install scripts (temporarily, since @krwq is working on a refactor of those) to use this new combined package so all the other workflows can be unblocked.

PR #1893, opened 2016-03-16
Add initial update-dependencies script.

eerhardt:
This script can be used to automatically update dotnet/cli's
dependencies on other repos like CoreFX, NuGet, etc.

Currently the script supports updating the CoreFX dependencies.

@piotrpMSFT @livarcocc @anurse @Sridhar-MS

### DIFF
--- a/scripts/update-dependencies/UpdateFilesTargets.cs
+++ b/scripts/update-dependencies/UpdateFilesTargets.cs
@@ -1,0 +1,233 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.DotNet.Cli.Build.Framework;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Scripts
+{
+    public static class UpdateFilesTargets
+    {
+        private static HttpClient s_client = new HttpClient();
+
+        [Target(nameof(GetDependencies), nameof(ReplaceVersions))]
+        public static BuildTargetResult UpdateFiles(BuildTargetContext c) => c.Success();
+
+        /// <summary>
+        /// Gets all the dependency information and puts it in the build properties.
+        /// </summary>
+        [Target]
+        public static BuildTargetResult GetDependencies(BuildTargetContext c)
+        {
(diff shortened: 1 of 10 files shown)


scripts/update-dependencies/UpdateFilesTargets.cs
```diff
@@ -0,0 +1,233 @@
...
+        [Target(nameof(GetDependencies), nameof(ReplaceVersions))]
+        public static BuildTargetResult UpdateFiles(BuildTargetContext c) => c.Success();
+
+        /// <summary>
+        /// Gets all the dependency information and puts it in the build properties.
+        /// </summary>
+        [Target]
+        public static BuildTargetResult GetDependencies(BuildTargetContext c)
+        {
+            string coreFxLkgVersion = s_client.GetStringAsync("https://raw.githubusercontent.com/eerhardt/versions/master/corefx/release/1.0.0-rc2/LKG.txt").Result;
```
eerhardt: Oops.  Missed this one that needs to be parameterized.  I'll update this in the morning to read this from an Environment Variable.

scripts/update-dependencies/project.json
```diff
@@ -0,0 +1,25 @@
...
+    "NuGet.Versioning": "3.4.0-rtm-0764",
+    "Newtonsoft.Json": "7.0.1",
+    "Octokit": "0.18.0",
+    "Microsoft.Net.Http": "2.2.29"
+  },
+  "frameworks": {
+    "netstandardapp1.5": {
+      "imports": [
+        "dnxcore50",
+        "portable-net45"
```
anurse: I think this has to be `portable-net45+win8`. If there's a package that has `portable-net45+[something else]` it is not guaranteed to work in `netstandardapp1.5`. Is there a package that falls into that bucket?

anurse:
LGTM. Minor concern about the portable import but otherwise good.

scripts/update-dependencies/project.json
eerhardt: Yes, the package that falls into this bucket is:  https://www.nuget.org/packages/Octokit/0.18.0, which we are using to create the PR on GitHub.  It has "portable-net45+wp80+win+wpa81" from looking at the package.

It happens to be working both on my machine and in the VSO build. I guess I haven't tried it on a Unix machine... Do you think I should add the full portable string?

eerhardt: Octokit also has "netcore451" as a target... FYI

anurse: How about `portable-net45+win`. That should be enough. For what it's worth, I think we've engaged with Octokit to get it ported to `netstandard` (but like most people they're waiting for things to stabilize)

eerhardt:
@dotnet-bot test this please

PR #1899, opened 2016-03-16
Revert "Update the .xproj files to reference the "DotNet" targets and props instead of "DNX"."

eerhardt:
This reverts commit 4304ca9848eced94a0f39829fd31b2a3cf601bec.

It turns out we can't make this change yet because the VS tooling is not public, so any external contributor wouldn't be able to open our projects.

@blackdwarf @anurse @livarcocc @evorine


anurse:
:shipit:

PR #1900, opened 2016-03-16
Trying to use corehost again

anurse:
I think corehost should be working. Let's find out why it's not in the debian package, if that's the case.

Woo! PRs to PRs! I just want to get a Ubuntu build going and this is the easiest way to do that. I'm also spinning up my docker machine to try it locally.


anurse:
@dotnet-bot test this please

PR #1913, opened 2016-03-16
update NuGet

anurse:
`type: platform`! At least I think so (/cc @emgarten)

/cc @piotrpMSFT @davidfowl @pakrym


pakrym:
Where is you `include` hack?

anurse:
It's there now.

anurse:
We're going to wait on some NuGet fixes first.

PR #1915, opened 2016-03-16
Revert merging relpaths and relax erroring

schellap:
@pakrym PTAL

Cc @anurse @gkhanna79


schellap:
@pakrym, this should unblock you. Basically, the merge should be done specific to an asset type, i.e., `native` can come from rid specific but `runtime` from non-rid specific asset. So if you just compared against rid specific asset for that type and returned empty, you will miss the other (the non-rid one.)

pakrym:
Look's good, merge it, and i'll try it tomorrow morning.

PR #1918, opened 2016-03-16
[WIP]  Remove DependencyModel csv reader, add lots of null checks everywhere

pakrym:
@anurse


test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextJsonWriterTests.cs
```diff
@@ -40,7 +40,7 @@ public JObject Save(DependencyContext dependencyContext)
             IReadOnlyList<RuntimeFallbacks> runtimeGraph = null)
         {
             return new DependencyContext(
-                            target ?? string.Empty,
+                            target ?? "DefaultTarget",
```
anurse: Hrm... I think I'd almost rather a `null`y value (like `string.Empty`) than some arbitrarily-named default we never expect someone to actually use...

anurse: Never mind, test code. Ignore me.

anurse:
:shipit:

PR #1920, opened 2016-03-17
Incorporate the new CLI layout in builds scripts and add Portable Project Tool commands

Sridhar-MS:
- Layout the Sdk, SharedFx and SharedHost in the stage1 and stage2 dirs.
- Create only 2 zips/tar files 
   - `dotnet` - contains shared fx and shared host.
   - `dotnet-dev` - contains shared fx, cli sdk and shared host.
- There is still work that needs to be done for the installers.
  - Need to create a `dotnet` redist for both osx and windows. This will have sharedfx and shared host.
  - Need to create a `dotnet-dev` installer for osx. This will contains sharedfx, shared host and cli sdk.

cc: @piotrpMSFT @brthor

PR #1930, opened 2016-03-17
Tools on the Shared FX and using NuGet tool restore

brthor:
Changes are all over the place here. 
Have this working locally but is blocked on #1920 before going in.

The weird commits here are because I needed latest from rel that #1920 hadn't picked up yet. As that PR gets closer to merging, these will shake out.

Putting up primarily for feedback in the meantime.

cc @joelverhagen @piotrpMSFT @Sridhar-MS @schellap @anurse @pakrym


src/dotnet/commands/dotnet-pack/PackageGenerator.cs
```diff
@@ -94,6 +94,7 @@ protected virtual void ProcessContext(ProjectContext context)
 
             TryAddOutputFile(context, inputFolder, outputName);
             TryAddOutputFile(context, inputFolder, $"{Project.Name}.xml");
+            TryAddOutputFile(context, inputFolder, $"{Project.Name}.runtimeconfig.json");
```
brthor: cc @piotrpMSFT @anurse 
Picking up runtimeconfig.json in pack if it is present to solve the dilemma with packing Portable Apps.
Deps Json is generated after restore.

brthor:
#1920 had a dependency on this so merging the PRs and closing this one

PR #1931, opened 2016-03-17
[WIP] Patch lock file with lock file fragments

cdmihai:
This refers to the scenario where VS drives the build and there are xproj projects referencing csproj projects.

When a project.json project references a csproj project, nuget cannot follow the dependency path toward the csproj. Those dependencies are only available at build time. Therefore, there will be a special msbuild target that generates a fragment lock file describing all the csproj projects that an xproj depends upon.

By convention, the fragment lock file is placed besides the master lock file. If the fragment is there, LockfileReader merges with it. 

A tentative fragment lock file format is here: https://gist.github.com/cdmihai/3ebda8bd9c52e33f235a

TODOs:
- [ ] make new library description for msbuild projects. The current code marks it as unresolved because path points to a csproj and not a project.json
- [ ] extract merging algorithm into new file and update to use an immutable LockFile
- [ ] write unit tests
- [ ] update documentation
- [ ] detect if lock file has holes and did not get merged with a fragment. Saves the user from weird unresolved dependencies errors.

@davidfowl 
@abpiskunov

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/Graph/IMergeable.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Graph/IMergeable.cs
@@ -1,0 +1,7 @@
+namespace Microsoft.DotNet.ProjectModel.Graph
+{
+    public interface IMergeable<T>
+    {
+        void MergeWith(T m);
+    }
+}
--- a/src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
@@ -4,13 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
-using Microsoft.DotNet.ProjectModel.Utilities;
-using NuGet.Versioning;
 
 namespace Microsoft.DotNet.ProjectModel.Graph
 {
-    public class LockFile
+    public class LockFile : IMergeable<LockFile>
     {
         public static readonly int CurrentVersion = 2;
         public static readonly string FileName = "project.lock.json";
@@ -86,6 +83,19 @@ namespace Microsoft.DotNet.ProjectModel.Graph
 
             message = null;
             return true;
+        }
+
+        public void MergeWith(LockFile fragmentLockFile)
+        {
+            if (Version != fragmentLockFile.Version)
+            {
+                throw new ArgumentException($"Cannot merge master lock file {LockFilePath} with fragment lock file {fragmentLockFile.LockFilePath} due to missmatching versions");
+            }
+
+            Targets.MergeWith(fragmentLockFile.Targets, l => l.TargetFramework.DotNetFrameworkName + l.RuntimeIdentifier);
+            ProjectLibraries.MergeWith(fragmentLockFile.ProjectLibraries, pl => pl.Name + pl.Version);
+            PackageLibraries.MergeWith(fragmentLockFile.PackageLibraries, pl => pl.Name + pl.Version);
+            ProjectFileDependencyGroups.MergeWith(fragmentLockFile.ProjectFileDependencyGroups, pdg => pdg.FrameworkName?.DotNetFrameworkName ?? "");
         }
     }
 }
--- a/src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
@@ -11,6 +11,7 @@ namespace Microsoft.DotNet.ProjectModel.Graph
     {
         public static readonly int CurrentVersion = 2;
         public static readonly string FileName = "project.lock.json";
+        public static readonly string FragmentFileName = "project.fragment.lock.json";
 
         public string LockFilePath { get; }
 
--- a/src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
@@ -116,14 +116,32 @@ namespace Microsoft.DotNet.ProjectModel.Graph
 
                 if (type == null || string.Equals(type, "package", StringComparison.OrdinalIgnoreCase))
                 {
-                    lockFile.PackageLibraries.Add(new LockFilePackageLibrary
+
+                    var msbuildProject = value.Value("msbuildProject");
+                    LockFilePackageLibrary packageLibrary;
+
+                    if (msbuildProject == null)
                     {
-                        Name = name,
-                        Version = version,
-                        IsServiceable = ReadBool(value, "serviceable", defaultValue: false),
-                        Sha512 = ReadString(value.Value("sha512")),
-                        Files = ReadPathArray(value.Value("files"), ReadString)
-                    });
+                        packageLibrary = new LockFilePackageLibrary
+                        {
+                            Name = name,
+                            Version = version,
+                            IsServiceable = ReadBool(value, "serviceable", defaultValue: false),
+                            Sha512 = ReadString(value.Value("sha512")),
+                            Files = ReadPathArray(value.Value("files"), ReadString)
+                        };
+                    }
+                    else
+                    {
+                        packageLibrary = new LockFilePackageLibrary
+                        {
+                            Name = name,
+                            Version = version,
+                            MSBuildProject = ReadString(msbuildProject)
+                        };
+                    }
+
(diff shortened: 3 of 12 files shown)


src/Microsoft.DotNet.ProjectModel/Graph/IMergeable.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.DotNet.ProjectModel.Graph
+{
+    public interface IMergeable<T>
+    {
+        void MergeWith(T m);
```
cdmihai: Should this just be MergeWithFragment, to reduce the scope of only merging with fragments?

src/Microsoft.DotNet.ProjectModel/Graph/MergeableExtensions.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NuGet.Packaging;
+
+namespace Microsoft.DotNet.ProjectModel.Graph
+{
+    internal static class CollectionExtensions
```
cdmihai: Match file and class name.

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -98,7 +98,7 @@ public static ProjectContextBuilder CreateBuilder(string projectPath, NuGetFrame
         /// <summary>
         /// Creates a project context for each framework located in the project at <paramref name="projectPath"/>
         /// </summary>
-        public static IEnumerable<ProjectContext> CreateContextForEachFramework(string projectPath, ProjectReaderSettings settings = null, IEnumerable<string> runtimeIdentifiers = null)
+        public static IEnumerable<ProjectContext> CreateContextForEachFramework(string projectPath, ProjectReaderSettings settings = null, IEnumerable<string> runtimeIdentifiers = null, IEnumerable<string> dependencyFragments = null)
```
cdmihai: Should I propagate the fragments to all the create methods, or just the ones that build and publish use? The fragment scenario is specific to VS builds only.

src/Microsoft.DotNet.ProjectModel/Graph/LockFileTargetLibrary.cs
```diff
@@ -34,5 +35,50 @@ public class LockFileTargetLibrary
         public IList<LockFileContentFile> ContentFiles { get; set; } = new List<LockFileContentFile>();
 
         public IList<LockFileRuntimeTarget> RuntimeTargets { get; set; } = new List<LockFileRuntimeTarget>();
+
+        public void MergeWith(LockFileTargetLibrary m)
+        {
+            TargetFramework = TargetFramework ?? m.TargetFramework;
+
+            if (Dependencies == null || !Dependencies.Any())
```
cdmihai: @davidfowl @abpiskunov 

Should I union these, blindly overwrite, or keep them as is?

src/Microsoft.DotNet.ProjectModel/Graph/IMergeable.cs
```diff
@@ -0,0 +1,7 @@
+namespace Microsoft.DotNet.ProjectModel.Graph
+{
+    public interface IMergeable<T>
```
davidfowl: Internal AF

cdmihai:
Pending update:
Account for implicit and explicit fragments.

Implicit fragments: referenced from the main lock file
Explicit fragments: referenced through command line args. Will add to the implicit ones.

src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
```diff
@@ -87,5 +84,18 @@ public bool IsValidForProject(Project project, out string message)
...
         }
+
+        public void MergeWith(LockFile fragmentLockFile)
+        {
+            if (Version != fragmentLockFile.Version)
+            {
+                throw new ArgumentException($"Cannot merge master lock file {LockFilePath} with fragment lock file {fragmentLockFile.LockFilePath} due to missmatching versions");
+            }
+
+            Targets.MergeWith(fragmentLockFile.Targets, l => l.TargetFramework.Framework + l.RuntimeIdentifier);
```
anurse: there's a `/` separating the framework and RID

src/Microsoft.DotNet.ProjectModel/Graph/LockFileReader.cs
```diff
@@ -105,12 +105,17 @@ private static void ReadLibrary(JsonObject json, LockFile lockFile)
...
-                    lockFile.ProjectLibraries.Add(new LockFileProjectLibrary
+                    var projectLibrary = new LockFileProjectLibrary
                     {
                         Name = name,
                         Version = version,
                         Path = ReadString(value.Value("path"))
-                    });
+                    };
+
+                    var buildTimeDependencyValue = value.Value("buildTimeDependency");
```
anurse: Didn't see this in the example. What's the expected value of this property?

src/Microsoft.DotNet.ProjectModel/Graph/LockFileTargetLibrary.cs
anurse: Overwrite I'd say, or don't change it at all. There should only be one source of truth for dependencies if possible.

pakrym:
Please not mutate a lock file, I'm gona do a change to make it immutable so we're able to cache it.

src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
anurse: I see now that this is just for de-duping, but still, I'd add a separator just in case, and you may as well use the one used in the lock file.

src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
```diff
@@ -87,5 +84,18 @@ public bool IsValidForProject(Project project, out string message)
...
+            ProjectLibraries.MergeWith(fragmentLockFile.ProjectLibraries, pl => pl.Name + pl.Version);
+            PackageLibraries.MergeWith(fragmentLockFile.PackageLibraries, pl => pl.Name + pl.Version);
```
anurse: Same here. Make the key match the Lock File property name by putting a `/` between Name and Version

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
anurse: Just the ones used by build and publish. Users can always use a `ProjectContextBuilder` directly

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -38,6 +38,8 @@ public class ProjectContextBuilder
 
         private ProjectReaderSettings Settings { get; set; } = ProjectReaderSettings.ReadFromEnvironment();
 
+        private IEnumerable<string> Dependencyfragments { get; set; }
```
anurse: Dependency**F**ragments

cdmihai:
@pakrym 
Well, since the LockFile now also needs to merge the contents of another one, I guess it no longer needs to be mutable.

Alternatively, in LockFile.MergeWith() I can clone the current object and mutate the clone.

src/Microsoft.DotNet.ProjectModel/Graph/LockFile.cs
abpiskunov: When main lock has this: "targets": {    ".NETFramework,Version=v4.6": {

and fragment has   "targets": {    ".NETFramework,Version=v4.5.2": {

after it merges it would create 2 targets sections? and how it would work after that? My target app has target net46, my reference app has target 4.5.2, they are compatible but different. Should main lock file add 4.5.2 under 4.6 ?

abpiskunov:
When you say fragment merges with main lock file, are you really trying to merge and update a main lock "file"? Or you are talking about updating a context when both files are read and merged in memory then build/publish uses merge data?

Otherwise if you are trying to update lock file, it would produce a warning form ProjectModel.Server that restore is needed and tooling would probably kick another restore which would mess up everything.

cdmihai:
@abpiskunov 
The two files are merged in memory. Both fragments are loaded into LockFile objects, and the objects get merged. Nothing happens at file level. When the project model reads the lock file, it will get transparently merged with the fragment at load time.

cdmihai:
Closing PR. Starting from scratch.

PR #1934, opened 2016-03-17

eerhardt:


eerhardt:
Oops.  This came from the wrong branch. Doing it again against the correct branch.

PR #1939, opened 2016-03-18
Chain in VC Redist

ellismg:
- Chain in the redist.
- Clean up how WiX is downloaded (use HttpClient instead of shelling out to PowerShell).

I validated that if the CRT is uninstalled, the bundle will install it.  I also validated that uninstalling the bundle does not remove the CRT.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1939)


packaging/windows/bundle.wxs
```diff
@@ -32,6 +32,13 @@
     <util:DirectorySearch Path="[PreviousInstallFolder]" Variable="DOTNETHOME" After="PreviousInstallFolderSearch" Condition="PreviousInstallFolder" />
 
     <Chain DisableSystemRestore="yes" ParallelCache="yes">
+      <ExePackage SourceFile="$(var.CrtRedistributableSourcePath)"
+                  DetectCondition='(VCRedist_KeyExists AND VCRedist_KeyValue &gt;= "v14.0.23506.00")'
+                  InstallCommand="/q /norestart"
+                  RepairCommand="/q/ /norestart /repair"
```
akoeplinger: extra slash after `/q`

scripts/dotnet-cli-build/MsiTargets.cs
```diff
@@ -14,6 +15,11 @@ public class MsiTargets
     {
         private const string ENGINE = "engine.exe";
 
+        private static Dictionary<string, string> CrtFwLinkMap { get; } = new Dictionary<string, string>(StringComparer.Ordinal) {
+            {"x86", "http://go.microsoft.com/fwlink/?LinkID=615459&amp;clcid=0x409"},
+            {"x64", "http://go.microsoft.com/fwlink/?LinkID=615460&amp;clcid=0x409"},
```
akoeplinger: https?

scripts/dotnet-cli-build/MsiTargets.cs
```diff
@@ -50,14 +64,42 @@ private static void AcquireWix(BuildTargetContext c)
...
 
             c.Info("Downloading WixTools..");
-            // Download Wix version 3.10.2 - https://wix.codeplex.com/releases/view/619491
-            Cmd("powershell", "-NoProfile", "-NoLogo",
-                $"Invoke-WebRequest -Uri https://wix.codeplex.com/downloads/get/1540241 -Method Get -OutFile {WixRoot}\\WixTools.zip")
-                    .Execute()
-                    .EnsureSuccessful();
+
+            // Download Wix version 3.10.2 - http://wix.codeplex.com/releases/view/619491
+            DownloadFile("http://wix.codeplex.com/downloads/get/1540241", Path.Combine(WixRoot, "WixTools.zip"));
```
akoeplinger: https to match what it was before

ellismg: Indeed :-(

packaging/windows/bundle.wxs
ellismg: Thanks.  Fixed,

scripts/dotnet-cli-build/MsiTargets.cs
ellismg: Same problem as with codeplex, sadly.

ellismg:
This will need some work once https://github.com/dotnet/cli/pull/1920 lands, but I don't expect the core of the idea changes much, we'll just be chaining into different bundles.

scripts/dotnet-cli-build/MsiTargets.cs
```diff
@@ -14,6 +15,11 @@ public class MsiTargets
     {
         private const string ENGINE = "engine.exe";
 
+        private static Dictionary<string, string> CrtFwLinkMap { get; } = new Dictionary<string, string>(StringComparer.Ordinal) {
+            {"x86", "http://go.microsoft.com/fwlink/?LinkID=615459&amp;clcid=0x409"},
```
nil4: Should `&amp;` be used here?

ellismg: Good call, fixed (I pulled these links out of some XML files).

ellismg:
I decided to just take the CRT and WiX versions we depend on and put them on our Azure blob storage instead of downloaded them each time.  This has the bonus of allowing us to use HTTPS instead of this questionable "HTTPS that 302's you to HTTP" stuff.

packaging/windows/bundle.wxs
```diff
@@ -32,6 +32,13 @@
     <util:DirectorySearch Path="[PreviousInstallFolder]" Variable="DOTNETHOME" After="PreviousInstallFolderSearch" Condition="PreviousInstallFolder" />
 
     <Chain DisableSystemRestore="yes" ParallelCache="yes">
+      <ExePackage SourceFile="$(var.CrtRedistributableSourcePath)"
+                  DetectCondition='(VCRedist_KeyExists AND VCRedist_KeyValue &gt;= "v14.0.23506.00")'
+                  InstallCommand="/q /norestart"
+                  RepairCommand="/q /norestart /repair"
```
robmen: I wouldn't add a `RepairCommand`. At first glance, it looks like the right thing to do but what it really means is that repairing your product means repairing the dependency that is really an entity all to its own. If the VC Redist is broken, better to let the user repair it directly.

scripts/dotnet-cli-build/MsiTargets.cs
```diff
@@ -60,6 +69,20 @@ private static void AcquireWix(BuildTargetContext c)
...
+            if (File.Exists(CrtRedistributablePath))
+            {
+                return;
+            }
+
+            Directory.CreateDirectory(Path.GetDirectoryName(CrtRedistributablePath));
+
+            c.Info($"Downloading VC_redist.{Arch}.{CrtVersion}.exe..");
+
+            DownloadFile($"https://dotnetcli.blob.core.windows.net/build/vc/{Arch}/VC_redist.{Arch}.{CrtVersion}.exe", CrtRedistributablePath);
```
robmen: Instead of this, you could consider just letting the bundle always download the VC Redist from this location or MSFT (depending which you trust more). Given most of your customers probably already have VC Redist installed, the download will usually get skipped and you won't need to have the bits as part of the build, just use the `RemotePayload` element.

ellismg: @piotrpMSFT @gkhanna79 @Petermarcu 

What do you think about the above?  I believe this means you'd need an internet connection in the case where you didn't have the redist installed and went to install the CLI.  It would save us about 10 megs in the bundle, I believe.

gkhanna79: I do agree that we should get VCRedist from the MSFt location and not our CLI Blob - otherwise, we will be responsible for updating it and could endup installing a version that is older than what is available on VC Website.

> I believe this means you'd need an internet connection in the case where you didn't have the redist installed and went to install the CLI

How is the CLI acquired when you dont have Internet connection?

ellismg: > I do agree that we should get VCRedist from the MSFt location and not our CLI Blob - otherwise, we will be responsible for updating it and could endup installing a version that is older than what is available on VC Website.

Is there a well known URL for "give me the latest version of the Dev 14 CRT?"  The FW Links I've found are specific to a version of the CRT.  Not re-hosting it also means we don't ensure the dependency is transmitted over HTTPS end to end, which is a little frustrating.  It also makes us dependent on more external resources, which we don't want.

> How is the CLI acquired when you dont have Internet connection?

Having an internet connection when I download the CLI is different from a requirement to have an internet connection when I go to install the product.  Maybe I downloaded the CLI before getting on an airplane and then go to install it after we've taken off.

gkhanna79: @ellismg I am not familiar with our setup technologies but would have thought there would be a way for CLI setup, when invoked, to download the VC++ redist, independent of our download source (the Azure blob). Is that not possible?

For the aircraft case, perhaps we could call out that we will need to download dependencies and would need an internet connection. This is similar to other setups (E.g. VS) where dependencies are downloaded over Internet and VS wouldnt just install if you only were to download the initial web installer.

The primary question is - how are we intending to align our setup?

piotrpMSFT:
For the download size question, @Petermarcu @richlander @blackdwarf. I think it's fine to download the CRT package live as this matches the Ubuntu and OS X experiences. What do others think?

As to wether or not we cache the installer on our own BLOB storage, my knee-jerk preference is to go with @ellismg's plan. I've not been able to find any documentation regarding a logical 'latest' link for CRT installers, promises about the retention policy for their binaries, or backwards-compatibility guarantees for their installers' command line parameters and behavior. Do such documents exist? Can we make such dependencies on the CRT's installers?

piotrpMSFT:
@ellismg has this been rebased? CI should be green at this point.

scripts/dotnet-cli-build/MsiTargets.cs
robmen: The bundle supports the `-layout` switch which will download everything that you can install later without internet connectivity.

Also, chances are most developers working with dotnet/cli will already have VC Redist, right?

ellismg:
> @ellismg has this been rebased? CI should be green at this point.

I am going to rebase shortly and update to use RemotePayload for the CRTs.

blackdwarf:
@piotrpMSFT what is the size question? I only see @ellismg comment about the 10MB. Is that it?

ellismg:
@piotrpMSFT @Sridhar-MS 

Rebased, moved to use RemotePayload for the CRT and added the CRT to both bundles.  I validated manually on a clean Windows 7 VM that the CRT is installed when not present when installing both bundles.

Can you take another look and also let me know if you want this for RC2?

gkhanna79:
This is required for RC2. @blackdwarf for folks who are downloading the zip archives of CLI, are we giving documentation to have VC++ 2015 REDIST preinstalled?

blackdwarf:
@gkhanna79 nope. Only for installers. We need a page that explains the install script usage in simple terms. Can you create an issue for that assign it to me?

ellismg:
@dotnet-bot test this please (Jenkins didn't actually launch many of the legs)

ellismg:
@mellinoe @Sridhar-MS @piotrpMSFT 

I think this is ready to go now, green CI, but it still needs a Code Review.  Once I get a LGTM, I'll start the ask mode process so I can merge this.

blackdwarf:
@ellismg I have to ask: what's with the closings and openings? :)

ellismg:
@blackdwarf The CI system was not launching some legs, even with a ask to restest, opening and closing makes the CI system see it as a new thing and start over.

packaging/windows/clisdk/bundle.wxs
```diff
@@ -45,6 +62,21 @@
...
+       Key="SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\$(var.Platform)"
+       Value="Version"
+       Result="exists" />
+    <util:RegistrySearch
+       Id="VCRedist_KeyValue"
+       Variable="VCRedist_KeyValue"
+       Root="HKLM"
+       Key="SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\$(var.Platform)"
+       Value="Version"
+       Result="value" />
```
Sridhar-MS: Do you need `Win64` attribute here (since we have an x86 installer too)? 

http://wixtoolset.org/documentation/manual/v3/xsd/wix/registrysearch.html

packaging/windows/sharedframework/bundle.wxs
```diff
@@ -38,6 +55,21 @@
...
+       Value="Version"
+       Result="exists" />
+    <util:RegistrySearch
+       Id="VCRedist_KeyValue"
+       Variable="VCRedist_KeyValue"
+       Root="HKLM"
+       Key="SOFTWARE\Microsoft\VisualStudio\14.0\VC\Runtimes\$(var.Platform)"
+       Value="Version"
+       Result="value" />
+
```
Sridhar-MS: Same question here for `Win64` attribute.

packaging/windows/clisdk/bundle.wxs
ellismg: Good catch.  We always want to search the 32-bit registry (even the 64 bit CRT writes there) so I will need to be explicit here.

ellismg:
@Sridhar-MS thanks for the review.  I addressed your feedback.  Can you take another look?

ellismg:
@dotnet-bot test Windows_NT x86 Debug Build please (file locking issue when trying to remove file)

Sridhar-MS:
:shipit:

packaging/windows/inc/crtvariables.wxi
```diff
@@ -0,0 +1,25 @@
...
+
+  <?define Crt_CertificatePublicKey="52868DFCA6E3AF2632389E6C1EE7D0468D3797D0"?>
+  <?define Crt_CertificateThumbprint="3BDA323E552DB1FDE5F4FBEE75D6D5B2B187EEDC"?>
+  <?define Crt_Description="Microsoft Visual C++ 2015 Redistributable ($(var.Platform)) - 14.0.23506"?>
+  <?define Crt_ProductName="Microsoft Visual C++ 2015 Redistributable ($(var.Platform)) - 14.0.23506"?>
+  <?define Crt_Version="14.0.23506.0"?>
+  <?define Crt_VersionRegistyKeyValue="v14.0.23506.00"?>
+
+  <?if $(var.Platform)=x86?>
+  <?define Crt_DownloadUrl="http://go.microsoft.com/fwlink/?LinkID=615459&amp;clcid=0x409"?>
```
joshfree: any reason these are hardcoded to English?

scripts/dotnet-cli-build/MsiTargets.cs
```diff
@@ -52,14 +55,24 @@ private static void AcquireWix(BuildTargetContext c)
             Directory.CreateDirectory(WixRoot);
 
             c.Info("Downloading WixTools..");
-            // Download Wix version 3.10.2 - https://wix.codeplex.com/releases/view/619491
-            Cmd("powershell", "-NoProfile", "-NoLogo",
-                $"Invoke-WebRequest -Uri https://wix.codeplex.com/downloads/get/1540241 -Method Get -OutFile {WixRoot}\\WixTools.zip")
-                    .Execute()
-                    .EnsureSuccessful();
+
+            DownloadFile($"https://dotnetcli.blob.core.windows.net/build/wix/wix.{WixVersion}.zip", Path.Combine(WixRoot, "WixTools.zip"));
```
joshfree: nit: consider Path.Combine for WixVersion

ellismg: We're building a URL here, not a path, so I don't think Path.Combine is the best thing to do here.

packaging/windows/inc/crtvariables.wxi
ellismg: @robmen Should I move either Description or ProductName (or both) into a wxl file so they can be localized?

I'll follow up with another PR if so.

robmen: If you want to localize them, definitely. In general, I minimize what goes in to .wxi files and preprocessor in general since people tend to very quickly abuse the whole thing. If a value isn't shared, just put it where it goes.

scripts/dotnet-cli-build/MsiTargets.cs
robmen: @ellismg :+1:

PR #1942, opened 2016-03-18
Add bsymbolic flags to linux linker

schellap:
The problem with not having these flags are:

`libhostfxr.so` and `libhostpolicy.so` would still keep using common functions that are defined in `dotnet.exe` effectively voiding a servicing override. Of course, servicing of common code would service both `dotnet.exe` and `libhostpolicy.so`, but this is needed to match Windows for consistency and is easier to reason about.

### DIFF
--- a/src/corehost/cli/setup.cmake
+++ b/src/corehost/cli/setup.cmake
@@ -37,6 +37,12 @@ else()
     add_compile_options(-Wno-unused-local-typedef)
 endif()
 
+# This is required to map a symbol reference to a matching definition local to the module (.so)
+# containing the reference instead of using definitions from other modules.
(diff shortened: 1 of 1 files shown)


schellap:
@gkhanna79 PTAL, kind of important for servicing.

gkhanna79:
How are we validating that with this flag, resolution of symbols is correct? Also, is similar things required for osx?
@janvorli what are your thoughts?

schellap:
I wrote hand test cases by adding some logging to common functions, without the flags, replacing just the hostfxr and hostpolicy had no effect wrt logging. With the flags as expected, they produced different logs.

schellap:
This is specific to GNU, not OSX binaries.

PR #1943, opened 2016-03-18
Runtime config knobs for CLR

schellap:
@gkhanna79 @adityamandaleeka @kouvel  @anurse


Documentation/specs/runtime-configuration-file.md
```diff
@@ -117,8 +122,8 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files.
 
 This section is copied verbatim from an identical section in the input `project.json` file (with the exception of the `target` parameter which is generated by the compilation process). The `runtimeConfig` section specifies parameters to be provided to the runtime during initialization. Known parameters include:
 
-* `gcServer` - Boolean indicating if the server GC should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/ms229357.aspx) setting)
-* `gcConcurrent` - Boolean indicating if background garbage collection should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).
+* `System.GC.Server` - Boolean indicating if the server GC should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/ms229357.aspx) setting)
```
kouvel: Instead of describing the options here, it might be better to have a link to https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/clr-configuration-knobs.md, which will be updated to describe each exposed option. A link in the default config file may also be useful.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -25,8 +25,13 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files.
...
-        "gcConcurrent": false,
+
+        "System.GC.Server": true,
+        "System.GC.Concurrent": false,
+        "System.Threading.ThreadPool.MinThreads": 4,
+        "System.Threading.ThreadPool.MaxThreads": 8,
+        "System.Threading.Thread.UseAllCpuGroups": true
+
         "framework": {
             "name": "Microsoft.DotNetCore",
```
schellap: Would it be better to put these under `properties` subsection, so the host will simply take all keys and `ToString` the values and pass it to coreclr. This bypasses type check on the values specified and catching errors early but helps a older version of the host to still pass newer properties to coreclr as it doesn't get into the business of parsing known names and typed values but simply calls `ToString` on them.

Cc @jkotas

Documentation/specs/runtime-configuration-file.md
```diff
@@ -117,8 +122,8 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files.
...
 
-* `gcServer` - Boolean indicating if the server GC should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/ms229357.aspx) setting)
-* `gcConcurrent` - Boolean indicating if background garbage collection should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).
+* `System.GC.Server` - Boolean indicating if the server GC should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/ms229357.aspx) setting)
+* `System.GC.Concurrent` - Boolean indicating if background garbage collection should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).
```
anurse: > Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).

This isn't really true anymore, at least not exactly. Originally the name I used was designed to match the `app.config` setting exactly. I do like the new name, but maybe rephrase to something like:

> This is equivalent to the `gcConcurrent` setting in the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx)

(Same for GC Server above, and the other new settings if they have equivalents).

Having the link between old and new settings will be useful for documenting later

Documentation/specs/runtime-configuration-file.md
kouvel: I think ToString would be fine, the component the config applies to can choose whether to ignore or how to issue an error.

Also, since we plan on exposing these properties eventually, is there a plan to remove reference to "runtime"? I think a "properties" section is fine instead "runtimeOptions", but what about the file name?

kouvel: Exposing through API for use by other components I mean

anurse: FYI the `runtimeOptions` section is also defined in `project.json` and the format is expected to match exactly. It is covered in some of the specs @DamianEdwards wrote, so if we're changing stuff **outside** of that section we should make sure to consider that. Within the `runtimeOptions` section, we can :tada: :arrow_down:

Documentation/specs/runtime-configuration-file.md
```diff
@@ -25,8 +25,15 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files.
 ```json
 {
     "runtimeOptions": {
-        "gcServer": true,
-        "gcConcurrent": false,
+
+        "configProperties": {
+            "System.GC.Server": true,
```
davidfowl: Why are these so verbose?

Documentation/specs/runtime-configuration-file.md
```diff
@@ -116,9 +123,12 @@ The files are both JSON files stored in UTF-8 encoding. Below are sample files.
...
-
-* `gcServer` - Boolean indicating if the server GC should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/ms229357.aspx) setting)
-* `gcConcurrent` - Boolean indicating if background garbage collection should be used (Default: _TBD_). Note: This is designed to mirror the existing [app.config](https://msdn.microsoft.com/en-us/library/yhwwzef8.aspx) setting).
+* `configProperties` - Indicates configuration properties that are available during the execution of the application.
+  * Examples:
+    * Full list of [configuration properties](https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/clr-configuration-knobs.md) for CoreCLR.
+    * `System.GC.Server` (old: `gcServer`) - Boolean indicating if the server GC should be used (Default: `true`).
+    * `System.GC.Concurrent` (old: `gcConcurrent`) - Boolean indicating if background garbage collection should be used (Default: `false`).
+    * `Fabrikam.App.TimeoutInSeconds` (_TBD_) - An application specified property that is available in managed code, during the execution of the application.
```
davidfowl: What is this?

Documentation/specs/runtime-configuration-file.md
schellap: I guess the idea is these might eventually be exposed as static properties for managed code to use, and namespacing is a good idea anyway so an app property doesn't interfere with runtime properties.

Documentation/specs/runtime-configuration-file.md
schellap: We are getting `GetData` to support `APP_CONTEXT_DEPS_FILES`. So theoretically an application can choose to put a config setting that is available at execution time. Cc @weshaggard

weshaggard: We added AppContext.GetData to enable passing things from the host to the application like APP_CONTEXT_DEPS_FILES but I wasn't aware we are planning to expose a full configuration system through here. Correct me if I'm wrong but the full framework doesn't expose such properties through AppDomain.GetData today does it? I would be a little hesitant to just add this without a scenario because it will end up getting abused as the new application configuration model and people will start asking for features which we never really intended to support here. 

cc @AlexGhiondea

schellap: I thought the `GetData` was going to be a generic purpose way of obtaining properties as we also use the API to initialize coreclr properties.

I will remove it if we do not intend to support custom properties.

kouvel: There was discussion on adding a way of exposing these config properties through a public API, but GetData may not be the right place for it. I think for now, GetData should not expose any of the new properties, and we should discuss further how to expose the new properties.

Documentation/specs/runtime-configuration-file.md
davidfowl: Who's idea?

jkotas: in full framework, there is a ton of random `COMPlus_` config switches (https://github.com/dotnet/coreclr/blob/master/src/inc/clrconfigvalues.h). These config switches can be controlled via config file as well. There is desire to bring some of these config switches to .NET Core.

The problem with what we have is:
- The names are very random (e.g. `gcConcurrent`, `Thread_UseAllCpuGroups`). My suggestion was to use the `Namespace.Name` to give it some reasonable naming convention.
- The boundaries between runtime and framework are blurry. E.g. Is threadpool runtime or framework? In CoreCLR, it is in coreclr.dll (-> runtime). In CoreRT, it is in separate .dll (-> framework). My suggestion was to not to differentiate between the two.

jkotas: @kouvel Is the doc with the proposal for these published somewhere?

piotrpMSFT:
@schellap, rebased to get rid of fixed build issues.

Documentation/specs/runtime-configuration-file.md
davidfowl: I don't want people to use this as a general config abstraction because it isn't that. More data about the naming scheme and why it's relevant would be good

piotrpMSFT:
@schellap why did this get merged without successful CI? We just finished picking builds up off the floor after last week's hackathon. We need to get back to keeping builds green.

I'll keep an eye on this CI run despite the merge. If it causes breaks I will revert the PR.

schellap:
This only changes documentation.

piotrpMSFT:
We're not doing YOLO merges anymore :)

Documentation/specs/runtime-configuration-file.md
schellap: @gkhanna79 @piotrpMSFT, See `Fabrikam.App.TimeoutInSeconds` above.

PR #1951, opened 2016-03-18
Pass DockerHub env vars to container used for build

naamunds:
These environment variables need to be set for the
TriggerDockerHubBuilds build target declared in PublishTargets.cs.

This fixes https://github.com/dotnet/dotnet-docker-preview/issues/4.

cc: @MichaelSimons @brthor @Sridhar-MS


MichaelSimons:
:shipit:

PR #1972, opened 2016-03-21
Use parts of nuget project and library model

pakrym:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1972)


piotrpMSFT:
What's the relationship between NuGet's project model and CLI's? Do we intend to keep both?

pakrym:
We are keeping both, but using lock file graph and some other classes from Nuget.

Microsoft.DotNet.Cli.sln
```diff
@@ -83,6 +83,50 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "update-dependencies", "scri
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.DotNet.Cli.Utils.Tests", "test\Microsoft.DotNet.Cli.Utils.Tests\Microsoft.DotNet.Cli.Utils.Tests.xproj", "{09C52F96-EFDD-4448-95EC-6D362DD60BAA}"
 EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "NuGet.Common", "..\..\nuget\NuGet.Client\src\NuGet.Core\NuGet.Common\NuGet.Common.xproj", "{98BEE375-A5A0-4FC2-9B7A-25DB41C8045D}"
```
brthor: Should these be in the solution?

If so, why 2 directories back vs `..\nuget` ?

global.json
```diff
@@ -1,3 +1,3 @@
 {
-  "projects": [ "src", "test" ]
+  "projects": [ "src", "test", "d:\\github\\nuget\\NuGet.Client\\src\\NuGet.Core\\" ]
```
brthor: This path shouldn't be absolute.

src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
```diff
@@ -95,7 +95,7 @@ public CommandSpec Resolve(CommandResolverArguments commandResolverArguments)
         }
 
         private CommandSpec ResolveCommandSpecFromToolLibrary(
-            LibraryRange toolLibrary,
+            ProjectLibraryDependency toolLibrary,
```
brthor: Why not still LibraryRange? NuGet seems to have a similar type.

Honestly should probably be https://github.com/NuGet/NuGet.Client/blob/3a0ee64c457cb9f4522da91812a48942a6122132/src/NuGet.Core/NuGet.LibraryModel/ToolDependency.cs

now that it's there.

pakrym:
Please don't review this, code is not cleaned up yet

pakrym:
Ready for review except nuget versions in `package.json` files

pakrym:
Ready for review except nuget versions in `package.json` files

src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
```diff
@@ -76,14 +75,14 @@ public CommandSpec Resolve(CommandResolverArguments commandResolverArguments)
         }
 
         private CommandSpec ResolveCommandSpecFromAllToolLibraries(
-            IEnumerable<LibraryRange> toolsLibraries,
+            IEnumerable<ProjectLibraryDependency> toolsLibraries,
```
brthor: Why `ProjectLibraryDependency`?

brthor:
LGTM :shipit:

src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
pakrym: Because `ProjectFile.Tools` are `ProjectLibraryDependency`, it was easier to do transformation to `LibraryRange` in this method.

pakrym:
Out of date

PR #1982, opened 2016-03-21
Update install scripts

krwq:
reopenning this PR as CI seems to be never checking the old PR

### DIFF
--- a/scripts/obtain/install.sh
+++ b/scripts/obtain/install.sh
@@ -397,7 +397,7 @@ extract_and_override_zip() {
     local out_path=$2
     
     local failed=false
-    tar -xzf "$zip_path" -C "$out_path" --overwrite > /dev/null || failed=true
+    tar -xzf "$zip_path" -C "$out_path" > /dev/null || failed=true
     if [ "$failed" == true ]; then
         say_err "Extraction failed"
         return 1
--- a/scripts/obtain/install.sh
+++ b/scripts/obtain/install.sh
@@ -334,8 +334,6 @@ get_user_share_path() {
     
     if [ ! -z "${DOTNET_INSTALL_DIR:-}" ]; then
         echo $DOTNET_INSTALL_DIR
-    elif [ ! -z "${PREFIX:-}" ]; then
-        echo $PREFIX
     else
         echo "/usr/local/share/dotnet"
     fi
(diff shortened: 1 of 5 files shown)


scripts/obtain/install.sh
```diff
@@ -23,44 +40,76 @@ if [ -t 1 ]; then
...
+    
+    which "$1" > /dev/null 2>&1
     return $?
 }
 
-#Not 100% sure at the moment that these checks are enough. We might need to take version into account or do something
-#more complicated. This seemed like a good beginning though as it should catch the default "clean" machine case and give
-#people an appropriate hint.
+check_min_reqs() {
+    if ! machine_has "curl"; then
```
ellismg: In a bunch of scripts in CoreFX and CoreCLR, we also try to use wget if it's present (but curl is not) which is common in some distros.  Can we do that here as well?

scripts/obtain/install.sh
```diff
@@ -4,8 +4,25 @@
 
 # Note: This script should be compatible with the dash shell used in Ubuntu. So avoid bashisms! See https://wiki.ubuntu.com/DashAsBinSh for more info
 
-#setup some colors to use. These need to work in fairly limited shells, like the Ubuntu Docker container where there are only 8 colors.
-#See if stdout is a terminal
+# Stop script on NZEC
```
ellismg: Have we considered putting a shebang at the top of this file so things work when we are using a non sh compatible shell?  You could at the same time use "bash" as the target of the shebang and not have to worry about doing dash compatible stuff.

scripts/obtain/install.sh
krwq: i agree it should be here but let's do these changes separately

scripts/obtain/install.sh
```diff
@@ -68,208 +117,493 @@ check_pre_reqs() {
             LDCONFIG_COMMAND="ldconfig"
         fi
 
-        [ -z "$($LDCONFIG_COMMAND -p | grep libunwind)" ] && say_err "Unable to locate libunwind. Install libunwind to continue" && _failing=true
-        [ -z "$($LDCONFIG_COMMAND -p | grep libssl)" ] && say_err "Unable to locate libssl. Install libssl to continue" && _failing=true
-        [ -z "$($LDCONFIG_COMMAND -p | grep libcurl)" ] && say_err "Unable to locate libcurl. Install libcurl to continue" && _failing=true
-        [ -z "$($LDCONFIG_COMMAND -p | grep libicu)" ] && say_err "Unable to locate libicu. Install libicu to continue" && _failing=true
-        [ -z "$($LDCONFIG_COMMAND -p | grep gettext)" ] && say_err "Unable to locate gettext. Install gettext to continue" && _failing=true
+        [ -z "$($LDCONFIG_COMMAND -p | grep libunwind)" ] && say_err "Unable to locate libunwind. Install libunwind to continue" && failing=true
```
brthor: Would it be preferable to give a more actionable message here?

Something like `Run apt-get install libunwind8` is what I have in mind

krwq: i think libunwind8 is Ubuntu specific. I will work on that separately

brthor: libunwind8 is Ubuntu specific and that was the point I was making, the acquisition for this on centos would be different, you ran into this yourself

scripts/obtain/install.sh
krwq: I'm not sure I understand, do you mean to create few copies of the script for different shells or completely not support dash and only leave bash?

ellismg: Neither.

If you put:

`#! /usr/bin/env bash`

As the first line of this script, when invoked, it will be run using bash.  Even on Ubuntu 14.04.

brthor: @ellismg what happens if bash isn't available on whatever distro?

Would `#!/usr/bin/env sh` be more portable?

ellismg: Assuming you're willing to not use bashims, that would be better.  FWIW, CoreCLR and CoreFX and WCF all require bash to build, but maybe taking some pain to have one less dependency to install is worth it.

PR #1987, opened 2016-03-22
Add support for runtime config

schellap:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1987)

### DIFF
--- a/src/corehost/cli/runtime_config.cpp
+++ b/src/corehost/cli/runtime_config.cpp
@@ -7,49 +7,55 @@
 #include "runtime_config.h"
 #include <cassert>
 
-typedef web::json::value json_value;
-
 runtime_config_t::runtime_config_t(const pal::string_t& path)
     : m_fx_roll_fwd(true)
     , m_path(path)
     , m_portable(false)
-    , m_gc_server(_X("0"))
 {
     m_valid = ensure_parsed();
 } 
 
-void parse_fx(const json_value& opts, pal::string_t* name, pal::string_t* version, bool* roll_fwd, bool* portable)
+bool runtime_config_t::parse_opts(const json_value& opts)
 {
-    name->clear();
-    version->clear();
-    *roll_fwd = true;
-    *portable = false;
-
     if (opts.is_null())
     {
-        return;
+        return true;
     }
 
     const auto& opts_obj = opts.as_object();
+    
+    auto properties = opts_obj.find(_X("configProperties"));
+    if (properties != opts_obj.end())
+    {
+        const auto& prop_obj = properties->second.as_object();
+        for (const auto& property : prop_obj)
+        {
+            m_properties[property.first] = property.second.is_string()
+                ? property.second.as_string()
+                : property.second.to_string();
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


schellap:
@gkhanna79 PTAL.

Cc @adityamandaleeka

src/corehost/cli/runtime_config.cpp
```diff
@@ -112,3 +112,18 @@ bool runtime_config_t::get_portable() const
...
+    for (const auto& kv : m_properties)
+    {
+        keys->push_back(pal::to_stdstring(kv.first));
+        values->push_back(pal::to_stdstring(kv.second));
+    }
+
+    if (m_properties.count(_STRINGIFY(System_GC_Server)) == 0)
+    {
+        keys->push_back(System_GC_Server);
+        values->push_back("true"); // Default true
```
gkhanna79: Why are we defaulting server GC to true?

src/corehost/cli/runtime_config.cpp
```diff
@@ -112,3 +112,18 @@ bool runtime_config_t::get_portable() const
...
+
+void runtime_config_t::config_kv(std::vector<std::string>* keys, std::vector<std::string>* values) const
+{
+    for (const auto& kv : m_properties)
+    {
+        keys->push_back(pal::to_stdstring(kv.first));
+        values->push_back(pal::to_stdstring(kv.second));
+    }
+
+    if (m_properties.count(_STRINGIFY(System_GC_Server)) == 0)
```
gkhanna79: This would mean we will need to update the host everytime a new property (e.g. min or max threads) would need to to be added. Is that right? Any reason we cannot just read the values and set them as configuration settings without knowing about them?

CC @sergiy-k @kouvel

schellap: @kouvel there was a request to turn on `SERVER_GC` at one point. I am wondering, can the coreclr.dll simply default to it being `true` if unspecified?

src/corehost/cli/runtime_config.cpp
kouvel: Server GC can't be on by default, it's not applicable for regular apps. The request was to turn it on by default for ASP.NET apps in the project template.

src/corehost/cli/runtime_config.cpp
adityamandaleeka: @gkhanna79 The host doesn't need to know about properties that are added. It can just pass the strings as they are unless (as in this case) it wants to explicitly force something if it wasn't set.

I don't know the details of that request, but if we really want to change the default behavior of CoreCLR  when server GC is not set explicitly, the place to do that would be in the definition of the CLRConfigValue for it (in clrconfigvalues.h). Once this system is in place, we will fall back on the defaults specified there if something is not set (for now, since we're respecting both methods of configuration for now).

schellap: Based on @kouvel 's comment, ASP.NET can configure this now the way want, so I am removing the host trying to know about any default init properties. Thanks guys!

PR #1989, opened 2016-03-22
AnsiConsole color handling

Peter-Schneider:
PR #1660 has alredy been merged - but there are actually two `AnsiConsole` classes (one in namespace `Microsoft.DotNet.Cli.Build.Framework` and the other in namespace `Microsoft.DotNet.Cli.Utils`.
With this PR both classes handle the colors in the same way.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1989)


Peter-Schneider:
Sorry.. this PR is not as tidy as is should be.. I'm closing it and send a new one..

PR #1991, opened 2016-03-22
AnsiConsole color handling

Peter-Schneider:
PR #1660 has alredy been merged - but there are actually two `AnsiConsole` classes (one in namespace `Microsoft.DotNet.Cli.Build.Framework` (which has been patched with the previous change) and the other in namespace `Microsoft.DotNet.Cli.Utils`.
With this PR both classes handle the colors in the same way.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1991)


brthor:
Pending CI :shipit:

PR #1999, opened 2016-03-22
Fix Manpage Path

brthor:
Dependent on #1890 but don't want to kill that test run, so opening up separately.

Puts the dotnet manage in the host deb package and the rest in the sdk package

cc @piotrpMSFT

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/1999)


brthor:
merged to #1890

PR #2016, opened 2016-03-23
Add runtime signature into deps.json

pakrym:
@anurse @schellap

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2016)

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
+++ b/src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
 using Microsoft.DotNet.ProjectModel;
 using Microsoft.DotNet.ProjectModel.Compilation;
 using Microsoft.DotNet.ProjectModel.Graph;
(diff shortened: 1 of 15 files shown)


src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
```diff
@@ -48,16 +50,38 @@ public DependencyContextBuilder(string referenceAssembliesPath)
...
+        private static string GenerateRuntimeSignature(IEnumerable<LibraryExport> runtimeExports)
+        {
+            var sha1 = SHA1.Create();
+            var builder = new StringBuilder();
+            var packages = runtimeExports
+                .Where(libraryExport => libraryExport.Library.Identity.Type == LibraryType.Package);
+            foreach (var libraryExport in packages)
+            {
+                builder.Append(libraryExport.Library.Identity.Name);
+                builder.Append(libraryExport.Library.Identity.Version);
```
anurse: Could you add a separator between each individual library, just to be safe? A newline would be sufficient, so just changing this to `AppendLine`

anurse:
:shipit: on the managed code. Native code looks fine but wait for @schellap 's sign off on that. Also wait for whatever craziness is going on with the builds :)

src/corehost/cli/deps_format.cpp
```diff
@@ -318,8 +318,11 @@ bool deps_json_t::load(bool portable, const pal::string_t& deps_path, const rid_
     {
         const auto json = json_value::parse(file);
 
-        const auto& runtime_target = json.at(_X("runtimeTarget"));
-        const pal::string_t& name = runtime_target.as_string();
+		const auto& runtime_target = json.at(_X("runtimeTarget"));
+
+        const pal::string_t& name = (runtime_target.type() == json_value::value_type::String)?
+			runtime_target.as_string():
+			runtime_target.at(_X("name")).as_string();
```
schellap: NIT: Whitespace issues.

`runtime_target.is_string() ?`

is simpler. Otherwise, .cpp looks good to me.

anurse:
FYI https://github.com/dotnet/cli/pull/2019 should fix the build issues

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextBuilderTests.cs
```diff
@@ -291,6 +291,19 @@ public void SkipsBuildDependencies()
...
+        [Fact]
+        public void GeneratesRuntimeSignatureOutOfPackageNamesAndVersions()
+        {
+            var context = Build(runtimeExports: new[]
+            {
+                Export(PackageDescription("Pack.Age", new NuGetVersion(1, 2, 3))),
+                Export(PackageDescription("Pack.Age", new NuGetVersion(1, 2, 3))),
+            });
+
+            var lib = context.Target.RuntimeSignature.Should().Be("85aa350d4f334de42348cdfc5e09f4ae9a328eeb");
```
davidfowl: lib?

schellap:
Just wanted to make sure you are adding signature of serviceable packages... a small optimization.

schellap:
@pakrym, @anurse, @davidfowl, any thoughts about just doing it for serviceable case or be generic.

anurse:
We just did it for all packages. It may mean occasionally over-breadcrumbing but I think that's OK.

PR #2021, opened 2016-03-24
Logging and case format fix

schellap:
Cc @eerhardt 

@gkhanna79 PTAL.

At one point we spec'ed the deps.json in the fx would be called `netcoreapp.deps.json`, so I cautiously did a `to_lower`, but now the name has changed to be a namespace `Microsoft.NETCore.App.deps.json`, which is case sensitive. This is an issue for OSX and Ubuntu.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2021)

### DIFF
--- a/src/corehost/cli/args.h
+++ b/src/corehost/cli/args.h
@@ -15,8 +15,6 @@ struct arguments_t
     pal::string_t app_dir;
     pal::string_t deps_path;
     pal::string_t dotnet_servicing;
-    pal::string_t dotnet_runtime_servicing;
-    pal::string_t dotnet_home;
     pal::string_t probe_dir;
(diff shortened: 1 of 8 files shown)


eerhardt:
Once the build errors on Windows are fixed :shipit: 

BTW - for my knowledge, how do I turn on the tracing?

schellap:
`set COREHOST_TRACE=1`

I'm going to spend some time during ask mode to really nail the logging, what we have now is not perfect.

eerhardt:
Test Windows_2016 x64 Release Build

eerhardt:
This looks good - only Win 2016 x64 test is left.  Merging in order to move #1990 along.

Issue #2009, opened 2016-03-23
Project directories containing a dot cannot resolve coreclr path


joelverhagen:
Project directories containing a dot cannot resolve coreclr path. A folder name without a dot (a.k.a. period, full stop) works just fine.

I used the `1.0.0-beta-001937` Windows x64 installer from the README:
https://dotnetcli.blob.core.windows.net/dotnet/beta/Installers/Latest/dotnet-dev-win-x64.latest.exe

This seems like a really basic problem, so perhaps it is a silly user issue :smile:.

## Steps to reproduce

```
mkdir foo.bar
cd foo.bar
dotnet new
dotnet restore
dotnet run
```

## Expected  behavior

```
Hello World!
```

## Actual behavior
```
Could not resolve coreclr path
```

## Environment data
`dotnet --version` output:

```
.NET Command Line Tools (1.0.0-beta-001937)

Product Information:
 Version:     1.0.0-beta-001937
 Commit Sha:  3616378130

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 Runtime Id:  win10-x64
```

colemickens:
I was experiencing this but was unaware that it was due to the period. This affects 1.0.1.001589-1 for Linux as well.

brthor:
fairly strange cc @schellap any ideas?

schellap:
This is a corehost bug. Thanks for reporting it. PR #2021 should address this.

mmalolepszy:
@schellap Will this also address #1988?

schellap:
Can this be closed now?

joelverhagen:
Yup, I've tried with newer builds and things seem to be working. Feel free to close.

PR #2023, opened 2016-03-24
Remove a needless dot elimination

schellap:
Fixes https://github.com/dotnet/cli/issues/2009
Cc @brthor, @joelverhagen


schellap:
Assumes a runtime config is not present and executes as a standalone app.

PR #2029, opened 2016-03-24
update to new nuget and turn inference on

anurse:
Updates us to NuGet 1068, which disables RID inference by default. So, for now `dotnet restore` turns it back on explicitly (this is part of staging the changes). The next phase will be to remove the defaulting and ensure that the CLI can be a portable app (but still provide the switch for other CLI consumers to use while converting).

/cc @emgarten @pakrym @piotrpMSFT @eerhardt

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2029)


pakrym:
:shipit:

eerhardt:
:+1:

emgarten:
:rocket:

anurse:
Merged without Windows 2016 because SLLLLLOOOOOWWWW (and we're turning it off anyway)

PR #2033, opened 2016-03-24
add back docker hub builds trigger

brthor:
cc @naamunds @MichaelSimons

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2033)


MichaelSimons:
:shipit: - Thanks for fixing this.

naamunds:
I think it would be a good idea to add dependencies on `PublishDebFilesToDebianRepo` and `PublishSdkInstallerFileToAzure` to `TriggerDockerHubBuilds`. Do you want to do that in this change, or should I do it in a separate change?

brthor:
build timed out
@dotnet-bot test Ubuntu x64 Debug Build please

naamunds:
Actually `TriggerDockerHubBuilds` is not dependent on `PublishDebFilesToDebianRepo`, but will be dependent on all three installer files being published to Azure (@MichaelSimons is planning to make a change to dotnet-docker-preview to install the other two Debian files in addition to the CLI SDK package). I can make a change to add the `TriggerDockerHubBuilds` dependencies separately.

This PR LGTM.

brthor:
@naamunds I don't know what we gain from moving the trigger, but am not opposed to it :smile:

brthor:
@MichaelSimons When installing the .debs via dpkg, you'll need to install the host, shared fx, then sdk in order. Sure you would've figured it out but we ran into in our testing so figured I'd call it out :smile:

MichaelSimons:
@brthor - yep already discovered this :smile: Are there plans to update the readme in this regard?  Having a link to the sdk .deps doesn't seem real helpful without the other two.

brthor:
yes adding this soon

brthor:
ahhhh hit update branch instead of merge......

PR #2040, opened 2016-03-24
Anurse/eerhardt update dependencies

anurse:
Relocated from https://github.com/dotnet/cli/pull/1990 so I could add commits ;)

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2040)

### DIFF
--- a/TestAssets/FSharpTestProjects/CompileFail/project.json
+++ b/TestAssets/FSharpTestProjects/CompileFail/project.json
@@ -9,7 +9,7 @@
   ],
   "dependencies": {
     "Microsoft.FSharp.Core.netcore": "1.0.0-alpha-151221",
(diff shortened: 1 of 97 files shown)


weshaggard:
The compile errors look correct. Currently Microsoft.Extensions.DependencyModel is targeting netstandard1.3 and in that generation System.Reflection.Assembly doesn't contain the GetEntryAssembly. That was added in netstandard1.5 version of System.Reflection. So that dependency looks like it will need to be tweaked.

anurse:
Yeah, I had changed it to `netstandard1.5` and then realized there was a merge issue and then when I fixed it I blasted away the change to 1.5 :). Fixed

anurse:
Windows 2016 was all that was left, so merged.

PR #2048, opened 2016-03-25
Windows Installer changes

Sridhar-MS:
- Make CLI SDK MSI non-upgradable. It must alwasy be installed SxS.
- The CLI bundle and SharedFx bundle are non-upgradeable. They are also installed SxS.
- Make host\muxer upgradeable. It will be upgradeable till v1.0.0 RTM. Post RTM will be installed SxS with v.1.0.0.
- SharedFx MSI was using the CLI MSI version. Fixing it to use the SharedFx version.
- Do not allow bundles to uninstall. User will be able to uninstall the  individual MSIs.

cc: @piotrpMSFT @Petermarcu @ellismg @joeloff

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2048)

### DIFF
--- a/packaging/windows/clisdk/bundle.wxs
+++ b/packaging/windows/clisdk/bundle.wxs
@@ -8,9 +8,7 @@
   <Bundle Name="$(var.ProductName)" Manufacturer="$(var.Manufacturer)"
           Version="$(var.DisplayVersion)" UpgradeCode="$(var.UpgradeCode)"
           AboutUrl="http://dotnet.github.io/"
-          Compressed="yes"
-          DisableModify="yes"
-          DisableRemove="yes">
+          Compressed="yes">
 
     <BootstrapperApplicationRef Id="WixStandardBootstrapperApplication.Foundation">
       <bal:WixStandardBootstrapperApplication
@@ -34,13 +32,13 @@
     <util:DirectorySearch Path="[PreviousInstallFolder]" Variable="DOTNETHOME" After="PreviousInstallFolderSearch" Condition="PreviousInstallFolder" />
 
     <Chain DisableSystemRestore="yes" ParallelCache="yes">
-      <MsiPackage SourceFile="$(var.CLISDKMsiSourcePath)" Visible="yes">
+      <MsiPackage SourceFile="$(var.CLISDKMsiSourcePath)">
         <MsiProperty Name="DOTNETHOME" Value="[DOTNETHOME]" />
       </MsiPackage>
-      <MsiPackage SourceFile="$(var.SharedFXMsiSourcePath)" Visible="yes">
+      <MsiPackage SourceFile="$(var.SharedFXMsiSourcePath)">
         <MsiProperty Name="DOTNETHOME" Value="[DOTNETHOME]" />
       </MsiPackage>
-      <MsiPackage SourceFile="$(var.SharedHostMsiSourcePath)" Visible="yes">
+      <MsiPackage SourceFile="$(var.SharedHostMsiSourcePath)">
         <MsiProperty Name="DOTNETHOME" Value="[DOTNETHOME]" />
       </MsiPackage>
     </Chain>
--- a/packaging/windows/clisdk/bundle.wxs
+++ b/packaging/windows/clisdk/bundle.wxs
@@ -6,9 +6,11 @@
   <?include "Variables.wxi" ?>
 
   <Bundle Name="$(var.ProductName)" Manufacturer="$(var.Manufacturer)"
-          Version="$(var.DisplayVersion)" UpgradeCode="70A1576F-63B6-4659-9E39-25C7B769DDE5"
+          Version="$(var.DisplayVersion)" UpgradeCode="$(var.UpgradeCode)"
           AboutUrl="http://dotnet.github.io/"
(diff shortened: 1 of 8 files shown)


Sridhar-MS:
cc @robmen 
I like to get your thoughts on this. The CLI sdk bundle currently has 3 MSIs

- .Net Core CLI in `<root>\sdk\<cliversion>` folder. Always installs SxS.
- .Net Core Runtime and Framework (SharedFx) in `<root>\shared\<.netversion>`. Always installs SxS.
-  .Net Core Host in `<root>\dotnet.exe` and `<root>\hostfxr.dll`. Upgrades build-to-build till RTM (v1.0). Post RTM (v1.1) will be installed SxS with RTM (v1.0).

I am planning to make the CLI bundle uninstallable. To make the ARP provide only option to uninstall the MSIs individually. Like below

bundle-arp


Because the CLI sdk bundle has the CLI version and it masks the SharedFx version inside it. This may lead to users uninstalling CLI to find that the SharedFx (which their app may depend upon) is also uninstalled. In other words, there is no way for a user to tell if I uninstall CLI X.0 will it also uninstall SharedFx Y.0.

packaging/windows/clisdk/bundle.wxs
```diff
@@ -32,13 +34,13 @@
     <util:DirectorySearch Path="[PreviousInstallFolder]" Variable="DOTNETHOME" After="PreviousInstallFolderSearch" Condition="PreviousInstallFolder" />
 
     <Chain DisableSystemRestore="yes" ParallelCache="yes">
-      <MsiPackage SourceFile="$(var.CLISDKMsiSourcePath)">
+      <MsiPackage SourceFile="$(var.CLISDKMsiSourcePath)" Visible="yes">
```
joeloff: The only way this will work is if we ensure that every other bundle that ships these MSIs also have them marked as visible, otherwise a bundle that carries a newer version of these MSIs and have them hidden will end up removing them from ARP.

Sridhar-MS: I reverted this change for now. The bundle is going to be in the ARP, not the MSIs.

robmen:
Sorry, GitHub never sent notification to me about this thread until it was merged in. I don't know what is up with their mail servers but they aren't working consistently.

There is no way to hide the Bundle from ARP. Burn isn't designed to allow that (on purpose). If everyone takes reference counts on the packages (Burn does this for you automatically) then you don't need to to any of this. It would be sad to have to pollute ARP with SharedFX the way we pollute it today with multiple versions of CRT, NETFX, etc.

How does someone get the SharedFX independent of the CLI? Is there a place to get it? Is it a redistributable? If so, what are the instructions given to developers to redistribute it?

If you must go this way, essentially declaring the Shared FX MSI to be "permanent package" (aka: ARP pollution) then make the Shared FX MsiPackage Visible and Permanent. Keep the CLI MsiPackage handeled by Burn. Then when the Bundle uninstalls, it will take the CLI with it, leave the Shared FX behind (but with an ARP entry so it can be removed).

Essentially, you're guaranteeing to clean up the CLI install, you have to uninstall two things. That's why I call it pollution and hope you won't go that way.

robmen:
Also, if you really want to display UI during uninstall about removing Shared FX that is possible. Not exactly sure how you explain to a user when it is safe so most user's would have to choose to not uninstall it, and back to pollution.

You really need to solve this problem for people. Also, if all this stuff is SxS, won't there be lots and lots of junk left around for people that upgrade regularly? I don't understand the scenario here. SxS is usually something you do per-release not per-build (because of the junk left behind for those build-to-build upgrading). But maybe I'm missing something.

Sridhar-MS:
@robmen After a discussion with @joeloff I decided not to make the individual MSIs available on ARP. Also these MSIs go into other bundles (not in this repo) like Asp.Net Core bundles. And for it to work we will have to author these MSIs as available on ARP in all the bundles that they ship in. Also you made a good point about polluting the ARP since we are going SxS from build-to-build.

> How does someone get the SharedFX independent of the CLI? Is there a place to get it? Is it a redistributable? 

SharedFx is available as an independent bundle. Users can download it from the table [here](https://github.com/dotnet/cli#sdk-installers). It is available in `.NET Core Installers` column. 

> If so, what are the instructions given to developers to redistribute it?

Developers can extract the MSIs from SharedFx bundle and redistribute them. But this is not documented yet. @blackdwarf You may be interested in it. Developers can chain the sharedfx MSI in their bundles along with their app and make it available for installation.

robmen:
@Sridhar-MS, yeah, sorry, I missed the "comments on outdated diff" above.

If your MSI packages are only available via Bundles (something I can wholeheartedly endorse), then you don't need to make any of them visible. Burn (the engine under Bundles) will correctly reference track the contained packages. That means if someone installs the SharedFX Bundle and installs the CLI SDK Bundle then the underlying MSI package will be shared and reference counted.

In the case where people ship the underlying MSI, then just have them follow the same reference counting system and don't expose the MSIs. IMHO, it's better to have the occasional "repair broken application after uninstalling other application" than to leave junk behind on the user's machine. One leads to a rare case of breakage (most people don't uninstall) and slows WinRot for those that do.

PR #2053, opened 2016-03-25
fix use latest version of fsc

enricosada:
This reverts to use latest version of fsc, downgraded by  43512b8973c89fa21a4236e4b250357c9fea767e

I think that was a merge conflict error right @anurse ?

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2053)

### DIFF
--- a/src/dotnet/project.json
+++ b/src/dotnet/project.json
@@ -24,7 +24,7 @@
 
     "Newtonsoft.Json": "7.0.1",
     "Microsoft.Net.Compilers.netcore": "1.3.0-beta1-20160225-02",
-    "Microsoft.FSharp.Compiler.netcore": "1.0.0-alpha-151218",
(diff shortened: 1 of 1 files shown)


enricosada:
i tried locally (both build and just replace files) and new version works, it's possibile to merge this asap?

Without that, f# doesnt work
The new version of `compile-fsc` use response files, old version doesnt support it ( + other bug like xplat ) so with latest there is the error

```
error FS0193: internal error: Exception of type 'System.Collections.Generic.KeyNotFoundException' was thrown.
```

because fsc ignore response file and doesnt see FSharp.Core, 

/cc @NumberByColors @brthor @piotrpMSFT

Sridhar-MS:

brthor:
:shipit: :shipit:

PR #2060, opened 2016-03-25
fix OSX issue in VSO build

anurse:
also, make it possible to have more verbose build runs when needed

/cc @eerhardt @Sridhar-MS @piotrpMSFT

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2060)

### DIFF
--- a/scripts/dotnet-cli-build/Utils/DotNetCli.cs
+++ b/scripts/dotnet-cli-build/Utils/DotNetCli.cs
@@ -1,9 +1,8 @@
-﻿using System.IO;
-using System.Linq;
-using System;
-using System.Runtime.InteropServices;
-using Microsoft.DotNet.Cli.Build.Framework;
+﻿using Microsoft.DotNet.Cli.Build.Framework;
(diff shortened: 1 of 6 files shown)


anurse:
Verified this on VSO manually. I cancelled the build to avoid publishing bogus packages, but it got well past the place it was failing before.

scripts/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -166,25 +165,30 @@ public static BuildTargetResult CheckPackageCache(BuildTargetContext c)
 
             if (ciBuild)
```
eerhardt: Does anyone set this anymore?  It used to be set in ci_build.cmd/sh. But those no longer exist.

anurse: If they aren't needed, let's remove them everywhere. This toggle was added on a whim later, but I didn't end up need it. I can remove it.

PR #2075, opened 2016-03-26
Fix example runtimeconfig.json.

adityamandaleeka:
The current example runtimeconfig.json in the documentation is invalid because it has trailing commas in the JSON objects. It fails to parse at run time with an error.

This change removes those commas and also removes the `UseAllCpuGroups` property which isn't supported for now.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2075)


adityamandaleeka:
cc: @schellap

PR #2077, opened 2016-03-26
Patch Lock File with extra build time dependencies

cdmihai:
This refers to the scenario where VS drives the build and there are xproj projects referencing csproj projects.

When a project.json project references a csproj project, nuget cannot follow the dependency path toward the csproj. Those dependencies are only available at build time. Therefore, there will be a special msbuild target that generates a fragment lock file describing all the csproj projects that an xproj depends upon.

By convention, the fragment lock file is placed besides the master lock file. If the fragment is there, LockfileReader merges with it.

Specs: https://github.com/dotnet/cli/blob/troy/tooling/Documentation/specs/xproj2csproj.md

Note that to add test samples containing xproj and csproj projects, I have to commit their lock, fragment, and csproj dlls. These assets cannot get restored and built by the CLI, so they are committed as is.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2077)


cdmihai:
Integration tests are failing for non windows machines, since they are for full framework only. I'll have to skip those tests on non windows machines.

brthor:
Why are you checking in binaries?

TestAssets/AsIsProjects/XprojCsprojProjects/BasicProject_invalid-missing-dll/Sample1-Xproj+Csproj/artifacts/src/ConsoleApp13/obj/Debug/net46/dotnet-compile.rsp
```diff
@@ -0,0 +1,20 @@
+--temp-output:C:\Users\micodoba\OneDrive for Business\tests\TestXprojCsproj-master\Sample1-Xproj+Csproj\artifacts\src\ConsoleApp13\obj\Debug\net46\
```
brthor: Why check this in?

troydai: The tooling end to end scenario is hard. We should clean up these info. We need the lock file and binary files since they are not created by `dotnet-cli`. They're VS artifact. but we need them for E2E test. We'll look for a better scenario.

brthor: If possible, we should try to find a better pattern here. 

Checking in all of NewtonSoft.Json :frowning:

cdmihai:
@brthor 

I checked in the binaries because they can only be outputted by Visual Studio. Same goes for the special lock files with dependency holes.

However, we decided that the CLI is not the place to have integration tests with VS projects, so I am reverting the commit that introduced these.

Also, closing the PR and reopening it to include @troydai's changes

PR #2080, opened 2016-03-26
Packaging work for corehost

schellap:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2080)

PR #2083, opened 2016-03-27
Resolve hostpolicy from deps file in exec mode

schellap:
@gkhanna79 PTAL.

For the `dotnet run` scenario, `hostpolicy.dll` itself has to be loaded from the additional probing path (nuget cache).

This PR can go in post #2080 and a new host package is uploaded.

Cc @brthor

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2083)

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -12,6 +12,7 @@
 #include "runtime_config.h"
 #include "cpprest/json.h"
 #include "error_codes.h"
+#include "deps_format.h"
 
 pal::string_t fx_muxer_t::resolve_fx_dir(const pal::string_t& muxer_dir, runtime_config_t* runtime)
 {
@@ -23,7 +24,7 @@ pal::string_t fx_muxer_t::resolve_fx_dir(const pal::string_t& muxer_dir, runtime
     fx_ver_t specified(-1, -1, -1);
     if (!fx_ver_t::parse(fx_ver, &specified, false))
     {
-        trace::info(_X("The specified runtimeconfig.json version [%s] could not be parsed"), fx_ver.c_str());
+        trace::error(_X("The specified runtimeconfig.json version [%s] could not be parsed"), fx_ver.c_str());
         return pal::string_t();
     }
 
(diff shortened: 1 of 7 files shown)


schellap:
@dotnet-bot test this please

schellap:

schellap:

schellap:

schellap:
Cc @piotrpMSFT @brthor merged to unblock.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -329,8 +330,20 @@ int fx_muxer_t::execute(const int argc, const pal::char_t* argv[])
             else
             {
                 trace::verbose(_X("Executing as a standalone app as per config file [%s]"), config_file.c_str());
-                corehost_init_t init(deps_file, probe_path, _X(""), host_mode_t::muxer, &config);
                 pal::string_t impl_dir = get_directory(app_or_deps);
+                if (!library_exists_in_dir(impl_dir, LIBHOSTPOLICY_NAME, nullptr) && !probe_path.empty() && !deps_file.empty())
```
gkhanna79: This does not sound right. Hostpolicy.dll should be loaded from the FX directory corresponding to the FX called out as app dependency. Why do we need this change?

schellap: For standalone apps with `dotnet build` and `dotnet run`

gkhanna79: How does one get a standalone app with "Dotnet build" and "dotnet run"? Unless they are published as standalone, all apps rely on SharedFX and should resolve the dependencies from there.

schellap: Yes, this is only for `dotnet build` which doesn't copy the hostpolicy into the bin directory. I was under the same impression, until @piotrpMSFT clarified this. The idea was, as per @piotrpMSFT, we'd be given a `deps file` and a `.dll` and a `additionalprobingpath` to resolve. Earlier this worked because corehost was copied to bin. Now, with muxer based activation of `dotnet build` there is no hostpolicy.dll which should be resolved against the standard meta package.

schellap: Cc @brthor

PR #2097, opened 2016-03-28
add trimming of published output based on sharedfx

anurse:
Any dependencies which **exactly** match the version requested in the
graph originating at the `type: platform` dependency (if any) are
trimmed from the publish output

/cc @davidfowl @piotrpMSFT @pakrym

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2097)

### DIFF
--- a/src/dotnet/commands/dotnet-publish/PublishCommand.cs
+++ b/src/dotnet/commands/dotnet-publish/PublishCommand.cs
@@ -14,6 +14,8 @@ using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.DotNet.Files;
 using Microsoft.DotNet.Tools.Common;
 using Microsoft.DotNet.ProjectModel.Utilities;
+using Microsoft.DotNet.ProjectModel.Graph;
+using NuGet.Versioning;
 
 namespace Microsoft.DotNet.Tools.Publish
 {
@@ -153,7 +155,14 @@ namespace Microsoft.DotNet.Tools.Publish
             var exporter = context.CreateExporter(configuration);
 
             var isPortable = string.IsNullOrEmpty(context.RuntimeIdentifier);
-            foreach (var export in exporter.GetAllExports())
+
+            // Collect all exports and organize them
+            var exports = exporter.GetAllExports()
+                .Where(e => e.Library.Identity.Type.Equals(LibraryType.Package))
+                .ToDictionary(e => e.Library.Identity.Name);
+            var collectExclusionList = isPortable ? GetExclusionList(context, exports) : new HashSet<string>();
+
+            foreach (var export in exporter.GetAllExports().Where(e => !collectExclusionList.Contains(e.Library.Identity.Name)))
             {
                 Reporter.Verbose.WriteLine($"Publishing {export.Library.Identity.ToString().Green().Bold()} ...");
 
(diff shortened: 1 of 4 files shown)


src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -153,7 +155,13 @@ private bool PublishProjectContext(ProjectContext context, string buildBasePath,
             var exporter = context.CreateExporter(configuration);
 
             var isPortable = string.IsNullOrEmpty(context.RuntimeIdentifier);
-            foreach (var export in exporter.GetAllExports())
+
+            // Collect all exports and organize them
+            var exports = exporter.GetAllExports()
+                .ToDictionary(e => e.Library.Identity.Name);
```
pakrym: This could be dangerous as proven by lots of `Duplicated key` errors in `DepencyModel`

pakrym:
Having a test would make me feel better, preferably with version overriden package in it.

src/dotnet/commands/dotnet-publish/PublishCommand.cs
anurse: I filtered by Package, so we shouldn't get any duplicates now.

src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -198,6 +207,40 @@ private bool PublishProjectContext(ProjectContext context, string buildBasePath,
...
+                return exclusionList;
+            }
+            else if (redistPackages.Count > 1)
+            {
+                throw new InvalidOperationException("Multiple packages with type: \"platform\" were specified!");
+            }
+            var redistExport = exports[redistPackages[0].Name];
+
+            exclusionList.Add(redistExport.Library.Identity.Name);
+            CollectDependencies(exports, redistExport.Library.Dependencies, exclusionList);
```
pakrym: Shouldn't `RequestedRanges` be used here?

pakrym: I thought dependencies get resolved versions of library.

brthor:
Is this for published PortableApps? 

If an app has rids, but type platform for Microsoft.NETCore.App, would we still expect this trimming to happen?

anurse:
> If an app has rids, but type platform for Microsoft.NETCore.App, would we still expect this trimming to happen?

Right now, yes. The behavior in that scenario is not fully defined and we can tidy up post RC2, but you would end up with a "standalone" publish, but without the runtime bits. It might fit the "shared-but-not-portable" app scenario we've spec'd out, but we'd need to go over that scenario since there may be edge cases we're missing.

src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -198,6 +207,40 @@ private bool PublishProjectContext(ProjectContext context, string buildBasePath,
...
+            return exclusionList;
+        }
+
+        private void CollectDependencies(Dictionary<string, LibraryExport> exports, IEnumerable<LibraryRange> dependencies, HashSet<string> exclusionList)
+        {
+            foreach (var dependency in dependencies)
+            {
+                var export = exports[dependency.Name];
+                if(export.Library.Identity.Version.Equals(dependency.VersionRange.MinVersion))
```
davidfowl: nit: spacing

src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -198,6 +207,40 @@ private bool PublishProjectContext(ProjectContext context, string buildBasePath,
...
+            exclusionList.Add(redistExport.Library.Identity.Name);
+            CollectDependencies(exports, redistExport.Library.Dependencies, exclusionList);
+            return exclusionList;
+        }
+
+        private void CollectDependencies(Dictionary<string, LibraryExport> exports, IEnumerable<LibraryRange> dependencies, HashSet<string> exclusionList)
+        {
+            foreach (var dependency in dependencies)
+            {
+                var export = exports[dependency.Name];
```
davidfowl: Can this ever throw? What about unresovled dependencies?

PR #2117, opened 2016-03-29
Add more overloads for assembly resolution in dependency model

pakrym:
Because @pranavkm made me

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2117)

### DIFF
--- a/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextTests.cs
+++ b/test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextTests.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


pranavkm:
:shipit:

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextTests.cs
```diff
@@ -11,9 +12,9 @@ namespace Microsoft.Extensions.DependencyModel.Tests
     public class DependencyContextTests
     {
         [Theory]
-        [InlineData("System.Banana.dll", "System.Banana")]
-        [InlineData("System.Banana.ni.dll", "System.Banana")]
-        [InlineData("FlibbidyFlob", "FlibbidyFlob")]
+        [InlineData("System.Collections.dll", "System.Collections")]
+        [InlineData("System.Collections.ni.dll", "System.Collections")]
+        [InlineData("mscorelib", "mscorelib")]
```
akoeplinger: is the typo (mscorlib vs mscorelib) intentional here?

anurse: Looks like the names don't actually matter for this test, but yeah, it would be best to be consistent and use `mscorlib`

anurse:
Psh, sounds like @pranavkm is just lazy.

:shipit:

test/Microsoft.Extensions.DependencyModel.Tests/DependencyContextTests.cs
pakrym: It matters, new AssemblyName on linux throws when it can't find assembly, nice catch!

pakrym:
@dotnet-bot test this please

PR #2138, opened 2016-03-30
Fix fsharp corehostify

enricosada:
fix f# with latest changes of .NET CLI

- use muxer (`dotnet exec`)
- build script changes:
    - `fsc.exe` -> `fsc.dll` (like `csc`)
    - add `fsc.deps.json` ( like `csc.deps.json` )
    - add  `fsc.runtimeconfig.json` ( like `csc.runtimeconfig.json` )
- fix path of `default.win32manifest`

/cc @brthor @piotrpMSFT @pakrym

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2138)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -20,6 +20,7 @@ namespace Microsoft.DotNet.Cli.Build
 
         public static readonly string[] BinariesForCoreHost = new[]
         {
+            "fsc",
             "csi",
             "csc",
             "vbc"
@@ -372,12 +373,22 @@ namespace Microsoft.DotNet.Cli.Build
             File.Copy(Path.Combine(Dirs.Corehost, $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}"), Path.Combine(outputDir, $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}"), overwrite: true);
             File.Copy(Path.Combine(Dirs.Corehost, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), Path.Combine(outputDir, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), overwrite: true);
 
-            var binaryToCorehostifyOutDir = Path.Combine(outputDir, "runtimes", "any", "native");
             // Corehostify binaries
             foreach (var binaryToCorehostify in BinariesForCoreHost)
             {
                 try
                 {
+                    string binaryToCorehostifyOutDir;
+                    if (binaryToCorehostify == "fsc")
+                    {
+                        //fsc.exe is in the same directory, not in runtimes/any/native subdirectory
+                        binaryToCorehostifyOutDir = outputDir;
+                    }
+                    else
+                    {
+                        binaryToCorehostifyOutDir = Path.Combine(outputDir, "runtimes", "any", "native");
+                    }
+
                     // Yes, it is .exe even on Linux. This is the managed exe we're working with
                     File.Copy(Path.Combine(binaryToCorehostifyOutDir, $"{binaryToCorehostify}.exe"), Path.Combine(outputDir, $"{binaryToCorehostify}.dll"));
                     File.Delete(Path.Combine(binaryToCorehostifyOutDir, $"{binaryToCorehostify}.exe"));
--- a/src/dotnet/commands/dotnet-compile-fsc/Program.cs
+++ b/src/dotnet/commands/dotnet-compile-fsc/Program.cs
@@ -170,7 +170,7 @@ namespace Microsoft.DotNet.Tools.Compiler.Fsc
                 //HACK we need default.win32manifest for exe
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
-                    var win32manifestPath = Path.Combine(AppContext.BaseDirectory, "runtimes", "any", "native", "default.win32manifest");
+                    var win32manifestPath = GetDefaultWin32ManifestPath();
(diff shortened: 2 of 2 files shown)


enricosada:
@dotnet-bot test Windows_NT x86 Release Build please

enricosada:
@dotnet-bot test Windows_NT x86 Release Build please

enricosada:

src/dotnet/commands/dotnet-compile-fsc/Program.cs
```diff
@@ -170,7 +170,7 @@ public static int Run(string[] args)
                 //HACK we need default.win32manifest for exe
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
-                    var win32manifestPath = Path.Combine(AppContext.BaseDirectory, "default.win32manifest");
+                    var win32manifestPath = Path.Combine(AppContext.BaseDirectory, "runtimes", "any", "native", "default.win32manifest");
```
brthor: AppContext.Basedirectory is redirected in tools scenario, I used location of the assembly to get around this.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -372,12 +373,22 @@ private static BuildTargetResult CompileCliSdk(BuildTargetContext c, DotNetCli d
...
             File.Copy(Path.Combine(Dirs.Corehost, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), Path.Combine(outputDir, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), overwrite: true);
 
-            var binaryToCorehostifyOutDir = Path.Combine(outputDir, "runtimes", "any", "native");
             // Corehostify binaries
             foreach (var binaryToCorehostify in BinariesForCoreHost)
             {
                 try
                 {
+                    string binaryToCorehostifyOutDir;
+                    if (binaryToCorehostify == "fsc")
```
pakrym: consider putting full relative paths into `BinariesForCoreHost`

src/dotnet/commands/dotnet-compile-fsc/Program.cs
enricosada: I'll fix it, removing BaseDirectory

brthor:
@enricosada I don't think we should take a dependency on the cli build here for the fsc runtimeconfig for a couple reasons:
1. dotnet-compile-fsc will be moving to a new repo, as it sits in the other PR, we can copy/paste the dotnet-compile-fsc to another repo and expect it to work (mostly)
2. Ideally the runtimeconfig.json would be part of the fsc.exe package eventually and none of this would be required.
3. The case for fsc.exe is slightly different than csc.exe since it is invoked via portable tool. Generating a deps.json which is different than that of dotnet-compile-fsc will break things. Additionally having a non-portable runtimeconfig.json will break things as well.

enricosada:
@brthor My first reason for this pr is i just want to fix f# , atm is broken with latest .NET CLI after `rc2-002071`.
I was thinking fsc needed runtimeconfig.json anyway ( and deps.json ), if compile-fsc is a tool or not, so i modified the build script like csc (to remain in sync).

Maybe i am wrong, i can close this pr, np.

I understood we can just `dotnet exec fsc.dll` from the `compile-fsc` (embedded command or in an extracted tool package). But maybe i understand too little of the problem. 

`dotnet-compile-fsc` (tool) call fsc bundled in .NET CLI (using corehost/dotnet), right? fsc should instead use same version of deps of `compile-fsc`? It's just a process execution (dotnet.exe + `exec fsc.dll`) or not?

I am :+1: about standalone `compile-fsc`, but we need `fsc.dll`, `fsc.runtimeconfig.json` (and deps )? 

Is fsc going to remain bundled inside .NET CLI after `compile-fsc` is extracted right? or is a dependency of `compile-fsc`?

About 2) I'll open an issue in visualfsharp about adding `runtimeconfig.json` ( /cc @KevinRansom ) 
Maybe we should move away from `dnxcore50` to netstandard too

If i am too confused, i can just wait for your pr to be merged
It's a lot of new stuff deps.json, runtimeconfig.json etc, i need to learn more about it, dotnetcli change a lot :smile:

KevinRansom:
Fsc.exe will not be deployed with dotnet cli.  Instead dotnet/cli will on-demand download it.

I believe Bryan is still working on the mechanism.

enricosada:
better having fsc outside bundled cli :+1: :+1: , easier to move

About fsc, @brthor i hope that can defined as tool too

Current:
```
{
    tools: {
       "Microsoft.FSharp.Compiler.netcore": { version: "1.0.0-alpha-160318",  namedtool: "fsc" }
    }
} 
```

Maybe we can tag the tool ( like resources ), so we can specify custom (version/package) from `project.json` or use latest `1.0.0-alpha-*`

enricosada:
Updated this pr to have only one place with AppContext.BaseDirectory, so it's easier to change in your pr ( default.win32 is relative to fsc ).

Btw fast question @brthor , i need to close this pr?

brthor:
@enricosada i think we've solved some of the same problems in our PRs, but this would need additional work to work as a packaged command. I've verified dotnet-compile-fsc is working in #1823 and have enabled tests testing compilation of the FSharp Test Projects (which are passing). For that reason I think it'd be best to rebase there and work on improving from there.

If fsc.exe was packaged like a PortableApp (with runtimeconfig.json), it could be used as a tool how you describe, but we'd need to do some more work to enable the `namedtool` feature. We were going to do this for RC2, but ended up pushing it back to RTM tools. At that point in time we could remove the step copying the runtimeconfig for fsc.exe in dotnet-compile-fsc.

enricosada:
thx @brthor for info, closing this in favor of your #1823 because is a lot more complete, and latest can be broken meanwhile (we have a new lkg `1.0.0-beta-002071` so np)

:+1: about run f# test proj inside cli, thx

Issue #2038, opened 2016-03-24
dotnet.exe fail to execute in Azure Web Site


troydai:
## Steps to reproduce
Try to deploy MVC to Azure Web Sites through Kudu. Here's the sample app:
https://github.com/troydai/HelloMvcKudu/tree/0950beaa48027a47496bd3426935cf0716f7e9be

Here's the deployment script:
https://github.com/troydai/HelloMvcKudu/blob/0950beaa48027a47496bd3426935cf0716f7e9be/deploy.ps1

## Expected  behavior
The deployment script will execute the `dotnet.exe --version`. It is expected to be successful. 

## Actual behavior
The command failed quietly. No exception was thrown and no output was generated. Turn on the trace, here the end portion of it:
```
remote: Adding to native path: D:\home\site\repository\.dotnet\sdk\1.0.0-beta-001986
remote: 
remote: Adding to resources path: D:\home\site\repository\.dotnet\sdk\1.0.0-beta-001986
remote: 
remote: Failed to load the dll from D:\home\site\repository\.dotnet\sdk\1.0.0-beta-001986\coreclr.dll, HRESULT: 0x8007007E
remote: 
remote: Failed to bind to coreclr
remote: 
remote: Tracing enabled
remote: 
remote: Own DLL path=[D:\home\site\repository\.dotnet\dotnet.dll]
remote: 
remote: Checking if CoreCLR path exists=[D:\home\site\repository\.dotnet\coreclr.dll]
remote: 
remote: Host operating in Muxer mode
remote: 
remote: --- Executing in muxer mode...
remote: 
remote: --- Resolving dotnet from working dir
remote: 
remote: Probing path [D:\home\site\repository\global.json] for global.json
remote: 
remote: Found global.json [D:\home\site\repository\global.json]
remote: 
remote: --- Resolving CLI version from global json [D:\home\site\repository\global.json]
remote: 
remote: CLI '/sdk/version' field not present/null in [D:\home\site\repository\global.json]
remote: 
remote: --- Resolving SDK version from SDK dir [D:\home\site\repository\.dotnet\sdk]
remote: 
remote: Considering version... [.]
remote: 
remote: Considering version... [..]
remote: 
remote: Considering version... [1.0.0-beta-001943]
remote: 
remote: Considering version... [1.0.0-beta-001986]
remote: 
remote: No production version found, so using latest prerelease
remote: 
[...]

PR #2144, opened 2016-03-30
Expand LoadLibrary search paths

schellap:
@gkhanna79 PTAL.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2144)

### DIFF
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -85,7 +85,10 @@ bool pal::getcwd(pal::string_t* recv)
 
 bool pal::load_library(const char_t* path, dll_t* dll)
 {
-    *dll = ::LoadLibraryW(path);
+    // LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR:
+    //   In portable apps, coreclr would come from another directory than the host,
+    //   so make sure coreclr dependencies can be resolved from coreclr.dll load dir.
+    *dll = ::LoadLibraryExW(path, NULL, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
     if (*dll == nullptr)
     {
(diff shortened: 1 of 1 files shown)


schellap:
Fixes https://github.com/dotnet/corefx/issues/7357

src/corehost/common/pal.windows.cpp
```diff
@@ -85,7 +85,7 @@ bool pal::getcwd(pal::string_t* recv)
 
 bool pal::load_library(const char_t* path, dll_t* dll)
 {
-    *dll = ::LoadLibraryW(path);
+    *dll = ::LoadLibraryExW(path, NULL, LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR | LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
```
gkhanna79: Please add a comment for this change. Otherwise, LGTM.

schellap:
Fixes https://github.com/dotnet/cli/issues/2038

@glennc @troydai

troydai:
@schellap can you give us some detail about how this would fix the #2038?

schellap:
The reason corehost in sdk folder worked was because dotnet.dll was a standalone app with the api-forwarders. LoadLibrary would pick up from own directories fine.

Once you used the muxer LoadLibrary couldn't find the api forwarders. In your case, an UI would've popped up saying api-... blah was missing. We couldn't get this important piece of diagnostic info in that min kernel. The fix expands the search path to use default search paths, and module load directory as well.

troydai:
Thanks. So what is it different in Antares from normal Windows environment. Because you theory only explains that the error message doesn't show up in Anates, which of course make the diagnosis harder but shouldn't the assembly searching logic the same?

schellap:
This never reproed on Win 10 boxes because they don't require api-forwarders. You require api-forwarders for min clients.

troydai:
Got it. Is possible to add test to automate this scenario on a platform other than windows 10?

Issue #2057, opened 2016-03-25
.NET Core Installer vs .NET Core Cli Installers is confusing


moozzyk:
I installed `.NET Core Installer` and wanted to confirm that I have the latest version and got:

```
PS C:\WINDOWS\system32> dotnet --version
Could not resolve SDK directory from [C:\Program Files\dotnet]
```

I did see that `--help` options in the new thing were limited but was not aware that I need a different installer to do my dev work. This is quite confusing since the tutorials say you should do `dotnet build`, `dotnet run` etc. but if you install just the Core Installer nothing of this works. The message and `--help` is very unhelpful either.

brthor:
@moozzyk The Core Installer is intended to place the shared framework and host only on your machine. 

In this way it can be used for running apps, but not building them.

The `.NET Core Cli` Installer is what you are looking for

moozzyk:
Yes, @anurse explained it too me. But users don't have @anurse or you around - the message should make it possible to figure it out on my own. At the moment is is not helpful.

brthor:
Totally agree! any suggestions there?

cc @blackdwarf

shahid-pk:
the .net core one should be called something like .net core runtime.

anurse:
I prefer "Runtime" and "SDK" as distinctions. Also, the muxer should be capable of responding to `--version` at some point before RTM :)

enricosada:
:+1: about `.NET Core Runtime`

And maybe `.NET Core CLI` can be just `.NET CLI`, because it can build all targets, not only `.NET Core`.
It's already a good name, short

![image](https://cloud.githubusercontent.com/assets/147243/14051122/cfefd8ea-f2c0-11e5-815b-57a5837878e7.png)

xoofx:
> I prefer "Runtime" and "SDK" as distinctions. Also, the muxer should be capable of responding to --version at some point before RTM :)

:+1:  for  `.NET Core Runtime` and `.NET Core SDK` installers.

piotrpMSFT:
@richlander @blackdwarf regarding branding. 

@schellap regarding --version in the muxer

brthor:
I think Tools is confusing because we have a separate concept of cli tools in the project.json.

It's not too bad, but every time we talk about tools we'll need to point out which tools we're talking about.

akoeplinger:
[...]

PR #2147, opened 2016-03-30
Clean up product naming

richlander:
I'd like to adopt new naming for the product. This is the proposal.

- .NET Core - .NET Core runtime and framework
- .NET Core SDK - .NET Core + CLI tools

/cc @blackdwarf, @piotrpMSFT, @Petermarcu

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2147)


blackdwarf:
Actually, I've made it similar as a proposal in: #2057. Here it is:

* .NET Core Runtime 
   * Shared runtime + host
* .NET Core Software Development Kit (SDK)
   * .NET Core Runtime vx.0.0.y
   * .NET Core CLI Tools va.0.0.b
   * In the future:
       * Docs
       * Samples

richlander:
Yes, very similar. I'd like to stick to what I wrote.

I'm also going to try to remove "shared fx" from our vocab.

blackdwarf:
@richlander any reason why yours is better? .NET Core on its own doesn't tell people what it is. .NET Core Runtime tells it cleanly.

richlander:
Two reasons:

- I don't want to use "runtime" in the name.
- I don't want to talk about "Shared runtime + host" as top-level concerns.

piotrpMSFT:
This should update installer branding as well

richlander:
Yes. I thought we'd start with this and then move to the product as a second step. Doesn't need to all go into one PR.

richlander:
I showed this to @coolcsh. I got the :+1: .

blackdwarf:
@richlander the "shared runtime" above was just a pointer as to what it is in it. It is a singular unit. But ok.

richlander:
@blackdwarf - I do want to stop using the term "shared runtime", so that's the other issue. I do like how you describe the SDK, however (minus the 'runtime' term).

blackdwarf:
@richlander whatever makes people don't doubt which link to click to get bits.

akoeplinger:
Hmm, using `.NET Core` as the moniker for the shared fx overloads an already overloaded term even more imho. 

I'd also add at least one sentence that tells people what they can actually _do_ with each package, e.g. run apps vs develop apps. Otherwise I'm left scratching my head which one I should pick. @blackdwarf's description is better in this case.

blackdwarf:
@akoeplinger for sure, we need descs on these. But, as it used to be said *nomen est omen* thus I think  being explicit about the "runtime" bit is better.

Petermarcu:
I'm going to commit this for now. We can update again if there are any more changes needed.

PR #2156, opened 2016-03-31
Probe shared locations

schellap:
Cc @gkhanna79

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2156)


src/corehost/cli/args.cpp
```diff
@@ -80,7 +80,10 @@ bool parse_arguments(const pal::string_t& deps_path, const pal::string_t& probe_
     pal::string_t opts_deps_file = _X("--depsfile");
     pal::string_t opts_probe_path = _X("--additionalprobingpath");
     pal::string_t deps_file = opts.count(opts_deps_file) ? opts[opts_deps_file] : deps_path;
-    pal::string_t probe_path = opts.count(opts_probe_path) ? opts[opts_probe_path] : probe_dir;
+    if (opts.count(opts_probe_path))
+    {
+        args.probe_paths.insert(opts[opts_probe_path]);
```
gkhanna79: Does this allow multiple probe paths to be added to the arg list? It does not seem so.

src/corehost/cli/args.cpp
```diff
@@ -103,6 +109,16 @@ bool parse_arguments(const pal::string_t& deps_path, const pal::string_t& probe_
     }
 
     pal::getenv(_X("DOTNET_PACKAGES_CACHE"), &args.dotnet_packages_cache);
```
gkhanna79: We should rename DOTNET_PACKAGES_CACHE to something like DOTNET_HOSTING_OPTIMIZATION_CACHE. @leecow @blackdwarf  What do you think?

src/corehost/cli/args.cpp
```diff
@@ -103,6 +109,16 @@ bool parse_arguments(const pal::string_t& deps_path, const pal::string_t& probe_
...
-    pal::getenv(_X("DOTNET_SERVICING"), &args.dotnet_servicing);
+    pal::getenv(_X("DOTNET_EXTENSIONS"), &args.dotnet_extensions);
```
gkhanna79: This is not required. Any shared location path will be specified at command-line (as --additionalProbingPath).

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -76,9 +77,16 @@ int execute_app(
 bool hostpolicy_exists_in_svc(pal::string_t* resolved_dir)
 {
     pal::string_t svc_dir;
-    if (!pal::getenv(_X("DOTNET_SERVICING"), &svc_dir))
+    if (!pal::getenv(_X("DOTNET_EXTENSIONS"), &svc_dir))
```
gkhanna79: Remove this env var check.

src/corehost/common/pal.windows.cpp
```diff
@@ -130,8 +130,26 @@ bool pal::get_default_packages_directory(string_t* recv)
...
+    append_path(recv, _X(".nuget"));
+    append_path(recv, _X("packages"));
+    return true;
+}
+
+bool pal::get_default_extensions_directory(string_t* recv)
+{
+    recv->clear();
+
+#ifdef _TARGET_X86_
```
gkhanna79: I think this check should instead use FOLDERID_ProgramFilesX86 value as described at https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457(v=vs.85).aspx. It will return PF-x86 for 32bit app on 64bit and PF for 32bit app on 32bit OS.

We shouldnt rely on environment variables for this.

src/corehost/cli/deps_entry.cpp
```diff
@@ -44,22 +32,56 @@ bool deps_entry_t::to_rel_path(const pal::string_t& base, pal::string_t* str) co
...
     }
     return exists;
 }
 
 // -----------------------------------------------------------------------------
+// Given a "base" directory, yield the local path of this file
+//
+// Parameters:
+//    base - The base directory to look for the relative path of this entry
+//    str  - If the method returns true, contains the file path for this deps
```
gkhanna79: Add comment about the "bool" arg as well.

gkhanna79: Also, the comment for "str" should be updated to indicate that this returns flat path for the file, within the base folder, as opposed to relative path.

src/corehost/cli/deps_format.cpp
```diff
@@ -124,9 +127,9 @@ pal::string_t get_own_rid()
 bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, const rid_fallback_graph_t& rid_fallback_graph)
 {
     pal::string_t host_rid = get_own_rid();
-    for (auto& package : *portable_assets)
+    for (auto& package : portable_assets->libs)
     {
-        pal::string_t matched_rid = package.second.count(host_rid) ? host_rid : _X("");
+        pal::string_t matched_rid = package.second.rid_assets.count(host_rid) ? host_rid : _X("");
```
gkhanna79: Is this a bug fix? I dont quite follow the change.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+    pal::string_t maj_min_star = lib_ver.substr(0, pat_start + 1) + _X('*');
+
+    pal::string_t path = probe_dir;
+    append_path(&path, entry.library_name.c_str());
+
+    pal::string_t cache_key = path;
+    append_path(&cache_key, maj_min_star.c_str());
+
+    pal::string_t max_str;
+    if (m_roll_forward_cache.count(cache_key))
```
gkhanna79: Why do we have .* based roll-forward check? We were only planning to roll-forward based upon patch version.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+    if (pal::directory_exists(m_fx_dir))
+    {
+        // FX probe
+        m_probes.push_back(probe_config_t::fx(m_fx_dir, m_fx_deps.get()));
+    }
+
+    for (const auto& probe : m_additional_probes)
+    {
+        // Additional paths
+        assert(pal::directory_exists(probe));
```
gkhanna79: User could specify an incorrect/non-existent path. Instead of the assert, please add a check for path existence, similar to the earlier checks.  Finally, please add an entry to tracelog for all the paths (FX, Dotnetextensions, probepaths, etc) that are found/not-found.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+        // FIXME: When removing additional probes add back config get fx roll fwd.
```
gkhanna79: Can you please elaborate on this FIXME and why we are not passing the rollforward status for them?

leecow:
@gkhanna79 , @blackdwarf DOTNET_HOSTING_OPTIMIZATION_CACHE seems to be more accurate.

blackdwarf:
@gkhanna79 @leecow agreed, though it is looooong...

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+        {
+            iter = m_additional_probes.erase(iter);
+        }
+    }
+    // FIXME: Remove nuget support with runtimeconfig.dev.json
+    // and setup roll forward to come from the config in setup_probe_config
+    if (m_additional_probes.empty())
+    {
+        pal::string_t probe_dir;
+        (void)pal::get_default_packages_directory(&probe_dir);
```
gkhanna79: Assuming this is the set of paths specified in runtimeconfig.dev.json, I believe they should be treated no differently from other probing paths and be unconditionally added to the probe path list.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+            trace::verbose(_X("    Skipping... not serviceable asset"));
+            continue;
+        }
+        if (config.only_runtime_assets && entry.asset_type != deps_entry_t::asset_types::runtime)
+        {
+            trace::verbose(_X("    Skipping... not runtime asset"));
+            continue;
+        }
+        pal::string_t probe_dir = config.probe_dir;
+        assert(pal::directory_exists(probe_dir));
```
gkhanna79: I would suggest making this conditional to allow handling of transient conditions that may make the folder inaccessible/absent.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+            {
+                assert(!config.roll_forward);
+                trace::verbose(_X("    Matched hash for [%s]"), candidate->c_str());
+                return true;
+            }
+            trace::verbose(_X("    Skipping... match hash failed"));
+        }
+        else if (config.probe_deps_json)
+        {
+            // If the deps json has it then someone has already done rid selection and put the right stuff in the dir.
```
gkhanna79: What happens if someone didnt place artifacts for all applicable RIDs? Is there any reason to not do the exact RID specific (or fallback) match?

src/corehost/cli/deps_resolver.cpp
```diff
@@ -119,6 +122,197 @@ void deps_resolver_t::get_dir_assemblies(
...
+        {
+            m_additional_probes.push_back(probe_dir);
+        }
+    }
+}
+
+bool deps_resolver_t::probe_entry_in_configs(const deps_entry_t& entry, pal::string_t* candidate)
+{
+    candidate->clear();
+    for (const auto& config : m_probes)
```
gkhanna79: How is the order maintained for probing? That is, for each package, how do we ensure the following remains true:

dotnetextension
hosting_optimization_Cache
FX
additionalprobingpaths at command line
additionalprobingpaths in runtimeconfig.dev.json

src/corehost/cli/deps_resolver.cpp
```diff
@@ -126,22 +320,17 @@ void deps_resolver_t::get_dir_assemblies(
...
     {
         pal::string_t candidate;
 
-        // Servicing override.
         if (deps->has_coreclr_entry())
         {
             const deps_entry_t& entry = deps->get_coreclr_entry();
-            trace::verbose(_X("Probing for CoreCLR package=[%s][%s] in servicing"), entry.library_name.c_str(), entry.library_version.c_str());
-            if (entry.is_serviceable && m_svc.find_redirection(entry.library_name, entry.library_version, entry.relative_path, &candidate))
+            if (probe_entry_in_configs(entry, &candidate))
```
gkhanna79: CoreCLR probe order, if I read correctly, looks wrong. We are going through the configs (which map to various probe paths) and then looking AppLocal and then FX instead of "dotnetextension;optimizationcache;fx;additionalprobepaths;applocal"

src/corehost/cli/args.cpp
schellap: That is separate work and ties with dev.json

src/corehost/cli/deps_resolver.cpp
```diff
@@ -151,68 +340,21 @@ pal::string_t deps_resolver_t::resolve_coreclr_dir(
...
-            if (coreclr_exists_in_dir(deps_dir))
-            {
-                return deps_dir;
-            }
-        }
-
-        // Packages dir.
-        pal::string_t coreclr_package;
-        if (!package_dir.empty())
+        // App/FX main dir or standalone app dir.
```
gkhanna79: FX dir is already probed above so, at this time, we should be probing applocal only (which will address standalone too)?

src/corehost/cli/deps_resolver.cpp
schellap: Already there in the setup additional probes function.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -230,44 +372,14 @@ pal::string_t deps_resolver_t::resolve_coreclr_dir(
...
-    pal::string_t ni_package_cache_dir;
-    if (!package_cache_dir.empty())
-    {
-        ni_package_cache_dir = package_cache_dir;
-        append_path(&ni_package_cache_dir, get_arch());
-    }
-
     // Obtain the local assemblies in the app dir.
-    get_dir_assemblies(app_dir, _X("local"), &m_local_assemblies);
+    get_dir_assemblies(m_app_dir, _X("local"), &m_local_assemblies);
```
gkhanna79: I have a question about this - lets talk offline.

src/corehost/cli/deps_format.cpp
schellap: Removes warnings and makes it easier to see in the debugger.

src/corehost/cli/runtime_config.cpp
```diff
@@ -38,6 +38,29 @@ bool runtime_config_t::parse_opts(const json_value& opts)
...
+        {
+            const auto& arr = probe_paths->second.as_array();
+            for (const auto& str : arr)
+            {
+                m_probe_paths.push_back(str.as_string());
+            }
+        }
+    }
+
+    auto roll_fwd = opts_obj.find(_X("rollForward"));
```
gkhanna79: applyPatches and not rollForward

src/corehost/cli/deps_resolver.cpp
schellap: It is major min star to query the file system. However, the parameter production only should be set to true.

src/corehost/cli/runtime_config.cpp
```diff
@@ -38,6 +38,29 @@ bool runtime_config_t::parse_opts(const json_value& opts)
         }
     }
 
+    auto probe_paths = opts_obj.find(_X("additionalProbingPaths"));
```
gkhanna79: To confirm, this is in runtimeconfig.json and not the dev one, right?

src/corehost/common/pal.windows.cpp
```diff
@@ -130,8 +130,26 @@ bool pal::get_default_packages_directory(string_t* recv)
     {
         return false;
     }
-    append_path(&*recv, _X(".nuget"));
-    append_path(&*recv, _X("packages"));
+    append_path(recv, _X(".nuget"));
```
gkhanna79: Shouldnt this be a FIXME to remove once you add runtimeconfig.dev.json support?

src/corehost/cli/args.cpp
gkhanna79: And order, in which the multiple probe paths are specified, should be maintained.

src/corehost/cli/deps_resolver.cpp
gkhanna79: Never mind - I got it.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -151,68 +340,21 @@ pal::string_t deps_resolver_t::resolve_coreclr_dir(
             trace::verbose(_X("Deps has no coreclr entry."));
         }
```
gkhanna79: rid specific lookup for portable fat if deps cotnains it and probe fails to contain.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -151,68 +340,21 @@ pal::string_t deps_resolver_t::resolve_coreclr_dir(
...
 
         return pal::string_t();
     };
 
-    trace::info(_X("--- Starting CoreCLR Proble from app deps.json"));
-    pal::string_t clr_dir = process_coreclr(m_portable, app_dir, m_deps.get());
+    trace::info(_X("--- Starting CoreCLR Probe from app deps.json"));
+    pal::string_t clr_dir = process_coreclr(m_portable, m_app_dir, m_deps.get());
     if (clr_dir.empty() && m_portable)
```
gkhanna79: If app deps does not have a CoreCLR entry AND app is portable should you check for CoreCLR in FX. Otherwise, you should fail.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -288,37 +399,21 @@ void deps_resolver_t::resolve_tpa_list(
...
-        // The app is portable so the rid asset should be picked up from relative subpath.
-        else if (is_portable && deps->try_ni(entry).to_rel_path(app_dir, &candidate))
+        // The rid asset should be picked up from app relative subpath.
+        else if (entry.is_rid_specific && deps->try_ni(entry).to_rel_path(m_app_dir, &candidate))
         {
             add_tpa_asset(entry.asset_name, candidate, &items, output);
         }
-        // The app is portable, but there could be a rid-less asset in the app base.
+        // The rid-less asset should be picked up from the app base.
         else if (dir_assemblies.count(entry.asset_name))
```
gkhanna79: If the entry is not found, then we should fail the start.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -278,7 +390,6 @@ void deps_resolver_t::resolve_tpa_list(
 
     auto process_entry = [&](bool is_portable, deps_json_t* deps, const dir_assemblies_t& dir_assemblies, const deps_entry_t& entry)
```
gkhanna79: Delete is_portable as it is no longer used.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -288,37 +399,21 @@ void deps_resolver_t::resolve_tpa_list(
...
-        else if (entry.to_hash_matched_path(package_cache_dir, &candidate))
+        // Try to probe from the shared locations.
+        if (probe_entry_in_configs(entry, &candidate))
         {
             add_tpa_asset(entry.asset_name, candidate, &items, output);
         }
-        // The app is portable so the rid asset should be picked up from relative subpath.
-        else if (is_portable && deps->try_ni(entry).to_rel_path(app_dir, &candidate))
+        // The rid asset should be picked up from app relative subpath.
+        else if (entry.is_rid_specific && deps->try_ni(entry).to_rel_path(m_app_dir, &candidate))
```
gkhanna79: depsdir and not m_app_dir

src/corehost/common/pal.windows.cpp
gkhanna79: File an issue for rtm

src/corehost/common/pal.unix.cpp
```diff
@@ -134,6 +134,14 @@ bool pal::get_default_packages_directory(pal::string_t* recv)
     return true;
 }
 
+bool pal::get_default_extensions_directory(string_t* recv)
+{
+    recv->clear();
+    append_path(recv, _X("opt"));
+    append_path(recv, _X("dotnet_extensions"));
```
gkhanna79: Remove the "_"

src/corehost/cli/args.cpp
schellap: Cc @glennc name change for antares cache, looping in.

gkhanna79:
@schellap Good to merge?

Issue #2159, opened 2016-03-31
dotnet muxer reports JSON syntax error when global.json is saved in UTF8+BOM


pakrym:
## Steps to reproduce
1. Save global.json in utf8 with byte order mark.
2. dotnet build

## Expected  behavior
dotnet build shows no errors

## Actual behavior
```
A JSON parsing exception occurred: * Line 1, Column 2 Syntax error: Malformed token
```

## Environment data
`dotnet --version` output:
```
A JSON parsing exception occurred: * Line 1, Column 2 Syntax error: Malformed token
1.0.0-beta-002171
```

pakrym:
//cc @piotrpMSFT @davidfowl @victorhurdugaci @pranavkm

pakrym:
This makes build to fail.
```
dotnet build
A JSON parsing exception occurred: * Line 1, Column 2 Syntax error: Malformed token
Project 1 (.NETStandardApp,Version=v1.5) will be compiled because the CLI changed
Compiling 1 for .NETStandardApp,Version=v1.5
D:\temp\1\A JSON parsing exception occurred: * Line 1, Column 2 Syntax error: Malformed token

Compilation failed.
    0 Warning(s)
    1 Error(s)

Time elapsed 00:00:05.4565174
```

gkhanna79:
@pakrym Why is this blocking given that JSON parsing was working all this while?

pakrym:
Because now its output makes `dotnet build` to fail

gkhanna79:
Have you changed the format in which the .json files are saved?

pakrym:
No, it reports false positive for files with BOM

gkhanna79:
> Save global.json in utf8 with byte order mark.

How was this saved earlier when this worked? I am not aware of any JSON parsing changes that would result in this.

pakrym:
Reporting changed in https://github.com/dotnet/cli/commit/1428eb81cfb1e81251b8241d08e5fefd95882153, it started printing line numbers into `stderr` and `dotnet build` thinks its `csc` errors.

gkhanna79:
Are you suggesting that if trace_error was not executed, the build would have succeeded as that would be strange given that json parsing failed?

pakrym:
Global.json is perfectly valid json file, it just has UTF8 BOM encoding which `corehost` can not process correctly. So yes, build would have succeeded if this error wasn't written to `stderr`.

pakrym:
I'm not fixing it

[...]

PR #2177, opened 2016-03-31
Do not write json error message to stderr

pakrym:
Workarounds https://github.com/dotnet/cli/issues/2159
@schellap @davidfowl @gkhanna79

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2177)


src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -106,7 +106,7 @@ pal::string_t fx_muxer_t::resolve_cli_version(const pal::string_t& global_json)
     catch (const web::json::json_exception& je)
     {
         pal::string_t jes = pal::to_palstring(je.what());
-        trace::error(_X("A JSON parsing exception occurred: %s"), jes.c_str());
+        trace::info(_X("A JSON parsing exception occurred: %s"), jes.c_str());
     }
     trace::verbose(_X("CLI version is [%s] in global json file [%s]"), retval.c_str(), global_json.c_str());
     return retval;
```
schellap: If you want to turn that into an info, that is fine as a workaround. Please add a file name while you are at it, so you can close out your other issue.

`trace::info(_X("A JSON parsing exception occurred: %s"), jes.c_str());`
`-->`
`trace::info(_X("A JSON parsing exception occurred in %s: %s"), config_file.c_str(), jes.c_str());`

pakrym: I'd like to unblock ASP.NET with this.

davidfowl:
:shipit: Need this to unblock

Petermarcu:
Take it to keep things unblocked assuming CI passes.

PR #2219, opened 2016-04-02
Add pre-reqs doc to the repo

blackdwarf:
This document outlines all of the native dependencies that are needed to run the CLI tools for Ubuntu, Windows, CentOS and OS X. 
As per chat with @leecow documentation does not need special approval to get merged. 

skip ci please

/cc @piotrpMSFT @brthor @ellismg

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2219)

### DIFF
--- a/Documentation/cli-prerequisites.md
+++ b/Documentation/cli-prerequisites.md
@@ -1,0 +1,81 @@
+CLI native prerequisites
+=========================
+
+This document outlines the dependencies needed to run .NET Core CLI tools. Most of these dependencies are also .NET Core's general dependencies, so installing them will make sure that you can run applications written for .NET Core in general.
+
+## Windows dependencies
+On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
+
+* Windows 10
+    * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
+* Windows 7 & 8, Windows Server 2008 & 2012
+    *  [Visual C++ Redistributable for Visual Studio 2012 Update 4](https://www.microsoft.com/en-us/download/confirmation.aspx?id=30679)
+
+## Ubuntu
+Ubuntu distributions require the following libraries installed:
+
+* libc6
+* libedit2
+* libffi6
+* libgcc1
+* libicu52
+* liblldb-3.6
+* libllvm3.6
+* liblttng-ust0
+* liblzma5
+* libncurses5
+* libpython2.7
+* libstdc++6
+* libtinfo5
+* libunwind8
+* liburcu1
+* libuuid1
+* zlib1g
+* libasn1-8-heimdal
+* libcomerr2
+* libcurl3
+* libgcrypt11
+* libgnutls26
+* libgpg-error0
+* libgssapi3-heimdal
+* libgssapi-krb5-2
+* libhcrypto4-heimdal
+* libheimbase1-heimdal
+* libheimntlm0-heimdal
+* libhx509-5-heimdal
+* libidn11
+* libk5crypto3
+* libkeyutils1
+* libkrb5-26-heimdal
+* libkrb5-3
+* libkrb5support0
+* libldap-2.4-2
+* libp11-kit0
+* libroken18-heimdal
+* librtmp0
+* libsasl2-2
+* libsqlite3-0
+* libssl1.0.0
+* libtasn1-6
+* libwind0-heimdal
+
+## CentOS
+CentOS distributions require the following libraries installed:
+
+* deltarpm
+* epel-release
+* unzip
+* libunwind
+* gettext 
+* libcurl-devel 
+* openssl-devel 
+* zlib 
+* libicu-devel
+
+## OS X 
+OS X requires the following libraries and versions installed:
+
+* libssl 1.1
+
+## Installing the dependencies
+Please follow the reccomended practices of each operating system in question. For Linux, we reccomend using your package manager such as `apt-get` for Ubuntu and `yum` for CentOS. For OS X and upgrading the libssl, we reccomend using [Homebrew](https://brew.sh/) 
--- a/Documentation/cli-prerequisites.md
+++ b/Documentation/cli-prerequisites.md
@@ -7,7 +7,7 @@ This document outlines the dependencies needed to run .NET Core CLI tools. Most 
 On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
 
 > **Note:** these dependencies are chained in the installer, you only need to install them manually if you are using
-> the installer script (`install.sh` or `install.ps1`). 
+> the installer script (`dotnet-install.sh` or `dotnet-install.ps1`). 
 
 * Windows 10
     * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
--- a/Documentation/cli-prerequisites.md
+++ b/Documentation/cli-prerequisites.md
@@ -6,6 +6,9 @@ This document outlines the dependencies needed to run .NET Core CLI tools. Most 
 ## Windows dependencies
 On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
 
+> **Note:** these dependencies are chained in the installer, you only need to install them manually if you are using
+> the installer script (`install.sh` or `install.ps1`). 
+
 * Windows 10
     * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
 * Windows 7 & 8, Windows Server 2008 & 2012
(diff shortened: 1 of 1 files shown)


Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,81 @@
...
+* libedit2
+* libffi6
+* libgcc1
+* libicu52
+* liblldb-3.6
+* libllvm3.6
+* liblttng-ust0
+* liblzma5
+* libncurses5
+* libpython2.7
```
ellismg: Really?

Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,81 @@
...
+* libkeyutils1
+* libkrb5-26-heimdal
+* libkrb5-3
+* libkrb5support0
+* libldap-2.4-2
+* libp11-kit0
+* libroken18-heimdal
+* librtmp0
+* libsasl2-2
+* libsqlite3-0
```
ellismg: This also seems suspect?  Is it some transitive dependency?

Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,81 @@
...
+* zlib 
+* libicu-devel
+
+## OS X 
+OS X requires the following libraries and versions installed:
+
+* libssl 1.1
+
+## Installing the dependencies
+Please follow the reccomended practices of each operating system in question. For Linux, we reccomend using your package manager such as `apt-get` for Ubuntu and `yum` for CentOS. For OS X and upgrading the libssl, we reccomend using [Homebrew](https://brew.sh/)
```
ellismg: Maybe mention the link dance you have to do?

Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,81 @@
...
+
+This document outlines the dependencies needed to run .NET Core CLI tools. Most of these dependencies are also .NET Core's general dependencies, so installing them will make sure that you can run applications written for .NET Core in general.
+
+## Windows dependencies
+On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
+
+* Windows 10
+    * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
+* Windows 7 & 8, Windows Server 2008 & 2012
+    *  [Visual C++ Redistributable for Visual Studio 2012 Update 4](https://www.microsoft.com/en-us/download/confirmation.aspx?id=30679)
```
ellismg: Why is this needed?  We don't chain this into the installers today.  Should we?

Documentation/cli-prerequisites.md
blackdwarf: Really. :) This was taken from the Dockerfile used in build. @brthor maybe we should clean these up?

Documentation/cli-prerequisites.md
blackdwarf: This is needed as documentation because of the script installers. We need to have this somewhere for that scenario.

Documentation/cli-prerequisites.md
blackdwarf: Sure, that makes sense.

Documentation/cli-prerequisites.md
ellismg: Sorry, my point was that while we chain the VC 2015 Redist into the bundle, we don't chain in VC 2012.  Is this actually still a requirement on pre Windows 10?  Can't everything be on VC 2015 (and if not, should we be installing VC 2012 on Win 7-8.1 machines when oh use the bundle?)

shahid-pk:
The debian distros dependencies seems transitive. Ideally development dependencies like ,lldb , clang etc should not be needed by the end user (consumer). It would have been great if their was distinction between runtime and development packages like that of windows.

blackdwarf:
@shahid-pk great feedback, I will make that change (or at least try :)).

Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,84 @@
...
+* zlib 
+* libicu-devel
+
+## OS X 
+OS X requires the following libraries and versions installed:
+
+* libssl 1.1
+
+## Installing the dependencies
+Please follow the reccomended practices of each operating system in question. For Linux, we reccomend using your package manager such as `apt-get` for Ubuntu and `yum` for CentOS. For OS X and upgrading the libssl, we reccomend using [Homebrew](https://brew.sh/); if you do use it, do not forget the link phase you need to do at the end of install (`brew link openssl --force`).
```
MichalStrehovsky: Typo: "recommend" spelled wrong twice on this line.

Documentation/cli-prerequisites.md
brthor: IIRC we took this list from the `ldd` magic command @ellismg 

Can't find that PR at the moment, but is it possible liblldb would have a dependency on this? I know they have a python wrapper.

ellismg: Possibly.  FWIW, It feels really weird that we have this giant matrix of dependencies for Ubuntu and very few for CentOS.  Especially given that things don't line up (e.g. unzip is not listed here, we require random RPM tools there?).  We also seem to require -revel dependencies on CentOS which also feels wrong.

Note that LLDB is not required for normal CLI operation (you won't be able to debug the runtime using the SOS plugin but that's not something most folks are going to want to do).

blackdwarf: Gentlemen, I can make this into whatever, but I would need the final list. :)

Documentation/cli-prerequisites.md
blackdwarf: I have no idea. @schellap @gkhanna79 can you comment?

Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,81 @@
+CLI native prerequisites
+=========================
+
+This document outlines the dependencies needed to run .NET Core CLI tools. Most of these dependencies are also .NET Core's general dependencies, so installing them will make sure that you can run applications written for .NET Core in general.
+
+## Windows dependencies
+On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
+
+* Windows 10
```
gkhanna79: This should be "pre-Win10" and not "Win10".

Documentation/cli-prerequisites.md
gkhanna79: VC++ 2015 redist is required on all pre-Win10 machines. 

I dont know who is requiring VC++ 2012.

Documentation/cli-prerequisites.md
gkhanna79: Pre-reqs on non-Windows seem a really long list. Have we considered authoring script files that folks can use to install the pre-reqs? Such a file can be part of the repo and folks wouldnt have to manually install them.

brthor: @gkhanna79 the deb packages fulfill that experience, and I'd be hesitant about adding it to our install scripts. I'm worried that it would make our install scripts need `sudo`

blackdwarf:
Guys, wouldn't the pre-reqs be the same libraries (different package names) between CentOS and Ubuntu?

Documentation/cli-prerequisites.md
blackdwarf: Could be old and should be removed.

brthor:
@blackdwarf 
Yes they should be unless the corefx native shims are different for CentOS (possible but I don't know)

blackdwarf:
@ellismg do you know the answer to above?

Documentation/cli-prerequisites.md
blackdwarf: @brthor @gkhanna79 they will *definitely* need `sudo`. There is no workaround for that.

gkhanna79: You can "sudo" the script that is the source of truth and central point for maintaining dependencies, as opposed to listing them here and updating them when they change.

Documentation/cli-prerequisites.md
blackdwarf: Fixed.

MichaelSimons:
What about the RHEL dependencies?  Are these going to be documented as well?

blackdwarf:
@MichaelSimons good point, I have no idea. Wouldn't they be similar/same to CentOS?

MichaelSimons:
I would imagine so but can't say definitively.

ellismg:
> Guys, wouldn't the pre-reqs be the same libraries (different package names) between CentOS and Ubuntu?

At a high level, the answer to this is "yes".  Different distros may have different names for the packages and they might be at different versions, but we haven't yet been at a place where on RHEL/CentOS we use completely different libraries to solve the same problem (e.g. openssl somewhere and libressl somewhere else).

> I would imagine so but can't say definitively.

The packages names should be the same between CentOS and RHEL.

brthor:
@ellismg 

> At a high level, the answer to this is "yes". Different distros may have different names for the packages and they might be at different versions, but we haven't yet been at a place where on RHEL/CentOS we use completely different libraries to solve the same problem (e.g. openssl somewhere and libressl somewhere else).

Is this true today between ubuntu and CentOS as well?

ellismg:
> Is this true today between ubuntu and CentOS as well?

Yes.  I meant to type "Ubuntu/RHEL" to make it a little more clear.  Across all the OSes we produce binaries for we have the same logical set of dependencies.

blackdwarf:
Yes, this is something that I remember from one of @stephentoub talks. OK, well, the latest update to this PR makes the two lists the same.

Documentation/cli-prerequisites.md
```diff
@@ -0,0 +1,50 @@
+CLI native prerequisites
+=========================
+
+This document outlines the dependencies needed to run .NET Core CLI tools. Most of these dependencies are also .NET Core's general dependencies, so installing them will make sure that you can run applications written for .NET Core in general.
+
+## Windows dependencies
+On Windows, the only dependency is the VC++ Redistributable. Depending on the version of Windows you are running on, the versions are changing.
+
+> **Note:** these dependencies are chained in the installer, you only need to install them manually if you are using
+> the installer script (`install.sh` or `install.ps1`).
```
piotrpMSFT: This dotnet-install.[sh|ps1]

piotrpMSFT:
:shipit:

PR #2220, opened 2016-04-02
Enable Cross Platform Publishing and add tests

### LINKED ISSUE #317: Enable cross-platform publishing
It looks like the ExeSuffix used by dotnet-publish is determined by the platform that the tool is currently being run on rather than the runtime you passed in to the publish operation.

### LINKED ISSUE #2141: Stand alone executables are published for the platform they are produced on, not the target
When performing a stand alone publish, the stand alone executable is published for whatever platform you are publishing from. For example, publishing for Ubuntu.14.04-x64 from my Windows machine, I still end up with myapp.exe, not just 'myapp'. @blackdwarf sees the same thing publishing for Windows from OSX, he ends up with 'myapp', not 'myapp.exe'.

To reproduce, publish Windows for Ubuntu, or on OSX/Ubuntu for Windows.

### LINKED ISSUE #2217: hostfxr library for the sharedfx is copied from the sdk directory during build.
cc @schellap @Sridhar-MS

brthor:
Fixes #317
Fixes #2141
Fixes #2217 

This enables Cross Publishing but is currently blocked on getting an updated host in the `Microsoft.NETCore.App` metapackage. (tests will fail) ( @schellap @weshaggard ) 

I was able to verify the changes locally by patching the dropped host with a recently built one, and also verified that appropriate files are dropped when publishing for a different platform.

After the metapackage is published we'll need to update the package references.

cc @gkhanna79 @piotrpMSFT @Sridhar-MS 

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2220)

### DIFF
--- a/TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
+++ b/TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
@@ -1,0 +1,35 @@
+{
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": [
+        "dnxcore50",
+        "portable-net45+win8"
+      ],
+      "dependencies": {
+        "Microsoft.NETCore.App": "1.0.0-rc2-*"
+      }
+    }
+  },
+  "runtimes": {
+    "win7-x64": {},
+    "win7-x86": {},
+    "osx.10.10-x64": {},
+    "osx.10.11-x64": {},
+    "ubuntu.14.04-x64": {},
+    "centos.7-x64": {},
+    "rhel.7.2-x64": {},
+    "debian.8.2-x64": {}
+  },
+
+  "runtimeOptions": {
+    "somethingString": "anything",
+    "somethingBoolean": true,
+    "someArray": ["one", "two"],
+    "someObject": {
+      "someProperty": "someValue"
+    }
+  }
+}
--- a/TestAssets/TestProjects/KestrelSample/KestrelStandalone/project.json
+++ b/TestAssets/TestProjects/KestrelSample/KestrelStandalone/project.json
@@ -12,9 +12,9 @@
     "../src/*.cs"
   ],
   "frameworks": {
-    "netstandardapp1.5": {
+    "netcoreapp1.0": {
       "dependencies": {
-        "Microsoft.NETCore.App": "1.0.0-rc2-24008"
+        "Microsoft.NETCore.App": "1.0.0-rc2-3002306"
       },
       "imports": [
         "dnxcore50",
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -422,6 +422,9 @@ namespace Microsoft.DotNet.Cli.Build
             File.Copy(
                 Path.Combine(Dirs.Corehost, HostPolicyBaseName),
                 Path.Combine(SharedFrameworkNameAndVersionRoot, HostPolicyBaseName), true);
+            File.Copy(
+                Path.Combine(Dirs.Corehost, DotnetHostFxrBaseName),
+                Path.Combine(SharedFrameworkNameAndVersionRoot, DotnetHostFxrBaseName), true);
 
             if (File.Exists(Path.Combine(SharedFrameworkNameAndVersionRoot, "mscorlib.ni.dll")))
             {
--- a/scripts/dotnet-cli-build/TestTargets.cs
+++ b/scripts/dotnet-cli-build/TestTargets.cs
@@ -58,7 +58,7 @@ namespace Microsoft.DotNet.Cli.Build
         [Target(nameof(RestoreTestAssetPackages), nameof(BuildTestAssetPackages))]
         public static BuildTargetResult SetupTestPackages(BuildTargetContext c) => c.Success();
 
-        [Target(nameof(RestoreTestAssetProjects), nameof(RestoreDesktopTestAssetProjects), nameof(BuildTestAssetProjects))]
+        [Target(nameof(RestoreTestAssetProjects), nameof(RestoreDesktopTestAssetProjects), nameof(RestoreCrossPublishTestAssetProjects), nameof(BuildTestAssetProjects))]
         public static BuildTargetResult SetupTestProjects(BuildTargetContext c) => c.Success();
 
         [Target]
@@ -135,6 +135,18 @@ namespace Microsoft.DotNet.Cli.Build
                 .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "TestAssets", "DesktopTestProjects"))
                 .Execute().EnsureSuccessful();
 
+            return c.Success();
+        }
+        
+        [Target]
+        public static BuildTargetResult RestoreCrossPublishTestAssetProjects(BuildTargetContext c)
+        {
+            var dotnet = DotNetCli.Stage2;
+
+            dotnet.Restore("--verbosity", "verbose")
+                .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "TestAssets", "CrossPublishTestProjects"))
+                .Execute().EnsureSuccessful();
+                
             return c.Success();
         }
 
--- a/src/dotnet/commands/dotnet-publish/PublishCommand.cs
+++ b/src/dotnet/commands/dotnet-publish/PublishCommand.cs
@@ -139,11 +139,10 @@ namespace Microsoft.DotNet.Tools.Publish
             var exports = exporter.GetAllExports();
             foreach (var export in exports.Where(e => !collectExclusionList.Contains(e.Library.Identity.Name)))
             {
-                Reporter.Verbose.WriteLine($"Publishing {export.Library.Identity.ToString().Green().Bold()} ...");
+                Reporter.Verbose.WriteLine($"publish: Publishing {export.Library.Identity.ToString().Green().Bold()} ...");
 
                 PublishAssetGroups(export.RuntimeAssemblyGroups, outputPath, nativeSubdirectories: false, includeRuntimeGroups: isPortable);
                 PublishAssetGroups(export.NativeLibraryGroups, outputPath, nativeSubdirectories, includeRuntimeGroups: isPortable);
-                export.RuntimeAssets.StructuredCopyTo(outputPath, outputPaths.IntermediateOutputDirectoryPath);
             }
 
             if (options.PreserveCompilationContext.GetValueOrDefault())
@@ -153,19 +152,9 @@ namespace Microsoft.DotNet.Tools.Publish
                     PublishRefs(export, outputPath, !collectExclusionList.Contains(export.Library.Identity.Name));
                 }
             }
-
-            if (context.ProjectFile.HasRuntimeOutput(configuration) && !context.TargetFramework.IsDesktop())
-            {
-                // Get the output paths used by the call to `dotnet build` above (since we didn't pass `--output`, they will be different from
-                // our current output paths)
-                var buildOutputPaths = context.GetOutputPaths(configuration, buildBasePath);
-                PublishFiles(
-                    new[] {
-                        buildOutputPaths.RuntimeFiles.DepsJson,
-                        buildOutputPaths.RuntimeFiles.RuntimeConfigJson
-                    },
-                    outputPath);
-            }
+            
+            var buildOutputPaths = context.GetOutputPaths(configuration, buildBasePath, null);
+            PublishBuildOutputFiles(buildOutputPaths, context, outputPath, Configuration);
 
             var contentFiles = new ContentFiles(context);
             contentFiles.StructuredCopyTo(outputPath);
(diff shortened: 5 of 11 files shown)


scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -397,6 +397,9 @@ public static void PublishSharedFramework(BuildTargetContext c, string outputDir
             File.Copy(
                 Path.Combine(Dirs.Corehost, HostPolicyBaseName),
                 Path.Combine(SharedFrameworkNameAndVersionRoot, HostPolicyBaseName), true);
+            File.Copy(
+                Path.Combine(Dirs.Corehost, DotnetHostFxrBaseName),
+                Path.Combine(SharedFrameworkNameAndVersionRoot, DotnetHostFxrBaseName), true);
```
Sridhar-MS: Good catch!!

schellap: There is the host executables in Constants which is an array of host dependencies.

test/dotnet-publish.Tests/PublishTests.cs
```diff
@@ -92,6 +92,29 @@ public void FailWhenNoRestoreTest()
             publishCommand.Execute().Should().Fail();
         }
 
+        [Theory]
+        [InlineData("ubuntu.14.04-x64", "", "libhostfxr.so", "libcoreclr.so", "libhostpolicy.so")]
```
gkhanna79: What about RHEL/CentOS?

brthor: My reasoning was that covering each of the expected asset types `.so, .dll, .dylib`, would cover most situations, but I can see that leaves a testing gap related to packaging issues for a specific platform.

Will add it in, thanks

brthor:
@weshaggard @schellap Any update for the `Microsoft.NETCore.App` package with the updated host dependency?

schellap:
@weshaggard could you please use the shared packages 1.0.1-beta-002200 (beta because corehost does some build version matching to locate its own serviced packages) and I would like it to be tied to the CLI build system.

weshaggard:
The new host packages are uploaded to dotnet-core feed but we don't have a Microsoft.NETCore.App meta-package that has the dependencies yet.

brthor:
I tested out with the new host package and it seems to be publishing the right things, and at least the current platform is runnable, need to do further testing with moving the published bits between machines.

Is the plan for #2274 to go in, wait for a build, and then get the new NETCore.App package?

piotrpMSFT:
@brthor this is both [do not merge] and `--pending-approval`. We should wait for the PR to be green before asking for approval. This tag is tracking `to do`s for approvers.

brthor:
Removed the label, I misunderstood its intention.

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -77,10 +77,16 @@ private void MakeCompilationOutputRunnableForCoreCLR()
             if (emitEntryPoint && !string.IsNullOrEmpty(_context.RuntimeIdentifier))
             {
                 // TODO: Pick a host based on the RID
-                CoreHost.CopyTo(_runtimeOutputPath, _compilerOptions.OutputName + Constants.ExeSuffix);
+                CopyHostTo(_runtimeOutputPath, _compilerOptions.OutputName + Constants.ExeSuffix);
+                // CoreHost.CopyTo(_runtimeOutputPath, _compilerOptions.OutputName + Constants.ExeSuffix);
             }
         }
 
+        private void CopyHostTo(string outputDirectory, string renameHost = "")
```
brthor: Will Delete this.

scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -137,6 +137,18 @@ public static BuildTargetResult RestoreDesktopTestAssetProjects(BuildTargetConte
 
             return c.Success();
         }
+        
+        [Target]
+        public static BuildTargetResult RestoreCrossPublishTestAssetProjects(BuildTargetContext c)
```
brthor: removing this, no longer needed.

TestAssets/TestProjects/KestrelSample/KestrelStandalone/project.json
```diff
@@ -12,9 +12,9 @@
     "../src/*.cs"
   ],
   "frameworks": {
-    "netstandardapp1.5": {
+    "netcoreapp1.0": {
       "dependencies": {
-        "Microsoft.NETCore.App": "1.0.0-rc2-24008"
+        "Microsoft.NETCore.App": "1.0.0-rc2-3002306"
```
eerhardt: How are we going to keep this version up to date?  Should this use *?

brthor: After generating the stub package for x86/x64 on windows we can use the *.

Will make this change in a separate PR.

TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
```diff
@@ -0,0 +1,35 @@
...
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": [
+        "dnxcore50",
+        "portable-net45+win8"
+      ],
+      "dependencies": {
```
eerhardt: I think the more standard is to put dependencies at the top level, isn't it?  At least for me it makes it more readable, since this only targets 1 TFM.

TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
```diff
@@ -0,0 +1,35 @@
...
+    "osx.10.10-x64": {},
+    "osx.10.11-x64": {},
+    "ubuntu.14.04-x64": {},
+    "centos.7-x64": {},
+    "rhel.7.2-x64": {},
+    "debian.8.2-x64": {}
+  },
+
+  "runtimeOptions": {
+    "somethingString": "anything",
```
eerhardt: Are these used?

TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
brthor: I copied the standalone app present in PortableTests. 

I agree we could use cleanup all around in the TestAssets but i think it's out of the scope of this PR.

krwq:
:ship: :shipit:

TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
brthor: They are not tested, no

eerhardt: I think they should be removed.

TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
eerhardt: I'm not asking to clean up other project.jsons.  I just meant this new one should look more "standard" - like how we generate it in the `dotnet new` template.

TestAssets/CrossPublishTestProjects/StandaloneAppCrossPublish/project.json
brthor: After merging the stub package fix, this whole project will go away.

I should've stated that first, but that is my primary reason for not wanting to take another trip through CI for cleanup.

eerhardt: I'm not even sure we should do the stub package fix, because I don't think standalone apps should be forced to reference Microsoft.NETCore.App.

brthor: As far as I know this is the current Plan of Record. 

If this is changing in the immediate term, this whole PR is not going to work because it depends on a host being present in the dependency graph for X-Publish.

brthor: Anyways let's move this discussion to #2438

src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -272,33 +316,60 @@ private static void PublishRefs(LibraryExport export, string outputPath, bool de
...
+            var publishedHostFile = ResolvePublishedHostFile(outputPath);
+            if (publishedHostFile == null)
             {
-                var hostBinaryPath = Path.Combine(AppContext.BaseDirectory, binaryName);
-                if (!File.Exists(hostBinaryPath))
-                {
-                    Reporter.Error.WriteLine($"Cannot find {binaryName} in the dotnet directory.".Red());
-                    return 1;
-                }
+                Reporter.Output.WriteLine($"publish: warning: host executable not available in dependencies, using host for current platform");
```
schellap: This should be an error, otherwise we are copying some random host and will be hard to debug issues.

brthor: I agree, that this should be a host. 

The host is not random however, it is copied from the shared fx directory (falls back to previous behavior). 

This will be a breaking change to remove the fallback, so I'm not too sure the best way to go about this.

schellap: Is it copied from the shared FX directory the app targeted? or the shared FX directory that CLI targeted? The difference could be problematic.

brthor: Hmmm, from the cli target.

schellap: Why are you relying on the old behavior? Your PR is fixing it as it is not the best thing to do. Even though they are binaries, there are testing, compatibility, servicing implications. They compiled the app with two CLIs and get different behaviors?

brthor: The primary reason I relied on old behavior was to temporarily limit just how breaking this change is (even in our own repo). 

I'm working on the changes in our own repo so we can remove the fallback behavior (it's a decent amount).

I do agree that we should remove this as it is completely inconsistent.

gkhanna79: @brthor I will be filing an issue to ensure that host/hostfxr/hostpolicy.dll get published from the packages part of the graph of FX that the app declares dependency upon.

brthor: @gkhanna79 should we add it to here? #2444

brthor: I want to make this clear that this fallback only happens when there is no host in the package graph.

gkhanna79: > I want to make this clear that this fallback only happens when there is no host in the package graph.

Is this a point in time thing and be removed for RTM? CLI, when publishing a standalone app, should be publishing the host/hostfxr/hostpolicy corresponding to what is referenced in the app's package graph.

schellap: @brthor, I wanted to understand why you are relying on the fallback, I am okay if you want to put it there for a good reason. Is it to support apps that target older netcoreapp and hence may not have it in the dependency closure?

schellap: My worry is this will mask issues from the publishing to feed side and complications like mismatch from the app dev side. I know you are making it better than what it is today, why not make it how it should be as we have everything in place to do so.

brthor: In our testbed we have a lot of tests which don't have the host in their dependency closure yet (and our partners would too). In trying to convert them I got blocked on our method of producing the NETCoreApp package only for the current rid (this breaks down on windows because it tries to restore both x86 and x64).

To get this scenario lit up, I left in the fallback behavior while I work on unblocking the issues with the package and convert the tests ( #2450 ). 

This is entirely a Point in Time issue and I will try to get it in for RC2.

brthor: Completely agree with you @schellap . The fallback was to get the scenario unblocked for our partners who need it.

schellap: That is a good reason to do this. Sure, temporarily... until we figure out how to solve this. As long as the graph resolved hostpolicy.dll gets coverage (with your added test), i.e., you should activate the test app on the main platform and it should work -- that doesn't have to block this PR. 

For cross what you have... just verifying the artifacts exist is okay.

:+1:

MartinJohns:
This does not seem to work. I tried to do a cross-platform publish from Windows to Ubuntu, but the resulting output still produces an **.exe** file.

Command used. `dotnet publish --runtime ubuntu.14.04-x64 --output out5`

dotnet-info:
```
.NET Command Line Tools (1.0.0-rc2-002374)

Product Information:
 Version:     1.0.0-rc2-002374
 Commit Sha:  f82da07fc4

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

src/dotnet/commands/dotnet-publish/PublishCommand.cs
brthor: I manually tested the mac -> Windows runnability as well

brthor:
@MartinJohns Please see the example app I posted on the relevant issue. 

Also please share a project.json

MartinJohns:
After a bunch of attempts with different packages I got it running, as it's supposed to be. Plenty of comments here: dotnet/cli#317

PR #2149, opened 2016-03-30
EULA update for telemetry/data collection

danquirk:
Updated the two copies of the OSX EULA. Not a super readable diff in GitHub unfortunately. Addition is this text as section 2:

_DATA. The software may collect information about you and your use of the software, and send that to Microsoft. Microsoft may use this information to improve our products and services. You can learn more about data collection and use in the help documentation and the privacy statement at http://go.microsoft.com/fwlink/?LinkId=528096.Your use of the software operates as your consent to these practices._

README updated to reflect acceptance of the EULA.

TODO:
- [x] Update the Windows copy of the EULA. Where is this?

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2149)

PR #2223, opened 2016-04-02
Telemetry notification on installation of debian package

danquirk:
To go along with https://github.com/dotnet/cli/pull/2149 we want to notify on install for the debian packages which otherwise don't show the EULA. This adds a simple postinstall script to the package which direct users to the GitHub copy of the ENU EULA like so:
image

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2223)

### DIFF
--- a/packaging/debian/postinst
+++ b/packaging/debian/postinst
@@ -1,3 +1,3 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 echo "This software may collect information about you and your use of the software, and send that to Microsoft."
 echo "Please visit http://aka.ms/dotnet-cli-eula for more information."
--- a/packaging/debian/postinst
+++ b/packaging/debian/postinst
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+echo "This software may collect information about you and your use of the software, and send that to Microsoft."
+echo "Please visit http://aka.ms/dotnet-cli-eula for more information."
--- a/scripts/package/package-debian.sh
+++ b/scripts/package/package-debian.sh
@@ -158,6 +158,7 @@ create_empty_debian_layout(){
     mkdir "$PACKAGE_LAYOUT_DIR/package_root"
     mkdir "$PACKAGE_LAYOUT_DIR/samples"
     mkdir "$PACKAGE_LAYOUT_DIR/docs"
+    mkdir "$PACKAGE_LAYOUT_DIR/debian"
 }
 
 copy_files_to_debian_layout(){
@@ -177,7 +178,7 @@ copy_files_to_debian_layout(){
     for manpage in "$PACKAGE_LAYOUT_DIR/docs/"*; do mv "$manpage" "${manpage}-${DOTNET_CLI_VERSION}.1"; done
 
     # Copy postinstall
-    cp "$PACKAGING_ROOT/postinst" "$PACKAGE_LAYOUT_DIR/$"
+    cp "$PACKAGING_ROOT/postinst" "$PACKAGE_LAYOUT_DIR/debian/postinst"
 }
 
 create_debian_package(){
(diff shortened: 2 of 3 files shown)


packaging/debian/postinst
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
ellismg: Does using `sh` instead make sense here?  I believe that on Ubuntu bash always exists (even if it's just a symlink to dash) but maybe it makes sense to say `sh` instead since it's more general?

tools/DebianPackageTool/package_tool
```diff
@@ -152,6 +152,9 @@ package_all(){
     package_absolute_placement
     package_samples
     package_docs
+    if [[ $PACKAGE_NAME == "dotnet-dev-"* ]]; then
+        package_debian
```
ellismg: FWIW, I think the name of this function is confusing.  I would not expect the guard given the general name (e.g. wouldn't I always want to package postinst if it was present?

packaging/debian/postinst
brthor: I concur with @ellismg here, we're not using any bashims and using sh will give the community greater freedom to try installing the package in other places.

scripts/package/package-debian.sh
```diff
@@ -171,6 +171,9 @@ copy_files_to_debian_layout(){
 
     # Copy Manpages
     cp -a "$MANPAGE_DIR/sdk/." "$PACKAGE_LAYOUT_DIR/docs"
+
+    # Copy postinstall
+    cp "$PACKAGING_ROOT/postinst" "$PACKAGE_LAYOUT_DIR/$"
```
brthor: This is fairly different than what we discussed, and this will end up placing the script in `/` (root of the filesystem)

What about copying the `postinst` script to a new directory called `debian` rather than `$` which will allow for additional customization rather than this special case only?

tools/DebianPackageTool/package_tool
brthor: We shouldn't be leaking semantic knowledge of the cli into the `package_tool` like this, it is intended to be general purpose. 

Why not remove the guard altogether and calling this function always?

tools/DebianPackageTool/package_tool
```diff
@@ -216,6 +219,12 @@ package_docs(){
     fi
 }
 
+package_debian(){
+    if [[ -d "$ABSOLUTE_PLACEMENT_DIR" ]]; then
```
brthor: To follow the design I mentioned above, this would copy all files from `$INPUT_DIR/debian` to `$DEBIAN_DIR`

That would allow for any kind of script or custom debian metadata.

tools/DebianPackageTool/package_tool
danquirk: I didn't initially have the guard but then it ended up trying to copy postinst into the sharedframework package which failed. Was being a little quick and dirty to get something working. Open to whatever the right thing to do is.

Agree the name sucks.

scripts/package/package-debian.sh
danquirk: I'll try again now that I know what I'm doing better. I initially was trying to put it into a folder like that but it wasn't showing up when I inspected the package. Will investigate.

packaging/debian/postinst
danquirk: Will do.

tools/DebianPackageTool/package_tool
danquirk: Done

tools/DebianPackageTool/package_tool
```diff
@@ -219,9 +217,10 @@ package_docs(){
     fi
 }
 
-package_debian(){
-    if [[ -d "$ABSOLUTE_PLACEMENT_DIR" ]]; then
-        cp "$ABSOLUTE_PLACEMENT_DIR/postinst" $DEBIAN_DIR/postinst
+package_install_scripts(){
+    # copy scripts for the package's control section like preinst, postint, etc
```
brthor: This will cover much more than install scripts but it's good for now :smile:

brthor:
:shipit: Nice thanks for doing this!

PR #2235, opened 2016-04-02
StreamForwarder shouldn't ignore empty lines

### LINKED ISSUE #2234: dotnet run hides empty lines from output
## Steps to reproduce
Per dotnet/corefx#7409, use `dotnet new`, modify the code to contain:

```c#
Console.Write("hello" + Environment.NewLine);
Console.Write(Environment.NewLine);
Console.WriteLine();
Console.Write("world" + Environment.NewLine);
```

Then run it using `dotnet run`.

## Expected  behavior
Two empty lines are shown:

```
hello

world
```

## Actual behavior
No empty lines:

```
hello
world
```

## Cause

This is caused by [this code in `StreamFormwarder`](https://github.com/dotnet/cli/blob/3188d0559bf8c7107ae03f5e2ffe98587045b709/src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs#L86-L89).

svick:
Fixes  #2234.

Also performed minor cleanup of `StreamForwarderTests`.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2235)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
@@ -83,11 +83,6 @@ namespace Microsoft.DotNet.Cli.Utils
 
         private void WriteBuilder()
         {
-            if (_builder.Length == 0)
-            {
-                return;
-            }
-
             WriteLine(_builder.ToString());
             _builder.Clear();
         }
--- a/test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
+++ b/test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
@@ -13,14 +13,13 @@ using Microsoft.DotNet.Tools.Test.Utilities;
 using Microsoft.Extensions.PlatformAbstractions;
 using System.Threading;
 
-namespace StreamForwarderTests
+namespace Microsoft.DotNet.Cli.Utils.Tests
 {
     public class StreamForwarderTests : TestBase
     {
-        private static readonly string s_rid = PlatformServices.Default.Runtime.GetLegacyRestoreRuntimeIdentifier();
         private static readonly string s_testProjectRoot = Path.Combine(AppContext.BaseDirectory, "TestAssets", "TestProjects");
 
-        private TempDirectory _root;
+        private readonly TempDirectory _root;
 
         public static void Main()
         {
@@ -101,6 +100,7 @@ namespace StreamForwarderTests
             TestCapturingAndForwardingHelper(ForwardOptions.WriteLine | ForwardOptions.Capture, inputStr, expectedCaptured, expectedWrites);
         }
 
+        [Flags]
         private enum ForwardOptions
         {
             None = 0x0,
--- a/test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
+++ b/test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
@@ -38,16 +38,18 @@ namespace Microsoft.DotNet.Cli.Utils.Tests
                 return new[]
                 {
                     new object[] { "123", new string[]{"123"} },
-                    new object[] { "123\n", new string[] {"123"} },
-                    new object[] { "123\r\n", new string[] {"123"} },
+                    new object[] { "123\n", new string[] {"123", ""} },
+                    new object[] { "123\r\n", new string[] {"123", ""} },
                     new object[] { "1234\n5678", new string[] {"1234", "5678"} },
                     new object[] { "1234\r\n5678", new string[] {"1234", "5678"} },
-                    new object[] { "1234\n5678\n", new string[] {"1234", "5678"} },
-                    new object[] { "1234\r\n5678\r\n", new string[] {"1234", "5678"} },
+                    new object[] { "1234\n\n5678", new string[] {"1234", "", "5678"} },
+                    new object[] { "1234\r\n\r\n5678", new string[] {"1234", "", "5678"} },
+                    new object[] { "1234\n5678\n", new string[] {"1234", "5678", ""} },
+                    new object[] { "1234\r\n5678\r\n", new string[] {"1234", "5678", ""} },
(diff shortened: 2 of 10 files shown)


test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
```diff
@@ -13,14 +13,13 @@
 using Microsoft.Extensions.PlatformAbstractions;
 using System.Threading;
 
-namespace StreamForwarderTests
+namespace Microsoft.DotNet.Cli.Utils
```
brthor: `.Tests` ?

svick: You're right, fixed.

test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
```diff
@@ -39,16 +38,18 @@ public static IEnumerable<object[]> ForwardingTheoryVariations
                 return new[]
                 {
                     new object[] { "123", new string[]{"123"} },
-                    new object[] { "123\n", new string[] {"123"} },
-                    new object[] { "123\r\n", new string[] {"123"} },
+                    new object[] { "123\n", new string[] {"123", ""} },
```
brthor: What is your reasoning here that this should produce an extra empty line?

svick: My reasoning is that if `\n` is added to the input, a line should be added to the output. `Console.WriteLine("123")` and `Console.WriteLine("123\n")` don't produce the same output, so I don't see why this should.

Another reason is that if this was the desired behavior, the fix of this issue would be more complicated.

svick:
@brthor I don't quite understand why this fails the tests. I created #2244, which fixes the issue in a different way and passes tests. Do you think #2244 is a better approach or should I look into the test failures here?

test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
brthor: Hmm this reasoning makes sense, but I'm wondering about this:

`s_expectedOutput = "Hello World!" + Environment.NewLine + Environment.NewLine;`

The app that one is executing just uses Console.WriteLine, so we should only expect to see one newline don't you think?

src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
```diff
@@ -83,11 +83,6 @@ public void Read(TextReader reader)
 
         private void WriteBuilder()
         {
-            if (_builder.Length == 0)
```
eerhardt: Isn't the real problem in the method directly above this one?

```C#
private static readonly char s_flushBuilderCharacter = '\n';

...

public void Read(TextReader reader)
{
...

    if (currentCharacter == s_flushBuilderCharacter)
    {
        WriteBuilder();
    }
    else if (! s_ignoreCharacters.Contains(currentCharacter))
    {
        _builder.Append(currentCharacter);
    }
```

The newlines aren't getting written to the builder.

brthor: The builder is written out per line using WriteLine so that shouldn't be necessary.

Seems like there is an edge case here when you have nothing but newlines though that isn't accounted for properly.

eerhardt: Won't you get multiple `\r` characters this way?

On Windows, the input will have `\r\n`. The Read method above will write the `\r` to the builder. Then when it reads the next character, it will call WriteLine when it sees `\n`.  Next the WriteLine method appends a Environment.NewLine to the end of the string, which then results in `\r\r\n`.

Doing it this way seems incorrect.  Instead I would imagine the Read method looking something like:

```C#
             while ((readCharacterCount = reader.Read(buffer, 0, bufferSize)) > 0)
             {
                 currentCharacter = buffer[0];
 
                 if (! s_ignoreCharacters.Contains(currentCharacter))
                 {
                     _builder.Append(currentCharacter);

                    if (currentCharacter == s_flushBuilderCharacter)
                    {
                           WriteBuilder();
                     }
                 }
             }
```

And then "WriteBuilder" just Writes the output, it doesn't do a "WriteLine".

brthor: @eerhardt The case you mentioned is handled by ignoring `\r` characters altogether in the input, and relying on WriteLine to insert the appropriate newline for the platform.

See: `! s_ignoreCharacters.Contains(currentCharacter)'

Your proposed change would end up omitting the `\r` altogether (since none is appended in the WriteBuilder function).

I don't think whether or not to include the newlines in the builder is the heart of the issue here though. Since we are doing a WriteLine per newline, the behavior should be identical. 

The heart of the issue is what do we do when we encounter `\n\n` or `\r\n\r\n` as an input? Currently we'd do two empty writelines, which I recently discovered won't write anything (not even a newline). 

Including the newlines in the builder is one solution to this problem (which your suggested change @eerhardt would fix). 

I think the ideal fix here is to stop special casing for `\r` and go with eric's suggestion, which would look like this:

```
            while ((readCharacterCount = reader.Read(buffer, 0, bufferSize)) > 0)
             {
                 currentCharacter = buffer[0];

                 _builder.Append(currentCharacter);

                    if (currentCharacter == s_flushBuilderCharacter)
                    {
                           WriteBuilder();
                     }
                 
             }
```

test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs
svick: I do, but I also think that `Console.Write(…)` and `Console.WriteLine(…)` should produce different output. Which is why I now think that #2244 is a better solution, it doesn't have this problem.

src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
svick: @brthor Ok, I have modified #2244 to just pass `\r` through.

eerhardt:
Closing this in favor of #2244.  We will use that fix going forward.

Issue #2200, opened 2016-04-01
ProjectModel.Server sends null versions when csproj is just added to project.json and there was no restore


abpiskunov:
## Steps to reproduce

          "ClassLibrary1": {
            "Name": "ClassLibrary1",
            "DisplayName": "ClassLibrary1",
            "Version": null,
            "Path": null,
            "Type": "Project",
            "Resolved": false,
            "Dependencies": [],
            "Errors": [
              {
                "ErrorCode": "NU1001",
                "SourceFilePath": "C:\\Users\\antonpis\\Documents\\Visual Studio 2015\\Projects\\ClassLibrary5\\src\\ConsoleApp1\\project.json",
                "Message": "The dependency ClassLibrary1 could not be resolved.",
                "Severity": 2,
                "StartLine": 13,
                "StartColumn": 28,
                "EndLine": 13,
                "EndColumn": 28,
                "FormattedMessage": "C:\\Users\\antonpis\\Documents\\Visual Studio 2015\\Projects\\ClassLibrary5\\src\\ConsoleApp1\\project.json(13,28): error NU1001: The dependency ClassLibrary1 could not be resolved.",
                "Source": {
                  "Name": "ClassLibrary1",
                  "Version": null
                }
              }
            ],
            "Warnings": []
          },

It is new , we never received nulls for versions and it breaks our code, since this is a Type=Project , could you please send a conventional default “1.0.0” for their version?

## Environment data
`dotnet --info` output:

abpiskunov:
@troydai

troydai:
/cc @muratg @Eilon @piotrpMSFT @davidfowl 

Since we're in ask mode, please help me to triage this bug. The fix is straightforward: filling a default version value in place to avoid the null value in project.json. The change will be low risk as well. On the other hand @abpiskunov here can help fill in the impact of this bug.

davidfowl:
Seems fine.

muratg:
Looks fine to me too. Email @Eilon and @DamianEdwards for RC2 approval.

troydai:
Fixed in cb14076b38a59c30176b887d0a0fb1318a58c54e

PR #2249, opened 2016-04-04
Use default version for unresolved MSBuild reference

troydai:
Issue #2200 

/cc @davidfowl @abpiskunov @piotrpMSFT 

Note: The fix is made a project model server. It is a conservative approach because the root cause is in the `ProjectReader` ([this line](https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/ProjectReader.cs#L271)). Here an unresolved `project` type dependencies is added the `null` version leak into the project model.

Now I bank my fix high on the stack just to ensure the impact will be limited to project model server.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2249)


troydai:
Rebase and waiting for a new round of CI build.

PR #2264, opened 2016-04-04
Optimize dependency context memory usage

pakrym:
1. Pool strings
2. Do not hold onto `DependencyContextJsonReader` in `DependencyContextLoader`
3. Use same reader with same string pool to load multiple contexts in case of portable app.

Brings `DependencyContext` size in ASP.NET app to ~250kb.

/cc @davidfowl

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2264)

### DIFF
--- a/src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
+++ b/src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Extensions.DependencyModel
 {
     public class DependencyContextJsonReader : IDependencyContextReader
     {
+        private readonly IDictionary<string, string> _stringPool = new Dictionary<string, string>();
+
         public DependencyContext Read(Stream stream)
         {
             if (stream == null)
@@ -26,6 +28,19 @@ namespace Microsoft.Extensions.DependencyModel
                     return Read(root);
                 }
             }
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                _stringPool.Clear();
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
         }
 
         private bool IsRuntimeTarget(string name) => name.Contains(DependencyContextStrings.VersionSeperator);
@@ -178,8 +193,8 @@ namespace Microsoft.Extensions.DependencyModel
 
             var seperatorPosition = nameWithVersion.IndexOf(DependencyContextStrings.VersionSeperator);
 
-            var name = nameWithVersion.Substring(0, seperatorPosition);
-            var version = nameWithVersion.Substring(seperatorPosition + 1);
+            var name = Pool(nameWithVersion.Substring(0, seperatorPosition));
+            var version = Pool(nameWithVersion.Substring(seperatorPosition + 1));
 
             var libraryObject = (JObject)property.Value;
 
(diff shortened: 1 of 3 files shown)


krwq:
What's the previous DependencyContext size?

pakrym:
~550 kb

pakrym:
@dotnet-bot Test CentOS7.1 x64 Debug Build  Please

src/Microsoft.Extensions.DependencyModel/DependencyContextJsonReader.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.Extensions.DependencyModel
 {
     public class DependencyContextJsonReader : IDependencyContextReader
     {
+        private readonly IDictionary<string, string> _stringPool = new Dictionary<string, string>();
```
eerhardt: Would HashSet work better?

pakrym: How would you use HashSet for this scenario?

eerhardt: Ya, good point.  I didn't fully think it through. You can check if the string is already cached, but there isn't a good way of getting the cached string out.

It just seemed odd that we have a Dictionary where the key == value. And that there isn't a more efficient data structure for this. You don't really need to track both the key and value. You just need to track the key...

davidfowl: Can you pass this into the reader and clear it when done?

pakrym: Don't want to expose internals? How would it help?

davidfowl: No this is just to aggressively clear the dictionary after call to Read.

pakrym: Done, reader is IDisposable.

pakrym:

pakrym:

eerhardt:
@pakrym - this has conflicts.  Are we still interested in taking this change?  If so, can you resolve the conflicts?  If not, please close.

pakrym:
Are you taking preview2 changes already?

eerhardt:
As of last night `rel/1.0.0` is open for business for preview2.  We have branched preview1 to `rel/1.0.0-preview1`.

pakrym:
I'll rebase this change then

src/Microsoft.Extensions.DependencyModel/DependencyContextLoader.cs
```diff
@@ -16,26 +16,26 @@ public class DependencyContextLoader
...
         public DependencyContextLoader() : this(
             DependencyContextPaths.Current.Application,
             DependencyContextPaths.Current.SharedRuntime,
             FileSystemWrapper.Default,
-            new DependencyContextJsonReader())
+            () => new DependencyContextJsonReader())
         {
         }
 
         internal DependencyContextLoader(
```
eerhardt: Is this constructor used anywhere else? Can it be made 'private' or even removed?

eerhardt:
:shipit:

src/Microsoft.Extensions.DependencyModel/DependencyContextLoader.cs
pakrym: Was going to be used in unit tests, but I think never did, I still want to keep it to keep separation between class dependencies and their defaults.

Issue #2267, opened 2016-04-04
`Failed to bind to coreclr` dotnet run failure on Windows 7 and Windows 2008 R2


pakrym:
## Steps to reproduce
`dotnet run` standalone app on Windows 7 or Windows 2008 R2

## Expected  behavior
Application runs

## Actual behavior
```
Running:
  program: dotnet  
  commandline: run -p ../../tools/Microsoft.AspNetCore.Server.Kestrel.GeneratedCode Http/FrameHeaders.Generated.cs Http/Frame.Generated.cs
  workingdir: D:\b\work\35116a8defcd3a21\Universe\KestrelHttpServer\src/Microsoft.AspNetCore.Server.Kestrel

Compiling Microsoft.AspNetCore.Server.Kestrel.GeneratedCode for .NETStandardApp,Version=v1.5

Compilation succeeded.
    0 Warning(s)
    0 Error(s)

Time elapsed 00:00:06.6573303
 

Failed to load the dll from C:\Users\asplab\.nuget\packages\runtime.win7-x64.Microsoft.NETCore.Runtime.CoreCLR\1.0.2-rc2-23931\runtimes\win7-x64\native\coreclr.dll, HRESULT: 0x8007007E
Failed to bind to coreclr
```

## Environment data
`dotnet --info` output:
`
1.0.0-beta-002202 
`
@piotrpMSFT @Petermarcu @davidfowl @muratg @anurse  @schellap

schellap:
On min OSes coreclr needs `api-*` dlls to be sxs `coreclr.dll` or windows `PATH` for `LoadLibraryExW` is setup correctly to point to the directories containing these api forwarder DLLs. Since `dotnet build` runs against the nuget cache, these api-forwarders are part of another directory. Though a bit ugly, it seems the fix would be to expand the DLL search paths to include the nuget cache native directories to dll search path.

PR #2271, opened 2016-04-05
Add ApiSets to process DLL search dirs

schellap:
@gkhanna79 PTAL
Fixes https://github.com/dotnet/cli/issues/2267

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2271)

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -22,7 +22,7 @@ namespace
 void add_tpa_asset(
     const pal::string_t& asset_name,
     const pal::string_t& asset_path,
-    std::set<pal::string_t>* items,
+    std::unordered_set<pal::string_t>* items,
     pal::string_t* output)
 {
     if (items->count(asset_name))
@@ -48,7 +48,7 @@ void add_tpa_asset(
 void add_unique_path(
     deps_entry_t::asset_types asset_type,
     const pal::string_t& path,
-    std::set<pal::string_t>* existing,
+    std::unordered_set<pal::string_t>* existing,
     pal::string_t* output)
 {
     // Resolve sym links.
@@ -377,7 +377,7 @@ void deps_resolver_t::resolve_tpa_list(
         get_dir_assemblies(m_fx_dir, _X("fx"), &m_fx_assemblies);
     }
 
-    std::set<pal::string_t> items;
+    std::unordered_set<pal::string_t> items;
 
     auto process_entry = [&](const pal::string_t& deps_dir, deps_json_t* deps, const dir_assemblies_t& dir_assemblies, const deps_entry_t& entry)
     {
@@ -474,7 +474,7 @@ void deps_resolver_t::resolve_probe_dirs(
         return get_directory(str);
     };
     std::function<pal::string_t(const pal::string_t&)>& action = is_resources ? resources : native;
-    std::set<pal::string_t> items;
+    std::unordered_set<pal::string_t> items;
 
     std::vector<deps_entry_t> empty(0);
     const auto& entries = m_deps->get_entries(asset_type);
@@ -482,15 +482,34 @@ void deps_resolver_t::resolve_probe_dirs(
 
     pal::string_t candidate;
 
+    bool track_api_sets = true;
     auto add_package_cache_entry = [&](const deps_entry_t& entry)
     {
         if (probe_entry_in_configs(entry, &candidate))
         {
-            add_unique_path(asset_type, action(candidate), &items, output);
+            // For standalone apps, on win7, coreclr needs ApiSets which has to be in the DLL search path.
+            const pal::string_t result_dir = action(candidate);
+
+            if (track_api_sets && pal::need_api_sets() &&
+                ends_with(entry.library_name, _X("Microsoft.NETCore.Windows.ApiSets"), false))
+            {
+                // For standalone and portable apps, get the ApiSets DLL directory,
+                // as they could come from servicing or other probe paths.
+                // Note: in portable apps, the API set would come from FX deps
+                // which is actually a standalone deps (rid specific API set).
+                // If the portable app relied on its version of API sets, then
+                // the rid selection fallback would have already been performed
+                // by the host (deps_format.cpp)
+                m_api_set_paths.insert(result_dir);
+            }
+
+            add_unique_path(asset_type, result_dir, &items, output);
         }
     };
     std::for_each(entries.begin(), entries.end(), add_package_cache_entry);
+    track_api_sets = m_api_set_paths.empty();
     std::for_each(fx_entries.begin(), fx_entries.end(), add_package_cache_entry);
+    track_api_sets = m_api_set_paths.empty();
 
     // For portable rid specific assets, the app relative directory must be used.
     if (m_portable)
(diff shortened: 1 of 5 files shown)


src/corehost/cli/deps_resolver.cpp
```diff
@@ -486,6 +486,16 @@ void deps_resolver_t::resolve_probe_dirs(
...
         if (probe_entry_in_configs(entry, &candidate))
         {
+            // For standalone apps, on win7, coreclr needs ApiSets which has to be in the DLL search path.
+            if (!m_portable && pal::need_api_sets()
+                    && ends_with(entry.library_name, _X("Microsoft.NETCore.Windows.ApiSets"), false))
+            {
+                const pal::string_t api_sets_dir = action(candidate);
+                if (api_sets_dir != clr_dir)
+                {
+                    m_api_set_paths.insert(api_sets_dir);
```
gkhanna79: Since all the apisets are in the same folder, we should only need to add the path once and skip adding it again.

Other than that, lgtm.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -486,6 +486,16 @@ void deps_resolver_t::resolve_probe_dirs(
     {
         if (probe_entry_in_configs(entry, &candidate))
         {
+            // For standalone apps, on win7, coreclr needs ApiSets which has to be in the DLL search path.
+            if (!m_portable && pal::need_api_sets()
```
pakrym: Are you sure we should keep this kind of logic in `corehost` itself and not in `hostpolicy`?

src/corehost/cli/deps_resolver.cpp
schellap: Didn't want to make package related assumptions, but the set would filter it out anyway. Make it more targeted?

src/corehost/cli/deps_resolver.cpp
schellap: Policy does loadlibrary on coreclr, and brings in all these dependencies and it also processes the deps file.

pakrym: I was talking about `Microsoft.NETCore.Windows.ApiSets` search logic, that is in `corehost`. What if newer `clr` would need `dll's` from another package?

schellap: Ideally, all native dirs will be added, this will make that worry go away. We decided to make a point fix because we don't expect this to change. Besides, hostpolicy is the most serviceable component and corehost the least.

src/corehost/cli/deps_resolver.cpp
gkhanna79: So long as we will have only a single path in the set, I am fine with the change.

src/corehost/cli/deps_resolver.cpp
gkhanna79: > What if newer clr would need dll's from another package?

APISets are the only CoreCLR dependency that are not present in the system (i.e. pre-req since they cannot be).

schellap:
@gkhanna79 PTAL. Fixes #7518

src/corehost/cli/deps_resolver.cpp
```diff
@@ -486,7 +486,15 @@ void deps_resolver_t::resolve_probe_dirs(
     {
         if (probe_entry_in_configs(entry, &candidate))
         {
-            add_unique_path(asset_type, action(candidate), &items, output);
+            // For standalone apps, on win7, coreclr needs ApiSets which has to be in the DLL search path.
+            const pal::string_t result_path = action(candidate);
+
+            if (!is_resources)
```
gkhanna79: Shouldn't we check if this is "native" as opposed to "!resources"?

schellap: Refresh this page pls.

schellap: See assertion above, it means one or the other.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -96,6 +96,9 @@ int run(const corehost_init_t* init, const runtime_config_t& config, const argum
     size_t property_size = property_keys.size();
     assert(property_keys.size() == property_values.size());
 
+    // Add API sets to the process DLL search
```
gkhanna79: "Add native paths"

src/corehost/common/pal.unix.cpp
```diff
@@ -84,6 +84,11 @@ bool pal::load_library(const char_t* path, dll_t* dll)
     return true;
 }
 
+bool pal::process_native_paths(const std::unordered_set<pal::string_t>& native_lib_files)
```
gkhanna79: This is a "void" and not a "bool" return type.

schellap:
@gkhanna79 PTAL. For Unix, we need to come up with a different solution. I don't know what that is, but we need to think about it and I need to chat with few experts within the CLR.

gkhanna79:
Modulo comments, LGTM.

gkhanna79:

gkhanna79:

gkhanna79:

schellap:
@gkhanna79 PTAL post your discussions with @davidfowl.

schellap:

src/corehost/common/pal.windows.cpp
```diff
@@ -54,6 +54,24 @@ bool pal::find_coreclr(pal::string_t* recv)
...
+    // some other NATIVE_DLL_SEARCH_DIRECTORIES and not necessarily side by side. However,
+    // CoreCLR.dll loads PInvoke DLLs with LOAD_WITH_ALTERED_SEARCH_PATH. Note that this
+    // option cannot be combined with LOAD_LIBRARY_SEARCH_USER_DIRS, so the AddDllDirectory
+    // doesn't help much in telling CoreCLR where to load the PInvoke DLLs from.
+    // So we resort to modifying the PATH variable on our own hoping Windows loader will do the right thing.
+
+    trace::verbose(_X("Setting PATH=%s"), path.c_str());
+
+    // Assuming no other thread modified PATH between Get and SetEnvironmentVariableW.
+    ::SetEnvironmentVariableW(_X("PATH"), path.c_str());
```
gkhanna79: You missed updating the path with apisets.

src/corehost/common/pal.windows.cpp
```diff
@@ -54,6 +54,24 @@ bool pal::find_coreclr(pal::string_t* recv)
...
+    // We need this ugly hack, as the PInvoked DLL's static dependencies can come from
+    // some other NATIVE_DLL_SEARCH_DIRECTORIES and not necessarily side by side. However,
+    // CoreCLR.dll loads PInvoke DLLs with LOAD_WITH_ALTERED_SEARCH_PATH. Note that this
+    // option cannot be combined with LOAD_LIBRARY_SEARCH_USER_DIRS, so the AddDllDirectory
+    // doesn't help much in telling CoreCLR where to load the PInvoke DLLs from.
+    // So we resort to modifying the PATH variable on our own hoping Windows loader will do the right thing.
+
+    trace::verbose(_X("Setting PATH=%s"), path.c_str());
+
+    // Assuming no other thread modified PATH between Get and SetEnvironmentVariableW.
```
gkhanna79: At this point in time, there should not be any other thread running since the host is prepping to start the runtime.

PR #2268, opened 2016-04-05
Enhance the update-dependencies script

eerhardt:
- Add the new version number to the commit and PR message.
- Check if anything has been updated and no-op if there wasn't anything updated.
- Add support to CC indviduals on the PR.

@piotrpMSFT @anurse

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2268)

PR #2274, opened 2016-04-05
Add Microsoft.NETCore.App meta-package build

weshaggard:
cc @ericstj @schellap 

This is the easiest way to get the Microsoft.NETCore.App meta-packaging building in CLI. At some point we should factor out the BuildTools part outside of corehost but that can be done at a later time.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2274)


scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -143,10 +143,10 @@ public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
         [Target(nameof(CompileCoreHost))]
         public static BuildTargetResult PackageCoreHost(BuildTargetContext c)
         {
-            if (!string.Equals(Environment.GetEnvironmentVariable("BUILD_COREHOST_PACKAGES"), "1"))
-            {
-                return c.Success();
-            }
+//            if (!string.Equals(Environment.GetEnvironmentVariable("BUILD_COREHOST_PACKAGES"), "1"))
```
eerhardt: ??  Do we want to keep this commented out?

src/corehost/packaging/dir.props
```diff
@@ -23,7 +23,7 @@
 
   <PropertyGroup>
     <PackageOutputPath>$(PackagesOutDir)</PackageOutputPath>
-    <RuntimeIdGraphDefinitionFile>$(ProjectDir)packages\Microsoft.NETCore.Platforms\1.0.1-rc2-23712\runtime.json</RuntimeIdGraphDefinitionFile>
+    <RuntimeIdGraphDefinitionFile>$(ProjectDir)packages\Microsoft.NETCore.Platforms\1.0.1-rc2-23931\runtime.json</RuntimeIdGraphDefinitionFile>
```
eerhardt: I don't have this line being updated in my UpdateDependencies script.  I will need to fix that...

src/corehost/packaging/pack.cmd
```diff
@@ -72,6 +72,9 @@ if not ERRORLEVEL 0 goto :Error
 "%__ProjectDir%\Tools\corerun" "%__ProjectDir%\Tools\MSBuild.exe" "%__ProjectDir%\projects\Microsoft.NETCore.DotNetHost.builds" /p:Platform=%__BuildArch% /p:DotNetHostBinDir=%__DotNetHostBinDir% /p:TargetsWindows=true /p:HostVersion=%__HostVer% /p:HostResolverVersion=%__FxrVer% /p:HostPolicyVersion=%__PolicyVer% /p:BuildNumberMajor=%__BuildMajor% /p:PreReleaseLabel=%__VersionTag% /verbosity:minimal
 if not ERRORLEVEL 0 goto :Error
 
+"%__ProjectDir%\Tools\corerun" "%__ProjectDir%\Tools\MSBuild.exe" "%__ProjectDir%\Microsoft.NETCore.App\Microsoft.NETCore.App.pkgproj" /p:Platform=x86 /p:CLIBuildVersion=%__BuildMajor% /verbosity:minimal
```
eerhardt: > /p:Platform=x86

does this need to be `%__BuildArch%`?  If not, i think a comment should be made describing why not since it isn't obvious to me.

scripts/dotnet-cli-build/CompileTargets.cs
weshaggard: I mainly left it commented out to make sure CI builds the packages but if all goes well I will remove these lines. I would like to get @schellap's opinion though since he was also working on this area.

src/corehost/packaging/dir.props
weshaggard: Does your UpdateDependencies script have a way to update non-project.json files? I actually did a fair amount of hacking in the meta-package building to try and make sure my dependencies were listed in a project.json.

src/corehost/packaging/pack.cmd
weshaggard: For meta-packages it really doesn't matter which architecture is used it should actually be architecture agnostic but our current package builds require a platform to be set. I will add a comment.

src/corehost/packaging/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -0,0 +1,47 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
schellap: I've put the pkgproj inside packaging/projects.

src/corehost/packaging/dir.props
eerhardt: > Does your UpdateDependencies script have a way to update non-project.json files?

Yes it does. It is just a C# .NET Core app, so it can do anything to the repo it needs to.

Here is the update:  https://github.com/dotnet/cli/pull/2268/commits/e0542adf51fed1554cb660fceaffcadfabc17438

src/corehost/packaging/pack.cmd
eerhardt: So why not just use `/p:Platform=%__BuildArch%` if it doesn't matter?

scripts/dotnet-cli-build/CompileTargets.cs
schellap: In x86, we hit MAX_PATH in VSO. Keep this uber packaging variable to guard against this. I am experimenting few layouts that would make VSO happy.

src/corehost/packaging/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
weshaggard: I didn't do that because you only have all your projects in the root of that directory. I will have a look and restructure this.

src/corehost/packaging/pack.cmd
weshaggard: I'm actually going to change the pkgproj to be AnyCPU and stop passing the parameter all together.

scripts/dotnet-cli-build/CompileTargets.cs
weshaggard: I'm going to remove the commented out code.

schellap: Don't checkin with that commented. People will get upset if VSO breaks.

src/corehost/packaging/dir.props
```diff
@@ -36,7 +36,7 @@
     <HostFullVersion>$(HostVersion)-$(PreReleaseLabel)-$(BuildNumberMajor)</HostFullVersion>
   </PropertyGroup>
 
-  <Target Name="ValidateArgs">
+  <Target Name="ValidateArgs" Condition="'$(CLIBuildVersion)'==''">
```
brthor: Why only check these values when the build version is empty?

src/corehost/packaging/Microsoft.NETCore.App/project.json
```diff
@@ -0,0 +1,46 @@
+{
+  "dependencies": {
+    "Microsoft.CodeAnalysis.CSharp": "1.3.0-beta1-20160321-04",
+    "Microsoft.CodeAnalysis.VisualBasic": "1.3.0-beta1-20160321-04",
+    "Microsoft.CSharp": "4.0.1-rc2-23931",
+    "Microsoft.NETCore.DotNetHostPolicy": "1.0.1-beta-002200",
```
brthor: As we rev the version of this package from the CLI, how can we update the dependency of the metapackage as well?

eerhardt: The long-term vision is that this will move out of the CLI. But for RC2 @gkhanna79 and @schellap have a short-term plan on how to make it work. Can you guys elaborate the plan?

brthor: I understand this, but with a hardcoded version here I don't understand how we can continuously publish these packages like we do with the others in our build. Or are we creating a new model here?

src/corehost/packaging/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
```diff
@@ -0,0 +1,47 @@
...
+      <Dependency Condition="'%(Dependency.Identity)' == 'Microsoft.CodeAnalysis.VisualBasic'">
+        <Exclude>Compile</Exclude>
+      </Dependency>
+      <Dependency Condition="'%(Dependency.Identity)' == 'System.Runtime.Loader'">
+        <Exclude>Compile</Exclude>
+      </Dependency>
+    </ItemGroup>
+
+    <!-- Use NETStandard.Library version suffix for version suffix of Microsoft.NETCore.App -->
+    <PropertyGroup>
```
brthor: If the host is revving at a faster pace than the pace at which we are taking new NETStandard.Library versions, does it make sense that the metapackage should have the same version as NETStandard.Library?

brthor: After reading the code, I can see the cli version is used as a part of the version, so the above is moot. Consider rewording this comment though?

src/corehost/packaging/dir.props
weshaggard: It is really there to have two modes and not force me to pass all these other host properties in when I'm building just the meta-package.

src/corehost/packaging/Microsoft.NETCore.App/project.json
weshaggard: If you see the message in my commit the my current versioning model for the meta-package is "1.0.0" + "NetStandardLibrary.BuildNumber" + "CLI build Number" See https://github.com/dotnet/cli/pull/2274/files#diff-a4a50058597e5e9e728f0ac284a74259R47

src/corehost/packaging/Microsoft.NETCore.App/Microsoft.NETCore.App.pkgproj
weshaggard: Oops I just responded to your other question pointing at this place. I can update the comment.

src/corehost/packaging/Microsoft.NETCore.App/project.json
brthor: I can see that from above, but how do we change the version of the host dependency to the version of the host package that we're building in the current build?

src/corehost/packaging/projects/Microsoft.NETCore.App/project.json
```diff
@@ -0,0 +1,45 @@
...
+    "System.Security.Cryptography.X509Certificates": "4.1.0-rc2-23931",
+    "System.Threading.Tasks.Dataflow": "4.6.0-rc2-23931",
+    "System.Threading.Tasks.Extensions": "4.0.0-rc2-23931",
+    "System.Threading.Tasks.Parallel": "4.0.1-rc2-23931",
+    "System.Threading.Thread": "4.0.0-rc2-23931",
+    "System.Threading.ThreadPool": "4.0.10-rc2-23931"
+  },
+  "frameworks": {
+    "netstandard1.5": { }
+  }
```
schellap: Not including the corehost package?

weshaggard: See https://github.com/dotnet/cli/pull/2274/files#diff-a4a50058597e5e9e728f0ac284a74259R11 which is a ProjectReference to the host package project, so it is always the live built version.

schellap: Oops yeah, saw that this morning.

weshaggard:
@schellap do you still want me to hold off on merging the change to re-enable the building of these packages? If we have issues I would like to work through them and fix them as opposed to holding off any further on this. If you don't have any more feedback I will squash and commit this change.

weshaggard:
I cleaned up and squashed my commits. @schellap could you please have one final look at sign-off on this and then I will merge it.

PR #2313, opened 2016-04-06
fix install.sh dry run to use exit instead of return

krwq:
`--dry-run` is currently printing errors with some versions of bash

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2313)


krwq:
This was already included in other PR

PR #2320, opened 2016-04-06
Update Roslyn to 1.3.0-beta1-20160405-05

eerhardt:
@weshaggard @piotrpMSFT 

Tomas says there may be a few more changes.  But I wanted to get this version in now to make sure tests pass with the newest Roslyn.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2320)

### DIFF
--- a/src/compilers/project.json
+++ b/src/compilers/project.json
@@ -8,8 +8,8 @@
       "type": "platform",
       "version": "1.0.0-rc2-23931"
     },
-    "Microsoft.CodeAnalysis.CSharp": "1.3.0-beta1-20160329-01",
-    "Microsoft.Net.Compilers.netcore": "1.3.0-beta1-20160329-01",
+    "Microsoft.CodeAnalysis.CSharp": "1.3.0-beta1-20160405-05",
+    "Microsoft.Net.Compilers.netcore": "1.3.0-beta1-20160405-05",
     "Microsoft.DiaSymReader.Native": "1.3.3"
   },
   "frameworks": {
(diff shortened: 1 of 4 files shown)


src/compilers/project.json
```diff
@@ -8,8 +8,8 @@
       "type": "platform",
       "version": "1.0.0-rc2-23931"
     },
-    "Microsoft.CodeAnalysis.CSharp": "1.3.0-beta1-20160329-01",
-    "Microsoft.Net.Compilers.netcore": "1.3.0-beta1-20160329-01",
+    "Microsoft.CodeAnalysis.CSharp": "1.3.0-beta1-20160405-05",
```
weshaggard: Can these be removed now given they are in Microsoft.NETCore.App?

weshaggard:
LGTM

src/compilers/project.json
eerhardt: I'm not sure what this project is for.  I am just doing replacements, not removing.  We can remove them separately if possible.

@brthor @pakrym - do you guys know?

brthor: Last I checked this project gets published into the sdk publish directory.

I'm not sure what it's original intention is but probably a hack to get unblocked in the hackathon.

Checkout out `CompileCliSdk` in CompileTargets.cs

pakrym: This project is here to generate portable deps files for `csc.exe` and `vbc.exe`. So in theory removing roslyn from here should work

Issue #2314, opened 2016-04-06
Is the reference to Microsoft.ApplicationInsights intentionally placed in Microsoft.Dotnet.Cli.Utils?


pranavkm:
`Microsoft.Dotnet.Cli.Utils` is referenced by most of the tools we https://github.com/aspnet. These now end up picking up a reference to ApplicationInsights despite not having any telemetry. Could we move the telemetry code into the dotnet project instead?

pranavkm:
:cake:

PR #2328, opened 2016-04-07
Move telemetry dependency out of Utils and into dotnet

danquirk:
Nothing fancy, just moving all the telemetry code into dotnet so that other packages which use Utils and do not need AppInsights do not have a dependency on AppInsights.

Fixes https://github.com/dotnet/cli/issues/2314

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2328)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/project.json
+++ b/src/Microsoft.DotNet.Cli.Utils/project.json
@@ -5,7 +5,6 @@
     "warningsAsErrors": true
   },
   "dependencies": {
-    "Microsoft.ApplicationInsights": "2.0.0-rc1",
     "Microsoft.DotNet.ProjectModel": "1.0.0-*",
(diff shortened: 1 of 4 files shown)


brthor:
:shipit:

eerhardt:
:shipit: 

I really like that we are now relying on the RTM version of AppInsights. :+1:

davidfowl:
> I really like that we are now relying on the RTM version of AppInsights. :+1:

Does it still require imports? Or does it actually support netstandard?

piotrpMSFT:
@danquirk did you send mail to Shiproom?

danquirk:
@piotrpMSFT sorry, I have not. I assumed the rest of the process was in your hands as far as what you guys wanted approved and for when. I was just making sure the PRs were green.

PR #2331, opened 2016-04-07
Consume Microsoft.NETCore.App

eerhardt:
Consuming Microsoft.NETCore.App in the CLI repo that is built inside the CLI repo.

This involves a few changes.

1. Need to generate the `dotnet new` templates in a pre-build step to get the correct Microsoft.NETCore.App version.
2. Need to generate the "sharedFramework" project to get the correct Microsoft.NETCore.App version and to only list the current RID.
3. All other references to Microsoft.NETCore.App are now `rc2-*`, and update all the `dotnet restore` calls to include `--fallbacksource Dirs.CoreHost` so the newly built packages are picked up. Also, cannot `--infer-runtimes` because on Win-x64, the x86 runtime is inferred but the x86 packages don't exist.
4. Change all the tests to be real portable apps, and fix them to work as real portable apps.
5. `dotnet test` and corehost need to specify the test assemblies runtimeconfig.json and not `dotnet-test-xunit`s runtimeconfig.json because the test assembly is a real portable app. Since the test assembly is portable, dotnet-test-xunit's runtimeconfig.json doesn't get restored.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2331)


src/corehost/cli/hostpolicy.cpp
```diff
@@ -246,9 +246,11 @@ SHARED_API int corehost_main(const int argc, const pal::char_t* argv[])
     }
     else
     {
-        pal::string_t dev_config_file;
-        auto config_path = get_runtime_config_from_file(args.managed_application, &dev_config_file);
-        runtime_config_t config(config_path, dev_config_file);
+        pal::string_t config_file, dev_config_file;
```
brthor: @schellap take a look at the changes here when you get a chance.

eerhardt:
test Ubuntu x64 Release Build

eerhardt:
OK - this change is ready for review.  The tests should be passing now.

@weshaggard @piotrpMSFT @brthor @Sridhar-MS @schellap 

/cc @mellinoe @ellismg - please review the "Shared Framework" creation change

eerhardt:

src/sharedframework/framework/project.json.template
```diff
@@ -0,0 +1,19 @@
...
+  "dependencies": {
+    "Microsoft.NETCore.App": "$(NetCoreAppVersion)"
+  },
+  "runtimes": {
+    "$(RID)": {}
+  },
+  "frameworks": {
+    "dnxcore50": {
+      "imports": [
+        "portable-net45+win8"
```
weshaggard: orthogonal but this import should no longer be needed. @ellismg @mellinoe

weshaggard:
LGTM thanks for getting this together @eerhardt. 

please wait for @schellap to review the host changes and @ellismg to review the shared framework changes.

brthor:
Took a look through and didn't see anything to publish these packages (but I may have missed it).

Is there work remaining there?

weshaggard:
We cannot publish the packages from a single leg of the builds. The actually publish them we need all the OS legs to be finished. For now I've created an custom VSO build that we can kick manually after a full set of builds have finished to publish the nuget packages for the corehost and NETCore.App.

brthor:
@weshaggard Where does it pick up the runtime nuget packages from the single leg of the builds?

weshaggard:
For the restore locally look for the various places @eerhardt has `"--fallbacksource", Dirs.Corehost` in the restore commands. The packages in an individual leg are currently dropped into the Dirs.Corehost folder, they are also uploaded to the Azure blob when the individual leg is complete.

test/ArgumentForwardingTests/ArgumentForwardingTests.cs
```diff
@@ -262,8 +263,8 @@ private string[] RawEvaluateArgumentString(string testUserArgument)
             {
                 StartInfo = new ProcessStartInfo
                 {
-                    FileName = ReflectorPath,
-                    Arguments = testUserArgument,
+                    FileName = Env.GetCommandPath("dotnet", ".exe", ""),
+                    Arguments = $"{ReflectorPath} {testUserArgument}",
```
brthor: This will break if we have a space in the path to the build.

brthor:
I see how we are managing that restore, but I don't see where we are uploading these pacakges to azure in the build. 

I would expect to see an addition in `PublishTargets` With a sibling to this:
```
[Target(
            nameof(PublishTargets.PublishCombinedHostFrameworkArchiveToAzure),
            nameof(PublishTargets.PublishCombinedHostFrameworkSdkArchiveToAzure),
            nameof(PublishTargets.PublishSDKSymbolsArchiveToAzure))]
        public static BuildTargetResult PublishArchivesToAzure(BuildTargetContext c) => c.Success();
```

What am I missing? Is the upload handled in msbuild?

test/ArgumentForwardingTests/ArgumentForwardingTests.cs
eerhardt: Arguments = $"\"{ReflectorPath}\" {testUserArgument}", to fix it?

brthor: It's unfortunately more complicated than that to fix this in all cases (which is exactly what this is testing :smile: )

But that should at least not break in the simple case.

There are other places where we have this issue as well in our build, so I wouldn't block on it.

weshaggard:
@brthor that part is currently handled here https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/PublishTargets.cs#L98

eerhardt:

ellismg:
I looked over the shared framework stuff and it looked good to me.

brthor:
Thanks Wes, I did just miss it.

:shipit: :shipit:

src/corehost/cli/hostpolicy.cpp
schellap: LGTM.

PR #2337, opened 2016-04-07
Bring has runtime output check back

pakrym:
`MakeRunnable` tries to create runtime project context even for class library projects which fails for net451 class libraries restored without rids and is totally unnecessary

@anurse @davidfowl

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2337)


anurse:
:shipit: (fix the test though :))

PR #2340, opened 2016-04-07
rename install.ps1/sh to dotnet-install.ps1/sh to match documentation

krwq:
The added files are a pure copy & paste of original ones. The old files are left so that existing consumers are not broken since the final location is still may change we are not removing them or adding any warning.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2340)

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -1,0 +1,338 @@
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+<#
+.SYNOPSIS
+    Installs dotnet cli
+.DESCRIPTION
+    Installs dotnet cli. If dotnet installation already exists in the given directory
+    it will update it only if the requested version differs from the one already installed.
+.PARAMETER Channel
+    Default: preview
+    Channel is the way of reasoning about stability and quality of dotnet. This parameter takes one of the values:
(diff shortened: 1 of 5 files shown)


krwq:
@dotnet-bot test CentOS7.1 x64 Debug

krwq:
seems like some jenkins issue

krwq:
@Sridhar-MS, could you take a look at the kestrel failures?

Sridhar-MS:
taking a look

Sridhar-MS:
@krwq I have a fix in https://github.com/dotnet/cli/pull/2378

scripts/obtain/dotnet-install.ps1
```diff
@@ -0,0 +1,338 @@
+#
```
eerhardt: Why are these two files diffing as completely new?  Did you remove the old install.ps1/sh?

krwq: no, too many people are using install.ps1 and install.sh so we are leaving two copies of it for now. once we get permanent links for them the content of install.ps1/sh will be replaced with a warning and NZEC so that people can act accordingly and update their script.

blackdwarf:

blackdwarf:

scripts/obtain/dotnet-install.ps1
natemcmaster: In the meantime, should we update our scripts to use dotnet-install.sh instead?

PR #2345, opened 2016-04-07
update nuget to 1160

anurse:
/cc @davidfowl @pakrym @eerhardt @piotrpMSFT

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2345)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -230,7 +230,6 @@ namespace Microsoft.DotNet.Cli.Build
         public static BuildTargetResult CompileStage1(BuildTargetContext c)
         {
             CleanBinObj(c, Path.Combine(c.BuildContext.BuildDirectory, "src"));
-            CleanBinObj(c, Path.Combine(c.BuildContext.BuildDirectory, "test"));
 
             if (Directory.Exists(Dirs.Stage1))
             {
@@ -258,7 +257,6 @@ namespace Microsoft.DotNet.Cli.Build
             var configuration = c.BuildContext.Get<string>("Configuration");
 
             CleanBinObj(c, Path.Combine(c.BuildContext.BuildDirectory, "src"));
-            CleanBinObj(c, Path.Combine(c.BuildContext.BuildDirectory, "test"));
 
             if (Directory.Exists(Dirs.Stage2))
             {
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -48,7 +48,7 @@ namespace Microsoft.DotNet.Cli.Build
             return c.Success();
         }
 
-        // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the 
+        // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the
         // Microsoft.NETCore.App package that is created during PackagePkgProjects.
         [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
         public static BuildTargetResult Compile(BuildTargetContext c)
@@ -565,7 +565,7 @@ namespace Microsoft.DotNet.Cli.Build
                 // in CompileTargets and the one in the shared library project.json match and are updated in lock step, but long term
                 // we need to be able to look at the project.lock.json file and figure out what version of Microsoft.NETCore.Runtime.CoreCLR
                 // was used, and then select that version.
-                ExecSilent(Crossgen.GetCrossgenPathForVersion(CompileTargets.CoreCLRVersion),
+                ExecSilent(Crossgen.GetCrossgenPathForVersion(CoreCLRVersion),
                     "-readytorun", "-in", file, "-out", tempPathName, "-platform_assemblies_paths", pathToAssemblies);
 
                 File.Delete(file);
--- a/scripts/dotnet-cli-build/PrepareTargets.cs
+++ b/scripts/dotnet-cli-build/PrepareTargets.cs
@@ -228,7 +228,7 @@ namespace Microsoft.DotNet.Cli.Build
         {
             var dotnet = DotNetCli.Stage0;
 
-            dotnet.Restore("--verbosity", "verbose", "--disable-parallel", "--infer-runtimes", "--fallbacksource", Dirs.Corehost)
+            dotnet.Restore("--verbosity", "verbose", "--disable-parallel", "--fallbacksource", Dirs.Corehost)
                 .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "src"))
                 .Execute()
                 .EnsureSuccessful();
--- a/src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
@@ -19,7 +19,7 @@ namespace Microsoft.DotNet.Cli.Utils
 {
     public class ProjectToolsCommandResolver : ICommandResolver
     {
-        private static readonly NuGetFramework s_toolPackageFramework = FrameworkConstants.CommonFrameworks.NetStandardApp15;
+        private static readonly NuGetFramework s_toolPackageFramework = FrameworkConstants.CommonFrameworks.NetCoreApp10;
         
         private static readonly CommandResolutionStrategy s_commandResolutionStrategy = 
             CommandResolutionStrategy.ProjectToolsPackage;
--- a/src/dotnet/commands/dotnet-build/CompileContext.cs
+++ b/src/dotnet/commands/dotnet-build/CompileContext.cs
@@ -12,6 +12,7 @@ using Microsoft.DotNet.ProjectModel;
 using Microsoft.DotNet.ProjectModel.Utilities;
 using Microsoft.DotNet.Tools.Compiler;
 using Microsoft.Extensions.PlatformAbstractions;
+using Microsoft.DotNet.ProjectModel.Compilation;
 
 namespace Microsoft.DotNet.Tools.Build
 {
@@ -79,7 +80,7 @@ namespace Microsoft.DotNet.Tools.Build
                 return true;
             }
 
-            foreach (var dependency in Sort(_rootProjectDependencies.ProjectDependenciesWithSources))
+            foreach (var dependency in Sort(_rootProjectDependencies))
             {
                 var dependencyProjectContext = ProjectContext.Create(dependency.Path, dependency.Framework, new[] { _rootProject.RuntimeIdentifier });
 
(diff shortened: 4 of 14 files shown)


eerhardt:
This is not the version on the Build Dependency Matrix.  We need to make sure this version NuGet wants shipped.  NuGet still says current build is 1130.

eerhardt:
@emgarten - Can NuGet update the Build Dependency Matrix with this build? Or is this not a build that is supposed to go into .NET Core RC2?

anurse:
This build contains fixes required to ship ASP.NET RC2, so whatever needs to happen to make this mergable should happen :).

emgarten:
There are a few extra fixes in this that haven't been fully tested. They should be helpful, but we're going to hold off on updating NuGet in the CLI until tomorrow.

anurse:
We can wait until then. This is blocking ASP.NET release, but not active ASP.NET work

anurse:
Just a reminder: We are waiting on a new NuGet build to come today, so don't merge this yet :)

eerhardt:
@emgarten @yishaigalatzer  - any update on this?  Is there a newer NuGet build that CLI should be using?

emgarten:
@anurse would you update this to use 3.5.0-beta-1160, it contains just the critical changes needed for the CLI
* NetStandardApp -> NetCoreApp for all nuget packages
* Allow projects to fill nupkg dependencies
* NetCoreApp as the default TFM for tools
* locale property in the lock file

eerhardt:
@anurse - did you just kick off like 5 Jenkins runs?

anurse:
@eerhardt I hope not, it should have a quiet period. I've noticed that before.

anurse:
So much for that... I'll clean it up...

eerhardt:
@emgarten - can you update the Build Dependency Matrix OneNote with the correct build?  That way I don't get questioned why CLI is using a version of NuGet that isn't on there...

emgarten:
NuGet XPlat has been updated to netcoreapp1.0, it looks like the build is failing because the project using that package need to add an imports or change framework.

```
11:06:58 Errors in D:\j\workspace\release_windo---4c3b899b\src\dotnet\project.json
11:06:58     Package NuGet.CommandLine.XPlat 3.5.0-beta-1160 is not compatible with netstandard1.5 (.NETStandard,Version=v1.5). Package NuGet.CommandLine.XPlat 3.5.0-beta-1160 supports:
11:06:58       - net46 (.NETFramework,Version=v4.6)
11:06:58       - netcoreapp1.0 (.NETCoreApp,Version=v1.0)
11:06:58     One or more packages are incompatible with .NETStandard,Version=v1.5.
```

emgarten:
It looks like ProjectToolsCommandResolver may be looking for the netstandard path instead of the netcoreapp path.

piotrpMSFT:
@brthor can you look at the resolver?

piotrpMSFT:
I just fixed ProjectToolsCommandResolver.

This exposed another bug with some interesting characteristics @cdmihai 

We are failing to find an output assembly for a p2p that is marked as "type"="project. The same project.json lists several other p2p's which don't have this label. The other p2p's are picked up by `dotnet build` and their ouputs are produced. This one is not picked up, output is not produced, and subsequent compile fails due to a missing dependency.

cdmihai:
Interesting. I'll look into it.

In the meantime can you please tell me what `type:project` means, and how does it affect the behaviour of things?

AFAIK, `type: build` means that the subgraph from that project should not be published, and `type:platform` means that the subgraph should not be published AND the root should be put into one of the deps files.

anurse:
`type:project` is not a thing :). It's actually `target: project` and it means that the project will always be selected, instead of a package. That's actually the default precedence anyway, but that constraint means that a package can **never** match that dependency. There is a similar `target: package` that can be used to restrict to packages even when a project would be able to match it.

anurse:
I'm also going to be taking a look, trying to get a local repro now

brthor:
:+1: for the constant change. 

Will this break existing tools that don't target netcoreapp though? 
May need to change fsc as well if that is the case.

cdmihai:
@piotrpMSFT 

How do I repro this? I applied this change: https://github.com/dotnet/cli/pull/2345/files#diff-d9b42908756d84c88d578d9980a4be37R22
And then I modified the project to project dependencies test asset to have `target: project` for p2p dependencies. However, it still builds successfully.

anurse:
It repros for me without any changes by just cloning this branch and running `.\build`

cdmihai:
OK, I was trying to get a minimal repro so I don't have to work with build.cmd :)

anurse:
Yeah, it looks like it's happening when trying to build the EndToEnd tests

cdmihai:
How very odd. I took the branch, built, and then modified this test asset https://github.com/dotnet/cli/tree/rel/1.0.0/TestAssets/TestProjects/TestProjectToProjectDependencies/src
to make some project depencies have target: project.

Stage1 still built it successfully.

Gonna debug to see what actually happens with EndToEnd.

anurse:
It's definitely unrelated to `target:project`. I'm having some trouble with debug symbols on my box but I'm getting ramped up to debug.

piotrpMSFT:
@anurse any updates?

anurse:
Not yet, still debugging. It's definitely just not building "Microsoft.Extensions.Testing.Abstractions" even though it is in the dependencies.

anurse:
I think I've identified the issue. The problem is that `dotnet-test-xunit` has a dependency on `Microsoft.Extensions.Testing.Abstractions` which is a project. The code in `dotnet-build` which sorts and selects projects to build only walks down projects so it stops at `dotnet-test-xunit` and schedules `Microsoft.Extensions.Testing.Abstractions` for build.

emgarten:
@anurse could this related to the change in nuget to allow projects to be used for nupkg depenencies?

cdmihai:
@anurse I'm seeing the same thing. Dotnet.Tools.Tests.Utilities -> dotnet-test-xunit -> MSFT.Extensions.Testing.Abstractions

So, in the past, ProjectDescription(Dotnet.Tools.Tests.Utilities).Dependencies() included the dependencies of dotnet-test-xunit, but now it does not do that anymore, and therefore Extensions.Testing.Abstractions is not seen anymore, so it's pushed back in the schedule because nobody seems to depend on it.

Questions is, why did ProjectDescription.Dependencies changed to not include transitive dependencies anymore? Is that a new design decision, or bug?

anurse:
@emgarten possibly.

@cdmihai I don't remember it ever having transitive dependencies, but I may be wrong.

anurse:
Ok, that failure is fixed, moving on to the next one :)

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -620,10 +618,17 @@ public static BuildTargetResult CrossgenSharedFx(BuildTargetContext c, string pa
                 // in CompileTargets and the one in the shared library project.json match and are updated in lock step, but long term
                 // we need to be able to look at the project.lock.json file and figure out what version of Microsoft.NETCore.Runtime.CoreCLR
                 // was used, and then select that version.
-                ExecSilent(Crossgen.GetCrossgenPathForVersion(CompileTargets.CoreCLRVersion),
+                ExecSilent(Crossgen.GetCrossgenPathForVersion(CoreCLRVersion),
                     "-readytorun", "-in", file, "-out", tempPathName, "-platform_assemblies_paths", pathToAssemblies);
 
-                File.Delete(file);
+                for (int i = 0; i < 10 && File.Exists(file); i++)
```
anurse: This reduced the frequency of the unauthorized access error moving KestrelDesktop, but it's still a band-aid

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -259,7 +258,6 @@ public static BuildTargetResult CompileStage2(BuildTargetContext c)
             var configuration = c.BuildContext.Get<string>("Configuration");
 
             CleanBinObj(c, Path.Combine(c.BuildContext.BuildDirectory, "src"));
-            CleanBinObj(c, Path.Combine(c.BuildContext.BuildDirectory, "test"));
```
anurse: Unnecessary and messes with rebuilding and re-running some tests.

scripts/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -228,7 +228,7 @@ public static BuildTargetResult RestorePackages(BuildTargetContext c)
         {
             var dotnet = DotNetCli.Stage0;
 
-            dotnet.Restore("--verbosity", "verbose", "--disable-parallel", "--infer-runtimes", "--fallbacksource", Dirs.Corehost)
+            dotnet.Restore("--verbosity", "verbose", "--disable-parallel", "--fallbacksource", Dirs.Corehost)
```
anurse: This was working around the issue with the NETCore App package, I'll revert it and rebase before merging.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -620,10 +618,17 @@ public static BuildTargetResult CrossgenSharedFx(BuildTargetContext c, string pa
...
                 // we need to be able to look at the project.lock.json file and figure out what version of Microsoft.NETCore.Runtime.CoreCLR
                 // was used, and then select that version.
-                ExecSilent(Crossgen.GetCrossgenPathForVersion(CompileTargets.CoreCLRVersion),
+                ExecSilent(Crossgen.GetCrossgenPathForVersion(CoreCLRVersion),
                     "-readytorun", "-in", file, "-out", tempPathName, "-platform_assemblies_paths", pathToAssemblies);
 
-                File.Delete(file);
+                for (int i = 0; i < 10 && File.Exists(file); i++)
+                {
+                    File.Delete(file);
```
eerhardt: Does this throw an exception?

anurse: No, though I'm probably just going to revert it since I don't think it's actually doing anything to help :)

src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.DotNet.Cli.Utils
 {
     public class ProjectToolsCommandResolver : ICommandResolver
     {
-        private static readonly NuGetFramework s_toolPackageFramework = FrameworkConstants.CommonFrameworks.NetStandardApp15;
+        private static readonly NuGetFramework s_toolPackageFramework = FrameworkConstants.CommonFrameworks.NetCoreApp10;
```
eerhardt: Wasn't there another change to do this that never got merged in?  https://github.com/dotnet/cli/pull/2350

src/dotnet/commands/dotnet-build/CompileContext.cs
```diff
@@ -514,31 +515,40 @@ private void MakeRunnable()
...
+            if (dependencies.ProjectDependenciesWithSources.TryGetValue(node.Library.Identity.Name, out projectDependency))
+            {
+                // Add to the list of projects to build
+                outputs.Add(projectDependency);
+            }
+        }
+
+        private class ProjectComparer : IEqualityComparer<ProjectDescription>
+        {
+            public bool Equals(ProjectDescription x, ProjectDescription y) => string.Equals(x.Identity.Name, y.Identity.Name);
```
eerhardt: Ordinal?

src/dotnet/project.json
```diff
@@ -55,13 +55,13 @@
...
-      "imports": [
-        "dnxcore50",
-        "netstandardapp1.5",
-        "portable-net45+win8",
-        "portable-net45+wp80+win8+wpa81+dnxcore50"
-      ]
+    "netcoreapp1.0": {
+        "imports": [
+            "dnxcore50",
+            "netstandardapp1.5",
```
eerhardt: Can this be removed now?  I thought only NuGet needed it.

test/dotnet-test.Tests/GivenThatWeWantToRunTestsInTheConsole.cs
```diff
@@ -21,14 +21,17 @@ public GivenThatWeWantToRunTestsInTheConsole()
...
-                testAssetManager.CreateTestInstance("ProjectWithTests", identifier: "ConsoleTests").WithLockFiles();
+                testAssetManager.CreateTestInstance("ProjectWithTests", identifier: "ConsoleTests");
 
             _projectFilePath = Path.Combine(testInstance.TestRoot, "project.json");
             var contexts = ProjectContext.CreateContextForEachFramework(
                 _projectFilePath,
                 null,
                 PlatformServices.Default.Runtime.GetAllCandidateRuntimeIdentifiers());
 
+            // Restore the project again in the destination to resolve projects
```
eerhardt: Can these give a little bit deeper of a reason of "why"?  That way when whatever the underlying issue is fixed, this restore can be removed.

src/dotnet/commands/dotnet-build/CompileContext.cs
anurse: That's the default, but yeah, I like being clear there as well :)

eerhardt:
I gave it a once over and added some minor comments.  Overall, LGTM.

test/dotnet-test.Tests/GivenThatWeWantToRunTestsInTheConsole.cs
anurse: There's no "issue" to be fixed. It's by design. Project paths are now stored in the lock file as relative paths and as a result, the project needs to be re-restored when it is moved to a different folder structure.

src/dotnet/commands/dotnet-build/CompileContext.cs
eerhardt: No, Ordinal is not the default string comparison.

src/dotnet/project.json
anurse: Trying now, we'll see if it works :)

src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
anurse: Yes, this supersedes that change

src/dotnet/commands/dotnet-build/CompileContext.cs
eerhardt: Oh I take that back... string.Equals does do ordinal.  I was thinking of StartsWith, IndexOf, etc.  My mistake.  Carry on, nothing to see here.

anurse:
```
17:14:56     Unable to resolve 'runtime.win7-x86.Microsoft.NETCore.DotNetHostPolicy (>= 1.0.1-rc2-002385-00)' for '.NETCoreApp,Version=v1.0 (win7-x86)'.
17:14:56     Unable to resolve 'runtime.win7-x86.Microsoft.NETCore.DotNetHostResolver (>= 1.0.1-rc2-002385-00)' for '.NETCoreApp,Version=v1.0 (win7-x86)'.
17:14:56     Unable to resolve 'runtime.win7-x86.Microsoft.NETCore.DotNetHost (>= 1.0.1-rc2-002385-00)' for '.NETCoreApp,Version=v1.0 (win7-x86)'.
```

Looks like we're still getting this :(

eerhardt:
:shipit:

piotrpMSFT:
wohoo!

PR #2350, opened 2016-04-07
[WIP] Switch tool TFM to netcoreapp1.0

pakrym:
/cc @moozzyk

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2350)


moozzyk:
Looks good to me but I am not expert in this area.

src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.DotNet.Cli.Utils
 {
     public class ProjectToolsCommandResolver : ICommandResolver
     {
-        private static readonly NuGetFramework s_toolPackageFramework = FrameworkConstants.CommonFrameworks.NetStandardApp15;
+        private static readonly NuGetFramework s_toolPackageFramework = FrameworkConstants.CommonFrameworks.NetCoreApp10;
```
brthor: Should we make this a list and have a preferred order

pakrym: Why? `netstandardapp` is going away and `netstandard` is compatible to `netcoreapp` so it will work out of the box.

piotrpMSFT: netstandardapp is going away but is not gone yet. Let's avoid a breaking change that requires rebuild of tools this close to lockdown.

pakrym: Nuget restores tools only for `netcoreapp` starting with https://github.com/NuGet/NuGet.Client/commit/9573f19ce94c84078e0d0e4f128bb24de7bfe140 so having two frameworks in CLI won't help anything. To make `netstandardadpp` tool work you just need to add `"import":"netstandardadpp"` to tool section.

brthor: @pakrym If you look here https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/CommandResolution/ToolPathCalculator.cs

You'll see that this variable is used to compute a directory name. 

So this won't just work out of the box for tools already targeting netstandard or netstandardapp

This is why I suggest a preference list.

anurse: A tool targeting `netstandardapp` will **already** be broken because of the upcoming NuGet change @pakrym referenced (https://github.com/NuGet/NuGet.Client/commit/9573f19ce94c84078e0d0e4f128bb24de7bfe140). NuGet only restores the tool for `netcoreapp1.0`. There's no point in having a preference order because NuGet will **never** put a `netstandardapp` tool, it will just fail with a compatibility error.

brthor: The existing tools actually target `netstandard1.5` (for the portable workaround) is `netcoreapp1.0` compatible with that?

pakrym: Yes

brthor: In which case the restore would still succeed but then we'd break on the CLI side?

brthor: Had some discussion around this.

When restoring a netstandard1.5 tool, will nuget still place the project.lock.json in the netcoreapp tfm directory?

Rather than answering this it'd be preferable to add a test that covers the scenario.

pakrym:
This is done in other PR

PR #2353, opened 2016-04-07
fix #745: dotnet new should not expose public types

### LINKED ISSUE #745: dotnet new template should not expose public types
The "new" template exposes a public type with a public method:

```C#
public class Program
    {
        {
            Console.WriteLine("Hello World!");
        }
    }
```
These should not be public. 

Also, maybe we should remove the namespace once there are no public types. it will simplify our "hello world"

krwq:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2353)

### DIFF
--- a/src/dotnet/commands/dotnet-new/CSharp_Console/Program.cs
+++ b/src/dotnet/commands/dotnet-new/CSharp_Console/Program.cs
@@ -2,7 +2,7 @@
 
 namespace ConsoleApplication
 {
-    public class Program
+    class Program
     {
         public static void Main(string[] args)
         {


krwq:
cc: @KrzysztofCwalina

davidfowl:
I have to ask, why does this matter?

brthor:
@dotnet-bot test this please

krwq:
@davidfowl, it's generally bad practice to expose public APIs

troydai:
I can offer one reason to make a program class publish. It could be useful for test. For example dotnet-projectmodel-server test initially "new" a program to run the server in process. (it is not the case now since the dotnet-projectmodel-server was wrapped into the dotnet project and the Program turn into a command).

eerhardt:
@krwq @davidfowl @blackdwarf Are we still interested in taking this PR for Preview2?  If not, it should be closed.

krwq:
I honestly think it actually doesn't matter, this is just a single word to change for user.

eerhardt:
Closing as this appears to have zero traction.

PR #2365, opened 2016-04-08
Update packages to rc2 24008

ellismg:
Pick up the new RC2 CoreCLR/CoreFX builds based on some fixes that went in today.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2365)


ellismg:
@eerhardt You will want to take a look at 0884efa (which fixed an issue with the update scripts) and 	3151698 (without this, the scripts would not update that project.json file, it seemed pretty bogus to have stuff listed twice at different versions).

e357fc7 is just me running update-dependencies.ps1 -Targets UpdateFiles, spot checking the results and then committing it.

ellismg:
/cc @piotrpMSFT @Petermarcu 

It looked like CI had not been super healthy on other PRs, so who knows if this will hit the same issues.

ellismg:
I validated by git grep there were no references to the old version except for in a comment in one of the pkgproj files that explained how we picked a version number that we did.

TestAssets/TestProjects/FSharpTestProjects/CompileFailApp/project.json
```diff
@@ -1,31 +1,32 @@
 {
-    "version": "1.0.0-*",
```
davidfowl: noooo formatting changes...

weshaggard:
FYI we once the fill CLI build finishes we will need to update the Microsoft.NETCore.App package version.

eerhardt:
> FYI we once the fill CLI build finishes we will need to update the Microsoft.NETCore.App package version.

That will conflict with my change:  https://github.com/dotnet/cli/pull/2331.  Could we hold off on something like that?

weshaggard:
Yes we should try to get your change in instead @eerhardt

Issue #2359, opened 2016-04-07
Put hostfxr.dll inside a versioned sub directory


schellap:
Currently, we have the following layout in the setup zips:

```
dotnet.exe
hostfxr.dll
shared\
sdk\
```

Do the following instead:

```
dotnet.exe
shared\
sdk\
host\
    fxr\
       1.0.0-rc2-002200\
             hostfxr.dll
       1.0.1-rc2-002201\
             hostfxr.dll
```

Currently, we are not overwriting hostfxr.dll during the installations.

Cc @piotrpMSFT @krwq @brthor @Sridhar-MS 

@gkhanna79, this is needed to mitigate the interface breaks and currently we are not copying hostfxr.dll if it is in use.

gkhanna79:
@schellap Doesnt this require change in the host as well?

schellap:
No, the host is already built with this logic baked in it.

I'm thinking once this model is in place for RC2, i.e., after @krwq implements it, we should use the side by side as the default first choice and then probe for the host/fxr as second choice. Right now as implemented we get sxs as the second choice in case host/fxr contains a higher version. This is to give the user a fallback.

gkhanna79:
> we should use the side by side as the default first choice and then probe for the host/fxr as second choice.

We expect to have a newer version be pushed to host\fxr and not side by side. Thus, making it fallback lookup does not sound right. I think we can remove hostfxr.dll from sxs to host and always keep it under host\fxr. this will always work and will always pickup new versions correctly without break.

What do you think @schellap ?

schellap:
> Thus, making it fallback lookup does not sound right. I think we can remove hostfxr.dll from sxs to host and always keep it under host\fxr. this will always work and will always pickup new versions correctly without break.

Once this change is done, there would be no hostfxr.dll by the side of dotnet.exe. I am just saying if user explicitly copied one next to it we'd pick it up as an override. This gives an user to explicitly copy something next to it as a workaround.

gkhanna79:
@schellap Why is that required? What is the user scenario (other than us) doing so?

piotrpMSFT:
We're kicking this out to rtm. There is significant installer work that needs to happen to appropriately cover the scenario.

schellap:
No issues since hostfxr.dll since last night should be backwards compatible with hostpolicy.

piotrpMSFT:
Tracked by https://github.com/dotnet/core-setup/issues/94

PR #2377, opened 2016-04-08
add hostfxr to a versioned folder

krwq:
Fixes https://github.com/dotnet/cli/issues/2359
@schellap @Sridhar-MS @brthor please validate

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2377)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -237,7 +237,7 @@ namespace Microsoft.DotNet.Cli.Build
             }
             Directory.CreateDirectory(Dirs.Stage1);
 
-            CopySharedHost(Dirs.Stage1);
+            CopySharedHost(c, Dirs.Stage1);
             PublishSharedFramework(c, Dirs.Stage1, DotNetCli.Stage0);
             var result = CompileCliSdk(c,
                 dotnet: DotNetCli.Stage0,
@@ -266,7 +266,7 @@ namespace Microsoft.DotNet.Cli.Build
             Directory.CreateDirectory(Dirs.Stage2);
 
             PublishSharedFramework(c, Dirs.Stage2, DotNetCli.Stage1);
-            CopySharedHost(Dirs.Stage2);
+            CopySharedHost(c, Dirs.Stage2);
             var result = CompileCliSdk(c,
                 dotnet: DotNetCli.Stage1,
                 outputDir: Dirs.Stage2);
@@ -317,15 +317,18 @@ namespace Microsoft.DotNet.Cli.Build
             FS.RmFilesInDirRecursive(directory, "*.pdb");
         }
 
-        private static void CopySharedHost(string outputDir)
+        private static void CopySharedHost(BuildTargetContext c, string outputDir)
         {
             // corehost will be renamed to dotnet at some point and then this can be removed.
             File.Copy(
                 Path.Combine(Dirs.Corehost, CoreHostBaseName),
                 Path.Combine(outputDir, $"dotnet{Constants.ExeSuffix}"), true);
+            
+            string versionedFxrDir = Path.Combine(outputDir, "host", "fxr", c.BuildContext.Get<BuildVersion>("BuildVersion").NuGetVersion);
+            Directory.CreateDirectory(versionedFxrDir);
             File.Copy(
                 Path.Combine(Dirs.Corehost, DotnetHostFxrBaseName),
-                Path.Combine(outputDir, DotnetHostFxrBaseName), true);
+                Path.Combine(versionedFxrDir, DotnetHostFxrBaseName), true);
         }
 
         public static void PublishSharedFramework(BuildTargetContext c, string outputDir, DotNetCli dotnetCli)
--- a/scripts/dotnet-cli-build/PackageTargets.cs
+++ b/scripts/dotnet-cli-build/PackageTargets.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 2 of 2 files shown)


scripts/dotnet-cli-build/PackageTargets.cs
```diff
@@ -97,6 +97,20 @@ public static BuildTargetResult CopySharedHostLayout(BuildTargetContext c)
...
                 File.Copy(file, destFile, true);
             }
+            
+            string versionedFxrDir = Path.Combine(sharedHostRoot, "host", "fxr", c.BuildContext.Get<BuildVersion>("BuildVersion").NuGetVersion);
+            Directory.CreateDirectory(versionedFxrDir);
+            foreach (var file in Directory.GetFiles(Dirs.Stage2, "*", SearchOption.TopDirectoryOnly))
+            {
+                if (!Path.GetFileName(file).StartsWith("hostfxr"))
+                {
+                    continue;
```
schellap: Contains() or use platform specific names.

scripts/dotnet-cli-build/PackageTargets.cs
```diff
@@ -97,6 +97,20 @@ public static BuildTargetResult CopySharedHostLayout(BuildTargetContext c)
                 var destFile = file.Replace(Dirs.Stage2, sharedHostRoot);
                 File.Copy(file, destFile, true);
             }
+            
+            string versionedFxrDir = Path.Combine(sharedHostRoot, "host", "fxr", c.BuildContext.Get<BuildVersion>("BuildVersion").NuGetVersion);
```
schellap: Also, where are you removing from root?

schellap:
This seems to pass CI because you haven't deleted the root one and on non-Windows platforms it hasn't been copied.

scripts/dotnet-cli-build/PackageTargets.cs
```diff
@@ -97,6 +97,20 @@ public static BuildTargetResult CopySharedHostLayout(BuildTargetContext c)
...
+            Directory.CreateDirectory(versionedFxrDir);
+            foreach (var file in Directory.GetFiles(Dirs.Stage2, "*", SearchOption.TopDirectoryOnly))
+            {
+                if (!Path.GetFileName(file).StartsWith("hostfxr"))
+                {
+                    continue;
+                }
+                
+                var destFile = file.Replace(Dirs.Stage2, versionedFxrDir);
+                File.Copy(file, destFile, true);
```
piotrpMSFT: Copy? or move?

krwq: Im repeating the existing logic. I think Copy is better since investigation of possible issues will be easier

schellap: No. Delete the one in the root. How will you avoid the unzip overwrite problem if you keep it in the root? Your unzip should not swallow overwrite errors except for dotnet.exe.

Sridhar-MS:
Why is this happening in `PackageTargets` ? We should put the hostfxr in the folders even in stage1 and stage2.

scripts/dotnet-cli-build/PackageTargets.cs
krwq: not removing, I though you wanted to keep both files, updated now

krwq:
I thought we are using output of this to create stage1 and stage2, will update this soon

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -317,15 +317,18 @@ private static void RemovePdbsFromDir(string directory)
...
 
-        private static void CopySharedHost(string outputDir)
+        private static void CopySharedHost(BuildTargetContext c, string outputDir)
         {
             // corehost will be renamed to dotnet at some point and then this can be removed.
             File.Copy(
                 Path.Combine(Dirs.Corehost, CoreHostBaseName),
                 Path.Combine(outputDir, $"dotnet{Constants.ExeSuffix}"), true);
+            
+            string versionedFxrDir = Path.Combine(outputDir, "host", "fxr", c.BuildContext.Get<BuildVersion>("BuildVersion").NuGetVersion);
```
brthor: Can we pass in the versionedFxrDir instead of the `BuildTargetContext`? 

Will make it less painful when we go to refactor this thing.

krwq: I think it will be the same..

eerhardt:
Do we still need this change? If so it will need to be made in `dotnet/core-setup`.

brthor:
@eerhardt These changes will likely need to be piped all the way through `core-setup` and `cli` because to do this right we need to add a fourth installer and zip.

@schellap Can you comment if this is needed for preview2?

brthor:
Thanks for pointing that out, you're right @eerhardt 

.NET Core RTM and cli preview 2.

schellap:
I think we should do this for preview2. Though RC2 hostfxr will have no problems selecting FX, few new changes have gone in. Even if we don't do a 4th installer, moving it to a versioned folder should be done. @gkhanna79, this would go away if we ask customers to remove rc2. What is the upgrade policy?  @blackdwarf, the problem is to move hostfxr.dll into a separate folder so RTM installer doesn't have to stomp on RC2 hostfxr.dll.

gkhanna79:
> this would go away if we ask customers to remove rc2. What is the upgrade policy?

There is no upgrade policy since they are SxS. This should move into its own folder.

brthor:
@schellap What is the benefit of a versioned folder without a 4th installer? I don't see how we could have SxS since the host installer is upgraded in place.

PR #2379, opened 2016-04-08
Fix equality issue in DependencyContext

### LINKED ISSUE #2381: DependencyContext.Default is null when in a `dotnet-` tool
When `DependencyContext.Default` is dereferenced by a tool (specifically `dotnet-ef` in this case), it always returns `null`.

This is because the code in the `DependencyContextLoader` that checks to see if the assembly is the "Entry Point Assembly" (i.e. the one with Program.Main), does so by comparing the `AssemblyName`s, which don't implement `Equals`, so it does reference equality and fails. As a result, it then tries to load the DependencyContext as though it were a standalone app, which fails because the tool does not have a `.deps.json` (and even if it succeeded it would be wrong because the tool is portable).

The net effect of this is that any tool that needs to perform assembly loading out of a project is missing data necessary for it to perform those loads. So in `dotnet-ef`'s case, it is unable to function, since it must load the project assembly.

/cc @natemcmaster @piotrpMSFT @pakrym

anurse:
/cc @pakrym @piotrpMSFT 

This is blocking EntityFramework RC2

Fixes #2381

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2379)

### DIFF
--- a/TestAssets/TestPackages/dotnet-dependency-context-test/Program.cs
+++ b/TestAssets/TestPackages/dotnet-dependency-context-test/Program.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Linq;
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.DotNet.Tools.DependencyContextTest
+{
+    public class Program
+    {
+        public static int Main(string[] args)
+        {
+            if(args.Length > 0 && args[0] == "--debug")
+            {
+                Console.WriteLine("Waiting for Debugger to attach, press ENTER to continue");
+                Console.WriteLine($"Process ID: {System.Diagnostics.Process.GetCurrentProcess().Id}");
+                Console.ReadLine();
+            }
+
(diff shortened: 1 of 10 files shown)


TestAssets/TestPackages/dotnet-dependency-context-test/Program.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Linq;
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.DotNet.Tools.DependencyInvoker
+{
+    public class Program
```
pakrym: There is DependencyContextValidator test project, maybe we can use it?

anurse: But seriously, it was tricky enough for me to work out one kind of tests. I'd just like to get this in :)

pakrym:
:shipit:

Issue #2405, opened 2016-04-11
Dev builds fail unless rebased 


brthor:
## Steps to reproduce
Clone rel
`git reset --soft HEAD~1`

(This puts HEAD one commit behind master)

`build.cmd -Targets Prepare,Compile`

## Expected  behavior
Builds

## Actual behavior
`Could not load host policy library from [C:\Users\brthor\code\cli\artifacts\win10-x64\stage1\shared\Microsoft.NETCore.App\1.0.0-rc2-3002340]
This may be because the targeted framework [Microsoft.NETCore.App 1.0.0-rc2-3002340] was not found.`

Interestingly I see Microsoft.NETCore.App version `1.0.0-rc2-3002338` in my shared directory.
It seems like there is some mismatch in where the version comes from throughout the build.

## Environment data
`dotnet --info` output:

Not applicable. Using a branch a 2 or 3 commits behind.
On Windows 10.

brthor:
@eerhardt Assigned you because you built this out. Feel free to redirect as necessary.

eerhardt:
I'm not 100% sure I understand why this is a problem that needs to be fixed... We've made many breaking changes from build to build that require getting the latest code in order to build with Stage0.

brthor:
This slows down the developer workflow. 

Right now every commit to rel will break the build in every branch until it's rebased. Sure we do make breaking changes, but not at that frequency.

If we get into a place where we are checking in 20 commits a day again (even if they're unrelated to the build), everyone will be spending half the day rebasing just to be productive.

eerhardt:
It looks like an issue is that Stage0 is using its current Shared FX version when writing the dotnet.runtimeconfig.json out to the Stage1 sdk folder:

```JSON
{
  "runtimeOptions": {
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "1.0.0-rc2-3002339"
    }
  }
}
```

What controls writing this .json file?

brthor:
The Executable class in Utils.

He's on vacation but maybe he'll see this @livarcocc

eerhardt:
Looks like the project.lock.json file for src\dotnet says:

```
 "Microsoft.NETCore.App/1.0.0-rc2-3002339":
```

probably that is the highest number in my nuget cache, and since we are using * versions, it uses that one.

brthor:
Ah, that's for sure an issue. Should we clean that package location at the beginning of the build? We do a similar thing for the test packages

[...]

PR #2410, opened 2016-04-11
Clean Microsoft.NETCore.* nuget packages from the cache during the build.

eerhardt:
This ensures the rest of the product will only use the packages it currently built, since it uses * versions.

@brthor @weshaggard @piotrpMSFT

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2410)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -67,9 +67,9 @@ namespace Microsoft.DotNet.Cli.Build
             return c.Success();
         }
 
-        // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the
-        // Microsoft.NETCore.App package that is created during PackagePkgProjects.
-        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(RestoreLockedCoreHost), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
+        // Moving PrepareTargets.RestorePackages after BuildCoreHost because managed code depends on the
+        // Microsoft.NETCore.App package that is created during BuildCoreHost.
+        [Target(nameof(PrepareTargets.Init), nameof(BuildCoreHost), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
         public static BuildTargetResult Compile(BuildTargetContext c)
         {
             return c.Success();
@@ -145,6 +145,11 @@ namespace Microsoft.DotNet.Cli.Build
 
             return c.Success();
         }
+
+        [Target(
+            nameof(CompileCoreHost),
+            nameof(BuildCoreHostPackages))]
+        public static BuildTargetResult BuildCoreHost(BuildTargetContext c) => c.Success();
 
         [Target(nameof(PrepareTargets.Init))]
         public static BuildTargetResult CompileCoreHost(BuildTargetContext c)
@@ -229,8 +234,15 @@ namespace Microsoft.DotNet.Cli.Build
             return c.Success();
         }
 
-        [Target(nameof(CompileTargets.GenerateStubHostPackages))]
-        public static BuildTargetResult PackagePkgProjects(BuildTargetContext c)
+        [Target(
+            nameof(GenerateStubHostPackages),
+            nameof(PackageHostPackages),
+            nameof(CleanHostPackagesFromCache),
+            nameof(RestoreLockedCoreHost))]
+        public static BuildTargetResult BuildCoreHostPackages(BuildTargetContext c) => c.Success();
+
+        [Target]
+        public static BuildTargetResult PackageHostPackages(BuildTargetContext c)
         {
             var arch = IsWinx86 ? "x86" : "x64";
             
@@ -289,6 +301,18 @@ namespace Microsoft.DotNet.Cli.Build
                     throw new BuildFailureException($"Nupkg for {fileFilter} was not created.");
                 }
             }
+            return c.Success();
+        }
+
+        [Target]
+        public static BuildTargetResult CleanHostPackagesFromCache(BuildTargetContext c)
+        {
+            IEnumerable<string> packagesToClean = NuGetUtils.GetPackageIds(
+                Dirs.CorehostLocalPackages, 
+                Dirs.CorehostDummyPackages);
+
+            NuGetUtils.CleanPackagesFromCache(packagesToClean);
+
             return c.Success();
         }
 
--- a/scripts/dotnet-cli-build/PackageTargets.cs
+++ b/scripts/dotnet-cli-build/PackageTargets.cs
@@ -217,7 +217,6 @@ namespace Microsoft.DotNet.Cli.Build
             var versionSuffix = c.BuildContext.Get<BuildVersion>("BuildVersion").CommitCountString;
             var configuration = c.BuildContext.Get<string>("Configuration");
 
-            var env = GetCommonEnvVars(c);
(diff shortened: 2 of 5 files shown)


eerhardt:
Fixes #2405

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -42,6 +42,14 @@ public class CompileTargets
         private static string DotnetHostFxrBaseName => $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}";
         private static string HostPolicyBaseName => $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}";
 
+        public static readonly string[] PkgProjects = new[]
```
brthor: I think this name is pretty confusing given the mac native installer is a PKG

weshaggard:
Do we ever using dotnet-core feed for any of the restores? That is where I'm publishing the meta-package today.

anurse:
This seems really scary. We shouldn't be depending upon the package cache being clean, it would break the ability to build the repo from different clones on the same machine, and it's just generally brittle...

weshaggard:
I agree with @anurse that we shouldn't depend on clearing the cache in a mainstream scenario like this. It would seem to me that that `*` versioning in the general has this problem so we either have to live with it or not use it.

scripts/dotnet-cli-build/CompileTargets.cs
eerhardt: I agree it is confusing, but this is already an issue with the repo, since that is the name used for these packages elsewhere:

https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/CompileTargets.cs#L149
https://github.com/dotnet/cli/tree/rel/1.0.0/pkg

I tried thinking of better names, but failed. Also, it would be confusing to name this variable something that didn't align with "PackagePkgProjects"

eerhardt:
> it would break the ability to build the repo from different clones on the same machine, and it's just generally brittle...

This scenario is already broken today since these package versions are based on the number of commits to the repo.  I could have one clone that I have 5 commits head of the tip, and another clone that is "sunk to the tip".  Building the first would make version "tip+5".  Building the second would make version "tip", but since there is a "tip+5" in the cache, that is what will get used.

eerhardt:
> Do we ever using dotnet-core feed for any of the restores? That is where I'm publishing the meta-package today.

No, the CLI repo is only ever dependent on the cli-deps feed and nuget.org.

eerhardt:
> It would seem to me that that * versioning in the general has this problem so we either have to live with it or not use it.

If we could have a P2P reference to the Microsoft.NETCore.App package, then this would all go away.  Also, if Microsoft.NETCore.App was built outside of the cli repo, this would go away.

But since we can't do those things, the only other option would be to generate the project.json files that consume Microsoft.NETCore.App during the build.

anurse:
> Also, if Microsoft.NETCore.App was built outside of the cli repo, this would go away.

This actually seems like the right approach to me. The redist shouldn't necessarily be built in the CLI repo, we just take a version and consume it. I realize there may be logistical issues there though.

eerhardt:
Do we have a consensus on what we should be doing here?  We are just deleting the nuget packages that we built using the local repo from the cache.  This is the same approach that we use in the TestTargets:

https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/TestTargets.cs#L215
https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/TestTargets.cs#L230

The idea being - if you generate a NuGet package using your local repo, you don't want that package hanging around in your cache. It is not the "official" version of that package since you just built it locally. And there may be other packages with the same Id and Version, but different contents.

weshaggard:
There are certainly discussions about where the correct place to build Microsoft.NETCore.App is but I don't see that changing before RTM of v1. So we should figure out what other reasonable options we have.

I'm not entirely sure how project references work in the case of a package that isn't a CLI base project like this but if we could get that to work then that seems like it might be a reasonable approach.

eerhardt:
> I'm not entirely sure how project references work in the case of a package that isn't a CLI base project like this but if we could get that to work then that seems like it might be a reasonable approach.

Who would know how to do something like this?  Would we need a fake project.json file for Microsoft.NETCore.App?

IMO - I think for now we should move forward with cleaning these packages out of the cache. It makes total sense to me that you don't want these packages in your cache. They are mutable.

We can log an investigation issue to see if we can use P2P references here, if we think that is a feasible approach.

scripts/dotnet-cli-build/CompileTargets.cs
pranavkm: `NuGetPackageProjects`?

eerhardt:
@brthor @Sridhar-MS @krwq I updated the strategy on how to clean the packages.  PTAL

eerhardt:
test OSX x64 Release Build

eerhardt:
test OSX x64 Release Build

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -149,6 +149,11 @@ public static BuildTargetResult RestoreLockedCoreHost(BuildTargetContext c)
             return c.Success();
         }
 
+        [Target(
+            nameof(CompileCoreHost),
+            nameof(BuildCoreHostPackages))]
+        public static BuildTargetResult BuildCoreHost(BuildTargetContext c) => c.Success();
```
Sridhar-MS: We had to separate CompileCoreHost and BuildCoreHostPackages into two unrelated targets to enable signing of host and dependent libraries. This change will break that.

eerhardt: What PR is that part of?  Depending on what you are doing there, the changes can merge pretty easily.

Sridhar-MS: Its the current state of the repo. Keep `CompileCoreHost` and `BuildCoreHostPackages` as unrelated targets, similar to `CompileCoreHost` and `PackagePkgProjects` before this change.

The original PR - https://github.com/dotnet/cli/pull/2439

eerhardt: They are still 2 independent targets.  You can call CompileCoreHost and then call BuildCoreHostPackages.

This just simplifies the Compile target to not have to know about 5 different "host" targets.

Sridhar-MS: Yeah.. That could still work. But since you have changed the name of the targets, this change warrants change in the VSO build definitions. 

This need for keeping VSO and our builds scripts in sync is horrible. We must think about it more.

eerhardt: Ah, I get your point now.

Yes, having VSO know about the internals of our build is not a good thing going forward.  :(

This change will actually clean VSO up a little bit.  Instead of saying:

```
build -Configuration $(BuildConfiguration) -Targets Prepare,CompileCoreHost
Sign
build -Configuration $(BuildConfiguration) -Targets Prepare,PackagePkgProjects,RestoreLockedCoreHost,...
```

It can now say:

```
build -Configuration $(BuildConfiguration) -Targets Prepare,CompileCoreHost
Sign
build -Configuration $(BuildConfiguration) -Targets Prepare,BuildCoreHostPackages,...
```

Either way I'm going to have to change VSO since I'm injecting a new step between `PackagePkgProjects` and `RestoreLockedCoreHost`

eerhardt:

eerhardt:
Closing.  Now that Microsoft.NETCore.App is getting built out of core-setup, we no longer need to do this cleaning.

PR #2411, opened 2016-04-11
Fix update deps

eerhardt:
Currently there are no CoreFX version numbers in our `dotnet new` templates, so we didn't need to update the version numbers in these files.

But with new `dotnet new` templates (#2106) coming online, we should include these files in our update script.

Also, refactored the Config class to lazy load the environment variables, so the UpdateFiles target can be run without setting any environment variables.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2411)


eerhardt:

piotrpMSFT:
:shipit:

PR #2415, opened 2016-04-12
Fix test project imports to be consistent.

eerhardt:
- Every test project.json needs portable-net451+win8 and dotnet5.4 imports (required by dotnet-test-xunit).
- If a test references NuGet, it also needs "netstandardapp1.5", because that the TFM NuGet uses currently.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2415)

### DIFF
--- a/test/ArgumentForwardingTests/project.json
+++ b/test/ArgumentForwardingTests/project.json
@@ -24,9 +24,8 @@
   "frameworks": {
     "netcoreapp1.0": {
       "imports": [
-        "netstandardapp1.5",
-        "dnxcore50",
-        "portable-net45+win8"
+        "dotnet5.4",
+        "portable-net451+win8"
       ]
     }
   },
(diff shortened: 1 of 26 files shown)


davidfowl:
@eerhardt Can you remove all imports then run restore, then see what fails, then add very specific imports?

eerhardt:
@davidfowl - I did do that and this is what came out of that exercise.

Xunit.abstractions needs the portable-net45+win8 import.

dotnet-test-xunit -> xunit.runner.reporter. And xunit.runner.reporter needs either dnxcore50 or dotnet5.4. Both TFMs are deprecated, but I choose dotnet5.4 simply because it doesn't say "dnx" and that is what we are using in our 'dotnet new -t test' template.

All test projects need those two imports no matter what.

Lastly, if the test references any NuGet assembly, it needs 'netstandardapp1.5' import since that is the TFM NuGet uses right now. (I logged a bug for NuGet to fix this).

piotrpMSFT:
:shipit:

eerhardt:

Issue #2418, opened 2016-04-12
[dotnet publish] publish fails with ne --no-build flag


abpiskunov:
## Steps to reproduce

from https://github.com/weidazhao/Hosting and follow the instruction in readme.md to setup Service Fabric SDK. After opening the Hosting.sln in VS, you can right-click Hosting.sfproj -> Package to repro the issue. 

## Expected  behavior

Publish to succeed.

## Actual behavior

1>------ Build started: Project: Microsoft.ServiceFabric.AspNetCore.Hosting, Configuration: Debug Any CPU ------
2>------ Build started: Project: Microsoft.ServiceFabric.AspNetCore.Gateway, Configuration: Debug Any CPU ------
1>  C:\Program Files\dotnet\dotnet.exe build "C:\Repos\Hosting\Microsoft.ServiceFabric.AspNetCore.Hosting" --configuration Debug --no-dependencies --build-base-path "C:\Repos\Hosting\artifacts"
2>  C:\Program Files\dotnet\dotnet.exe build "C:\Repos\Hosting\Microsoft.ServiceFabric.AspNetCore.Gateway" --configuration Debug --no-dependencies --build-base-path "C:\Repos\Hosting\artifacts"
1>  Compiling Microsoft.ServiceFabric.AspNetCore.Hosting for .NETFramework,Version=v4.5.2
2>  Compiling Microsoft.ServiceFabric.AspNetCore.Gateway for .NETFramework,Version=v4.5.2
1>  Compilation succeeded.
1>      0 Warning(s)
1>      0 Error(s)
1>  Time elapsed 00:00:04.3782632
1>
3>------ Build started: Project: Counter, Configuration: Debug Any CPU ------
4>------ Build started: Project: Sms, Configuration: Debug Any CPU ------
4>  C:\Program Files\dotnet\dotnet.exe build "C:\Repos\Hosting\Sms" --configuration Debug --no-dependencies --build-base-path "C:\Repos\Hosting\artifacts"
2>  Compilation succeeded.
2>      0 Warning(s)
2>      0 Error(s)
2>  Time elapsed 00:00:05.2997171
2>
5>------ Build started: Project: Gateway, Configuration: Debug Any CPU ------
4>  Compiling Sms for .NETFramework,Version=v4.5.2
4>  Compilation succeeded.
4>      0 Warning(s)
4>      0 Error(s)
4>  Time elapsed 00:00:07.7961812
4>
[...]

PR #2437, opened 2016-04-12
Fix publish when build was ran using build-base-path

pakrym:
Fixes: https://github.com/dotnet/cli/issues/2418
@anurse

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2437)


anurse:
:shipit:

Issue #2413, opened 2016-04-12
Some files are missing AuthentiCode signatures


joeloff:
Some of the runtime.win7-x64.* packages seem to have DLLs that are missing Authenticode signatures, e.g. runtime.win7-x64.Microsoft.NETCore.DotNetHost.1.0.0-rc2-00001, \runtimes\win7-x64\native\dotnet.exe is missing an Authenticode signature. Not sure if this is intentional.
Other files are hostfxr.dll, hostpolicy.dll and Microsoft.DotNet.Watcher.Core.dll and runtimes\win7-x64\lib\netstandard1.0\mscorlib.dll

gkhanna79:
CC @weshaggard @joperezr @markwilkie - shouldn't the Windows mscorlib.dll be signed?

@piotrpMSFT @schellap can you please look into signing of hostfxr.dll and hostpolicy.dll?

@Sridhar-MS dotnet.exe signing sounds like https://github.com/dotnet/cli/issues/1103.

ericstj:
@joeloff can you point us at the exact package you're seeing an unsigned mscorlib.dll in?  If I check https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.win7-x64.Microsoft.NETCore.Runtime.CoreCLR/1.0.2-rc2-24008 I see that mscorlib.dll is SHA256 authenticode signed.

weshaggard:
Yes I expect that mscorlib should be properly signed but the host assemblies may not be getting signed correctly. @schellap please follow-up on that.

joeloff:
Talked to @ericstj The mscorlib came from a package we don't want to ship: Microsoft.TargetingPack.Private.CoreCLR.1.0.0-rc2-23931.nupkg

gkhanna79:
@Sridhar-MS were you able to fix signing of hostfxr.dll and hostpolicy.dll?

Sridhar-MS:
I have a PR out for this - https://github.com/dotnet/cli/pull/2439

piotrpMSFT:
#2439 has been merged. This bug will get closed once VSO definition is updated by @Sridhar-MS

Petermarcu:
@Sridhar-MS , have the VSTS definitions been updated?

Sridhar-MS:
I have updated the VSTS definitions. Waiting for a build to complete successfully. I can close this issue after that.

Sridhar-MS:
[...]

PR #2439, opened 2016-04-12
Bifurcate the compile CoreHost and nuget packing step into two.

Sridhar-MS:
This is needed to insert a signing step between them where hostfxr.dll and
hostpolicy.dll will be signed.

Fixes - #2413 

cc: @schellap

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2439)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -51,7 +51,7 @@ namespace Microsoft.DotNet.Cli.Build
 
         // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the 
         // Microsoft.NETCore.App package that is created during PackagePkgProjects.
-        [Target(nameof(PrepareTargets.Init), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
+        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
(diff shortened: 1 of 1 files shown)


brthor:
:shipit:

schellap:
:ship:

gkhanna79:
@brthor What does the Mouse icon mean?

schellap:
Squirrel = nimble (ship quickly) type `:shipit:`

schellap:
Also https://en.wikipedia.org/wiki/Crasher_Squirrel

brthor:
I think of it as the more hip equivalent of LGTM :smile:

weshaggard:
How is the signing step going to be inserted? It looks like you are still always building them wouldn't you need to remove PackagePkgProjects from the list of Compile targets to run?

Sridhar-MS:
I will be doing something like this in the VSO build definition
>
`build.cmd -Targets Prepare,CompileCorehost`
`VSO build step to sign hostfxr.dll and hostpolicy.dll`
`build.cmd -Targets Prepare,PackagePkgProjects`

Earlier `PackagePkgProjects` did both compilation and packaging.

weshaggard:
I understood that part the piece I was missing is that we don't just call the Compile target. I kind of still think we should remove PackagePkgProjects from the list of targets Compile builds but it isn't absolutely necessary.

Sridhar-MS:
Well. It is not possible to remove `PackagePkgProjects` from `Compile` targets. Building the CLI SDK and SharedFx is dependent on `Microsoft.NetCore,App` which is generated by `PackagePkgProjects`.

weshaggard:
Fair enough.

gkhanna79:
Ready2Merge?

Sridhar-MS:
Have not got a replay from shiproom yet. Just pinged them.

PR #2443, opened 2016-04-12
Servicing for hostpolicy

schellap:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2443)

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -378,24 +378,32 @@ int fx_muxer_t::read_config_and_execute(
         trace::verbose(_X("Executing as a portable app as per config file [%s]"), config_file.c_str());
         pal::string_t fx_dir = (mode == host_mode_t::split_fx) ? own_dir : resolve_fx_dir(own_dir, &config);
         corehost_init_t init(deps_file, probe_paths, fx_dir, mode, config);
-        return execute_app(fx_dir, &init, new_argc, new_argv);
+
+        pal::string_t impl_dir;
+
+        // First lookup hostpolicy.dll in servicing with the version of hostpolicy.dll that was compiled lock step with hostfxr.
+        if (!hostpolicy_exists_in_svc(&impl_dir))
+        {
+            impl_dir = fx_dir;
+        }
+        return execute_app(impl_dir, &init, new_argc, new_argv);
     }
     else
     {
         pal::string_t impl_dir;
         trace::verbose(_X("Executing as a standalone app as per config file [%s]"), config_file.c_str());
-        if (mode == host_mode_t::standalone)
+
+        // First lookup hostpolicy.dll in servicing with the version of hostpolicy.dll that was compiled lock step with hostfxr.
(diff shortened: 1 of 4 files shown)


schellap:
@gkhanna79 can you take a look?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -378,24 +378,28 @@ int fx_muxer_t::read_config_and_execute(
         trace::verbose(_X("Executing as a portable app as per config file [%s]"), config_file.c_str());
         pal::string_t fx_dir = (mode == host_mode_t::split_fx) ? own_dir : resolve_fx_dir(own_dir, &config);
         corehost_init_t init(deps_file, probe_paths, fx_dir, mode, config);
-        return execute_app(fx_dir, &init, new_argc, new_argv);
+
+        pal::string_t impl_dir;
+        if (!hostpolicy_exists_in_svc(&impl_dir))
```
gkhanna79: Can you please add a comment on which version will be picked up?

src/corehost/cli/setup.cmake
```diff
@@ -71,5 +71,5 @@ if(${CLI_CMAKE_PKG_RID} STREQUAL "")
     message(FATAL_ERROR "A minimum supported package rid is not specified (ex: win7-x86 or ubuntu.14.04-x64, osx.10.10-x64, rhel.7-x64)")
 endif()
 
-add_definitions(-DHOST_POLICY_PKG_NAME="Microsoft.NETCore.DotNetHostPolicy")
+add_definitions(-DHOST_POLICY_PKG_NAME="runtime.${CLI_CMAKE_PKG_RID}.Microsoft.NETCore.DotNetHostPolicy")
```
gkhanna79: Is this a bug fix? If not, why do we need this change now?

schellap: This is the package that will be looked up in the servicing location.

gkhanna79: I get that - but shouldnt this change have been present earlier as well?

schellap: Sorry, this is a bug fix. The architecture discussion prompted me to think about this.

schellap: Is msil decided, or should it be msft?

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -87,6 +87,7 @@ bool hostpolicy_exists_in_svc(pal::string_t* resolved_dir)
 {
     pal::string_t svc_dir;
     pal::get_default_extensions_directory(&svc_dir);
+    append_path(&svc_dir, _X("msil"));
```
gkhanna79: I think we should call the subfolder containing non-R2R packages as "pkgs", and not "msil", since it will contain packages that are not just MSIL (e.g. CoreCLR).

Please fix the other location where "msil" was committed.

CC @leecow @ericstj

schellap:
This should be fixed broadly with #2452

PR #2450, opened 2016-04-13
Convert TestAssets to `Microsoft.NETCore.App`

### LINKED ISSUE #2438: Move TestAsset Apps from NETStandard.Library to Microsoft.NETCore.App
This is cleaning up Test Debt.

To do this we must generate stub NuGet packages for each of the host packages for each rid they support other than the rid of the current platform.

This is needed for 2 reasons:
- `--infer-runtimes` in nuget restores both x86 and x64 rids on windows platforms, but we only build a host for one of those rids
- `--infer-runtimes` does not override or filter rids which exist in a project.json. As a result of this, our standalone test assets are restored against all rids they list, and we need stub host packages for each of these rids.

cc @eerhardt @piotrpMSFT

brthor:
Fixes #2438

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2450)


brthor:
I'm seeing that rids in a project.json override `--infer-runtimes` 

@emgarten is this expected behavior?

emgarten:
@brthor yes, the inferred runtimes are considered fallbacks and only used if the project.json file does not specify any runtimes under the runtimes section.

brthor:
Thanks @emgarten I am surprised about that, but can work around it.

brthor:
15 test failures, down from 69

brthor:
9 test failures now

brthor:
@pakrym @anurse 

I am seeing some weird failures in the DependencyContextValidator.
```
Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'System, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e, Retargetable=Yes' or one of its dependencies. The system cannot find the file specified.
```

Any ideas what might be going on?

pakrym:
Did you change imports? Looks like you are getting wrong target of Newtonsoft.

brthor:
Yes I probably did, was trying to clean those up while I am going through these projects.

Thanks!

brthor:
What tipped you off that might be the issue?

pakrym:
I've seen this error to many times and System version is 2 which doesn't look right.

brthor:
Alright, so I didn't remove any imports it turns out. Only changed from `netstandard1.5` -> `netcoreapp1.0` in the DependencyContextValidator.

It seems to be when it's using the default DependencyContextLoader, but reading through the code isn't telling me much about what might be going on here.

brthor:
I've switched it back to see if the tests pass without it, and am building on windows so i can debug the failure. Let me know if you have any ideas though.

brthor:
@Sridhar-MS @piotrpMSFT @krwq 

Please review this PR.

Most of it is TfM switches.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -52,12 +74,37 @@ public static BuildTargetResult UpdateBuild(BuildTargetContext c)
 
         // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the
         // Microsoft.NETCore.App package that is created during PackagePkgProjects.
-        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
+        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(CompileTargets.GenerateStubHostPackages), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
```
Sridhar-MS: Can you merge `GenerateStubHostPackages` and `PackagePkgProjects` into one?

scripts/dotnet-cli-build/Utils/BuildVersion.cs
```diff
@@ -13,6 +13,7 @@ public class BuildVersion
         public string VersionSuffix => $"{ReleaseSuffix}-{CommitCountString}";
         public string NuGetVersion => $"{Major}.{Minor}.{Patch}-{VersionSuffix}";
         public string NetCoreAppVersion => $"{Major}.{Minor}.{Patch}-rc2-3{CommitCountString}";
+        public string HostVersion => $"{Major}.{Minor}.1-rc2-{CommitCountString}-00";
```
Sridhar-MS: Why is the host version different from the CLI version? You will have to make changes in the installers too to use this version.

Sridhar-MS: One example - https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/PkgTargets.cs#L169

krwq: ???

scripts/dotnet-cli-build/CompileTargets.cs
brthor: Yea sounds good

scripts/dotnet-cli-build/Utils/BuildVersion.cs
brthor: This is the version of the host nuget package as it exists. (https://dotnet.myget.org/gallery/dotnet-core)

We should unify these...

@schellap any context on why this is different?

TestAssets/TestProjects/DependencyContextFromTool/project.json
```diff
@@ -4,15 +4,16 @@
   },
   "dependencies": {},
   "frameworks": {
-    "netstandardapp1.5": {
+    "netcoreapp1.0": {
       "imports": [
+        "netstandardapp1.5",
```
eerhardt: Do you know why this is needed?

TestAssets/TestProjects/TestProjectToProjectDependencies/src/L21/project.json
```diff
@@ -1,10 +1,10 @@
 {
   "version": "1.0.0-*",
   "dependencies": {
-    "NETStandard.Library": "1.5.0-rc2-24018"
+    "Microsoft.NETCore.App": "1.0.0-rc2-*"
   },
   "frameworks": {
-    "netstandardapp1.5": {
+    "netcoreapp1.0": {
```
eerhardt: Should be netstandard right?  L1 and L2 projects are libraries and should be netstandard.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -52,12 +74,37 @@ public static BuildTargetResult UpdateBuild(BuildTargetContext c)
...
+            var buildVersion = c.BuildContext.Get<BuildVersion>("BuildVersion");
+
+            foreach (var hostPackageId in HostPackages)
+            {
+                foreach (var rid in NetCoreAppRids)
+                {
+                    if (! rid.Equals(currentRid))
+
+                    {
+                        CreateDummyRuntimeNuGetPackage(
```
brthor: @weshaggard we are creating these dummy nuget packages so we can restore standalone apps successfully. 

Given your knowledge of how we upload these to the feed, will they get picked up in that logic?

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -35,6 +36,27 @@ public class CompileTargets
...
+        {
+            "win7-x64",
+            "win7-x86",
+            "osx.10.10-x64",
+            "osx.10.11-x64",
+            "ubuntu.14.04-x64",
+            "centos.7-x64",
+            "rhel.7-x64",
+            "rhel.7.2-x64",
+            "debian.8.2-x64",
```
eerhardt: @MichaelSimons can correct me if I'm wrong but I think we only have debian.8 rid. Not Debian.8.2

scripts/dotnet-cli-build/Utils/BuildVersion.cs
schellap: Host evolves independently, as bug fixes happen. CLI is not a nupkg. These two are different components. Host is part of netcore.app. Don't make both the same and do not use cli version for host.

TestAssets/TestProjects/TestProjectToProjectDependencies/src/L21/project.json
brthor: There are significant changes needed to the incremental tests in order to do this.

I agree that is technically correct, but given the intention of these tests is testing incremental build I don't believe it will have any impact.

krwq:
Make sure that the dummy nupkgs won't get uploaded during the official build. Other than that LGTM :+1:

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -52,12 +74,37 @@ public static BuildTargetResult UpdateBuild(BuildTargetContext c)
...
         // Microsoft.NETCore.App package that is created during PackagePkgProjects.
-        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
+        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(CompileTargets.GenerateStubHostPackages), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
         public static BuildTargetResult Compile(BuildTargetContext c)
         {
             return c.Success();
         }
 
+        [Target]
+        public static BuildTargetResult GenerateStubHostPackages(BuildTargetContext c)
```
eerhardt: Can you stick a comment in here explaining why we do this?

TestAssets/TestProjects/DependencyContextFromTool/project.json
brthor: fairly certain I added this back in explicitly but I can't remember, I'll take it out and we'll see if it fails again

scripts/dotnet-cli-build/CompileTargets.cs
brthor: What value will the comment add?

scripts/dotnet-cli-build/CompileTargets.cs
brthor: I tried to find where I got the debian 8.2 rid, but can't find it anymore. 

After @MichaelSimons confirms, will remove it.

scripts/dotnet-cli-build/Utils/BuildVersion.cs
brthor: When is the host nuget package published? Is it during the build?

schellap: Post build on all platforms with a VSO definition and a util program that @weshaggard authored.

schellap: Sorry that turned out to be too convoluted.

1. All build machines produce nupkg
2. A separate definition scans azure blob storage and pushes the full list of packages.

scripts/dotnet-cli-build/CompileTargets.cs
eerhardt: It isn't really a naming issue, it is a "why" issue.  Someone coming into this code would easily ask themselves, "why in the world are they generating a "stub host package"?". This would tell them the answer.

A link to an issue would suffice.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -305,6 +353,52 @@ public static BuildTargetResult CompileStage2(BuildTargetContext c)
...
+            Directory.CreateDirectory(tempPjDirectory);
+
+            var tempPjFile = Path.Combine(tempPjDirectory, "project.json");
+            var tempSourceFile = Path.Combine(tempPjDirectory, "Program.cs");
+
+            File.WriteAllText(tempPjFile, projectJson.ToString());
+            File.WriteAllText(tempSourceFile, programCs.ToString());
+
+            var oldDir = Directory.GetCurrentDirectory();
+            Directory.SetCurrentDirectory(tempPjDirectory);
```
eerhardt: Would it work better to set the directory on each command, like we normally do?

 dotnet.Restore("params")
  **.WorkingDirectory(tempPjDirectory)**
   .Execute()

scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -96,10 +96,9 @@ public static BuildTargetResult RestoreTestAssetProjects(BuildTargetContext c)
                 .Execute()
                 .EnsureSuccessful();
 
-            // The 'ProjectWithTests' is a portable test app. Cannot call --infer-runtimes on it, since on win x64 machines,
-            // the x86 runtime is being inferred, and there are no x86 DotNetHost packages
             dotnet.Restore(
```
eerhardt: This project can now be folded back into TestAssets\TestProjects.  The reason it was separate is because of the problem you are solving. Now there is no reason for it to be separate anymore.

Although it isn't critical for your work to do it, and may just make it longer.  So if you aren't going to do this, can you either open an issue, or put a TODO here?

scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -178,7 +165,7 @@ public static BuildTargetResult BuildTestAssetPackages(BuildTargetContext c)
                 {
                     "netstandard1.5",
                     "netstandard1.3",
-                    "netstandardapp1.5"
+                    "netcoreapp1.0"
```
eerhardt: Here's a classic case of having "why" we are doing things listed in comments.  I look at this code and I ask myself, "why are we doing this?", but there is no reason listed.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: Cool, I will add a link to an issue.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: Yes it would be better 😄 will do

scripts/dotnet-cli-build/CompileTargets.cs
weshaggard: Yes every *.nupkg under Dirs.Corehost gets pushed to our Azure storage (see https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/dotnet-cli-build/PublishTargets.cs#L100) and then will end up getting published because we just publish all *.nupkg files we find in the azure storage for a given build and push them to nuget.

We should put these dummy packages in another location or filter them out. I've not read through this full review yet but I'm curious why we need these at all, it is probably not a good idea to create them at all.

scripts/dotnet-cli-build/TestTargets.cs
brthor: The comment above is intended to answer that. It's to be able to ignore net451 build failures on non-windows. 

If we don't do it this way, a net451 failure will fail builds for all tfms.

Anyways, I see your point, the comment could be better. 
I do think better factoring is the more ideal answer to that though.

eerhardt: I still don't think I get it.  We build all these packages on all these frameworks just so we can ignore net451 build failures on non-windows?

Maybe it would be better if the tests that needed/used these projects would do the building during the test, instead of building them during this step in the build?

scripts/dotnet-cli-build/CompileTargets.cs
weshaggard: I have that exact question while reviewing this... so I agree we should add some comments as to why we are doing this.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: Okay, we will need to do some filtering here, or add another fallback source when we do our restores. 

The dummy packages are needed so that we can restore all of our tests against the Microsoft.NETCore.App metapackage we just built. Our standalone tests require these dummy packages to be present, and on any windows platform `--infer-runtimes` causes nuget to look for both x86 and x64 packages, which is initially how this need started.

eerhardt: Unfortunately given

1. That Microsoft.NETCore.App is built during this same build
2. That we can only build 1 runtime specific version of DotNetHost during the build
3. NuGet wants to restore all runtimes for a project.json

We have to generate these 'dummy' packages.  I've tried to change NuGet with https://github.com/NuGet/Home/issues/2534, but that doesn't seem to be going anywhere. Hopefully in RTM we move DotNetHost and Microsoft.NETCore.App packages out of dotnet/cli, then we won't need to do this anymore.

weshaggard: Can we limit this to just the x86 or x64 windows packages then and use --infer-runtimes everywhere else?

eerhardt:
Overall this looks good, @brthor.  Thanks for going through all this trouble, our tests look like they are in a lot better shape now.

scripts/dotnet-cli-build/CompileTargets.cs
weshaggard: ultimately I think we need to add back the --runtimes switch to dotnet restore, do you know if that is more or less work then doing this stub generation and dealing with that?

scripts/dotnet-cli-build/TestTargets.cs
brthor: We build the projects on all of these frameworks so we can ignore net451 failures on non-windows.

The package is only built once, with or without net451 binaries depending on platform.

The test assets that use these packages need to be restored against them just after this step, so it would be complex (and ugly) to orchestrate that in individual tests.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: @weshaggard `--infer-runtimes` does not win over rids in the project.json, so we cannot do that currently.

Completely agreed on --runtimes being the better solution.

The stub generation is working, the only concern left is publishing appropriately. The nuget work is probably more.

eerhardt: > ultimately I think we need to add back the --runtimes switch to dotnet restore, do you know if that is more or less work then doing this stub generation and dealing with that?

@yishaigalatzer @piotrpMSFT @davidfowl and I all had this conversation last week.

@yishaigalatzer's response was "it is a very easy task to add back in --runtimes" but he wants us (.NET and ASP.NET) to come to consensus on this topic, instead of putting it in and taking it out, and putting it back in.
There were also concerns that no other customer would need this, and that this is a CLI-only issue.

scripts/dotnet-cli-build/TestTargets.cs
brthor: Given that I have seen this path being referenced in test cases which will need to change, I'd like to put this for a separate PR.

I will put a todo and file an issue.

scripts/dotnet-cli-build/TestTargets.cs
brthor: The problem is that if we don't loop through frameworks like this, and just call build, the first build failure (for net451) causes the other tfm builds not to happen and we get empty packages.

It is fairly ugly and needs cleanup I agree, but it is outside the scope of this PR.

eerhardt: > It is fairly ugly and needs cleanup I agree, but it is outside the scope of this PR.

Totally agree outside of this PR. My uber point was the reason for "why" comments.

scripts/dotnet-cli-build/CompileTargets.cs
schellap: +1 for the `--runtime` switch (corert repo uses this to restore the ILC flavor we just built (x86 or x64) and we haven't upgraded to latest CLI)

TestAssets/TestProjects/TestAppWithLibrary/TestApp/project.json
```diff
@@ -9,11 +9,9 @@
       "target": "project",
       "version": "1.0.0-*"
     },
-    "NETStandard.Library": "1.5.0-rc2-24018"
+    "Microsoft.NETCore.App": "1.0.0-rc2-*"
```
davidfowl: why the wildcard versions?

brthor: We restore these against the package that we built earlier in the build. 

The version is determined at build time.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: I'm going to add a small explanation with a link to an issue with more information

scripts/dotnet-cli-build/TestTargets.cs
brthor: I went ahead and moved this, changed the tests, and removed where it was getting built separately.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: done.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: couldn't find this rid anywhere in the repo so I removed it

TestAssets/TestProjects/TestProjectToProjectDependencies/src/L21/project.json
brthor: Went ahead and implemented these changes. Required piping through the full framework name to the command result assertions that exist for testing whether build output actually compiled a project.

TestAssets/TestProjects/DependencyContextFromTool/project.json
brthor: removed.

TestAssets/TestProjects/TestProjectToProjectDependencies/src/L21/project.json
eerhardt: If this is causing problems for your PR, you can feel free to ignore or log a separate issue for this comment.  It was wrong before (using netstandardapp), so you aren't making it any worse by targeting netcoreapp.

scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -91,25 +94,18 @@ public static BuildTargetResult RestoreTestAssetProjects(BuildTargetContext c)
...
                 "--infer-runtimes",
                 "--fallbacksource", Dirs.TestPackages,
-                "--fallbacksource", Dirs.Corehost)
+                "--fallbacksource", Dirs.Corehost,
+                "--fallbacksource", Dirs.CorehostDummyPackages)
                 .WorkingDirectory(Path.Combine(c.BuildContext.BuildDirectory, "TestAssets", "TestProjects"))
                 .Execute()
                 .EnsureSuccessful();
 
-            // The 'ProjectWithTests' is a portable test app. Cannot call --infer-runtimes on it, since on win x64 machines,
```
eerhardt: Nice!

Just a heads up, this is going to conflict with #2527, where @pranavkm added more tests to this `ProjectWithTests` directory. Last one in will have to merge ;).

scripts/dotnet-cli-build/CompileTargets.cs
MichaelSimons: @brthor, @eerhardt is correct, we only have the debian.8 rid.

scripts/dotnet-cli-build/TestTargets.cs
piotrpMSFT: I agree with the intent, but would prefer to see a refactoring. Comments are impossible to keep up to date.

scripts/dotnet-cli-build/TestTargets.cs
brthor: thanks!

scripts/dotnet-cli-build/TestTargets.cs
brthor: Agreed on the refactoring direly needed.

brthor:
I took time last night to go over a couple areas that were not quite right and fix them (incremental tests, ProjectWithTests), I figure we may as well get all tests inline with the new model. 

Please take a minute to go over the PR again, test failures should be solved in the latest push, and I am hoping to merge this soon.

test/dotnet-build.Tests/IncrementalTestBase.cs
```diff
@@ -12,6 +12,9 @@ namespace Microsoft.DotNet.Tools.Builder.Tests
 {
     public class IncrementalTestBase : TestBase
     {
+        protected readonly string _libraryFrameworkFullName = ".NETStandard,Version=v1.5";
+        protected readonly string _appFrameworkFullName = ".NETCoreApp,Version=v1.0";
```
eerhardt: `const` instead?

src/dotnet-compile-fsc/Program.cs
```diff
@@ -85,6 +85,7 @@ public static int Main(string[] args)
             bool targetNetCore = 
                 commonOptions.Defines.Contains("DNXCORE50") ||
                 commonOptions.Defines.Where(d => d.StartsWith("NETSTANDARDAPP1_")).Any() ||
```
eerhardt: Can/should this be removed?

test/dotnet-build.Tests/IncrementalTestBase.cs
brthor: I prefer readonly for no good reason 😄 , is there a meaningful difference here?

eerhardt:
:shipit:

src/dotnet-compile-fsc/Program.cs
brthor: Yeah it probably should, but I'll let the FSC guys take care of that, they have already begun taking this to a new repo.

FYI @enricosada

test/dotnet-build.Tests/IncrementalTestBase.cs
eerhardt: The meaningful difference is that every instance of IncrementalTestBase will get a reference to this string, which bloats the object. And every time you access it, it goes back to the reference to get the value. When a member is marked `const` each object won't get its own reference. Instead the constant will get in-lined wherever it is used.

Obviously this isn't a huge deal in tests. In general .NET programming it is better to use `const` when possible for non-public/protected constants. When you have a public/protected constant that could possibly change in the future, then you should use readonly instead of const.

brthor: Oh nice, I did not know this. Thanks for the explanation!

Since it's not critical I won't kick off another build for this change, but if CI comes back red will make the change.

src/dotnet-compile-fsc/Program.cs
enricosada: I'll do that. I am testing the new package with that fix.

Issue #212, opened 2015-11-17
dotnet should support Debian distros


MichaelSimons:
# Repro Steps
1. Create the following Dockerfile.

FROM buildpack-deps:jessie-scm

RUN echo "deb [arch=amd64] http://apt-mo.trafficmanager.net/repos/dotnet/ trusty main" > /etc/apt/sources.list.d/dotnetdev.list \
    && apt-key adv --keyserver apt-mo.trafficmanager.net --recv-keys 417A0893 \
    && echo "deb [arch=amd64] http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.6 main" > /etc/apt/sources.list.d/llvm.list \
    && wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key|apt-key add - \
    && apt-get update \
    && apt-get install -y dotnet

RUN git clone https://<GitUsername>:<GitToken>@github.com/dotnet/cli.git cli

WORKDIR /cli/test/TestLibrary

RUN dotnet restore

WORKDIR /cli/test/TestApp

RUN dotnet restore

2. Replace <GitUsername> and <GitToken> with your credentials and build the docker image (using a Ubuntu 14.04-x64 machine) -  docker build -t restoreBug .
3. Run the built image - docker run -it restoreBug
4. Inspect the project.lock.json file that dotnet restore created.  Notice what runtimes it pulled packages for.
5.  Notice the error when you run dotnet run

# Actual Results
Notice that the two runtimes it pulled packages for are 8.debian-x64 and 8.debian-x86.   When you run step 6 you get the following error "Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'System.Console, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified."  To correct this behavior you can specify ubuntu.14.04-x64 as the runtime for commands like restore and publish.  Not sure of a workaround for run though.

# Expected Results
Previously whenever I ran these steps the dotnet was defaulting the runtime to as Ubuntu.14.04-x64.  It was not necessary to specify the --runtime parameter.

## Note
At the time this bug was logged, the version of the dotnet debian package used is 0.0.1-alpha-20151115-171248-1

brthor:
Is this a debian docker image?

If so, it seems like 8.debian-x64 as the RID might be the correct behavior and we need to account for it in other places.

Petermarcu:
[...]

PR #2472, opened 2016-04-13
Add support for Debian CLI builds

MichaelSimons:
These are the necessary min bar changes needed to produce Debian CLI builds.  This does not include the changes to produce installers/Debian packages as this is not necessary for RC2 given the time constraints we are facing.

Two builds have been run on these changes in order to get the Debian artifacts to do some verification with.  The interesting artifacts are:

- https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-dev-debian-x64.latest.tar.gz which is currently the same as https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/1.0.0-rc2-002365/dotnet-dev-debian-x64.1.0.0-rc2-002365.tar.gz
- Since this isn't official yet, the associated packages weren't uploaded to NuGet therefore you will need use the fallbacksource restore option with the associated nupkg such as https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/1.0.0-rc2-002365/Microsoft.NETCore.App.1.0.0-rc2-3002365.nupkg

#212 
dotnet/dotnet-docker#11

The remaining work after this PR is merged:
- point my VSO build definition that is targeting my fork to cli rel/1.0.0
- update Wes' PublishTargets to publish the Debian packages
- update the readme with the Debian drop links

Post RC2 remaining work
- Produce Debian packages 

@brthor @piotrpMSFT @Sridhar-MS

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2472)


Sridhar-MS:
@MichaelSimons Couple of questions
- Does your change produce debian tarballs and publish them to azure blob storage?
- If not then the debian readme links will be outdated and fall behind other OSes?
- If things are working then why don't we add debian builds to Jenkins?

TestAssets/TestProjects/PortableTests/StandaloneApp/project.json
```diff
@@ -21,7 +21,7 @@
     "ubuntu.14.04-x64": {},
     "centos.7-x64": {},
     "rhel.7.2-x64": {},
-    "debian.8.2-x64": {}
+    "debian.8-x64": {}
```
brthor: This should probably also be changed in the CrossPublishStandaloneApp

MichaelSimons: @brthor Absolutely, I didn't see that change come in yesterday.

scripts/docker/debian/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+# This could become a "microsoft/coreclr" image, since it just installs the dependencies for CoreCLR (and stdlib)
+RUN echo "deb http://llvm.org/apt/jessie/ llvm-toolchain-jessie-3.6 main"  | tee /etc/apt/sources.list.d/llvm.list && \
+    curl http://llvm.org/apt/llvm-snapshot.gpg.key | apt-key add - && \
+    apt-get update && apt-get -qqy install\
+	libc6 \
+	libedit2 \
+	libffi6 \
+	libgcc1 \
+	libicu52 \
+	liblldb-3.6 \
```
brthor: @ellismg I remember having some conversation about this library on debian but can't recall what you had said.

IIRC this library isn't available from the default feeds on debian:jessie. Is there an alternate library we could pull down to satisfy this requirement?

ellismg: I assume you are talking about `liblldb-3.6`?

This library is used when you want to debug the runtime using LLVM and use SOS (which is a helper plugin that understands how to do runtime specific things like dump method tables, converted the addresses of jit'ed methods back to what method they came from, get information about the GCHeap, etc).

It's not used at runtime by CoreCLR.  If this package doesn't exist, CoreCLR will still work.

MichaelSimons: @Sridhar-MS 
- Yes with these changes debian tarballs are produced by the builds.  As I eluded to in my summary comments, I have a VSO build definition hooked up to my fork that is uploading these to azure.  This build definition will need to be modified when this PR is merged.
- Jenkins - I wasn't sure the Jenkins integration was necessary for RC2 as I listed in my summary.  Regardless can debian builds be added to Jenkins prior to merging this PR?

Sridhar-MS: > Jenkins - I wasn't sure the Jenkins integration was necessary for RC2 as I listed in my summary. Regardless can debian builds be added to Jenkins prior to merging this PR?

If we are generating debian tarballs regularly then we need Jenkins CI to make sure no one breaks it. It is just a oneline change at https://github.com/dotnet/cli/blob/rel/1.0.0/netci.groovy#L12 assuming the builds are started the same way as here - https://github.com/dotnet/cli/blob/rel/1.0.0/netci.groovy#L39. The change is to replace `Ubuntu:x64:Debug` with `Debian8.2:x64:Debug`. 

Jenkins will not picked it up immediately, but after merging the netci.groovy changes any new PR will include the Debian build check.

src/corehost/build.sh
```diff
@@ -18,7 +18,7 @@ init_distro_name_and_rid()
         export __rid_plat=rhel.7
     elif [ "$(cat /etc/*-release | grep -cim1 debian)" -eq 1 ]; then
         export __distro_name=debian
-        export __rid_plat=
+        export __rid_plat=debian.8
     else
```
schellap: Just making sure that this contains the value of: `runtimes\{debian.*}\native\libhostpolicy.so`

MichaelSimons: @schellap, I'm not 100% clear on what you are asking.  I verified a runtime.debian.8-x64.Microsoft.NETCore.DotNetHostPolicy.1.0.1-rc2-002365-00.nupkg package is being produced which contains a runtimes\debian.8-x64\native\libhostpolicy.so.

netci.groovy
```diff
@@ -9,7 +9,7 @@ def project = GithubProject
 def branch = GithubBranchName
 def isPR = true
 
-def platformList = ['Ubuntu:x64:Debug', 'Ubuntu:x64:Release', 'OSX:x64:Release', 'Windows_NT:x64:Release', 'Windows_NT:x86:Debug', 'RHEL7.2:x64:Release', 'CentOS7.1:x64:Debug']
+def platformList = ['Debian8.2:x64:Debug', 'Ubuntu:x64:Release', 'OSX:x64:Release', 'Windows_NT:x64:Release', 'Windows_NT:x86:Debug', 'RHEL7.2:x64:Release', 'CentOS7.1:x64:Debug']
```
MichaelSimons: @Sridhar-MS, Added Jenkins as you suggested - thanks

scripts/docker/debian/Dockerfile
MichaelSimons: @brthor did you want me to remove `liblldb-3.6` or leave it?

brthor: @MichaelSimons 
I would like to remove it in the long run, because it requires some additional steps to install on debian Jessie. We'd need to look at how that impacts the debugging experience though.

I don't think it's going to be an issue until we make a debian package for the debian distro, so let's leave it for now.

ellismg: The impact to the debugging experience is it makes it harder for runtime developers and escalation engineers to debug runtime issues.  While you can use the plugin to debug any sort of managed code, it's a very rough experience and something we wouldn't expect customers to do.

It not being present will not break VS debugging, it's only needed when you debug the runtime with lldb and try to load the plugin.

src/corehost/build.sh
schellap: Perfect, that's what I asked about. The relative path gets wired into the host code at compile time through -D and the relative path should match package sub-path.

tmds:
README.md still has Debian 8**.2** and CentOS 7**.1**.

PR #2491, opened 2016-04-14
Account for packages not built with serviceable attribute

schellap:
@gkhanna79 @ericstj, CoreCLR packages are not authored with the "serviceable" attribute, the host optimizes a probe for anything that is not marked serviceable from the servicing location.

I realize that it might be too much to expect package authors to put this flag, so I made the host ignore this flag and not perform a probe optimization.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2491)

### DIFF
--- a/src/corehost/cli/deps_resolver.cpp
+++ b/src/corehost/cli/deps_resolver.cpp
@@ -427,7 +427,7 @@ void deps_resolver_t::resolve_tpa_list(
         else
         {
             // FIXME: Consider this error as a fail fast?
-            trace::verbose(_X("Error: Could not resolve path to assembly: [%s, %s, %s]"), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+            trace::verbose(_X("Warning: Could not resolve path to assembly: [%s, %s, %s]"), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
         }
     };
     
--- a/src/corehost/cli/args.h
+++ b/src/corehost/cli/args.h
@@ -57,12 +57,18 @@ struct probe_config_t
 
     static probe_config_t svc_ni(const pal::string_t& dir, bool patch_roll_fwd, bool prerelease_roll_fwd)
     {
-        return probe_config_t(dir, false, patch_roll_fwd, prerelease_roll_fwd, nullptr, true, true);
+        bool runtime_assets_only = true;  // Since this is ngen based package location, assets should be managed.
+        bool serviceable_only = false;    // Ignore deps specified "serviceable" flag since all packages
+                                          // are not expected to be authored with this attribute.
+        return probe_config_t(dir, false, patch_roll_fwd, prerelease_roll_fwd, nullptr, serviceable_only, runtime_assets_only);
     }
 
     static probe_config_t svc(const pal::string_t& dir, bool patch_roll_fwd, bool prerelease_roll_fwd)
     {
-        return probe_config_t(dir, false, patch_roll_fwd, prerelease_roll_fwd, nullptr, true, false);
+        bool runtime_assets_only = false; // Include both native and managed assemblies.
+        bool serviceable_only = false;    // Ignore deps specified "serviceable" flag since all packages
+                                          // are not expected to be authored with this attribute.
+        return probe_config_t(dir, false, patch_roll_fwd, prerelease_roll_fwd, nullptr, serviceable_only, runtime_assets_only);
     }
 
     static probe_config_t cache_ni(const pal::string_t& dir)


ericstj:
The attribute is not "on the package" its on the assembly.  Serviceable attribute is added automatically by buildtools for all of our assemblies, I believe it is nuget that is not setting it in the lockfile.  As far as I can tell it never reads it from the assembly: https://github.com/NuGet/NuGet.Client/search?utf8=%E2%9C%93&q=Serviceable.

Previously this was the opt-in/opt-out mechanism so removing this means everyone gets servicing.  Are we OK with that?

gkhanna79:
I concur with @ericstj that we should not blindly turn on servicing for every package, unless it is marked as serviceable.

That said, @ericstj, if Nuget is not marking our packages as serviceable, then how are we intending to service the packages?

CC @weshaggard

ericstj:
That's my point, we need to file a bug on NuGet.  Maybe for RC2 it's reasonable to turn it on for everything but for RTM we need to get the right state plumbed through.

schellap:
@gkhanna79, yes ideally the serviceable flag has exactly that purpose to hint the host. Our packages are marked serviceable, is it guaranteed that all our partners are aware of this flag (the folks who synced with us to opt into our servicing policy)?


@anurse, this is a nuget bug if @ericstj indicates that all our assemblies are marked serviceable, the deps file for coreclr doesn't mark it so.

anurse:
It sounds like this is definitely a NuGet bug. DNX would scan the assemblies in the package for that attribute and mark the package as serviceable in the lock file. It appears we'd need to port that code over to NuGet.

gkhanna79:
CC @leecow 

@ericstj Where is the Serviceable attribute specified for mscorlib? I dont see it in the open build.

gkhanna79:
LGTM.

gkhanna79:
@yishaigalatzer Is this a known issue for Nuget that is fixed (or, will be fixed) for RTM?

ericstj:
@gkhanna79 here's where it is set: https://github.com/dotnet/buildtools/blob/e22e82015f66d24c7e43ed93501cac6e6e9e169a/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Build.Common.targets#L27.

You need to make a change in the CoreCLR repro to match corefx:
https://github.com/dotnet/corefx/blob/4c403bca0e33a74423a955fcfb1461e69b4b4a9d/dir.props#L26

gkhanna79:
I will get a PR in to fix this for CoreCLR. @ericstj Can you confirm if this is a bug in Nuget and other packages are affected as well, or is this CoreCLR specific only? The fix can then scoped appropriately.

ericstj:
@gkhanna79 confirmed, if I upgrade to RC3 packages (required for https://github.com/dotnet/corefx/commit/d64277feac26a9dd0611c776589b9a7e55b5f619) I still don't see NuGet adding Serviceable attribute in the project.lock.json.  This is indeed a NuGet bug.

gkhanna79:
@ericstj Thank you for the confirmation. I am good with this PR - how about you @ericstj?

yishaigalatzer:
That doesn't sound like a bug, more like a missing feature/underspec'd feature. Please file a bug on nuget, also please let us know if this is an rc2 blocking issue?

gkhanna79:
@ericstj Once the dir.props is updated, should I see the attribute in mscorlib.dll? I ask because I dont see it after making the change.

weshaggard:
mscorlib may not be using our assembly version generation targets, I would need to double check. @joperezr should be able to confirm that.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -427,7 +427,7 @@ void deps_resolver_t::resolve_tpa_list(
         else
         {
             // FIXME: Consider this error as a fail fast?
-            trace::verbose(_X("Error: Could not resolve path to assembly: [%s, %s, %s]"), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
+            trace::verbose(_X("Warning: Could not resolve path to assembly: [%s, %s, %s]"), entry.library_name.c_str(), entry.library_version.c_str(), entry.relative_path.c_str());
```
ericstj: Why did you change this to a warning?

joperezr:
> mscorlib may not be using our assembly version generation targets

It does use our assembly version targets, but it is not using the common targets which are the ones that contain the attribute. This is because mscorlib [directly references the versioning targets](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/mscorlib.csproj#L185) instead of referencing them through the common targets.

src/corehost/cli/deps_resolver.cpp
schellap: Only printed in logs. When turning on logging, _._ gets error and compilation looks for Error string. We punted _._ probe fixing to RTM. There is an active issue on that.

gkhanna79:
@schellap Do you want to create an issue for this and send an email about it for commit?

schellap:
Closing, as workaround https://github.com/dotnet/cli/issues/2569 belongs where the bug is.

PR #2516, opened 2016-04-15
Fix download link for RHEL 7.2 .NET Core Binaries

naamunds:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2516)


Sridhar-MS:
:shipit:

Issue #2474, opened 2016-04-13
CLI Binaries are Not Pre-Compiled


brianrob:
CLI binaries JIT for over 1 second when doing a dotnet build.  There is also significant JIT time in dotnet new and dotnet run.

This can be resolved by pre-compiling the CLI binaries using crossgen.

piotrpMSFT:
We're going to give this a shot, but may need to revert the change if we see stability issues.

brianrob:
Awesome.  Thanks!

piotrpMSFT:
Fixed by #2517

PR #2517, opened 2016-04-15
Enable crossgen for CLI SDK binaries.

Sridhar-MS:
TODO: Write tests to make sure assemblies are crossgened.

@piotrpMSFT @brthor @brianrob 

Fixes https://github.com/dotnet/cli/issues/2474

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2517)

### DIFF
--- a/scripts/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs
+++ b/scripts/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs
@@ -36,11 +36,6 @@ namespace Microsoft.DotNet.Cli.Build.Framework
                 cmd.CaptureStdErr().CaptureStdOut();
             }
 
-            if (env == null)
-            {
-                env = new Dictionary<string, string>();
-            }
-
             var result = cmd.Environment(env).Execute();
 
             result.EnsureSuccessful();
--- a/scripts/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs
+++ b/scripts/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs
@@ -5,16 +5,17 @@ namespace Microsoft.DotNet.Cli.Build.Framework
     public static class BuildHelpers
     {
         public static int ExecInSilent(string workingDirectory, string command, params string[] args) => ExecInSilent(workingDirectory, command, (IEnumerable<string>)args);
-        public static int ExecInSilent(string workingDirectory, string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory, silent: true);
+        public static int ExecInSilent(string workingDirectory, string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory, silent: true, env: null);
 
         public static int ExecIn(string workingDirectory, string command, params string[] args) => ExecIn(workingDirectory, command, (IEnumerable<string>)args);
-        public static int ExecIn(string workingDirectory, string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory, silent: false);
+        public static int ExecIn(string workingDirectory, string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory, silent: false, env: null);
 
         public static int ExecSilent(string command, params string[] args) => ExecSilent(command, (IEnumerable<string>)args);
-        public static int ExecSilent(string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory: null, silent: true);
+        public static int ExecSilent(string command, IEnumerable<string> args) => ExecSilent(command, args, env: null);
+        public static int ExecSilent(string command, IEnumerable<string> args, IDictionary<string, string> env) => ExecCore(command, args, workingDirectory: null, silent: true, env: null);
 
         public static int Exec(string command, params string[] args) => Exec(command, (IEnumerable<string>)args);
-        public static int Exec(string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory: null, silent: false);
+        public static int Exec(string command, IEnumerable<string> args) => ExecCore(command, args, workingDirectory: null, silent: false, env: null);
 
         public static Command Cmd(string command, params string[] args) => Cmd(command, (IEnumerable<string>)args);
         public static Command Cmd(string command, IEnumerable<string> args)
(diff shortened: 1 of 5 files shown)


scripts/dotnet-cli-build/Utils/PEUtils.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public static bool HasMetadata(string pathToFile)
+        {
+            try
+            {
+                using (var inStream = File.OpenRead(pathToFile))
+                {
+                    using (var peReader = new PEReader(inStream))
+                    {
+                        return peReader.HasMetadata;
```
brthor: Nice! HasMetadata only returns true for managed binaries?

Sridhar-MS: Yup

scripts/dotnet-cli-build/Utils/PEUtils.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection.PortableExecutable;
+
+namespace Microsoft.DotNet.Cli.Build
+{
+    public class PEUtils
```
eerhardt: can this be a static class?

scripts/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs
```diff
@@ -5,38 +5,47 @@ namespace Microsoft.DotNet.Cli.Build.Framework
...
                 cmd.CaptureStdErr().CaptureStdOut();
             }
-            var result = cmd.Execute();
+
+            if (env == null)
+            {
+                env = new Dictionary<string, string>();
+            }
+
+            var result = cmd.Environment(env).Execute();
```
eerhardt: Maybe cmd.Environment should allow passing in `null` and just no-oping?  That way we don't waste creating an empty dictionary for no reason.

scripts/dotnet-cli-build/Utils/Crossgen.cs
```diff
@@ -42,9 +59,59 @@ public static string GetCrossgenPathForVersion(string coreClrVersion)
...
+                var paths = new List<string>() { sharedFxPath, pathToAssemblies };
+                paths.AddRange(addtionalPaths);
+                var platformAssembliesPaths = string.Join(";", paths.Distinct());
+
+                IList<string> crossgenArgs = new List<string> {
+                    "-readytorun", "-in", file, "-out", tempPathName,
+                    "-platform_assemblies_paths", platformAssembliesPaths
+                };
+
+                var env = new Dictionary<string, string>()
```
eerhardt: You could declare this dictionary outside of the loop, since it stays the same for all iterations.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -530,6 +532,8 @@ private static BuildTargetResult CompileCliSdk(BuildTargetContext c, DotNetCli d
             File.Delete(compilersDeps);
             File.Delete(compilersRuntimeConfig);
 
+            CrossgenUtil.CrossgenDirectory(c, outputDir);
```
eerhardt: Does crossgenn'ing Stage1 take a lot of time?  If we only crossgenn'd Stage2, would that save some time in our build?

eerhardt:
This looks really good @Sridhar-MS.  I added a few minor comments and questions.

scripts/dotnet-cli-build/Utils/PEUtils.cs
Sridhar-MS: Fixed.

scripts/Microsoft.DotNet.Cli.Build.Framework/BuildHelpers.cs
Sridhar-MS: Fixed.

scripts/dotnet-cli-build/Utils/Crossgen.cs
Sridhar-MS: Fixed.

scripts/dotnet-cli-build/CompileTargets.cs
Sridhar-MS: @eerhardt I like to keep stage1 and stage2 similar since we sort of use it for testing CLI itself.   There are plenty of other optimizations we can do
- crossgen in parallel
- run tests in parallel, actually was experimenting with this. The time for test execution is down to less than 5 mins on my Macbook Pro.

brthor: @piotrpMSFT originally set up the crossgenning of stage1 and we saw dramatic performance improvement.

scripts/dotnet-cli-build/Utils/Crossgen.cs
```diff
@@ -42,9 +59,61 @@ public static string GetCrossgenPathForVersion(string coreClrVersion)
...
+            // else pass the current directory of assembly as input to crossgen.
+            var addtionalPaths = Directory.GetDirectories(pathToAssemblies, "*", SearchOption.AllDirectories).ToList();
+            var paths = new List<string>() { sharedFxPath, pathToAssemblies };
+            paths.AddRange(addtionalPaths);
+            var platformAssembliesPaths = string.Join(Path.PathSeparator.ToString(), paths.Distinct());
+
+            var env = new Dictionary<string, string>()
+            {
+                // disable partial ngen
+                { "COMPLUS_ZapDisable", "0" }
```
Sridhar-MS: @brianrob is this the right env variable for disabling partial ngen?

scripts/dotnet-cli-build/CompileTargets.cs
eerhardt: I'm fine with this.  I just wanted to call it out as something that may be unnecessary.

As our build keeps getting longer and does more things, we just need to ensure we aren't doing unnecessary things.

eerhardt:
:shipit:

scripts/dotnet-cli-build/Utils/Crossgen.cs
brianrob: Yes, that's the right one.

Issue #1574, opened 2016-02-25
Remove compile as a top level command


davidfowl:
It's no longer supposed to be used by anyone but `dotnet build` and since it's intrinsic, shouldn't be replaceable.

/cc @piotrpMSFT @blackdwarf

blackdwarf:
@davidfowl agreed, people are getting confused. Who do we assign this to?

davidfowl:
I'll do it. Let's see what breaks!

blackdwarf:
Go for it! Ya canna fix it unless its broken. :)

davidfowl:
Done

PR #2518, opened 2016-04-15
Invoke compile-csc in-proc

### LINKED ISSUE #2542: dotnet compile should invoke compile-csc in the same process
See https://github.com/dotnet/cli/issues/2483

This issue is for the first fix in order to improve the performance of dotnet build.

We should invoke compile-csc in the same process instead of spinning up a new process. This saves 1 second on my machine when compiling the "dotnet new" project.

eerhardt:
Since processes are expensive, suppress spawning a new process when dotnet-compile is invoking dotnet-compile-csc.

Fix #2542

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2518)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
@@ -1,0 +1,141 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+
+namespace Microsoft.DotNet.Cli.Utils
+{
+    /// <summary>
+    /// A Command that is capable of running in the current process.
+    /// </summary>
+    public class BuiltInCommand : ICommand
+    {
+        private readonly IEnumerable<string> _commandArgs;
+        private readonly Func<string[], int> _builtInCommand;
+        private readonly StreamForwarder _stdOut;
+        private readonly StreamForwarder _stdErr;
+
+        public string CommandName { get; }
+        public string CommandArgs => string.Join(" ", _commandArgs);
+
+        public BuiltInCommand(string commandName, IEnumerable<string> commandArgs, Func<string[], int> builtInCommand)
+        {
+            CommandName = commandName;
+            _commandArgs = commandArgs;
+            _builtInCommand = builtInCommand;
+
+            _stdOut = new StreamForwarder();
+            _stdErr = new StreamForwarder();
+        }
+
+        public CommandResult Execute()
+        {
+            TextWriter originalConsoleOut = Console.Out;
+            TextWriter originalConsoleError = Console.Error;
+
+            try
+            {
+                // redirecting the standard out and error so we can forward
+                // the output to the caller
+                using (BlockingMemoryStream outStream = new BlockingMemoryStream())
+                using (BlockingMemoryStream errorStream = new BlockingMemoryStream())
+                {
+                    Console.SetOut(new StreamWriter(outStream) { AutoFlush = true });
+                    Console.SetError(new StreamWriter(errorStream) { AutoFlush = true });
+
+                    // Reset the Reporters to the new Console Out and Error.
+                    Reporter.Reset();
+
+                    Thread threadOut = _stdOut.BeginRead(new StreamReader(outStream));
+                    Thread threadErr = _stdErr.BeginRead(new StreamReader(errorStream));
+
+                    int exitCode = _builtInCommand(_commandArgs.ToArray());
+
+                    outStream.DoneWriting();
+                    errorStream.DoneWriting();
+
+                    threadOut.Join();
+                    threadErr.Join();
+             
+                    // fake out a ProcessStartInfo using the Muxer command name, since this is a built-in command
+                    ProcessStartInfo startInfo = new ProcessStartInfo(new Muxer().MuxerPath, $"{CommandName} {CommandArgs}");
+                    return new CommandResult(startInfo, exitCode, null, null);
+                }
+            }
+            finally
+            {
+                Console.SetOut(originalConsoleOut);
+                Console.SetError(originalConsoleError);
+
+                Reporter.Reset();
+            }
+        }
+
+        public ICommand OnOutputLine(Action<string> handler)
+        {
+            if (handler == null)
+            {
+                throw new ArgumentNullException(nameof(handler));
+            }
+
+            _stdOut.ForwardTo(writeLine: handler);
+
+            return this;
+        }
+
+        public ICommand OnErrorLine(Action<string> handler)
+        {
+            if (handler == null)
+            {
+                throw new ArgumentNullException(nameof(handler));
+            }
+
+            _stdErr.ForwardTo(writeLine: handler);
+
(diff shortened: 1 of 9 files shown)


eerhardt:
This change is ready to be reviewed.

/cc @piotrpMSFT @Sridhar-MS @brthor @cdmihai @anurse 

/bcc @Petermarcu

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Cli.Utils
+{
+    /// <summary>
+    /// A Command that is capable of running in the current process.
+    /// </summary>
+    public class BuiltInCommand : ICommand
```
brthor: How come we don't subclass command to remove the duplication of code?

brthor: Nevermind it's totally different :smile:

src/Microsoft.DotNet.Cli.Utils/LineNotificationTextWriter.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public override Encoding Encoding
+        {
+            get { return _encoding; }
+        }
+
+        // Write(char) gets called for all overloads of Write
+        public override void Write(char value)
+        {
+            lock (_currentString)
```
brthor: What's the scenario where we'd have multiple of these running in the same process? Are we launching built in commands with threads?

I guess the real question is why do we need to lock this?

src/Microsoft.DotNet.Cli.Utils/Reporter.cs
```diff
@@ -14,16 +14,34 @@ public class Reporter
...
-            new Reporter(AnsiConsole.GetOutput()) : 
-            NullReporter;
+        public static Reporter Output { get; private set; }
+        public static Reporter Error { get; private set; }
+        public static Reporter Verbose { get; private set; }
+
+        /// <summary>
+        /// Resets the Reporters to write to the current Console Out/Error.
+        /// </summary>
+        public static void Reset()
```
brthor: :+1:

src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
```diff
@@ -181,31 +184,15 @@ public override bool Compile(ProjectContext context, CompilerCommandApp args)
...
-                    if (diagnostic != null)
-                    {
-                        diagnostics.Add(diagnostic);
-                    }
-                    else
-                    {
-                        Reporter.Output.WriteLine(line);
-                    }
-                }).Execute();
+            // Cache the reporters before invoking the command in case it is a built-in command, which replaces
```
brthor: Can this be handled in the command factory? Feels like it would be cleaner than having to do this wherever we use the factory.

test/Microsoft.DotNet.Cli.Utils.Tests/BuiltInCommandTests.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Linq;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using Xunit;
+
+namespace Microsoft.DotNet.Cli.Utils
+{
+    public class BuiltInCommandTests : TestBase
+    {
+        [Fact]
+        public void TestExecute()
```
brthor: Can you put the expectation of the result of this test in the method name? 

Having to go through and change a bunch of tests this week, I've found being able to determine what the test is supposed to do from the method name is immensely helpful.

eerhardt: Yes, this is a good idea.  Is it acceptable to put the expectation of the test in a `<summary>` comment on the test?  Or does it have to be in the name?

src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
eerhardt: Unfortunately it can't.

The reason this place is special is because we are calling Reporter.Write during OnErrorLine and OnOutputLine handlers. At this point, Reporter.Out is going to write to the interception TextWriter, which will then raise OnOutputLine again, and you'll get into an infinite recursion.

BuiltInCommand writes to console -> OnOutputLine handler is invoked -> Writes to Reporter.Out -> OnOutputLine handler invoked -> ...

Since this callsite expicitly is handling the BuiltInCommand's output and wants to re-write that output to the console, it needs to cache the Reporter that is pointing to the real stdout

src/Microsoft.DotNet.Cli.Utils/LineNotificationTextWriter.cs
eerhardt: It is for the case where a BuiltInCommand uses multiple threads that write to Console.Out or Console.Error. It probably doesn't happen currently, but if this starts getting reused in other places, or compile-csc starts doing more things, it felt better to be safe since this object is going to be held in the static Console.Out property.

src/dotnet/Program.cs
```diff
@@ -6,26 +6,37 @@
...
+using NuGet.Frameworks;
 
 namespace Microsoft.DotNet.Cli
 {
     public class Program
     {
+        private static Dictionary<string, Func<string[], int>> s_builtIns = new Dictionary<string, Func<string[], int>>
+        {
+            ["build"] = BuildCommand.Run,
+            ["compile-csc"] = CompileCscCommand.Run,
```
Sridhar-MS: Any reason why this list does not include `dotnet-compile`?

Sridhar-MS:
@eerhardt Is it possible to remove the process hop between `compile-csc` and `csc` usng reflection? It is already done for `restore` command - https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-restore/NuGet3.cs#L30.

src/dotnet/Program.cs
eerhardt: @davidfowl removed it with https://github.com/dotnet/cli/commit/c7fe191f0bbae27f729a4b1fa183a16e59e61098 for issue #1574.

eerhardt:
> Is it possible to remove the process hop between compile-csc and csc usng reflection? 

We will work on implementing that in a different change.  I'm currently running into issues with loading `csc` in proc. For more details, see:  https://github.com/dotnet/cli/issues/2483#issuecomment-210272316

eerhardt:
@Petermarcu approved this verbally in a triage meeting.  Merging.

cdmihai:
@eerhardt 
Can you please add dotnet-compile and dotnet-compile-{compilerName} to the path probing preconditions?

eerhardt:
> Can you please add dotnet-compile and dotnet-compile-{compilerName} to the path probing preconditions?

I'm not sure I know which path probing preconditions you mean (I'm relatively new to this product). If you feel this is needed, can you open an issue for it?

cdmihai:
Here: https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-compile/CompilerUtil.cs#L123

In the returned list, add dotnet-compile-{compilerName}.

cdmihai:
Actually, I can do that and also add tests for the incremental preconditions.

cdmihai:
Opened issue here: #2581

PR #2546, opened 2016-04-17
Breadcrumbs for servicing

schellap:
@gkhanna79 PTAL.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2546)

### DIFF
--- a/src/corehost/cli/breadcrumbs.cpp
+++ b/src/corehost/cli/breadcrumbs.cpp
@@ -1,0 +1,91 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <thread>
+#include "pal.h"
+#include "utils.h"
+#include "trace.h"
+#include "breadcrumbs.h"
+
+breadcrumb_writer_t::breadcrumb_writer_t(const std::unordered_set<pal::string_t>* files)
+    : m_status(false)
+    , m_files(files)
+{
+    if (!pal::get_default_breadcrumb_store(&m_breadcrumb_store))
+    {
+        m_breadcrumb_store.clear();
+    }
+}
+
+// Begin breadcrumb writing: write synchronously or launch a
+// thread to write breadcrumbs.
+void breadcrumb_writer_t::begin_write()
+{
+    trace::verbose(_X("--- Begin breadcrumb write"));
+    if (m_breadcrumb_store.empty())
+    {
+        trace::verbose(_X("Breadcrumb store was not obtained... skipping write."));
+        m_status = false;
+        return;
+    }
+
+    trace::verbose(_X("Number of breadcrumb files to write is %d"), m_files->size());
+    if (m_files->empty())
+    {
+        m_status = true;
+        return;
+    }
+    m_thread = std::thread(write_worker_callback, this);
+    trace::verbose(_X("Breadcrumbs will be written using a background thread"));
+}
+
+// Write the breadcrumbs. This method should be called
+// only from the background thread.
+void breadcrumb_writer_t::write_callback()
+{
+    bool successful = true;
+    for (const auto& file : *m_files)
+    {
+        pal::string_t file_path = m_breadcrumb_store;
+        pal::string_t file_name = _X("netcore,") + file;
+        append_path(&file_path, file_name.c_str());
+        if (!pal::file_exists(file_path))
+        {
+            if (!pal::touch_file(file_path))
+            {
+                successful = false;
+            }
+        }
+    }
+    // m_status should not be modified by anyone else.
+    m_status = successful;
+}
+
+// ThreadProc for the background writer.
+void breadcrumb_writer_t::write_worker_callback(breadcrumb_writer_t* p_this)
+{
+    try
+    {
+        trace::verbose(_X("Breadcrumb thread write callback..."));
+        p_this->write_callback();
+    }
+    catch (...)
+    {
+        trace::warning(_X("An unexpected exception was thrown while leaving breadcrumbs"));
+    }
+}
+
+// Wait for completion of the background tasks, if any.
+bool breadcrumb_writer_t::end_write()
+{
+    if (m_thread.joinable())
+    {
+        trace::verbose(_X("Waiting for breadcrumb thread to exit..."));
+
+        // Block on the thread to exit.
+        m_thread.join();
+    }
+    trace::verbose(_X("--- End breadcrumb write %d"), m_status);
+    return m_status;
+}
+
--- a/src/corehost/cli/fxr/hostfxr.cpp
+++ b/src/corehost/cli/fxr/hostfxr.cpp
@@ -60,13 +60,17 @@ int execute_app(
 
     if (code != StatusCode::Success)
     {
-        trace::error(_X("Could not load host policy library from [%s]"), impl_dll_dir.c_str());
+        trace::error(_X("Expected to load %s from [%s]"), LIBHOSTPOLICY_NAME, impl_dll_dir.c_str());
         if (init->fx_dir() == impl_dll_dir)
         {
             pal::string_t name = init->fx_name();
             pal::string_t version = init->fx_version();
             trace::error(_X("This may be because the targeted framework [\"%s\": \"%s\"] was not found."),
                 name.c_str(), version.c_str());
+        }
+        else
+        {
+            trace::error(_X("This may be because of an invalid .NET Core FX configuration in the folder."));
         }
         return code;
     }
@@ -86,7 +90,7 @@ int execute_app(
 bool hostpolicy_exists_in_svc(pal::string_t* resolved_dir)
 {
     pal::string_t svc_dir;
-    pal::get_default_extensions_directory(&svc_dir);
+    pal::get_default_servicing_directory(&svc_dir);
     append_path(&svc_dir, _X("pkgs"));
 
     pal::string_t version = _STRINGIFY(HOST_POLICY_PKG_VER);
--- a/src/corehost/common/pal.unix.cpp
+++ b/src/corehost/common/pal.unix.cpp
@@ -12,7 +12,7 @@
 #include <sys/types.h>
 #include <pwd.h>
 #include <unistd.h>
-
+#include <fcntl.h>
 #include <fnmatch.h>
 
 #if defined(__APPLE__)
@@ -32,6 +32,18 @@ pal::string_t pal::to_lower(const pal::string_t& in)
     pal::string_t ret = in;
     std::transform(ret.begin(), ret.end(), ret.begin(), ::tolower);
     return ret;
+}
+
+bool pal::touch_file(const pal::string_t& path)
+{
+    int fd = open(path.c_str(), (O_CREAT | O_EXCL), (S_IRUSR | S_IRGRP | S_IROTH));
+    if (fd == -1)
+    {
+        trace::warning(_X("Failed to open() file descriptor in %s(%s)"), __FUNCTION__, path.c_str());
+        return false;
+    }
+    (void) close(fd);
+    return true;
 }
 
 bool pal::getcwd(pal::string_t* recv)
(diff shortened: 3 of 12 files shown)


src/corehost/common/pal.windows.cpp
```diff
@@ -54,6 +55,17 @@ bool pal::find_coreclr(pal::string_t* recv)
...
 
+bool pal::touch_file(const pal::string_t& path)
+{
+    HANDLE hnd = ::CreateFileW(path.c_str(), 0, 0, NULL, CREATE_NEW, FILE_ATTRIBUTE_NORMAL, NULL);
+    if (hnd == INVALID_HANDLE_VALUE)
+    {
+        trace::verbose(_X("Failed to leave breadcrumb"), HRESULT_FROM_WIN32(GetLastError()));
+        return false;
+    }
+    return true;
```
schellap: CloseHandle

src/corehost/common/pal.unix.cpp
```diff
@@ -112,11 +124,47 @@ bool pal::is_path_rooted(const pal::string_t& path)
...
+    recv->clear();
+    pal::string_t ext;
+    if (!get_default_extensions_directory(&ext))
+    {
+        return false;
+    }
+
+    append_path(&ext, _X("breadcrumbs"));
+    if (access(ext.c_str(), (R_OK | W_OK)) != 0)
+    {
```
schellap: Dir exists check

schellap:
@gkhanna79 PTAL

Cc @ericstj @leecow

schellap:
@ellismg for changes to `pal.unix.cpp`

ericstj:
Can you share sample paths this creates?

schellap:
Samples:

```
C:\ProgramData\Microsoft\NetFramework\BreadcrumbStore\netcore,System.AppContext

C:\ProgramData\Microsoft\NetFramework\BreadcrumbStore\netcore,System.AppContext,1.0.0
```

ericstj:
Those are package IDs and versions: correct?

schellap:
Right, `netcore,PackageId,PackageVersion`. I used comma as separator to match existing breadcrumbs and to not confuse with package or version namespacing.

src/corehost/common/pal.unix.cpp
```diff
@@ -112,11 +124,64 @@ bool pal::is_path_rooted(const pal::string_t& path)
...
-    append_path(recv, _X("opt"));
-    append_path(recv, _X("dotnetextensions"));
+    pal::string_t ext;
+    if (pal::getenv(_X("DOTNET_EXTENSIONS"), &ext) && pal::realpath(&ext))
+    {
+        // We should have the path in ext.
+        trace::info(_X("Realpath $DOTNET_EXTENSIONS [%s]"), ext.c_str());
+    }
+    
+    if (!pal::directory_exists(ext))
```
gkhanna79: If DOTNET_EXTENSIONS is not defined, then will pal::string_t be zero initialized? If not, then this check is bogus. We should zero init ext and check it to be zero length here instead.

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        m_breadcrumb_store.clear();
+    }
+}
+
+// Begin breadcrumb writing: write synchronously or launch a
+// thread to write breadcrumbs.
+void breadcrumb_writer_t::begin_write()
+{
+    trace::verbose(_X("--- Begin breadcrumb write"));
```
gkhanna79: This trace should be moved to be after the empty check.

src/corehost/common/pal.unix.cpp
schellap: It will be zero inited by the default constructor and directory exists will early out on an empty string.

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    trace::verbose(_X("Number of breadcrumb files to write is %d"), m_files->size());
+    if (m_files->size() <= SYNCHRONOUS_WRITE_SIZE)
+    {
+        trace::verbose(_X("Writing breadcrumbs on the calling thread"));
+        m_status = write();
+        return;
+    }
+    m_thread = std::thread(thread_callback, this);
```
gkhanna79: The callback should be named something like "breadcrumb_write_callback".

src/corehost/cli/breadcrumbs.cpp
schellap: No, this is simply tracking when the call was made.

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -0,0 +1,105 @@
...
+    trace::verbose(_X("--- Begin breadcrumb write"));
+    if (m_breadcrumb_store.empty())
+    {
+        trace::verbose(_X("Breadcrumb store was not obtained... skipping write."));
+        m_status = false;
+        return;
+    }
+
+    trace::verbose(_X("Number of breadcrumb files to write is %d"), m_files->size());
+    if (m_files->size() <= SYNCHRONOUS_WRITE_SIZE)
```
gkhanna79: How did we arrive at this? Do we have reason to believe that higher counts will be serviced better via another thread.

gkhanna79: I would suggest doing all writes via a thread callback or none. If goal is to not block the main app thread, then move everything to thread callback.

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -0,0 +1,105 @@
...
+        trace::warning(_X("An unexpected exception was thrown while leaving breadcrumbs"));
+    }
+    try
+    {
+        // Signal the waiting thread.
+        p_this->m_promise.set_value(retval);
+    }
+    catch (...)
+    {
+        trace::warning(_X("Could not signal waiting threads..."));
```
gkhanna79: When can we be here?

src/corehost/cli/breadcrumbs.cpp
schellap: When CLI runs we emit 300 files i.e., roughly 300 probes and 300 writes in the worst case. We can always toggle the knob based on our perf studies. CLR leaves about 40 files at various points during assembly loads. Since we leave it eagerly I arrived at half that number. This is open to tuning.

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -0,0 +1,105 @@
...
+        m_status = write();
+        return;
+    }
+    m_thread = std::thread(thread_callback, this);
+    trace::verbose(_X("Breadcrumbs will be written using a background thread"));
+}
+
+// Write the breadcrumbs. This method could be called fom the
+// calling thread and the background thread. But at any given
+// time only 1 thread will be here.
```
gkhanna79: > But at any given time only 1 thread will be here.

There is no synchronization ensuring this. Is that expected or implied?

src/corehost/cli/breadcrumbs.cpp
```diff
@@ -0,0 +1,105 @@
...
+    for (const auto& file : *m_files)
+    {
+        pal::string_t file_path = m_breadcrumb_store;
+        pal::string_t file_name = _X("netcore,") + file;
+        append_path(&file_path, file_name.c_str());
+        if (!pal::file_exists(file_path))
+        {
+            if (!pal::touch_file(file_path))
+            {
+                successful = false;
```
gkhanna79: @ericstj Is it reasonable to expect breadcrumb writing to fail?

gkhanna79: What is the behavior on desktop for similar situation?

src/corehost/cli/breadcrumbs.cpp
schellap: It is implied since we launch a thread and forget about doing anything else in the main thread. I can add an assert we don't call begin write twice.

src/corehost/cli/fxr/hostfxr.cpp
```diff
@@ -60,14 +60,18 @@ int execute_app(
 
     if (code != StatusCode::Success)
     {
-        trace::error(_X("Could not load host policy library from [%s]"), impl_dll_dir.c_str());
+        trace::error(_X("Expected to load %s from [%s]"), LIBHOSTPOLICY_NAME, impl_dll_dir.c_str());
```
gkhanna79: I would suggest a more user-understandable message - "Invalid .NET Core FX configuration in folder <path to folder>"

src/corehost/common/pal.unix.cpp
```diff
@@ -112,11 +124,64 @@ bool pal::is_path_rooted(const pal::string_t& path)
...
+    }
+
+    append_path(&ext, _X("breadcrumbs"));
+    if (!pal::directory_exists(ext))
+    {
+        trace::info(_X("Breadcrumb store does not exist in [%s]"), ext.c_str());
+        return false;
+    }
+
+    if (access(ext.c_str(), (R_OK | W_OK)) != 0)
```
ericstj: Is it possible that we can't enumerate perms but can actually write?

src/corehost/cli/breadcrumbs.cpp
schellap: It is already namespaced within `breadcrumb_writer_t`, so `breadcrumb_writer_t::breadcrumb_write_callback` is redundant. I've renamed it to `write_worker_callback`

src/corehost/cli/breadcrumbs.cpp
schellap: I like that suggestion. I've changed it to always launch a thread.

src/corehost/cli/breadcrumbs.cpp
schellap: This is my being extra careful, as an uncaught exception means a fail fast. Breadcrumb design philosophy is to not break the app in doing so.

schellap: The downside is, the caller might block on the shared state which is in theory breaking as well. An exception at this point should not happen unless the app is already in an invalid state like a memory corruption or some unforeseen situation.

src/corehost/cli/fxr/hostfxr.cpp
schellap: Okay done.

src/corehost/common/pal.unix.cpp
schellap: @ericstj, I haven't addressed your PR yet, as I wanted to test this out first. Will send another commit for this one specifically.

schellap:

schellap:

src/corehost/cli/fxr/hostfxr.cpp
gkhanna79: Is there a PR with this change?

gkhanna79: Never mind - saw the change.

src/corehost/common/pal.unix.cpp
```diff
@@ -112,11 +124,68 @@ bool pal::is_path_rooted(const pal::string_t& path)
...
+        // We should have the path in ext.
+        trace::info(_X("Realpath CORE_BREADCRUMBS [%s]"), ext.c_str());
+    }
+
+    if (!pal::directory_exists(ext))
+    {
+        trace::info(_X("Directory core breadcrumbs [%s] was not specified or found"), ext.c_str());
+        ext.clear();
+        append_path(&ext, _X("opt"));
+        append_path(&ext, _X("corebreadcrumbs"));
```
gkhanna79: Is it "breadcrumb" or "breadcrumbs" on Windows? We should ensure that whatever follows "CORE" is consistent across platforms.

schellap: In Windows it is called BreadcrumbStore. On Unix this is abbreviated partially as BreadcrumbS ;) On Windows there is no env variable. We are consistent on Unix on the plural for both env and dir.

schellap:
@gkhanna79 the name core_servicing needs to be locked down for muxer lock down. We are okay on the branding etc.? @ericstj brought up a point that this is too generic as a name and I agree -- effectively, if we screw up inside this folder, customers won't know who to talk to.

gkhanna79:
We do not expect the name to be known publicly - it is an infra-use artifact. And I expect that @leecow document on servicing to reference it as well.

schellap:
Fixes #2607

PR #2554, opened 2016-04-18
Update README.md to include Debian build status and download links

MichaelSimons:
skip ci please

/cc @Sridhar-MS

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2554)


README.md
```diff
@@ -35,6 +35,7 @@ You can download .NET Core as either an installer (MSI, PKG) or a zip (zip, gzip
...
 |**Ubuntu 14.04**|[![](https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/Ubuntu_x64_Release_version_badge.svg)](https://dotnetcli.blob.core.windows.net/dotnet/beta/dnvm/latest.ubuntu.x64.version)|*See Below*|*See Below*|[Download](https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-ubuntu-x64.latest.tar.gz)|[Download](https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-dev-ubuntu-x64.latest.tar.gz)|
+|**Debian 8.2**|[![](https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/Debian_x64_Release_version_badge.svg)](https://dotnetcli.blob.core.windows.net/dotnet/beta/dnvm/latest.debian.x64.version)|N/A|N/A|[Download](https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-debian-x64.latest.tar.gz)|[Download](https://dotnetcli.blob.core.windows.net/dotnet/beta/Binaries/Latest/dotnet-dev-debian-x64.latest.tar.gz)|
```
Petermarcu: Do we also need to update the section below where we break out the layers of the Ubuntu installs to answer the N/A case in the table?

MichaelSimons: @Petermarcu, No I don't think so.  The Debian installer work was out of scope for RC2.  Your only option for installation at this time is to use the tarballs.  This is why N/A was used in the installer columns.

Sridhar-MS:
:shipit:

PR #2559, opened 2016-04-18
Fixes #2512

### LINKED ISSUE #2512: ProjectDependenciesCommandFactory.Create does not respect Configuration
## Steps to reproduce
- Instantiate a `ProjectDependenciesCommandFactory` with a specific configuration:
```C#
var commandFactory = new ProjectDependenciesCommandFactory(
    framework,
    "Release",
    outputPath: null,
    buildBasePath: null,
    projectDirectory: "....");
```

- Create and execute `ICommand`: `commandFactory.Create("some name", Enumerable.Empty<string>()).Execute()`

## Expected  behavior
- Executes command in the `"Release"` configuration.

## Actual behavior
- Executes command in the `"Debug"` configuration.

## Why
Looks like if the configuration is not provided to the command factories `Create` method defaults to `Constants.DefaultConfiguration` instead of inheriting the `ProjectDependenciesCommandFactory `_configuration` private variable. See [here](https://github.com/dotnet/cli/blob/880dcf2fc820e8db6f7f9efd639b06a552a6f414/src/Microsoft.DotNet.Cli.Utils/ProjectDependenciesCommandFactory.cs#L36-L41).

Seems like an inconsistency amongst the other provided params like `framework` where it'll default to the `ProjectDependenciesCommandFactory` private variable value if it's not provided to `Create`.

Can currently work around this by providing the configuration to the `ProjectDependenciesCommandFactory` ctor and its `Create` method. (Or just the `Create` method, haven't fiddled with it too much to see if it's absolutely required to be set as a private variable).

brthor:
This PR changes the behavior of the `ProjectDependenciesCommandFactory` to not always use the default configuration. 

I think the pattern of passing in the configuration should change to one place instead of two, but to minimize code changes this will fix the issue we are seeing.

@NTaylorMullen PTAL

cc @livarcocc

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2559)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/ProjectDependenciesCommandFactory.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/ProjectDependenciesCommandFactory.cs
@@ -27,13 +27,18 @@ namespace Microsoft.DotNet.Cli.Utils
             _outputPath = outputPath;
             _buildBasePath = buildBasePath;
             _projectDirectory = projectDirectory;
+
+            if (_configuration == null)
+            {
+                _configuration = Constants.DefaultConfiguration;
+            }
         }
 
         public ICommand Create(
             string commandName,
             IEnumerable<string> args,
             NuGetFramework framework = null,
-            string configuration = Constants.DefaultConfiguration)
+            string configuration = null)
         {
             if (string.IsNullOrEmpty(configuration))
             {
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Looks good. But side question, do you guys typically not add tests for these types of things?

brthor:
Yeah you are right, we need a test, I was going for the quick fix, but let's make sure it doesn't regress

brthor:
@dotnet-bot test this please

NTaylorMullen:
Should add a variation where you call `.Create` with the configuration non-null and expect it to override the value that it was constructed with.

brthor:
added the variation, that's a good idea.

NTaylorMullen:
:shipit:

PR #2560, opened 2016-04-18
Add crossgen tests

Sridhar-MS:
Simple tests which does static analysis of managed assemblies metadata to
make sure that they are crossgened. Currently it verifies that all the
assemblies in CLI SDK and SharedFx directory are crossgened.

cc @brianrob @JohnChen0 @piotrpMSFT 

I am not sure if there is a way to make sure that the assemblies are not partially crossgened. @brianrob @JohnChen0 do you guys know of a reliable way to do that?

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2560)

### DIFF
--- a/test/Microsoft.DotNet.Tools.Tests.Utilities/PeReaderUtils.cs
+++ b/test/Microsoft.DotNet.Tools.Tests.Utilities/PeReaderUtils.cs
@@ -11,6 +11,19 @@ namespace Microsoft.DotNet.Tools.Test.Utilities
 {
     public static class PeReaderUtils
     {
+        public static bool IsCrossgened(this PEReader peReader)
+        {
+            bool isCrossgened = false;
+
+            if (peReader.HasMetadata)
+            {
+                // 4 is the magic numbers that is set in the CLR header's flags when crossgened.
+                isCrossgened = (int)peReader.PEHeaders.CorHeader.Flags == 4;
+            }
+
+            return isCrossgened;
+        }
+
         public static string GetAssemblyAttributeValue(string assemblyPath, string attributeName)
         {
             if (!File.Exists(assemblyPath))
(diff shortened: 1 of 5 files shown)


brianrob:
I will defer to @JohnChen0 here as he is the NGEN expert.

test/Microsoft.DotNet.Tools.Tests.Utilities/PeReaderUtils.cs
```diff
@@ -11,6 +11,19 @@ namespace Microsoft.DotNet.Tools.Test.Utilities
...
     public static class PeReaderUtils
     {
+        public static bool IsCrossgened(this PEReader peReader)
+        {
+            bool isCrossgened = false;
+
+            if (peReader.HasMetadata)
+            {
+                // 4 is the magic numbers that is set in the CLR header's flags when crossgened.
+                isCrossgened = (int)peReader.PEHeaders.CorHeader.Flags == 4;
```
JohnChen0: This field is a bit flag, so shouldn't do an equal comparison. Also, we have an enum that defines its values. Better to use something like:

`isCrossgened = (peReader.PEHeaders.CorHeader.Flags & CorFlags.ILLibrary) != 0;`

Unfortunately the term `ILLibrary` won't make sense to most people, so a comment that it means crossgened would be helpful.

JohnChen0:
There is no simple way to tell if an assembly is partially crossgened.

piotrpMSFT:
nice!

piotrpMSFT:
:shipit:

brianrob:
Looks great!

piotrpMSFT:
It would be great if we could test for partial crossgen, though. Could these checks be made by the crossgen tool itself?

piotrpMSFT:
Can we rebase and merge this?

Sridhar-MS:
rebased

Issue #2521, opened 2016-04-15
Dotnet Build accepts multiple projects in single invocation


piotrpMSFT:

piotrpMSFT:
@davidfowl when is this coming in?

piotrpMSFT:
@davidfowl ping

pakrym:
@piotrpMSFT I'm working on it, should have a PR tomorrow or day after.

piotrpMSFT:
@pakrym:

> ![image](https://cloud.githubusercontent.com/assets/2379589/14802672/8538d36a-0b07-11e6-8358-390b29bfd064.png)

Is this ready to go?

anurse:
This is done.

PR #2574, opened 2016-04-18
[WIP] Add multiple input project support

pakrym:
First step of https://github.com/dotnet/cli/issues/2521

This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2574)


src/dotnet/commands/dotnet-build/ProjectGraphNode.cs
```diff
@@ -0,0 +1,26 @@
...
+        Task<ProjectContext> _projectContextCreator;
+
+        public ProjectGraphNode(Task<ProjectContext> projectContext, IEnumerable<ProjectGraphNode> dependencies, bool isRoot = false)
+        {
+            _projectContextCreator = projectContext;
+            Dependencies = dependencies.ToList();
+            IsRoot = isRoot;
+        }
+
+        public ProjectContext ProjectContext { get { return _projectContextCreator.GetAwaiter().GetResult(); } }
```
anurse: Should we be blocking here?

pakrym: Yes, consumer needs `ProjectContext`, there is nothing we can do.

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                return RunCompile(projectNode);
+            }
+        }
+
+        protected abstract CompilationResult RunCompile(ProjectGraphNode projectNode);
+
+        protected abstract bool SkipIncremental(ProjectGraphNode projectNode);
+
+        private class ProjectContextIdentity : Tuple<string, NuGetFramework>
```
anurse: Can we promote this to ProjectContext, structify it and make it a property directly on the context?

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -0,0 +1,257 @@
...
+using Microsoft.DotNet.Tools.Compiler;
+using Microsoft.Extensions.PlatformAbstractions;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Tools.Build
+{
+    class DotnetProjectBuilder : ProjectBuilder
```
anurse: Dot**N**et

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -0,0 +1,91 @@
...
+            }
+            else
+            {
+                return RunCompile(projectNode);
+            }
+        }
+
+        protected abstract CompilationResult RunCompile(ProjectGraphNode projectNode);
+
+        protected abstract bool SkipIncremental(ProjectGraphNode projectNode);
```
anurse: I'd call this something like "PreviousBuildIsNotOutOfDateAndCanBeReusedWithoutRecompilingTheProject"? :trollface: 

Troll names aside, it wasn't exactly clear what this method was supposed to do until I reviewed the code in `DotNetProjectBuilder`. It seemed like maybe it was supposed to return true if we skip the incremental build? But I don't even 100% know what that means :P. Maybe just call it "NeedsRebuilding" and merge it with the existing method on DotNetProjectBuilder? Or "IsUpToDate" or something?

src/dotnet/commands/dotnet-build/ProjectGraphNode.cs
anurse: Hrm... still makes me a little queasy :). We're going to flow async up eventually, right?

pakrym: With parallel build - yes.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -315,7 +315,7 @@ private static void CleanOutputDir(string directory)
 
         private static void RemovePdbsFromDir(string directory)
         {
-            FS.RmFilesInDirRecursive(directory, "*.pdb");
+            //FS.RmFilesInDirRecursive(directory, "*.pdb");
```
anurse: Remove :)

anurse:
:shipit:

src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
```diff
@@ -0,0 +1,71 @@
...
+            foreach (var context in contexts)
+            {
+                var libraries = context.LibraryManager.GetLibraries();
+                var lookup = libraries.ToDictionary(l => l.Identity.Name);
+                var root = lookup[context.ProjectFile.Name];
+                yield return TraverseProject(root, lookup, context);
+            }
+        }
+
+        private ProjectGraphNode TraverseProject(LibraryDescription library, IDictionary<string, LibraryDescription> lookup, ProjectContext context = null)
```
cdmihai: Change type to ProjectDescription and rename to project?

src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
```diff
@@ -0,0 +1,71 @@
...
+                else
+                {
+                    deps.AddRange(TraversePackage(libraryDescription, lookup));
+                }
+            }
+            var task = context != null ? Task.FromResult(context) : Task.Run(() => _projectContextFactory(library.Path, library.Framework));
+            return new ProjectGraphNode(task, deps, context != null);
+        }
+
+        private IEnumerable<ProjectGraphNode> TraversePackage(LibraryDescription root, IDictionary<string, LibraryDescription> lookup)
```
cdmihai: Change type to TargetLibraryWithAssets and rename to libraryWithAssets? There's also MSBuildProjectDescription which should behave just like a package (get ignored but collect sub projects).

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -0,0 +1,88 @@
...
+            result = CompileWithDependencies(projectNode);
+
+            _compilationResults[projectNode.ProjectContext.Identity] = result;
+
+            return result;
+        }
+
+        private CompilationResult CompileWithDependencies(ProjectGraphNode projectNode)
+        {
+            var skipIncrementalCheck = false;
```
cdmihai: what is the use of this variable? Does not appear to be read anywhere.

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -0,0 +1,243 @@
...
+        private readonly BuilderCommandApp _args;
+        private readonly IncrementalPreconditionManager _preconditionManager;
+        private readonly CompilerIOManager _compilerIOManager;
+        private readonly ScriptRunner _scriptRunner;
+        private readonly DotNetCommandFactory _commandFactory;
+
+        public DotNetProjectBuilder(BuilderCommandApp args) : base(args.ShouldSkipDependencies)
+        {
+            _args = (BuilderCommandApp)args.ShallowCopy();
+            _preconditionManager = new IncrementalPreconditionManager(
```
cdmihai: _preconditionManager should be used to compute preconditions and then have incrementality switched on / off based on its results

I'd make two visitors on the graph: a visitor that computes incremental preconditions (aggregates preconditions from all nodes) and another visitor that builds (mostly what's in ProjectBuilder) the graph. The result of the first visitor (whether the entire graph is safe or not for incrementality) affects the way the second visitor is setup (incremental on / off)

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
pakrym: Yeah, forgot to use it, its a bug.

src/dotnet/commands/dotnet-build/CompilerIOManager.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Tools.Compiler;
+using Microsoft.DotNet.ProjectModel;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Cli.Compiler.Common;
+
+namespace Microsoft.DotNet.Tools.Build
+{
+    class CompilerIOManager
```
davidfowl: nit: public

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -0,0 +1,88 @@
...
+                    {
+                        continue;
+                    }
+                    var result = Build(dependency);
+                    switch (result)
+                    {
+                        case CompilationResult.IncrementalSkip:
+                            break;
+                        case CompilationResult.Success:
+                            skipIncrementalCheck = true;
```
cdmihai: rename to something that shows the meaning on why it was skipped: "dependencyChanged" or "dependentProjectCompiled"

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -0,0 +1,88 @@
...
+                            break;
+                        case CompilationResult.Failure:
+                            return CompilationResult.Failure;
+                        default:
+                            throw new NotSupportedException();
+                    }
+                }
+            }
+            if (skipIncrementalCheck || NeedsRebuilding(projectNode))
```
cdmihai: This makes it a bit hard for my current work on caching inputs between runs. If NeedsRebuilding is skipped, it cannot cache inputs anymore, meaning that when NeedsRebuild is called, it might get a stale cache. This would cause a cache miss.

I could cache the inputs at some other time (maybe at the same time the project is stamped with the SDK version), but that would smear the code responsible with incrementality all over. Or we could add fancy 

I'd vote for removing the short circuit. What's your opinion on that?

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -0,0 +1,253 @@
...
+            return !versionsAreEqual;
+        }
+
+        protected override bool NeedsRebuilding(ProjectGraphNode  graphNode)
+        {
+            if (_args.ShouldNotUseIncrementality)
+            {
+                return true;
+            }
+            var preconditions = _preconditionManager.GatherIncrementalPreconditions(graphNode);
```
cdmihai: The decision on whether to turn incrementality on / off needs to be taken for the entire graph, not subgraphs. I'd go for my suggestion on having 2 visitors.

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
pakrym: Saving inputs on compile and reading in `NeedsRebuild` should work. I can't see how it misses cache?

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
pakrym: This check is disabling incrementality for whole graph. I don't want to overcomplicate code at this moment. If we see a need later we can always add visitors.

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
pakrym: Yeah, lost this part during one of PR reactions.

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
cdmihai: Visitors aside, even though the args are the same, by taking the decision on each node rather than at the whole graph level, you can end up with a scenario where part of the graph passes the preconditions and part of the graph does not pass. This would be an incorrect state.

---
In reply to: [60173858](https://github.com/dotnet/cli/pull/2574#discussion_r60173858) [](ancestors = 60173858)

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
cdmihai: Yes, if inputs are written outside of NeedsRebuild (like besides SDK stamping) then there's no stale cache with the `skipIncrementalCheck` short circuit. 

The downside to separating caching inputs from the actual incremental check is that it scatters the responsibility around. For example, responding to the --no-incremental is harder, since more parts of the code would need to branch on it, not just NeedsRebuild.

---
In reply to: [60173794](https://github.com/dotnet/cli/pull/2574#discussion_r60173794) [](ancestors = 60173794)

cdmihai:
The incremental precondition check needs to be done for the entire graph: if any node in the graph fails the preconditions, then the entire build should not be incremental.

pakrym:
Build `MVC` successfully with this branch.

src/dotnet/commands/dotnet-build/IncrementalPreconditionManager.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Linq;
+using Microsoft.DotNet.Tools.Compiler;
+using Microsoft.DotNet.ProjectModel;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Cli.Compiler.Common;
+using Microsoft.DotNet.ProjectModel.Utilities;
+
+namespace Microsoft.DotNet.Tools.Build
+{
+    class IncrementalPreconditionManager
```
eerhardt: Nit: can you explicitly declare the visibility on these classes? It's #5 in https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

src/Microsoft.DotNet.ProjectModel/ProjectContextIdentity.cs
```diff
@@ -0,0 +1,36 @@
+using Microsoft.Extensions.Internal;
```
eerhardt: Copyright header on this and all new files.

src/dotnet/ProjectGlobbingResolver.cs
```diff
@@ -0,0 +1,61 @@
...
+    internal class ProjectGlobbingResolver
+    {
+        private readonly DirectoryInfoBase _baseDirectory;
+
+        public ProjectGlobbingResolver(DirectoryInfoBase baseDirectory)
+        {
+            _baseDirectory = baseDirectory;
+        }
+
+        public ProjectGlobbingResolver():this(new DirectoryInfoWrapper(new DirectoryInfo(Directory.GetCurrentDirectory())))
```
eerhardt: Nit: spacing.

src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs
```diff
@@ -82,17 +67,7 @@ private void AddCompileParameters()
...
-            _ilcArgsOption = _app.Option("--ilcarg <ARG>", "Command line option to be passed directly to ILCompiler.", CommandOptionType.MultipleValue);
-            _ilcPathOption = _app.Option("--ilcpath <PATH>", "Path to the folder containing custom built ILCompiler.", CommandOptionType.SingleValue);
-            _ilcSdkPathOption = _app.Option("--ilcsdkpath <PATH>", "Path to the folder containing ILCompiler application dependencies.", CommandOptionType.SingleValue);
-            _appDepSdkPathOption = _app.Option("--appdepsdkpath <PATH>", "Path to the folder containing ILCompiler application dependencies.", CommandOptionType.SingleValue);
-            _cppModeOption = _app.Option("--cpp", "Flag to do native compilation with C++ code generator.", CommandOptionType.NoValue);
-            _cppCompilerFlagsOption = _app.Option("--cppcompilerflags <flags>", "Additional flags to be passed to the native compiler.", CommandOptionType.SingleValue);
+            _projectArgument = _app.Argument("<PROJECT>", "The project to compile, defaults to the current directory. Can be a path to a project.json or a project directory", multipleValues: true);
```
cdmihai: Update to reflect multiple values and globbing?

cdmihai:
>                 ProjectPathValue = _projectArgument.Value;

I forget what value returns when there's multiple values. I think it's the first one, right? If that's the case, you can just delete this line and check for has values

---
Refers to: src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs:84 in 2c5b506. [](commit_id = 2c5b5069e39b9ec5a3a927dc0f0440a751f72575, deletion_comment = False)

src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs
```diff
@@ -118,52 +93,13 @@ public int Execute(OnExecute execute, string[] args)
...
-
-                    if (targets.Count == 0)
-                    {
-                        // We filtered everything out
-                        Reporter.Error.WriteLine($"Project does not support framework: {fx.DotNetFrameworkName}.");
-                        return 1;
-                    }
-
-                    Debug.Assert(targets.Count == 1);
+                    frameworks = new [] { NuGetFramework.Parse(_frameworkOption.Value()) };
```
cdmihai: FrameworkOption is single value, yet the var is plural. Should the frameworks option also be multiple value? Also, should users be able to specify frameworks for each project, or is there just a single set of frameworks and all projects must support the entire set?

In extension to this, how do the other command args interact with multiple projects, like output, buildbasepath, etc?

cdmihai:
>                 ProjectPathValue = _projectArgument.Value;

I forget what value returns when there's multiple values. I think it's the first one, right? If that's the case, you can just delete this line and check for has values

---
Refers to: src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs:84 in 2c5b506. [](commit_id = 2c5b5069e39b9ec5a3a927dc0f0440a751f72575, deletion_comment = False)

cdmihai:
>                 ProjectPathValue = _projectArgument.Value;

I forget what value returns when there's multiple values. I think it's the first one, right? If that's the case, you can just delete this line and check for has values



---
Refers to: src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs:84 in 2c5b506. [](commit_id = 2c5b5069e39b9ec5a3a927dc0f0440a751f72575, deletion_comment = False)

src/dotnet/commands/dotnet-build/Program.cs
```diff
@@ -34,13 +35,67 @@ public static int Run(string[] args)
...
+                var allFrameworks = project.GetTargetFrameworks().Select(f => f.FrameworkName);
+                if (!allFrameworks.Any())
+                {
+                    throw new InvalidOperationException(
+                        $"Project '{file}' does not have any frameworks listed in the 'frameworks' section.");
+                }
+                IEnumerable<NuGetFramework> selectedFrameworks;
+                if (frameworks != null)
+                {
+                    selectedFrameworks = allFrameworks.Where(f => frameworks.Any(ff => ff.Equals(f)));
```
cdmihai: Won't this be confusing for the user? He has to punch in the union of all frameworks, and each projects picks what it supports. If the user's intention is actually build for frameworks A and B he might get some projects built for A, some built for B, some built for A and B, and some will fail.

anurse: I agree. I think that if the user specifies `--framework`, then either all the projects can build with that framework, or none of them build. It's a little less flexible but I think it's easier to understand. Of course, when `--framework` is not specified (as I would expect from most bulk builds), we just build everything.

src/dotnet/commands/dotnet-build/IncrementalPreconditionManager.cs
```diff
@@ -0,0 +1,120 @@
...
+            if (_skipDependencies)
+            {
+                return new List<ProjectContext>(1) { projectNode.ProjectContext };
+            }
+
+            // include initial root project
+            var contextsToCheck = new List<ProjectContext>(1 + projectNode.Dependencies.Count) { projectNode.ProjectContext };
+
+            // TODO: not traversing deeper than 1 level of dependencies
+            contextsToCheck.AddRange(projectNode.Dependencies.Select(n => n.ProjectContext));
```
cdmihai: projectNode.Dependencies returns only the direct dependencies right? If yes, this should be transitive to avoid having mixed cones. Besides the mixed cones issues, PrintSummary in DotnetProjectBuilder won't show all the preconditions for all the nodes in the graph (it will only print dependencies for the root nodes and their direct children)

src/dotnet/commands/dotnet-build/Program.cs
pakrym: We do not support having multiple frameworks as `dotnet build` argument, its either null or one. I just made this class to support a little more just in case.

src/dotnet/commands/dotnet-compile/CompilerCommandApp.cs
pakrym: No `-f` is singular. All other arguments work exactly as before as if user executed multiple builds with same arguments.

src/dotnet/commands/dotnet-build/IncrementalPreconditionManager.cs
pakrym: I think we should print preconditions for each project, because we are calculating them for projects not graph. But I'll check how it looks right now.

pakrym:

pakrym:
I'm going to merge this one, @davidfowl will file design questions regarding incremental compilation that need to be solved.

Issue #2505, opened 2016-04-14
Binding redirect file is not generated for referenced desktop packages that contain executables.


NTaylorMullen:
## Steps to reproduce
- Create a project that has a net451 TFM (I'll refer to the project as `UserProject`)
- Add a reference to a package that contains an executable that expects `Newtonsoft.Json 7.0.1`. (I'll refer to the added package as `ToolX`)
- Add a reference to `Newtonsoft.Json 8.0.3` to `UserProject`.
- Build `UserProject`
- Execute `ToolX.exe` in the `bin/Debug/...` folder of `UserProject`

## Expected  behavior
- `ToolX.exe` runs as expected

## Actual behavior
- Type load exception for `Newtonsoft.Json 7.0.1`

## Reason
There's a binding redirect generated for `UserProject.exe` but not for the added `ToolX.exe`. We should be generating a binding redirect file for the added exe in the case that it's being brought into a desktop tfm.

We (ASP.NET team) are building tools that are capable of loading desktop targeting user projects; we do this by referencing the tool once in the tools node and again in the dependencies node. When you run `dotnet my-tool` it launches the portable version of `my-tool` which will then build the users project and invoke itself in the built output since it was referenced also in the dependencies node. This is what made the above issue apparent, we were trying to invoke a net451 tool in a user project that targeted a higher version of `Newtonsoft.Json`.

/cc @DamianEdwards @Eilon @davidfowl @muratg @danroth27 @piotrpMSFT @brthor @natemcmaster @prafullbhosale @pranavkm

piotrpMSFT:
@Sridhar-MS will have a fix today.

piotrpMSFT:
Fix is in PR, but needs test cases.

piotrpMSFT:
@Sridhar-MS can you create an RTM issue to track generating the redirects for publish output?

piotrpMSFT:
Fixed by #2586

PR #2586, opened 2016-04-19
Generate binding redirects for all executable package dependencies.

Sridhar-MS:
Binding redirects are already generated for the current project. This
change is to add these same binding redirects to all executable package
dependencies.

Fixes - #2505

Meanwhile @piotrpMSFT @anurse @pakrym, can you guys please take a look?
I still need to do some more testing but I like to get some feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2586)

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -4,6 +4,7 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="coreclr-xunit" value="https://www.myget.org/F/coreclr-xunit/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -4,7 +4,6 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
-    <add key="coreclr-xunit" value="https://www.myget.org/F/coreclr-xunit/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
@@ -34,27 +34,46 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         private static SHA1 Sha1 { get; } = SHA1.Create();
 
-        public static XDocument GenerateBindingRedirects(this IEnumerable<LibraryExport> dependencies, XDocument document)
+        public static void GenerateBindingRedirects(this IEnumerable<LibraryExport> dependencies, IEnumerable<string> configFiles)
         {
             var redirects = CollectRedirects(dependencies);
 
             if (!redirects.Any())
             {
                 // No redirects required
-                return document;
+                return;
             }
-            document = document ?? new XDocument();
 
-            var configuration = GetOrAddElement(document, ConfigurationElementName);
+            foreach (var configFile in configFiles)
+            {
+                GenerateBindingRedirects(configFile, redirects);
+            }
+        }
+
+        internal static void GenerateBindingRedirects(string configFile, AssemblyRedirect[] bindingRedirects)
+        {
+            XDocument configRoot = null;
+
+            if (File.Exists(configFile))
+            {
+                configRoot = XDocument.Load(configFile);
+            }
+
+            configRoot = configRoot ?? new XDocument();
+
+            var configuration = GetOrAddElement(configRoot, ConfigurationElementName);
             var runtime = GetOrAddElement(configuration, RuntimeElementName);
             var assemblyBindings = GetOrAddElement(runtime, AssemblyBindingElementName);
 
-            foreach (var redirect in redirects)
+            foreach (var redirect in bindingRedirects)
             {
                 AddDependentAssembly(redirect, assemblyBindings);
             }
 
-            return document;
+            using (var fileStream = File.Open(configFile, FileMode.OpenOrCreate, FileAccess.ReadWrite))
+            {
+                configRoot.Save(fileStream);
+            }
         }
 
         private static void AddDependentAssembly(AssemblyRedirect redirect, XElement assemblyBindings)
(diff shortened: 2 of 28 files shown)


src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -254,30 +254,34 @@ public void WriteDeps(LibraryExporter exporter)
...
+            List<string> configFiles = new List<string>();
+            configFiles.Add(configFile);
 
-            var path = outputName + ".config";
-            using (var stream = File.Create(path))
+            foreach (var export in exporter.GetDependencies())
             {
-                appConfig.Save(stream);
+                var dependencyExecutables = export.RuntimeAssemblyGroups.GetDefaultAssets()
+                                                .Where(asset => asset.FileName.ToLower().EndsWith(".exe"))
```
pakrym: We have a constant somewhere with `.exe` value, use it. And change `.config` to use one to.

src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
```diff
@@ -34,27 +34,46 @@ public static class BindingRedirectGenerator
...
+        internal static void GenerateBindingRedirects(string configFile, AssemblyRedirect[] bindingRedirects)
+        {
+            XDocument configRoot = null;
+
+            if (File.Exists(configFile))
+            {
+                configRoot = XDocument.Load(configFile);
+            }
+
+            configRoot = configRoot ?? new XDocument();
```
pranavkm:
```

File: NuGet.Config [hunk 3]
```diff
@@ -4,6 +4,7 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="coreclr-xunit" value="https://www.myget.org/F/coreclr-xunit/api/v3/index.json" />
```
Sridhar-MS: @eerhardt I had to include this feed as a workaround to run desktops tests? Can we get the latest (1.0.0-rc2-157751-46) `dotnet-test-xunit` package to the cli-deps feed?

test/binding-redirects.Tests/BindingRedirectTests.cs
```diff
@@ -0,0 +1,286 @@
...
+            var redirects = GetRedirects(_executableDependencyBuildOutput);
+            VerifyRedirects(redirects, ExpectedBindingsAppNoConfig);
+
+            var commandResult = new TestCommand(_executableDependencyBuildOutput)
+                                    .Execute();
+            commandResult.Should().Pass();
+        }
+
+        //[WindowsOnlyFact]
+        public void Publish_Generates_Redirects_For_Executable_Dependency()
```
Sridhar-MS: @NTaylorMullen @anurse @pakrym @davidfowl I like to hear your thoughts on this scenario. Tools package is generally a development scenario. Currently (with this fix) `dotnet build` generates binding redirects for any packages containing executables. Do you think that `dotnet publish` (which is a production scenario) must also support generate binding redirects for executable dependencies?

src/Microsoft.DotNet.Compiler.Common/Executable.cs
Sridhar-MS: Fixed

src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
Sridhar-MS: Fixed

test/binding-redirects.Tests/BindingRedirectTests.cs
NTaylorMullen: Eh, if you're going to publish the exe to the output directory then it'd make sense to also publish the binding redirect file (otherwise there's no reason for the exe being published at all). Soo in my opinion the publish scenario isn't important for rc2 but for correctness should be done prior to RTM.

The alternative would be never to publish package based exe's which just seems wrong.

Sridhar-MS: I have created an issue to track this in RTM - #2632

Sridhar-MS:
@dotnet-bot test this please

Sridhar-MS:
@mmitche CI is not getting triggered for this PR? Any idea?

test/binding-redirects.Tests/BindingRedirectTests.cs
```diff
@@ -0,0 +1,287 @@
...
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using Microsoft.DotNet.TestFramework;
+using Microsoft.Extensions.PlatformAbstractions;
+using Xunit;
+using FluentAssertions;
+
+namespace Microsoft.DotNet.Tests
+{
+    public class TestSetupFixture : TestBase
```
piotrpMSFT: This should be its own cs file

test/binding-redirects.Tests/BindingRedirectTests.cs
```diff
@@ -0,0 +1,287 @@
...
+            var configFile = ConfigurationManager.OpenExeConfiguration(exePath);
+            foreach (var appSetting in ExpectedAppSettings)
+            {
+                var value = configFile.AppSettings.Settings[appSetting.Key];
+                value.Should().NotBeNull($"AppSetting with key '{appSetting.Key}' not found in config file.");
+                value.Value.Should().Be(appSetting.Value, $"For AppSetting '{appSetting.Key}' - Expected Value '{appSetting.Value}', Actual '{ value.Value}'");
+            }
+        }
+
+        [Fact]
```
piotrpMSFT: Can we have the tests above the infra?

test/binding-redirects.Tests/BindingRedirectTests.cs
```diff
@@ -0,0 +1,287 @@
...
+        {
+            var redirects = GetRedirects(_appWithConfigPublishOutput);
+            VerifyRedirects(redirects, ExpectedBindingsAppWithConfig);
+            VerifyAppSettings(_appWithConfigPublishOutput);
+
+            var commandResult = new TestCommand(_appWithConfigPublishOutput)
+                                    .Execute();
+            commandResult.Should().Pass();
+        }
+    }
```
piotrpMSFT: We need two more tests. We should invoke the app via the Command Resolver, ideally through the dotnet-dependency-tool-invoker, and show that the binding redirects are working correctly in both the config and no-config case.

piotrpMSFT:
Once we add the end-to-end scenario tests, :shipit:

Sridhar-MS:

piotrpMSFT:
@Petermarcu approved this offline.

Issue #2552, opened 2016-04-18
Typo in TFM gives Microsoft.CSharp error


Ciantic:
Put this in the project.json:

```
{
    "compilationOptions": {
        "emitEntryPoint": true
    },
    "dependencies": {
        "Microsoft.AspNetCore.Mvc.Core": "1.0.0-*",
        "Microsoft.AspNetCore.Mvc.DataAnnotations": "1.0.0-*",
        "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*"
    },
    "frameworks": {
        "net999": {}                                 <-- Offending Line
    },
    "runtimes": {
        "win81-x64": {}
    },
    "version": "1.0.0"
}
```

NuGet.config
```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="AspNetCI" value="https://www.myget.org/F/aspnetcirelease/api/v3/index.json" />
    <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```

Startup.cs
```C#
using System.Reflection;
using Microsoft.AspNetCore.Builder;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Configuration;
using Microsoft.AspNetCore.Hosting;
using System.IO;

namespace TestApp
{
    public class Startup
    {
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvcCore();
        }
        
        public void Configure(IApplicationBuilder app)
        {
            app.UseMvcWithDefaultRoute();
        }
		
        public static void Main(string[] args)
        {
            var host = new WebHostBuilder()
                        .UseKestrel()
                        .UseDefaultHostingConfiguration(args)
                        .UseStartup<Startup>()
                        .Build();

            host.Run();
        }
    }
}
```

```
dotnet restore
dotnet run
An item with the same key has already been added. Key: Microsoft.CSharp
```

So if you use e.g. net462 without having it, dotnet gives the above non-descriptive error.

```
dotnet --version
1.0.0-rc2-002416
```

Ciantic:
@piotrpMSFT this is not *just* "net999" it also requires that there is dependency: `"Microsoft.AspNetCore.Mvc.DataAnnotations": "1.0.0-*",` otherwise you get different error.

davidfowl:
I can repro.

davidfowl:
[...]

PR #2597, opened 2016-04-20
Fix duplicate dependency issue

davidfowl:
If a package has the same name as a framework assembly in the dependency
graph, we usually replace it with the framework assembly if the package
provides no assets. If the framework assembly wasn't resolved, it would
skip this logic and end up adding dupes to the list, which blows up later on.
This is a tactical fix to solve the issue, we need to do some more thinking
to determine how we want to resolve conflicts between framework assemblies,
packages and dlls with the same name.

#2552

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2597)

### DIFF
--- a/test/Microsoft.DotNet.ProjectModel.Tests/PackageDependencyProviderTests.cs
+++ b/test/Microsoft.DotNet.ProjectModel.Tests/PackageDependencyProviderTests.cs
@@ -65,5 +65,18 @@ namespace Microsoft.DotNet.ProjectModel.Tests
 
             Assert.Equal(4, context.RootProject.Dependencies.Count());
         }
+        
+        [Fact]
+        public void NoDuplicateReferencesWhenFrameworkMissing()
+        {
+            var instance = TestAssetsManager.CreateTestInstance("TestMicrosoftCSharpReferenceMissingFramework")
+                                            .WithLockFiles();
+
+            var context = new ProjectContextBuilder().WithProjectDirectory(instance.TestRoot)
+                                                     .WithTargetFramework("net99")
+                                                     .Build();
+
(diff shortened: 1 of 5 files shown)


troydai:
Related: https://github.com/dotnet/cli/issues/2602

davidfowl:

pakrym:
:shipit:

test/Microsoft.DotNet.ProjectModel.Tests/PackageDependencyProviderTests.cs
```diff
@@ -65,5 +65,18 @@ public void TestDuplicateDefaultDesktopReferences(string sampleName, bool withLo
...
+        public void NoDuplicateReferencesWhenFrameworkMissing()
+        {
+            var instance = TestAssetsManager.CreateTestInstance("TestMicrosoftCSharpReferenceMissingFramework")
+                                            .WithLockFiles();
+
+            var context = new ProjectContextBuilder().WithProjectDirectory(instance.TestRoot)
+                                                     .WithTargetFramework("net99")
+                                                     .Build();
+
+            context.LibraryManager.GetLibraries().ToDictionary(l => l.Identity.Name);
```
eerhardt: Should this test assert something?

pakrym: `ToDictionary` is assert that there is no duplications, maybe comment is needed here.

davidfowl: Seems xunit removed `Assert.DoesNotThrow`, which doesn't do anything but it looks nice

PR #2601, opened 2016-04-20
Fixed broken Equals

bkoelman:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2601)

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/AnalyzerOptions.cs
+++ b/src/Microsoft.DotNet.ProjectModel/AnalyzerOptions.cs
@@ -25,13 +25,17 @@ namespace Microsoft.DotNet.ProjectModel
 
         public override bool Equals(object obj)
         {
-            if (obj == null)
+            if (ReferenceEquals(obj, null))
             {
                 return false;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.ProjectModel/AnalyzerOptions.cs
```diff
@@ -25,13 +25,17 @@ public class AnalyzerOptions
...
             {
                 return false;
             }
+            if (ReferenceEquals(obj, this))
+            {
+                return true;
+            }
 
             var options = obj as AnalyzerOptions;
-            return obj != null && (this == options);
```
eerhardt: Fixing the `options != null` part is a good change, but I think the second half should continue being `(this == options)`.  The reasoning being that the "rules" for what makes two AnalyzerOptions equal should be in a single spot.  Here the rule that the "LanguageIds should be the same" is duplicated into 2 spots.

bkoelman:
Replaced by https://github.com/dotnet/cli/pull/2998.

PR #2603, opened 2016-04-20
Reduce size of test names to fix long path breaks on VSO builds

brthor:
cc @piotrpMSFT @Sridhar-MS @eerhardt 

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2603)

### DIFF
--- a/test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectDependenciesCommandFactory.cs
+++ b/test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectDependenciesCommandFactory.cs
@@ -25,7 +25,7 @@ namespace Microsoft.DotNet.Cli.Utils.Tests
         private static readonly NuGetFramework s_desktopTestFramework = FrameworkConstants.CommonFrameworks.Net451;  
         
         [WindowsOnlyFact]
-        public void It_resolves_desktop_apps_from_the_output_directory_using_Debug_configuration_when_configuration_is_null()
+        public void It_resolves_desktop_apps_defaulting_to_Debug_Configuration()
         {
             var configuration = "Debug";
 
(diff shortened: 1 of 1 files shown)


eerhardt:
:shipit:

brthor:
This will fix the current issues we are seeing but I'm wondering if we should change the key for test asset copying from test method names, to something which has a consistent length.

Thoughts on this?

eerhardt:
> This will fix the current issues we are seeing but I'm wondering if we should change the key for test asset copying from test method names, to something which has a consistent length.

I think that could break things like multiple tests running at the same time that use the same test asset. If you don't isolate them per test, they could be stomping on each other. (I remember hearing a rumor that @Sridhar-MS is working on making our tests fully parallel.)

brthor:
> I think that could break things like multiple tests running at the same time that use the same test asset. If you don't isolate them per test, they could be stomping on each other. (I remember hearing a rumor that @Sridhar-MS is working on making our tests fully parallel.)

It need not be the same name it's using to isolate tests, it could be something like the truncated method name, with collision handling (happen _1 or similar)

eerhardt:
> It need not be the same name it's using to isolate tests, it could be something like the truncated method name, with collision handling (happen _1 or similar)

Yes, I think something like that could work.  But I'd think that would be work for RTM. Your current change is all that's needed for RC2.

brthor:
Fair enough, just want to get this on Folks' mind.

Sridhar-MS:
When creating the `TestInstance` you can specify your own name too - https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestAssetsManager.cs#L87

PR #2594, opened 2016-04-20
Remove System.CommandLine and Microsoft.Extensions.CommandLineUtils.Sources dependencies

eerhardt:
- Checking in the source files for Microsoft.Extensions.CommandLineUtils.Sources instead of relying on the NuGet package. There were two changes to these files on the 1st commit:

 1. Change the copyright header.  (Working with @richlander and legal to exactly how we should approach this).
 2. Change the namespace to Microsoft.DotNet.Cli.CommandLine.

 No actual command line utils logic changed in these files in the first commit.

- Removed all usages of System.CommandLine in the CLI and replaced them with the copied Microsoft.DotNet.Cli.CommandLine APIs.
 - This required some enhancements to the command line utils code. You can see all the enhancements in the 2nd commit.
 - The 3rd commit is uptaking the new enhancements.

Some notes:
- I changed resgen to not use a command line util at all, and instead compile explicitly passes the parameters. This is because resgen can't be invoked by a user at all, so there was no reason to write and read an .rsp file in this scenario. If we need to add a command line interface to resgen, it can be added on top of the current code in the future.
- F# compiler shares the CommandLine files and two common compiler utility files with the C# compiler. This is needed to keep our CommandLine APIs internal. If/when F# moves to a different repo, these files will need to be copied.

Fix #2526 

This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2594)

PR #2606, opened 2016-04-20
Move ASP.NET package references to rc2-20581 from the aspnetrelease feed.

### LINKED ISSUE #2567: CLI must reference the aspnet packages from 'aspnetrelease' feed
@eerhardt The 1.0.0-rc2* aspnet packages in the `cli-deps` feed seem to be from `aspnetcidev` (the assemblies are not signed correctly). We must be picking those packages from the `aspnetrelease` feed.

eerhardt:
The places where I removed a dependency is because a lower package (InternalAbstrations) was already bringing in that dependency.

Fix #2567

@Sridhar-MS

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2606)

### DIFF
--- a/scripts/Microsoft.DotNet.Cli.Build.Framework/project.json
+++ b/scripts/Microsoft.DotNet.Cli.Build.Framework/project.json
@@ -3,7 +3,7 @@
   "dependencies": {
     "NETStandard.Library": "1.5.0-rc2-24018",
     "System.Diagnostics.Process": "4.1.0-rc2-24018",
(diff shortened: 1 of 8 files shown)


eerhardt:
FYI - I didn't update CommandLineUtils because my other PR #2594 is removing that dependency. I didn't update the other "Sources" dependencies because we should only be getting the shared source files from them, and not the actual .dll.

piotrpMSFT:
:shipit:

brthor:
:shipit:

PR #2614, opened 2016-04-20
Speedup build of CLI repo

pakrym:
Remove excessive `dotnet restore` and `dotnet build` calls from scripts.

This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2614)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -87,18 +87,21 @@ namespace Microsoft.DotNet.Cli.Build
         {
             string currentRid = GetRuntimeId();
             var buildVersion = c.BuildContext.Get<BuildVersion>("BuildVersion");
-
+            PrepareDummyRuntimeNuGetPackage(
+                            DotNetCli.Stage0,
+                            buildVersion.HostNuGetPackageVersion,
+                            Dirs.CorehostDummyPackages);
             foreach (var hostPackageId in HostPackages)
             {
                 foreach (var rid in HostPackageSupportedRids)
                 {
-                    if (! rid.Equals(currentRid))
+                    if (!rid.Equals(currentRid))
                     {
                         CreateDummyRuntimeNuGetPackage(
-                            DotNetCli.Stage0, 
+                            DotNetCli.Stage0,
                             hostPackageId,
-                            rid, 
-                            buildVersion.HostNuGetPackageVersion, 
+                            rid,
+                            buildVersion.HostNuGetPackageVersion,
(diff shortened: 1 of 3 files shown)


pakrym:
@davidfowl @anurse @brthor

pakrym:
@dotnet-bot test OSX x64 Release Build  please

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -384,11 +384,28 @@ private static void CreateDummyRuntimeNuGetPackage(DotNetCli dotnet, string base
...
+            projectJson.Append("  \"dependencies\": { \"NETStandard.Library\": \"1.5.0-rc2-24008\" },");
+            projectJson.Append("  \"frameworks\": { \"netcoreapp1.0\": { \"imports\": [\"netstandard1.5\", \"dnxcore50\"] } },");
+            projectJson.Append("}");
+
+            var tempPjDirectory = Path.Combine(Dirs.Intermediate, "dummyNuGetPackageIntermediate");
+            var tempPjFile = Path.Combine(tempPjDirectory, "project.json");
+
+            File.WriteAllText(tempPjFile, projectJson.ToString());
 
             dotnet.Pack(
```
eerhardt: Do we even need to pack each project?  What if we just made 1 unplug and then copied the same file 7 times for all the other rids?

eerhardt: *nupkg

scripts/dotnet-cli-build/TestTargets.cs
```diff
@@ -162,17 +162,11 @@ public static BuildTargetResult BuildTestAssetPackages(BuildTargetContext c)
                 var fullPath = Path.Combine(c.BuildContext.BuildDirectory, relativePath.Replace('/', Path.DirectorySeparatorChar));
                 c.Info($"Packing: {fullPath}");
 
-                // build and ignore failure, so net451 fail on non-windows doesn't crash the build
```
eerhardt: :+1:

eerhardt:
:shipit:

scripts/dotnet-cli-build/CompileTargets.cs
eerhardt: (If only we could restore for just the RID we were currently on, we wouldn't need to do this at all. ;))

pakrym: NuGet reads `.nuspec` inside of `.nupkg` so this would not work unfortunately.

Issue #2489, opened 2016-04-14
CLI uses typeof(object).GetTypeInfo().Assembly.Location to determine FX location


schellap:
If CoreCLR gets serviced, mscorlib's path will be in the servicing root and not the actual FX location.

## Steps to reproduce
`copy /y %USERPROFILE%\.nuget\packages\runtime.win7-x64.Microsoft.NETCore.Runtime.CoreCLR\[Matching CLI Version]\ "C:\Program Files (x86)\dotnetextensions\pkgs\runtime.win7-x64.Microsoft.NETCore.Runtime.CoreCLR\[Matching CLI Version]\"`

`dotnet publish`

## Expected  behavior
FX dir determination should be derived from property `APP_CONTEXT_DEPS_FILES` and path containing Microsoft.NETCore.App.deps.json. See:

https://github.com/dotnet/cli/blob/803fef6a8b4838afd0e5f77e9e342d3e616d6c96/src/Microsoft.Extensions.DependencyModel/DependencyContextLoader.cs#L156

## Actual behavior
https://github.com/dotnet/cli/blob/ac4c6702d5d345dc9c397f80290f105fedebe98b/src/Microsoft.DotNet.Cli.Utils/Muxer.cs#L38

## Environment data
`dotnet --info` output:

schellap:
@brthor @piotrpMSFT @Sridhar-MS @gkhanna79

Petermarcu:
@piotrpMSFT , who should pick this one up?

schellap:
I can pick this one...

schellap:
This would require host change to plumb the fx dir.

gkhanna79:
I believe there are two changes required:

1) Add the FX dir to AppContext - @schellap Please file an issue to fi this.
2) Plumb the CLI to consume the FX dir - which is this issue. @brthor Should look into this.

brthor:
Once we have this in AppContext, this should be a one line change:

https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/CoreHost.cs#L31

I believe @sokket was picking this one up?

schellap:
No issues guys, I'll make the fix.

gkhanna79:
Filed https://github.com/dotnet/cli/issues/2558 against @schellap

piotrpMSFT:
This change needs to account for a number of codepaths
 - [ ] Muxer.cs
 - [ ] CoreHost.cs

[...]

PR #2617, opened 2016-04-21
Determine FX dir using FX_DEPS_FILE

schellap:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2617)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/CoreHost.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/CoreHost.cs
@@ -28,7 +28,8 @@ namespace Microsoft.DotNet.Cli.Utils
             {
                 if (_hostDir == null)
                 {
-                    _hostDir = Path.GetDirectoryName(typeof(object).GetTypeInfo().Assembly.Location);
+                    var fxDepsFile = Muxer.GetDataFromAppDomain("FX_DEPS_FILE");
+                    _hostDir = Path.GetDirectoryName(fxDepsFile);
                 }
 
                 return _hostDir;
--- a/src/Microsoft.DotNet.Cli.Utils/Muxer.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/Muxer.cs
@@ -33,11 +33,24 @@ namespace Microsoft.DotNet.Cli.Utils
             }
         }
 
+        public static string GetDataFromAppDomain(string propertyName)
+        {
+            var appDomainType = typeof(object).GetTypeInfo().Assembly?.GetType("System.AppDomain");
+            var currentDomain = appDomainType?.GetProperty("CurrentDomain")?.GetValue(null);
+            var deps = appDomainType?.GetMethod("GetData")?.Invoke(currentDomain, new[] { propertyName });
+
+            return deps as string;
+        }
+
         private bool TryResolveMuxerFromParentDirectories()
         {
-            var appBase = new FileInfo(typeof(object).GetTypeInfo().Assembly.Location);
-            var muxerDir = appBase.Directory?.Parent?.Parent?.Parent;
+            var fxDepsFile = GetDataFromAppDomain("FX_DEPS_FILE");
+            if (string.IsNullOrEmpty(fxDepsFile))
+            {
+                return false;
+            }
 
+            var muxerDir = new FileInfo(fxDepsFile).Directory?.Parent?.Parent?.Parent;
             if (muxerDir == null)
             {
                 return false;


schellap:
@brthor, PTAL.

schellap:

src/Microsoft.DotNet.Cli.Utils/Muxer.cs
```diff
@@ -33,11 +33,24 @@ public Muxer()
...
+
+            return deps as string;
+        }
+
         private bool TryResolveMuxerFromParentDirectories()
         {
-            var appBase = new FileInfo(typeof(object).GetTypeInfo().Assembly.Location);
-            var muxerDir = appBase.Directory?.Parent?.Parent?.Parent;
+            var fxDepsFile = GetDataFromAppDomain("FX_DEPS_FILE");
+            if (fxDepsFile == null)
```
schellap: Use null or empty...

src/Microsoft.DotNet.Cli.Utils/CoreHost.cs
```diff
@@ -28,7 +28,8 @@ private static string HostDir
             {
                 if (_hostDir == null)
                 {
-                    _hostDir = Path.GetDirectoryName(typeof(object).GetTypeInfo().Assembly.Location);
+                    var fxDepsFile = Muxer.GetDataFromAppDomain("FX_DEPS_FILE");
+                    _hostDir = Path.GetDirectoryName(fxDepsFile);
```
schellap: Throw exception if null or empty.

schellap:
Fixes #2489

schellap:
Build please

schellap:
@piotrpMSFT, this is the last RC bug. I am going to update CLI builds to use a known Muxer next.

schellap:
@Sridhar-MS could you PTAL?

src/Microsoft.DotNet.Cli.Utils/Muxer.cs
```diff
@@ -33,11 +33,24 @@ public Muxer()
             }
         }
 
+        public static string GetDataFromAppDomain(string propertyName)
+        {
+            var appDomainType = typeof(object).GetTypeInfo().Assembly?.GetType("System.AppDomain");
+            var currentDomain = appDomainType?.GetProperty("CurrentDomain")?.GetValue(null);
+            var deps = appDomainType?.GetMethod("GetData")?.Invoke(currentDomain, new[] { propertyName });
```
Sridhar-MS: Why use reflection? Why not `AppDomain.CurrentDomain`?

Sridhar-MS:
LGTM modulo one comment

src/Microsoft.DotNet.Cli.Utils/Muxer.cs
```diff
@@ -33,11 +33,24 @@ public Muxer()
...
+
+            return deps as string;
+        }
+
         private bool TryResolveMuxerFromParentDirectories()
         {
-            var appBase = new FileInfo(typeof(object).GetTypeInfo().Assembly.Location);
-            var muxerDir = appBase.Directory?.Parent?.Parent?.Parent;
+            var fxDepsFile = GetDataFromAppDomain("FX_DEPS_FILE");
```
Sridhar-MS: One more question. If the current app is executing in `standalone` app what is the value that is returned? Empty?

schellap: It is empty. This property is always set to a directory or empty string. An app executing in standalone mode should not use this function, if it did it would return false or be prepared to deal with exceptions.

src/Microsoft.DotNet.Cli.Utils/Muxer.cs
schellap: I couldn't find a dependency that supports both net45 and netstandard1.5 that can do `AppContext.GetData`. `System.AppContext` supports net46.

src/Microsoft.DotNet.Cli.Utils/Muxer.cs
brthor: What is the location of the FX_DEPS_FILE, the shared fx directory?

schellap: Yes, it is the path to Microsoft.NETCore.App.deps.json

brthor: okay cool, makes sense

schellap: Note however, standalone apps if they call into this, they would get "Could not resolve muxer"

brthor: Standalone apps can still resolve from the path so that should be fine.

PR #2619, opened 2016-04-21
Cache inputs for globbing change detection

cdmihai:
Build caches the inputs and outputs into a `.IncrementalCache` file that's kept in the intermediate output directory. It uses it to find out when the globbing patterns change, thus making incremental react to file additions, deletion, and renames.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2619)

### DIFF
--- a/src/Microsoft.DotNet.Compiler.Common/ProjectContextExtensions.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/ProjectContextExtensions.cs
@@ -37,6 +37,12 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
             return Path.Combine(intermediatePath, ".SDKVersion");
         }
 
+        public static string IncrementalCacheFile(this ProjectContext context, string configuration, string buildBasePath, string outputPath)
+        {
+            var intermediatePath = context.GetOutputPaths(configuration, buildBasePath, outputPath).IntermediateOutputDirectoryPath;
+            return Path.Combine(intermediatePath, ".IncrementalCache");
+        }
+
         // used in incremental compilation for the key file
         public static CommonCompilerOptions ResolveCompilationOptions(this ProjectContext context, string configuration)
         {
--- a/src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
+++ b/src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
@@ -11,6 +11,9 @@ namespace Microsoft.DotNet.TestFramework
 {
     public class TestInstance
     {
+        // made tolower because the rest of the class works with normalized tolower strings
+        private static readonly IEnumerable<string> BuildArtifactBlackList = new List<string>() {".IncrementalCache", ".SDKVersion"}.Select(s => s.ToLower()).ToArray();
+
         private string _testDestination;
         private string _testAssetRoot;
 
@@ -105,8 +108,13 @@ namespace Microsoft.DotNet.TestFramework
                                  .Where(file =>
                                  {
                                      file = file.ToLower();
-                                     return file.Contains($"{Path.DirectorySeparatorChar}bin{Path.DirectorySeparatorChar}") 
+
+                                     var isArtifact = file.Contains($"{Path.DirectorySeparatorChar}bin{Path.DirectorySeparatorChar}") 
                                             || file.Contains($"{Path.DirectorySeparatorChar}obj{Path.DirectorySeparatorChar}");
+
+                                     var isBlackListed = BuildArtifactBlackList.Any(b => file.Contains(b));
+
+                                     return isArtifact && !isBlackListed;
                                  });
 
             foreach (string binFile in binFiles)
--- a/src/dotnet/commands/dotnet-build/CompilerIO.cs
+++ b/src/dotnet/commands/dotnet-build/CompilerIO.cs
@@ -2,18 +2,43 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.DotNet.Tools.Build
 {
-    internal struct CompilerIO
+    internal class CompilerIO
     {
-        public readonly List<string> Inputs;
-        public readonly List<string> Outputs;
+        public readonly IEnumerable<string> Inputs;
+        public readonly IEnumerable<string> Outputs;
 
-        public CompilerIO(List<string> inputs, List<string> outputs)
+        public CompilerIO(IEnumerable<string> inputs, IEnumerable<string> outputs)
         {
             Inputs = inputs;
             Outputs = outputs;
         }
+
+        public DiffResult DiffInputs(CompilerIO other)
+        {
+            var myInputSet = new HashSet<string>(Inputs);
+            var otherInputSet = new HashSet<string>(other.Inputs);
+
+            var additions = myInputSet.Except(otherInputSet);
+            var deletions = otherInputSet.Except(myInputSet);
+
+            return new DiffResult(additions, deletions);
+        }
+
+        internal class DiffResult
+        {
+            public IEnumerable<string> Additions { get; private set; }
+            public IEnumerable<string> Deletions { get; private set; }
+
+            public DiffResult(IEnumerable<string> additions, IEnumerable<string> deletions)
+            {
+                Additions = additions;
+                Deletions = deletions;
+            }
+        }
+
     }
 }
--- a/src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
+++ b/src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
@@ -21,14 +21,17 @@ namespace Microsoft.DotNet.Tools.Build
         private readonly CompilerIOManager _compilerIOManager;
         private readonly ScriptRunner _scriptRunner;
         private readonly DotNetCommandFactory _commandFactory;
+        private readonly IncrementalManager _incrementalManager;
 
         public DotNetProjectBuilder(BuildCommandApp args)
         {
             _args = args;
+
             _preconditionManager = new IncrementalPreconditionManager(
                 args.ShouldPrintIncrementalPreconditions,
                 args.ShouldNotUseIncrementality,
                 args.ShouldSkipDependencies);
+
             _compilerIOManager = new CompilerIOManager(
                 args.ConfigValue,
                 args.OutputValue,
@@ -36,7 +39,19 @@ namespace Microsoft.DotNet.Tools.Build
                 args.GetRuntimes(),
                 args.Workspace
                 );
+
+            _incrementalManager = new IncrementalManager(
+                this,
+                _compilerIOManager,
+                _preconditionManager,
+                _args.ShouldSkipDependencies,
+                _args.ConfigValue,
+                _args.BuildBasePathValue,
+                _args.OutputValue
+                );
+
             _scriptRunner = new ScriptRunner();
+
             _commandFactory = new DotNetCommandFactory();
         }
 
@@ -52,7 +67,7 @@ namespace Microsoft.DotNet.Tools.Build
                     Directory.CreateDirectory(parentDirectory);
                 }
 
-                string content = ComputeCurrentVersionFileData();
+                string content = DotnetFiles.ReadAndInterpretVersionFile();
 
                 File.WriteAllText(projectVersionFile, content);
             }
@@ -60,14 +75,6 @@ namespace Microsoft.DotNet.Tools.Build
             {
                 Reporter.Verbose.WriteLine($"Project {project.GetDisplayName()} was not stamped with a CLI version because the version file does not exist: {DotnetFiles.VersionFile}");
             }
-        }
-
-        private static string ComputeCurrentVersionFileData()
-        {
-            var content = File.ReadAllText(DotnetFiles.VersionFile);
-            content += Environment.NewLine;
-            content += PlatformServices.Default.Runtime.GetRuntimeIdentifier();
-            return content;
         }
 
         private void PrintSummary(ProjectGraphNode projectNode, bool success)
@@ -83,17 +90,6 @@ namespace Microsoft.DotNet.Tools.Build
             Reporter.Output.WriteLine();
         }
 
-        private void CreateOutputDirectories()
-        {
-            if (!string.IsNullOrEmpty(_args.OutputValue))
-            {
-                Directory.CreateDirectory(_args.OutputValue);
-            }
-            if (!string.IsNullOrEmpty(_args.BuildBasePathValue))
-            {
-                Directory.CreateDirectory(_args.BuildBasePathValue);
-            }
-        }
 
         private void CopyCompilationOutput(OutputPaths outputPaths)
         {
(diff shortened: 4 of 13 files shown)


test/dotnet-build.Tests/IncrementalTests.cs
```diff
@@ -100,20 +100,77 @@ public void TestNoVersionFile()
...
+            var buildResult = BuildProject();
+            buildResult.Should().HaveCompiledProject(MainProject, _appFrameworkFullName);
+
+            var sourceFile = Path.Combine(GetProjectDirectory(MainProject), "Program2.cs");
+            File.Delete(sourceFile);
+            Assert.False(File.Exists(sourceFile));
+
+            // second build; should get rebuilt since we deleted a source file
+            buildResult = BuildProject();
+            buildResult.Should().HaveCompiledProject(MainProject, _appFrameworkFullName);
```
cdmihai: Add a third build to assert cache was updated and thus the build skipped (must have missed have lost that change during the merge).

test/dotnet-build.Tests/IncrementalTests.cs
```diff
@@ -100,20 +100,77 @@ public void TestNoVersionFile()
...
+
+            var sourceFile = Path.Combine(GetProjectDirectory(MainProject), "Program2.cs");
+            var destinationFile = Path.Combine(Path.GetDirectoryName(sourceFile), "ProgramNew.cs");
+            File.Move(sourceFile, destinationFile);
+            Assert.False(File.Exists(sourceFile));
+            Assert.True(File.Exists(destinationFile));
+
+            // second build; should get rebuilt since we renamed a source file
+            buildResult = BuildProject();
+            buildResult.Should().HaveCompiledProject(MainProject, _appFrameworkFullName);
```
cdmihai: Add third build that should skip

cdmihai:
@pakrym @anurse @davidfowl

cdmihai:

piotrpMSFT:
@eerhardt please review

test/dotnet-build.Tests/IncrementalTests.cs
```diff
@@ -100,17 +101,78 @@ public void TestNoVersionFile()
...
+
+            // third build; incremental cache should have been regenerated and project skipped
             BuildProject().Should().HaveSkippedProjectCompilation(MainProject, _appFrameworkFullName);
+        }
+
+        [Fact]
+        public void TestRebuilDeletedSourceAfterCliChanged()
```
eerhardt: Type-o in the name. TestRebuil

test/dotnet-build.Tests/IncrementalTests.cs
```diff
@@ -100,17 +101,78 @@ public void TestNoVersionFile()
...
+
+            // second build; should get rebuilt since we deleted a source file
+            BuildProject().Should().HaveCompiledProject(MainProject, _appFrameworkFullName);
+
+            // third build; incremental cache should have been regenerated and project skipped
+            BuildProject().Should().HaveSkippedProjectCompilation(MainProject, _appFrameworkFullName);
+        }
+
+        [Fact]
+        public void TestRebuilRenamedSource()
```
eerhardt: same type-o TestRebuil

src/dotnet/commands/dotnet-build/IncrementalResult.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.Tools.Build
+{
+    internal class IncrementalResult
+    {
+        public static IncrementalResult DoesNotNeedRebuild = new IncrementalResult(false, "", Enumerable.Empty<string>());
```
eerhardt: readonly?

src/dotnet/commands/dotnet-build/IncrementalResult.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        private IncrementalResult(bool needsRebuilding, string reason, IEnumerable<string> items)
+        {
+            NeedsRebuilding = needsRebuilding;
+            Reason = reason;
+            Items = items;
+        }
+
+        public IncrementalResult(string reason)
+            : this(true, reason, Enumerable.Empty<string>())
```
eerhardt: (minor) This is totally a personal preference kind of thing, but just throwing it out there:

I like having the defaults of booleans to be 'false'.  It makes a very consistent experience for consumers of classes that take bools.  So usually I invert the name so the default value is false.  In this case "SkipRebuild".

src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
```diff
@@ -105,8 +107,13 @@ public TestInstance WithBuildArtifacts()
                                  .Where(file =>
                                  {
                                      file = file.ToLower();
-                                     return file.Contains($"{Path.DirectorySeparatorChar}bin{Path.DirectorySeparatorChar}") 
+
+                                     var isArtefact = file.Contains($"{Path.DirectorySeparatorChar}bin{Path.DirectorySeparatorChar}")
```
eerhardt: should this be isArtifact?

cdmihai: http://grammarist.com/spelling/artefact-artifact/

I'll switch to artifact, since the CLI is a US product :)

cdmihai:
@eerhardt 
Renamed NeedsRebuild to SkipBuild https://github.com/dotnet/cli/pull/2619/commits/db3f626e30c97a31d808c1b4684c9a468d245e7e

Does this look better or worse? I can't tell :)

cdmihai:
@dotnet-bot test this please

cdmihai:
@dotnet-bot test this please

eerhardt:
> Tried to rename NeedsRebuild to SkipRebuild, but it involves inverting so many conditions that I gave up :)

Totally fine.  It's an internal api, so I'm not pushing real hard for it.  Just wanted to call it out.

src/Microsoft.DotNet.Compiler.Common/ProjectContextExtensions.cs
```diff
@@ -37,6 +37,12 @@ public static string GetSDKVersionFile(this ProjectContext context, string confi
             return Path.Combine(intermediatePath, ".SDKVersion");
         }
 
+        public static string IncrementalCacheFile(this ProjectContext context, string configuration, string buildBasePath, string outputPath)
+        {
+            var intermediatePath = context.GetOutputPaths(configuration, buildBasePath, outputPath).IntermediateOutputDirectoryPath;
+            return Path.Combine(intermediatePath, ".IncrementalCache");
```
eerhardt: Is there a good place to put ".IncrementalCache" as a const string somewhere?

src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
```diff
@@ -105,8 +107,13 @@ public TestInstance WithBuildArtifacts()
                                  .Where(file =>
                                  {
                                      file = file.ToLower();
-                                     return file.Contains($"{Path.DirectorySeparatorChar}bin{Path.DirectorySeparatorChar}") 
+
+                                     var isArtifact = file.Contains($"{Path.DirectorySeparatorChar}bin{Path.DirectorySeparatorChar}") 
                                             || file.Contains($"{Path.DirectorySeparatorChar}obj{Path.DirectorySeparatorChar}");
+
+                                     var isBlackListed = BuildArtifactBlackList.Any(b => file.Contains(b.ToLower()));
```
eerhardt: Is doing a ToLower here correct on all platforms?  Some file systems are case sensitive.

eerhardt: Also, if you are going to ToLower it, you might as well do that in the definition of BuildArtifactBlackList, so you don't pay the cost to lower the constant string every time this is called.

src/dotnet/commands/dotnet-build/CompilerIO.cs
```diff
@@ -2,18 +2,43 @@
...
+            var myInputSet = new HashSet<string>(Inputs);
+            var otherInputSet = new HashSet<string>(other.Inputs);
+
+            var additions = myInputSet.Except(otherInputSet);
+            var deletions = otherInputSet.Except(myInputSet);
+
+            return new DiffResult(additions, deletions);
+        }
+
+        internal struct DiffResult
```
eerhardt: Any specific reason why this is a struct?

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -150,118 +146,39 @@ protected override CompilationResult RunCompile(ProjectGraphNode projectNode)
...
-
-            // find the output with the earliest write time
-            var minOutputPath = compilerIO.Outputs.First();
-            var minDateUtc = File.GetLastWriteTimeUtc(minOutputPath);
-
-            foreach (var outputPath in compilerIO.Outputs)
-            {
-                if (File.GetLastWriteTimeUtc(outputPath) >= minDateUtc)
+                Reporter.Output.WriteLine($"Project {projectName} will be compiled because {result.Reason}");
+                foreach (var item in result.Items)
```
eerhardt: Maybe it makes sense to check for Verbose before even getting into this loop?  That way if Verbose isn't turned on, you don't need to loop through the Items at all.

src/dotnet/commands/dotnet-build/IncrementalCache.cs
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+
```
eerhardt: (nit) double new line

src/dotnet/commands/dotnet-build/IncrementalCache.cs
```diff
@@ -0,0 +1,93 @@
...
+        private const string OutputsKeyNane = "outputs";
+
+        public CompilerIO CompilerIO { get; }
+
+        public IncrementalCache(CompilerIO compilerIO)
+        {
+            CompilerIO = compilerIO;
+        }
+
+        public static void WriteToFile(string cacheFile, IncrementalCache cache)
```
eerhardt: Should this just be a instance method?  Then you can just call it cache.WriteToFile(fileName), instead of IncrementalCache.WriteToFile(fileName, cache)

src/dotnet/commands/dotnet-build/IncrementalCache.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                throw new InvalidDataException($"Could not write the incremental cache file: {cacheFile}", e);
+            }
+        }
+
+        private static void CreatePathIfAbsent(string filePath)
+        {
+            var parentDir = Path.GetDirectoryName(filePath);
+
+            if (!File.Exists(parentDir))
```
eerhardt: Directory.Exists?

src/dotnet/commands/dotnet-build/IncrementalManager.cs
```diff
@@ -0,0 +1,201 @@
...
+            }
+
+            result = TimestampsChanged(compilerIO);
+            if (result.NeedsRebuilding)
+            {
+                return result;
+            }
+
+            return IncrementalResult.DoesNotNeedRebuild;
+
```
eerhardt: (nit) unnecessary blank line.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
```diff
@@ -0,0 +1,201 @@
...
+                return result;
+            }
+
+            // check cache against input glob pattern changes
+            var incrementalCacheFile = graphNode.ProjectContext.IncrementalCacheFile(_configuration, _buildBasePath, _outputPath);
+
+            if (!File.Exists(incrementalCacheFile))
+            {
+                // no cache present; cannot tell if anything changed
+                return IncrementalResult.DoesNotNeedRebuild;
```
eerhardt: ?? If we don't find a cache file we say that we don't need to rebuild?  I would think it would be the opposite.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        private IncrementalResult TimestampsChanged(CompilerIO compilerIO)
+        {
+            // find the output with the earliest write time
+            var minDateUtc = File.GetLastWriteTimeUtc(compilerIO.Outputs.First());
+
+            foreach (var outputPath in compilerIO.Outputs)
+            {
+                if (File.GetLastWriteTimeUtc(outputPath) >= minDateUtc)
```
eerhardt: Two calls to File.GetLastWriteTimeUtc(outputPath).  This can be cached in a variable and only checked once.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            return missingItems.Any()
+                ? new IncrementalResult($"expected {itemsType} are missing", missingItems)
+                : IncrementalResult.DoesNotNeedRebuild;
+        }
+
+        private IncrementalResult TimestampsChanged(CompilerIO compilerIO)
+        {
+            // find the output with the earliest write time
+            var minDateUtc = File.GetLastWriteTimeUtc(compilerIO.Outputs.First());
```
eerhardt: What if there are no Outputs?  First() would throw.  Is this guaranteed to always have at least one Output?

eerhardt:
Sorry I reviewed this in two rounds.  I'm done now. :)

Overall looks good.  Just some mostly minor comments.

src/Microsoft.DotNet.Compiler.Common/ProjectContextExtensions.cs
cdmihai: We could all the file names from the intermediary output directory somewhere in utils, or in the common compiler (so there's no cyclic dependencies).

This would be the names for:
- sdk version file,
- incremental cache,
- response files (compile.rsp, compile-csc.rsp)
etc

src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
cdmihai: was doing tolower becase `file` is also made toLower. I could make the blacklist tolower and introduce this implicit dependency that all names are tolower.

src/dotnet/commands/dotnet-build/CompilerIO.cs
cdmihai: No :)

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
cdmihai: Good point

src/dotnet/commands/dotnet-build/IncrementalCache.cs
cdmihai: OMG! that might be a bug.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
cdmihai: Quite a good question. I was also not very happy with the decision, and see pros and cons for either way.

If it _did_ trigger a rebuild when the cache file is absent, it would mean that the incremental manager is implicitly coupled to the precondition that some other piece of code generates it.
When that precondition gets broken and no other piece of code generates that file, then the IncrementalManager will always answer with rebuilding, killing incrementality off. I did not like that implicit assumption.

If it _does not_ trigger a rebuild when the cache file is absent, it puts the responsibility in the hands of the caller, and if the calling code decides not to generate the file, then incremental proceeds with the other heuristics. 

First option is error proof, choosing to be conservative, but kills all incrementality. Second option may lead to false negative builds (turns the feature that this PR introduces off).

There's pros and cons for each side. If you still think it should trigger a rebuild, then I'll change it. The same decision is taken for the .SDKVersion file as well, so I'll change it there as well.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
cdmihai: True. I considered that the perf hit is too insignificant to warrant extra code :) I'll have the enumerable skip once before the loop.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
cdmihai: Initially the check happened just before that line, but I moved it to a previous method that handles incremental logic at the input / output item level (existence, additions, deletions, etc): https://github.com/dotnet/cli/pull/2619/files#diff-cd21e60f3dbf89c5314c79ce85a1b39cR119

I thought about turning each tiny bit of incremental code that does a check into it's own small object implementing some interface and then composing the entire incremental logic in some chain of command  thingie, but it seemed a bit of overkill.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
eerhardt: Going to the file system (or any other I/O) is never an "insignificant" performing action.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
eerhardt: I can see your reasoning now now that it is explained.  Don't change this on account of me, if you think this is the right approach I don't disagree.

Maybe sticking a summary of this reasoning in a comment would help the next person reading the code.

src/dotnet/commands/dotnet-build/IncrementalManager.cs
eerhardt: If you're really going for no extra code, you can invert the condition and remove the continue:

var outputDateUtc = File.GetLastWriteTimeUtc(outputPath);
if (outputDateUtc < minDateUtc)
{
    minDateUtc = outputDateUtc;
}

src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
eerhardt: When things are case sensitive (like file systems in Linux), it is not a good practice to use ToLower when comparing strings.

cdmihai: True. However, looking at the rest of that class, there's toLower everywhere. I'm going to leave it like this and open an issue about it. Maybe there's other parts of the test framework that do that too. On one hand, it gives you cheap normalization, on the other hand, it may hide bugs.

src/Microsoft.DotNet.Compiler.Common/ProjectContextExtensions.cs
cdmihai: I looked into it and it's not worth it. Ideally, there should be an IntermediaryOutputFiles from the OutputCalculator, and a pluginable mechanism where unknown verbs plugin new files (to know about compile-csc.rsp, compile-fsc.rsp. etc). Too much of a change for this PR.

eerhardt: OK - fine with me.  I was just looking to cut down on the number of places these strings are duplicated.

src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
```diff
@@ -11,6 +11,9 @@ namespace Microsoft.DotNet.TestFramework
 {
     public class TestInstance
     {
+        // made tolower because the rest of the class works with normalized tolower strings
+        private static readonly IEnumerable<string> BuildArtifactBlackList = new List<string>() {".IncrementalCache", ".SDKVersion"}.Select(s => s.ToLower());
```
eerhardt: The issue with this code is that every time you enumerate the collection, it will ToLower the strings again and again.  You could call ToArray() at the end to fix this.

eerhardt: I also still want to call out that I don't think calling ToLower on file paths is correct cross-platform, since some platforms have case sensitive file systems.

test/dotnet-build.Tests/out.xml
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
eerhardt: Probably don't want to check this in.

src/Microsoft.DotNet.TestFramework/Microsoft.DotNet.TestFramework.TestInstance.cs
eerhardt: Ah, I see you logged https://github.com/dotnet/cli/issues/2646.  Thanks!  That was the right call here.  The ToLower is outside of the scope of your changes.

eerhardt:
:shipit:

Issue #2520, opened 2016-04-15
Static Caching for ProjectContext


piotrpMSFT:

davidfowl:
- Refactor dotnet build to use the a single build graph structure
- Support passing multiple projects to build (and globs)
- Add ProjectContext caching (use WorkspaceContext)
- Make ProjectContext lighter (based on measurements)
 - Pool strings during lock file parsing
 - LibraryExporter needs to cache exports 
- Pass project context to dotnet-compile

## Measure
- MVC 
- Single hello world console (dotnet new)

piotrpMSFT:
@davidfowl when is this coming in?

piotrpMSFT:
@davidfowl ping

anurse:
I have a PR coming that addresses some of these. We're going to be iterating quickly on this, producing smaller changes that keep working so this may trigger multiple PRs.

anurse:
The bulk of this is now done. We're going to keep investigating build perf and iterating as needed.

PR #2626, opened 2016-04-21
Memory usage improvements in build

anurse:
* Use a WorkspaceContext in dotnet-build to cache project data across
multiple compilations in a single build action
* Dramatically reduce string and object duplication by introducing a
"Symbol Table" that shares instances of NuGetVersion, NuGetFramework,
VersionRange and string across multiple lock-file parses

Test Results:
* Testing was done by compiling Microsoft.AspNetCore.Mvc (and it's
dependencies) and taking memory snapshots after each compilation in
dotMemory
* We used to allocate ~3MB and deallocate ~2.5MB on EACH compilation in
a single build action. This has been reduced to ~120KB
allocated/deallocated
* After introducing WorkspaceContext, total memory usage spiked from 6MB
across the whole build action to about 13MB, introducing the symbol
table dropped it back to about 5-6MB.

Part of fixes for #2520 

/cc @davidfowl @pakrym @troydai (for DTH-related changes) @cdmihai

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2626)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/Command.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/Command.cs
@@ -7,8 +7,9 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
-using NuGet.Frameworks;
+using System.Threading.Tasks;
 using Microsoft.DotNet.ProjectModel;
+using NuGet.Frameworks;
 
 namespace Microsoft.DotNet.Cli.Utils
 {
@@ -131,12 +132,11 @@ namespace Microsoft.DotNet.Cli.Utils
 
             Reporter.Verbose.WriteLine($"Process ID: {_process.Id}");
 
-            var threadOut = _stdOut.BeginRead(_process.StandardOutput);
-            var threadErr = _stdErr.BeginRead(_process.StandardError);
-
+            var taskOut = _stdOut.BeginRead(_process.StandardOutput);
+            var taskErr = _stdErr.BeginRead(_process.StandardError);
             _process.WaitForExit();
-            threadOut.Join();
-            threadErr.Join();
+
+            Task.WaitAll(taskOut, taskErr);
 
             var exitCode = _process.ExitCode;
 
--- a/src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
@@ -1,8 +1,11 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.IO;
-using System.Text;
-using System.Threading;
 using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.Cli.Utils
 {
@@ -17,7 +20,7 @@ namespace Microsoft.DotNet.Cli.Utils
 
         public string CapturedOutput
         {
-            get 
+            get
             {
                 return _capture?.GetStringBuilder()?.ToString();
             }
@@ -43,18 +46,16 @@ namespace Microsoft.DotNet.Cli.Utils
             return this;
         }
 
-        public Thread BeginRead(TextReader reader)
+        public Task BeginRead(TextReader reader)
         {
-            var thread = new Thread(() => Read(reader)) { IsBackground = true };
-            thread.Start();
-            return thread;
+            return Task.Run(() => Read(reader));
         }
 
         public void Read(TextReader reader)
         {
             var bufferSize = 1;
 
-            int readCharacterCount; 
+            int readCharacterCount;
             char currentCharacter;
 
             var buffer = new char[bufferSize];
@@ -93,7 +94,7 @@ namespace Microsoft.DotNet.Cli.Utils
         }
 
         private void WriteLine(string str)
-        { 
+        {
             if (_capture != null)
             {
                 _capture.WriteLine(str);
--- a/src/Microsoft.DotNet.ProjectModel/Graph/LockFileSymbolTable.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Graph/LockFileSymbolTable.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Concurrent;
+using NuGet.Frameworks;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.ProjectModel.Graph
+{
+    public class LockFileSymbolTable
+    {
+        private ConcurrentDictionary<string, NuGetVersion> _versionTable = new ConcurrentDictionary<string, NuGetVersion>(Environment.ProcessorCount * 4, 1000, StringComparer.Ordinal);
+        private ConcurrentDictionary<string, VersionRange> _versionRangeTable = new ConcurrentDictionary<string, VersionRange>(Environment.ProcessorCount * 4, 1000, StringComparer.Ordinal);
+        private ConcurrentDictionary<string, NuGetFramework> _frameworksTable = new ConcurrentDictionary<string, NuGetFramework>(Environment.ProcessorCount * 4, 1000, StringComparer.Ordinal);
+        private ConcurrentDictionary<string, string> _stringsTable = new ConcurrentDictionary<string, string>(Environment.ProcessorCount * 4, 1000, StringComparer.Ordinal);
+
+        public NuGetVersion GetVersion(string versionString) => _versionTable.GetOrAdd(versionString, (s) => NuGetVersion.Parse(s));
+        public VersionRange GetVersionRange(string versionRangeString) => _versionRangeTable.GetOrAdd(versionRangeString, (s) => VersionRange.Parse(s));
+        public NuGetFramework GetFramework(string frameworkString) => _frameworksTable.GetOrAdd(frameworkString, (s) => NuGetFramework.Parse(s));
+        public string GetString(string frameworkString) => _stringsTable.GetOrAdd(frameworkString, frameworkString);
+    }
+}
--- a/src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
@@ -154,39 +154,6 @@ namespace Microsoft.DotNet.ProjectModel
                         .BuildAllTargets();
         }
 
-
-        /// <summary>
-        /// Creates a project context based on existing context but using runtime target
-        /// </summary>
-        /// <param name="context"></param>
-        /// <param name="runtimeIdentifiers"></param>
-        /// <returns></returns>
-
-        public ProjectContext CreateRuntimeContext(IEnumerable<string> runtimeIdentifiers)
-        {
-            // Temporary until we have removed RID inference from NuGet
-            if(TargetFramework.IsCompileOnly)
-            {
-                return this;
-            }
-
-            var context = CreateBuilder(ProjectFile.ProjectFilePath, TargetFramework)
-                .WithRuntimeIdentifiers(runtimeIdentifiers)
-                .WithLockFile(LockFile)
-                .Build();
-
-            if (!context.IsPortable && context.RuntimeIdentifier == null)
-            {
-                // We are standalone, but don't support this runtime
-                var rids = string.Join(", ", runtimeIdentifiers);
-                throw new InvalidOperationException($"Can not find runtime target for framework '{TargetFramework}' compatible with one of the target runtimes: '{rids}'. " +
-                                                    "Possible causes:" + Environment.NewLine +
-                                                    "1. The project has not been restored or restore failed - run `dotnet restore`" + Environment.NewLine +
-                                                    $"2. The project does not list one of '{rids}' in the 'runtimes' section.");
-            }
-            return context;
-        }
-
         public OutputPaths GetOutputPaths(string configuration, string buidBasePath = null, string outputPath = null)
         {
             return OutputPathsCalculator.GetOutputPaths(ProjectFile,
--- a/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
@@ -132,16 +132,42 @@ namespace Microsoft.DotNet.ProjectModel
             EnsureProjectLoaded();
             LockFile = LockFile ?? LockFileResolver(ProjectDirectory);
 
-            if (LockFile != null)
+            if (LockFile != null && LockFile.Targets.Any())
             {
+                var deduper = new HashSet<string>();
                 foreach (var target in LockFile.Targets)
                 {
-                    yield return new ProjectContextBuilder()
-                    .WithProject(Project)
-                    .WithLockFile(LockFile)
-                    .WithTargetFramework(target.TargetFramework)
-                    .WithRuntimeIdentifiers(new[] { target.RuntimeIdentifier })
-                    .Build();
+                    var id = $"{target.TargetFramework}/{target.RuntimeIdentifier}";
+                    if (deduper.Add(id))
+                    {
+                        var builder = new ProjectContextBuilder()
+                            .WithProject(Project)
+                            .WithLockFile(LockFile)
+                            .WithTargetFramework(target.TargetFramework)
+                            .WithRuntimeIdentifiers(new[] { target.RuntimeIdentifier });
+                        if (IsDesignTime)
+                        {
+                            builder.AsDesignTime();
+                        }
+
+                        yield return builder.Build();
+                    }
+                }
+            }
+            else
+            {
+                // Build a context for each framework. It won't be fully valid, since it won't have resolved data or runtime data, but the diagnostics will show that
+                // (Project Model Server needs this)
+                foreach (var framework in Project.GetTargetFrameworks())
+                {
+                    var builder = new ProjectContextBuilder()
+                        .WithProject(Project)
+                        .WithTargetFramework(framework.FrameworkName);
+                    if (IsDesignTime)
+                    {
+                        builder.AsDesignTime();
+                    }
+                    yield return builder.Build();
                 }
             }
         }
@@ -165,8 +191,17 @@ namespace Microsoft.DotNet.ProjectModel
 
             RootDirectory = GlobalSettings?.DirectoryPath ?? RootDirectory;
             PackagesDirectory = PackagesDirectory ?? PackageDependencyProvider.ResolvePackagesPath(RootDirectory, GlobalSettings);
-            ReferenceAssembliesPath = ReferenceAssembliesPath ?? FrameworkReferenceResolver.GetDefaultReferenceAssembliesPath();
-            var frameworkReferenceResolver = new FrameworkReferenceResolver(ReferenceAssembliesPath);
+
+            FrameworkReferenceResolver frameworkReferenceResolver;
+            if (string.IsNullOrEmpty(ReferenceAssembliesPath))
+            {
+                // Use the default static resolver
+                frameworkReferenceResolver = FrameworkReferenceResolver.Default;
+            }
+            else
+            {
+                frameworkReferenceResolver = new FrameworkReferenceResolver(ReferenceAssembliesPath);
+            }
 
             LockFileLookup lockFileLookup = null;
             EnsureProjectLoaded();
@@ -202,7 +237,7 @@ namespace Microsoft.DotNet.ProjectModel
             if (mainProject != null)
             {
                 platformDependency = mainProject.Dependencies
-                    .Where(d =>  d.Type.Equals(LibraryDependencyType.Platform))
+                    .Where(d => d.Type.Equals(LibraryDependencyType.Platform))
                     .Cast<LibraryRange?>()
                     .FirstOrDefault();
             }
@@ -274,7 +309,7 @@ namespace Microsoft.DotNet.ProjectModel
             {
                 var frameworkInfo = Project.GetTargetFramework(TargetFramework);
 
-                if (string.IsNullOrEmpty(ReferenceAssembliesPath))
+                if (frameworkReferenceResolver == null || string.IsNullOrEmpty(frameworkReferenceResolver.ReferenceAssembliesPath))
                 {
                     // If there was an attempt to use reference assemblies but they were not installed
                     // report an error
@@ -373,13 +408,13 @@ namespace Microsoft.DotNet.ProjectModel
             {
                 var library = pair.Value;
 
-                // The System.* packages provide placeholders on any non netstandard platform 
+                // The System.* packages provide placeholders on any non netstandard platform
                 // To make them work seamlessly on those platforms, we fill the gap with a reference
                 // assembly (if available)
                 var package = library as PackageDescription;
-                if (package != null && 
-                    package.Resolved && 
-                    package.HasCompileTimePlaceholder && 
+                if (package != null &&
+                    package.Resolved &&
+                    package.HasCompileTimePlaceholder &&
                     !TargetFramework.IsPackageBased())
                 {
                     var newKey = new LibraryKey(library.Identity.Name, LibraryType.ReferenceAssembly);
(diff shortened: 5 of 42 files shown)


src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -21,6 +21,7 @@ internal class DotNetProjectBuilder : ProjectBuilder
         private readonly CompilerIOManager _compilerIOManager;
         private readonly ScriptRunner _scriptRunner;
         private readonly DotNetCommandFactory _commandFactory;
+        private readonly WorkspaceContext _workspace;
```
cdmihai: Is this used?

anurse: Nope! It was but I removed it since the BuildCommandArgs had it. Good catch :)

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -154,39 +154,6 @@ public static IEnumerable<ProjectContext> CreateContextForEachTarget(string proj
...
 
-
-        /// <summary>
-        /// Creates a project context based on existing context but using runtime target
-        /// </summary>
-        /// <param name="context"></param>
-        /// <param name="runtimeIdentifiers"></param>
-        /// <returns></returns>
-
-        public ProjectContext CreateRuntimeContext(IEnumerable<string> runtimeIdentifiers)
```
cdmihai: I'd still make this available from the project context. Seems more intuitive to me. Maybe it could delegate to the workspace context?  This brings the question, should a ProjectContext know about its WorkspaceContext?

anurse: Long-term, I want to restructure things so that there is a proper hierarchy: Workspace -> Project -> ProjectContext. At that point, it'll be much easier to do this. Until then, I was really concerned about not duplicating the code. It makes it harder to generate a runtime context, but even this was just a helper method around ProjectContextBuilder.

I'd rather just leave the code in one place for RC2 and then restructure things post-RC2

piotrpMSFT:
@Sridhar-MS can you review this?

src/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -131,12 +132,11 @@ public CommandResult Execute()
...
-            var threadOut = _stdOut.BeginRead(_process.StandardOutput);
-            var threadErr = _stdErr.BeginRead(_process.StandardError);
-
+            var taskOut = _stdOut.BeginRead(_process.StandardOutput);
+            var taskErr = _stdErr.BeginRead(_process.StandardError);
             _process.WaitForExit();
-            threadOut.Join();
-            threadErr.Join();
+
+            Task.WhenAll(taskOut, taskErr);
```
Sridhar-MS: Is it `Task.WhenAll(taskOut, taskErr).Wait()`? Because you are not storing the returned `Task` for use anywhere else.

pakrym: Good catch, it should be `WaitAll` there

src/dotnet/commands/dotnet-build/Program.cs
```diff
@@ -15,13 +15,19 @@ namespace Microsoft.DotNet.Tools.Build
 {
     public class BuildCommand
     {
-        public static int Run(string[] args)
+        public static int Run(string[] args) => Run(args, null);
+
+        public static int Run(string[] args, WorkspaceContext workspace)
```
piotrpMSFT: This change means that in-box commands have a distinct performance advantage over extensions. We need to get to a model where this WorkspaceContext can be accessed from across process boundaries. How will we make this perf benefit available to other commands?

src/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -9,6 +9,7 @@
 using System.Runtime.CompilerServices;
 using NuGet.Frameworks;
 using Microsoft.DotNet.ProjectModel;
+using System.Threading.Tasks;
```
troydai: order! order!

src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
```diff
@@ -3,6 +3,7 @@
 using System.Text;
 using System.Threading;
```
troydai: still needed?

src/Microsoft.DotNet.ProjectModel/Graph/LockFileSymbolTable.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
troydai: license and registration please.

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -134,14 +134,40 @@ public IEnumerable<ProjectContext> BuildAllTargets()
...
-                    .WithLockFile(LockFile)
-                    .WithTargetFramework(target.TargetFramework)
-                    .WithRuntimeIdentifiers(new[] { target.RuntimeIdentifier })
-                    .Build();
+                    var builder = new ProjectContextBuilder()
+                        .WithProject(Project)
+                        .WithLockFile(LockFile)
+                        .WithTargetFramework(target.TargetFramework)
+                        .WithRuntimeIdentifiers(new[] { target.RuntimeIdentifier });
+                    if(IsDesignTime)
```
troydai: format

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -134,14 +134,40 @@ public IEnumerable<ProjectContext> BuildAllTargets()
...
+                    {
+                        builder.AsDesignTime();
+                    }
+
+                    var context = builder.Build();
+                    if (deduper.Add($"{context.TargetFramework}/{context.RuntimeIdentifier}"))
```
troydai: could this be done earlier before the context is built?

src/Microsoft.DotNet.ProjectModel/ProjectContextCollection.cs
```diff
@@ -51,6 +55,15 @@ public bool HasChanged
             }
         }
 
+        public ProjectContext GetTarget(NuGetFramework targetFramework) => GetTarget(targetFramework, string.Empty);
+        public ProjectContext GetTarget(NuGetFramework targetFramework, string runtimeIdentifier)
```
troydai: one empty line before

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -41,33 +49,42 @@ private WorkspaceContext(IEnumerable<string> projectPaths)
...
+                return new WorkspaceContext(Enumerable.Empty<string>(), ProjectReaderSettings.ReadFromEnvironment(), designTime);
             }
 
-            var context = new WorkspaceContext(projectPaths);
+            var context = new WorkspaceContext(projectPaths, ProjectReaderSettings.ReadFromEnvironment(), designTime);
             return context;
         }
 
-        public static WorkspaceContext Create()
+        public static WorkspaceContext Create(bool designTime) => Create(ProjectReaderSettings.ReadFromEnvironment(), designTime);
```
troydai: would be nicer if we leave some comments about usage

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -41,33 +49,42 @@ private WorkspaceContext(IEnumerable<string> projectPaths)
...
-            return new WorkspaceContext(Enumerable.Empty<string>());
+            var settings = ProjectReaderSettings.ReadFromEnvironment();
+            if (!string.IsNullOrEmpty(versionSuffix))
+            {
+                settings.VersionSuffix = versionSuffix;
+            }
+            return Create(settings, designTime);
         }
 
+        public static WorkspaceContext Create(ProjectReaderSettings settings, bool designTime) => new WorkspaceContext(Enumerable.Empty<string>(), settings, designTime);
```
troydai: wrap

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -128,31 +145,93 @@ public void Refresh()
             _needRefresh = false;
         }
 
-        public IReadOnlyList<ProjectContext> GetProjectContexts(string projectPath)
+        public IReadOnlyList<ProjectContext> GetProjectContexts(string projectPath) => (IReadOnlyList<ProjectContext>)GetProjectContextCollection(projectPath)?.ProjectContexts.AsReadOnly() ?? EmptyArray<ProjectContext>.Value;
+        public ProjectContext GetProjectContext(string projectPath, NuGetFramework framework) => GetProjectContext(projectPath, framework, EmptyArray<string>.Value);
+        public ProjectContext GetProjectContext(string projectPath, NuGetFramework framework, IEnumerable<string> runtimeIdentifier)
```
troydai: wrap and empty lines

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -128,31 +145,93 @@ public void Refresh()
             _needRefresh = false;
         }
 
-        public IReadOnlyList<ProjectContext> GetProjectContexts(string projectPath)
+        public IReadOnlyList<ProjectContext> GetProjectContexts(string projectPath) => (IReadOnlyList<ProjectContext>)GetProjectContextCollection(projectPath)?.ProjectContexts.AsReadOnly() ?? EmptyArray<ProjectContext>.Value;
```
troydai: now this is an excessive use, not really readable at this level.

anurse: Whatstheproblem?Don'tyoulikenicecompactlanguage?!

troydai: why `IReadOnlyList`? is `IEnumerable` enough?

anurse: Maybe, not my code originally :)

src/dotnet/commands/dotnet-build/Program.cs
davidfowl: You can't replace build/compile anyways, the only extensibility point here is `dotnet-compile-{compilername}` and csc is almost as second class as everything else 😄.

src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
```diff
@@ -43,11 +44,9 @@ public StreamForwarder ForwardTo(Action<string> writeLine)
             return this;
         }
 
-        public Thread BeginRead(TextReader reader)
+        public Task BeginRead(TextReader reader)
         {
-            var thread = new Thread(() => Read(reader)) { IsBackground = true };
-            thread.Start();
-            return thread;
+            return Task.Run(() => Read(reader));
```
davidfowl: Any reason we don't do read async reads?

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -128,31 +145,93 @@ public void Refresh()
...
+        {
+            var contexts = GetProjectContextCollection(context.ProjectDirectory);
+            if (contexts == null)
+            {
+                return null;
+            }
+
+            var runtimeContext = runtimeIdentifiers
+                .Select(r => contexts.GetTarget(context.TargetFramework, r))
+                .FirstOrDefault(c => c != null);
```
troydai: The combination of `Select` and `FirstOrDefault` clauses feels like a bit waste. Only the first element of the `Select` clause result got used. This can be done with a loop cheaper.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -128,31 +145,93 @@ public void Refresh()
...
         }
 
+        public Project GetProject(string projectDirectory) => GetProjectCore(projectDirectory)?.Model;
+
         public ProjectContextCollection GetProjectContextCollection(string projectPath)
         {
+            var normalizedPath = NormalizeProjectPath(projectPath);
+            if (normalizedPath == null)
+            {
+                return null;
```
troydai: I'm not really sure about this. Returning `null` feels like a new contract. Doesn't it break any client?

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -128,31 +145,93 @@ public void Refresh()
...
+            }
             return _projectContextsCache.AddOrUpdate(
```
troydai: add empty line to improve readability.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -128,31 +145,93 @@ public void Refresh()
...
             return _projectsCache.AddOrUpdate(
-                projectDirectory,
+                normalizedPath,
                 key => AddProjectEntry(key, null),
                 (key, oldEntry) => AddProjectEntry(key, oldEntry));
         }
 
         private LockFile GetLockFile(string projectDirectory)
         {
+            var normalizedPath = NormalizeProjectPath(projectDirectory);
```
troydai: why do we suddenly needs to normalize path? the previous contract is it accept absolution path.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -251,17 +330,9 @@ private FileModelEntry<LockFile> AddLockFileEntry(string projectDirectory, FileM
...
-                    var builder = new ProjectContextBuilder()
-                        .WithProjectResolver(path => GetProject(path).Model)
-                        .WithLockFileResolver(path => GetLockFile(path))
-                        .WithProject(project)
-                        .WithTargetFramework(framework.FrameworkName)
-                        .AsDesignTime();
-
-                    currentEntry.ProjectContexts.Add(builder.Build());
-                }
+                var builder = InitializeProjectContextBuilder(project);
```
troydai: why do we need the `builder` as an intermediate here? can we have the function return all contexts together?

src/dotnet/commands/dotnet-build/BuildCommandApp.cs
```diff
@@ -4,8 +4,10 @@
 using System.Collections.Generic;
 using Microsoft.DotNet.Cli.CommandLine;
 using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.ProjectModel;
 using Microsoft.Extensions.PlatformAbstractions;
 using NuGet.Frameworks;
+using Microsoft.DotNet.ProjectModel;
```
troydai: order! order!

src/dotnet/commands/dotnet-build/BuildCommandApp.cs
```diff
@@ -44,11 +46,15 @@ public class BuildCommandApp
...
         public bool ShouldSkipDependencies { get; set; }
 
+        public WorkspaceContext Workspace { get; private set; }
+
         // workaround: CommandLineApplication is internal therefore I cannot make _app protected so baseclasses can add their own params
         private readonly Dictionary<string, CommandOption> baseClassOptions;
 
-        public BuildCommandApp(string name, string fullName, string description)
+        public BuildCommandApp(string name, string fullName, string description) : this(name, fullName, description, workspace: null) { }
+        public BuildCommandApp(string name, string fullName, string description, WorkspaceContext workspace)
```
troydai: condense code make my eyes bleed. add empty lines please.

test/dotnet-projectmodel-server.Tests/DthTests.cs
```diff
@@ -422,7 +422,7 @@ public void WaitForLockFileReleased(int occupyFileFor, bool expectSuccess)
             {
                 var lockFilePath = Path.Combine(testProject, LockFile.FileName);
                 var lockFileContent = File.ReadAllText(lockFilePath);
-                var fs = new FileStream(lockFilePath, FileMode.Create, FileAccess.Write, FileShare.None);
+                var fs = new FileStream(lockFilePath, FileMode.Open, FileAccess.Write, FileShare.None);
```
troydai: why?

test/dotnet-projectmodel-server.Tests/DthTests.cs
```diff
@@ -506,13 +506,13 @@ public void MSBuildReferenceTest()
...
-                                .AssertProperty("Name", classLibrary)
-                                .AssertProperty<JArray>("Errors", array => array.Count == 0)
-                                .AssertProperty<JArray>("Warnings", array => array.Count == 0);
+                    var dependency = dependencies.RetrieveDependency(classLibrary);
+                    dependency.AssertProperty("Type", LibraryType.MSBuildProject.ToString());
+                    dependency.AssertProperty("Path", NormalizePathString(Path.Combine(testProjectRoot, classLibrary, $"{classLibrary}.csproj")));
+                    dependency.AssertProperty<bool>("Resolved", true);
+                    dependency.AssertProperty("Name", classLibrary);
+                    dependency.AssertProperty<JArray>("Errors", array => array.Count == 0);
+                    dependency.AssertProperty<JArray>("Warnings", array => array.Count == 0);
```
troydai: what's wrong with my beautify elegant superior fluent assertions?

test/dotnet-projectmodel-server.Tests/DthTests.cs
troydai: if the open mode worth test then both mode should be added not change from one to another.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
troydai: ok, I was asking the Troy who wrote these code.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -8,6 +8,7 @@
 using System.Linq;
```
troydai: After the change it feels like the `WorkspaceContext` code become a little more messy and clumsy. It is understandable since it now suites more usages.  I'm wondering is it possibly this class can be broken down further to produce a cleaner solution. For example the function `GetRuntimeContext`, can it be moved to a utility class or a helper. And the `WorkspaceContext` can be really a place simply to cache objects other that filter and extract contexts suits different conditions.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -400,5 +485,11 @@ private static IEnumerable<ProjectDescription> GetProjectReferences(ProjectConte
                 yield return description;
             }
         }
+
+        private bool RidsMatch(string rid, IEnumerable<string> compatibleRids)
```
troydai: static? helper class?

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -280,6 +351,20 @@ private FileModelEntry<LockFile> AddLockFileEntry(string projectDirectory, FileM
             return currentEntry;
         }
 
+        private ProjectContextBuilder InitializeProjectContextBuilder(Project project)
```
troydai: In both cases where this function is used, the direct following action is to build the `ProjectContext`. So how about return the `ProjectContext` collection here directly?

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -280,6 +351,20 @@ private FileModelEntry<LockFile> AddLockFileEntry(string projectDirectory, FileM
...
+        {
+            var builder = new ProjectContextBuilder()
+                .WithProjectResolver(path => GetProjectCore(path)?.Model)
```
troydai: this doesn't feel right. when path can't be normalized that is when the project returned is null. then the builder will call a project resolver which always return null. doesn't it fail anything? we have test case cover it? 

i think the flow should end here, the `GetProjectCore` shouldn't return `null` at all.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
```diff
@@ -238,7 +317,7 @@ private FileModelEntry<LockFile> AddLockFileEntry(string projectDirectory, FileM
                 currentEntry = new ProjectContextCollection();
             }
 
-            var projectEntry = GetProject(projectDirectory);
+            var projectEntry = GetProjectCore(projectDirectory);
             if (projectEntry.Model == null)
```
troydai: `GetProjectCore` can return null!

src/dotnet/commands/dotnet-build/Program.cs
anurse: Yeah, this isn't even really used right now. It's also totally an RC2 quick-fix. We have some thoughts about refactoring a lot of this code post-RC2 into something cleaner.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
troydai: still, please break down this one line magic tricks into plain c#

src/Microsoft.DotNet.Cli.Utils/StreamForwarder.cs
anurse: Not sure, wasn't my code originally :).

anurse: I'm not even sure we need the forwarder at all any more...

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
anurse: Sure, I can refactor to a loop but since we're already dealing with `IEnumerable` (and thus heap allocations for enumerators) it wouldn't improve performance. All of this is lazy-executed, so it wouldn't execute the actual code any more times than a loop.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
anurse: No existing client is broken. DTH only ever provides valid paths.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
anurse: All other places in WorkspaceContext accepted either the project.json path or the project directory. It seemed inconsistent to have this be different.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
anurse: It's clearer for debugging, and the optimizer will elide it anyway.

test/dotnet-projectmodel-server.Tests/DthTests.cs
anurse: You mean your difficult to debug single-line fluent assertion? :trollface:

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
anurse: Yeah, this definitely makes WorkspaceContext a little gross. I think we're just going to have to live with that for RC2, we don't have time to do the good refactoring we want to do. Post-RC2, I have thoughts on a better way to structure these so that we can get a cleaner API.

src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs
anurse: This can happen in scenarios when the project being resolved can't be found. Before this would cause an exception. Now, it causes the project dependency to be marked as unresolved (since ProjectContext already has code to handle a null result from the resolver).

anurse:

pakrym:

anurse:

anurse:
This is ready to go, just waiting on approval... /cc @Petermarcu

src/dotnet/commands/dotnet-build/Program.cs
piotrpMSFT: I agree that this immediate PR does not create a disadvantage. It does, however introduce a pattern which then leads to bifurcation. We're all acknowledging that this is a quick fix. We need to be diligent about preventing the pattern from growing and about developing a solid cross-platform solution.

PR #2630, opened 2016-04-21
Fixed netcore50 package based frameworks

pakrym:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2630)

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/Utilities/FrameworksExtensions.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Utilities/FrameworksExtensions.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Linq;
+using System;
 using System.Runtime.Versioning;
 
 namespace NuGet.Frameworks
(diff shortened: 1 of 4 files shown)


pakrym:
@dotnet-bot test this please

piotrpMSFT:
Is there an issue related to this? What's the problem that this fixes?

pakrym:
This was regressed by https://github.com/dotnet/cli/pull/2597

pakrym:
@dotnet-bot build Windows_NT x86 Debug Build  please

pakrym:

pakrym:

src/Microsoft.DotNet.ProjectModel/Utilities/FrameworksExtensions.cs
```diff
@@ -56,5 +56,12 @@ public static string GetFrameworkString(this NuGetFramework self)
             }
             return name + "-" + frameworkName.Profile;
         }
+        
+        public static bool IsPackageBased(this NuGetFramework self)
```
eerhardt: IMO this is a very confusing method name, especially since there is already a property on NuGetFramework with the same name.  How would consumers know which to call? This extension method or the property?

pakrym: This is getting fixed in NuGet so method is temporary/

eerhardt: Maybe let's not make a new public method in DotNet.ProjectModel for it then, so we can remove it later.  Also, putting that as a comment linked to the Nuget issue would be nice so we can know when it gets fixed in NuGet we can remove it.

davidfowl: this entire class could be internal

davidfowl:
@eerhardt https://github.com/NuGet/NuGet.Client/pull/529

PR #2631, opened 2016-04-21
WIP: Enable building for multiple runtimes

krwq:
for preview

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2631)


src/dotnet/buildandcopy.bat
```diff
@@ -1,4 +1,4 @@
 @echo off
 set clirepo=x:\oss\cli-2343-redo
-set cliver=1.0.0-rc2-002462
+set cliver=1.0.0-rc2-002470
```
krwq: this will be removed, using this for now

brthor:
@dotnet-bot test this please

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -72,6 +73,35 @@ private void MakeCompilationOutputRunnableForFullFramework()
...
         }
+        
+        private IEnumerable<string> FindCoreHostFiles()
+        {
+            if (string.IsNullOrEmpty(_context.RuntimeIdentifier))
+            {
+                yield break;
+            }
+
+            string[] allowedCoreHostFileNames = new string[] {
```
brthor: What do you think about moving this to Corehost.cs under a FileNames property

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -72,6 +73,35 @@ private void MakeCompilationOutputRunnableForFullFramework()
...
+                "dotnet",
+                "hostpolicy",
+                "hostfxr",
+                "libhostpolicy",
+                "libhostfxr"
+            };
+            foreach (var export in _exporter.GetAllExports())
```
brthor: What if we pass in the exports here, so this function is less dependent on class state.

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -154,6 +155,100 @@ public static IEnumerable<ProjectContext> CreateContextForEachTarget(string proj
                         .BuildAllTargets();
         }
 
+        public static bool IsRidCompatibleWith(string rid, string otherRid)
```
brthor: What if we create a dict of rids -> compatible rids.

This won't work for example for `aot` rids in netcore50

src/Microsoft.DotNet.ProjectModel/RuntimeOutputFiles.cs
```diff
@@ -24,7 +24,15 @@ public string Executable
...
             {
-                var extension = FileNameSuffixes.CurrentPlatform.Exe;
+                string extension;
+                if (string.IsNullOrEmpty(_runtimeIdentifier))
+                {
+                    extension = FileNameSuffixes.CurrentPlatform.Exe;
+                }
+                else
+                {
+                    extension = _runtimeIdentifier.ToLower().StartsWith("win") ? ".exe" : string.Empty;
```
brthor: we should create a common function for this, it's pretty specific, may potentially need to change, and is in multiple places

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
krwq: i think we might be able to remove this function by using the changed version of CurrentRid (the one with legacy rid in the name), if not let's do the tree

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -116,17 +116,27 @@ private void CopyCompilationOutput(OutputPaths outputPaths)
 
         private void MakeRunnable(ProjectGraphNode graphNode)
         {
-            var runtimeContext = graphNode.ProjectContext.ProjectFile.HasRuntimeOutput(_args.ConfigValue) ?
-                _args.Workspace.GetRuntimeContext(graphNode.ProjectContext, _args.GetRuntimes()) :
-                graphNode.ProjectContext;
+            foreach (var runtimeContext in CreateRuntimeContexts(graphNode))
```
cdmihai: Extract loop body in a "MakeRunnableForRuntime" method

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -154,6 +155,100 @@ public static IEnumerable<ProjectContext> CreateContextForEachTarget(string proj
...
+        /// <returns>List of project contexts - one for each runtime identifier and one for runtime independent IL code (RuntimeIdentifier == null)</returns>
+
+        public IEnumerable<ProjectContext> CreateAllRuntimeContexts(string buildConfiguration)
+        {
+            if (!ProjectFile.HasRuntimeOutput(buildConfiguration))
+            {
+                return new[] { this };
+            }
+
+            IEnumerable<ProjectContext> allContexts = CreateContextForEachTarget(ProjectFile.ProjectFilePath);
```
cdmihai: Does this go through the workspace context cache or not? (can't tell :))

AFAIK, @anurse moved the runtime context creation to the workspace context. Could you reuse what he did?

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -147,19 +157,19 @@ protected override CompilationResult RunCompile(ProjectGraphNode projectNode)
             finally
             {
                 StampProjectWithSDKVersion(projectNode.ProjectContext);
-                _incrementalManager.CacheIncrementalState(projectNode);
+                _incrementalManager.CacheIncrementalState(projectNode, CreateRuntimeContexts(projectNode));
```
cdmihai: I don't like piping the runtime context everywhere, just so that at some point the compiler IO manager receives them. It couples more code to the knowledge of runtime contexts than need be.

I'd remove them and let only the compiler IO manager create them (it has the graph node)

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -116,17 +116,27 @@ private void CopyCompilationOutput(OutputPaths outputPaths)
...
 
-            var outputPaths = runtimeContext.GetOutputPaths(_args.ConfigValue, _args.BuildBasePathValue, _args.OutputValue);
-            var libraryExporter = runtimeContext.CreateExporter(_args.ConfigValue, _args.BuildBasePathValue);
+                var executable = new Executable(runtimeContext, outputPaths, libraryExporter, _args.ConfigValue);
+                executable.MakeCompilationOutputRunnable();
+            }
+        }
 
-            CopyCompilationOutput(outputPaths);
+        public IEnumerable<ProjectContext> CreateRuntimeContexts(ProjectGraphNode graphNode)
```
cdmihai: I'd move this to the workspace context, near the other runtime context factory method

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
brthor: I haven't worked too much its the WorkspaceContext, could either you or @anurse comment on what that would look like?

cdmihai: There's this method for creating / caching runtime contexts: [https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs#L187](https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.ProjectModel/WorkspaceContext.cs#L187)

I'd change that factory method to return a collection rather than (if I'm reading it right) the first runtime it finds and delete the one in ProjectContext (and also the one in DotnetProjectBuilder)

@anurse, does that sound right?

---
In reply to: [61349348](https://github.com/dotnet/cli/pull/2631#discussion_r61349348) [](ancestors = 61349348)

scripts/dotnet-cli-build/Utils/NuGetRidUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NuGet.RuntimeModel;
+
+namespace Microsoft.DotNet.Cli.Build
+{
+    public class NuGetRidUtils
```
brthor: @krwq I was seeing that our dummy package host was getting published in some of the tests and causing failures. 

This utility uses the NuGet Apis to determine what is and isn't a compatible rid. We only generate a stub package if the dummy package rid is not compatible with the host rid. Since we generate the real package for the host rid I think this makes sense.

scripts/dotnet-cli-build/runtime-graph.json
```diff
@@ -0,0 +1,338 @@
+{
```
brthor: This is the legit runtime graph, I think we'll need to find another way to load it instead of checking it in, but this might work.

davidfowl:
What does this desire actually do? Does it produce outputs for multiple runtimes? I see lots of things in here that don't look right and make a bunch of assumptions. Where's the bug describing the problem being solved and where's the explanation of the solution?

anurse:
A lot of the test projects are getting a big `runtimes` section as a result of this change. Wouldn't it be more stable to just make those projects portable? We need some standalone apps for testing, but I don't think we need all our test projects to be standalone...

brthor:
@anurse We have a lot of assumptions baked into test infrastructure that these things produce some runnable output with an exe. It's possible but would take some time. 

Why is portable inherently more stable though?

brthor:
@davidfowl The issue is here: https://github.com/dotnet/cli/issues/2343

The solution has been mostly conversations between @piotrpMSFT @Sridhar-MS @krwq and I, and what we're going for is simply to pull the host from the package graph when building standalone apps, so that we can support cross-building.

src/Microsoft.DotNet.ProjectModel/ProjectContext.cs
```diff
@@ -157,32 +162,20 @@ public static IEnumerable<ProjectContext> CreateContextForEachTarget(string proj
...
-            if (rid == null)
-            {
-                return false;
-            }
-
-            if (RuntimeIdentifier == rid)
+            if (RuntimeGraph == null)
             {
-                return true;
+                throw new Exception("IsRidCompatible requires a ProjectContext with a loaded RuntimeGraph.");
```
krwq: InvalidOperationException

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -528,6 +545,36 @@ private static string ComposeMessageFromInnerExceptions(Exception exception)
...
         }
 
+        private RuntimeGraph LoadRuntimeGraph(LockFile lockFile)
+        {
+            if (lockFile == null)
+            {
+                throw new Exception($"{nameof(ProjectContextBuilder)}: lock file is null, unable to load runtime graph");
+            }
+
+            // Question: can P2P references define a rid graph?
```
krwq: bump the question

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -528,6 +545,36 @@ private static string ComposeMessageFromInnerExceptions(Exception exception)
             }
         }
 
+        private RuntimeGraph LoadRuntimeGraph(LockFile lockFile)
+        {
+            if (lockFile == null)
+            {
+                throw new Exception($"{nameof(ProjectContextBuilder)}: lock file is null, unable to load runtime graph");
```
krwq: ArgumentException

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -528,6 +545,36 @@ private static string ComposeMessageFromInnerExceptions(Exception exception)
             }
         }
 
+        private RuntimeGraph LoadRuntimeGraph(LockFile lockFile)
```
brthor: Taking a look at the apis available in NuGet this seems right. 

@emgarten would you mind taking a quick look at this?

@anurse , @pakrym I am looking through the WorkspaceContext but not too sure how to integrate this, could you provide some guidance there?

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
brthor: cc @anurse

src/dotnet/commands/dotnet-build/DotnetProjectBuilder.cs
```diff
@@ -116,17 +116,27 @@ private void CopyCompilationOutput(OutputPaths outputPaths)
...
+        {
+            var allRuntimeContexts = _args.Workspace.GetAllRuntimeContexts(graphNode.ProjectContext);
+            if (!string.IsNullOrEmpty(_args.RuntimeValue))
```
cdmihai: Could you make an overload for GetAllRuntimeContext that gets the _args.RuntimeValue and cuts down on needless computation of runtime contexts? If there's 10 RIDs in project.json and the user specifies one, it would be nice to create and cache only one context

cdmihai:
Are you going to eventually call csc for each RID? Or just the make runnable part is per RID?

anurse:
FYI, it looks like we're not going to need this for RC2, so we should hold this for now to avoid major churn. The review can continue, but I want to make sure everyone is aware of the timing :)

krwq:
sounds like we might not need it for neither rc2 or rtm. ill hold until we get to the final decisions

brthor:
This PR is no longer applicable with the move to MSBuild

PR #2636, opened 2016-04-22
Muxer lock down

schellap:
@piotrpMSFT once this makes through the CI and a build is generated, we need to lock to that version with another commit. I need to resolve the `restore` `publish` of `RestoreLockedCoreHost` target on non-Windows as they want the exact RID apparently. Otherwise, it is close to done.

We have three folders:
`artifacts/win7-x64/corehost/latest`
`artifacts/win7-x64/corehost/locked`
`artifacts/win7-x64/corehost`

1. Binaries get dumped into `latest` by build process.
2. Nupkgs are produced into `corehost` from `latest`.
3. `locked` is populated with what we `locked` on. You can be `locked` on either `latest` available in `corehost` from step 2 or a known version available on the feed.

Working on the CI issues, @Sridhar-MS you can take a look...

Cc @leecow @gkhanna79 

This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2636)

### DIFF
--- a/scripts/dotnet-cli-build/CompileTargets.cs
+++ b/scripts/dotnet-cli-build/CompileTargets.cs
@@ -36,31 +36,30 @@ namespace Microsoft.DotNet.Cli.Build
             "vbc.exe"
         };
 
-        public static readonly string[] HostPackageSupportedRids = new[]
-        {
-            "win7-x64",
-            "win7-x86",
-            "osx.10.10-x64",
-            "osx.10.11-x64",
-            "ubuntu.14.04-x64",
-            "centos.7-x64",
-            "rhel.7-x64",
-            "rhel.7.2-x64",
-            "debian.8-x64"
-        };
+        public static string HostPackagePlatformRid => HostPackageSupportedRids[
+                             (PlatformServices.Default.Runtime.OperatingSystemPlatform == Platform.Windows)
+                             ? $"win7-{PlatformServices.Default.Runtime.RuntimeArchitecture}"
+                             : PlatformServices.Default.Runtime.GetRuntimeIdentifier()];
 
-        public static readonly string[] HostPackages = new[]
+        public static readonly Dictionary<string, string> HostPackageSupportedRids = new Dictionary<string, string>()
         {
-            "Microsoft.NETCore.DotNetHost",
-            "Microsoft.NETCore.DotNetHostPolicy",
-            "Microsoft.NETCore.DotNetHostResolver"
+            // Key: Current platform RID. Value: The actual publishable (non-dummy) package name produced by the build system for this RID.
+            { "win7-x64", "win7-x64" },
+            { "win7-x86", "win7-x86" },
+            { "osx.10.10-x64", "osx.10.10-x64" },
+            { "osx.10.11-x64", "osx.10.10-x64" },
+            { "ubuntu.14.04-x64", "ubuntu.14.04-x64" },
+            { "centos.7-x64", "rhel.7-x64" },
+            { "rhel.7-x64", "rhel.7-x64" },
+            { "rhel.7.2-x64", "rhel.7-x64" },
+            { "debian.8-x64", "debian.8-x64" }
         };
 
         public const string SharedFrameworkName = "Microsoft.NETCore.App";
 
         public static Crossgen CrossgenUtil = new Crossgen(CoreCLRVersion);
 
-        private static string CoreHostBaseName => $"corehost{Constants.ExeSuffix}";
+        private static string DotnetHostBaseName => $"dotnet{Constants.ExeSuffix}";
         private static string DotnetHostFxrBaseName => $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}";
         private static string HostPolicyBaseName => $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}";
 
@@ -73,7 +72,7 @@ namespace Microsoft.DotNet.Cli.Build
 
         // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the
         // Microsoft.NETCore.App package that is created during PackagePkgProjects.
-        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
+        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(RestoreLockedCoreHost), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
         public static BuildTargetResult Compile(BuildTargetContext c)
         {
             return c.Success();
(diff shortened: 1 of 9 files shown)


scripts/dotnet-cli-build/Utils/Dirs.cs
```diff
@@ -22,7 +22,9 @@ public static class Dirs
         public static readonly string Stage2 = Path.Combine(Output, "stage2");
         public static readonly string Stage2Compilation = Path.Combine(Output, "stage2compilation");
         public static readonly string Stage2Symbols = Path.Combine(Output, "stage2symbols");
-        public static readonly string Corehost = Path.Combine(Output, "corehost");
+        public static readonly string CorehostLatest = Path.Combine(Output, "corehost", "latest");
```
schellap: May be just use Corehost but call it latest to keep signing working or it can be updated to read from this folder instead?

schellap:

schellap:
@piotroMSFT you guys okay if I merge this? This PR's build is the locked muxer. Next PR will lock CLI to this PR's build.

schellap:
@Sridhar-MS PTAL, the muxer versioning is broken and this PR fixes it.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -36,31 +36,30 @@ public class CompileTargets
...
-            "rhel.7.2-x64",
-            "debian.8-x64"
-        };
+        public static string HostPackagePlatformRid => HostPackageSupportedRids[
+                             (PlatformServices.Default.Runtime.OperatingSystemPlatform == Platform.Windows)
+                             ? $"win7-{PlatformServices.Default.Runtime.RuntimeArchitecture}"
+                             : PlatformServices.Default.Runtime.GetRuntimeIdentifier()];
 
-        public static readonly string[] HostPackages = new[]
+        public static readonly Dictionary<string, string> HostPackageSupportedRids = new Dictionary<string, string>()
```
brthor: Can you give some context on why this is needed?

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -73,7 +72,7 @@ public static BuildTargetResult UpdateBuild(BuildTargetContext c)
 
         // Moving PrepareTargets.RestorePackages after PackagePkgProjects because managed code depends on the
         // Microsoft.NETCore.App package that is created during PackagePkgProjects.
-        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
+        [Target(nameof(PrepareTargets.Init), nameof(CompileCoreHost), nameof(PackagePkgProjects), nameof(RestoreLockedCoreHost), nameof(PrepareTargets.RestorePackages), nameof(CompileStage1), nameof(CompileStage2))]
```
brthor: cc @Sridhar-MS this will impact VSO?

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -85,45 +84,78 @@ public static BuildTargetResult Compile(BuildTargetContext c)
...
-                            Dirs.CorehostDummyPackages);
-            foreach (var hostPackageId in HostPackages)
+            string currentRid = GetRuntimeId();
+            PrepareDummyRuntimeNuGetPackage(DotNetCli.Stage0);
+            foreach (var hostPackage in buildVersion.LatestHostPackages)
             {
-                foreach (var rid in HostPackageSupportedRids)
+                foreach (var rid in HostPackageSupportedRids.Keys)
                 {
                     if (!rid.Equals(currentRid))
```
brthor: Would it make sense to check the value here instead of the keys?

For example if we're on osx10.10, do we still need to generate a osx10.10 package for osx10.11?

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -85,45 +84,78 @@ public static BuildTargetResult Compile(BuildTargetContext c)
...
                 }
             }
             return c.Success();
         }
 
-        private static string HostVer = "1.0.1";
-        private static string HostPolicyVer = "1.0.1";
-        private static string HostFxrVer = "1.0.1";
+        [Target(nameof(PrepareTargets.Init))]
+        public static BuildTargetResult RestoreLockedCoreHost(BuildTargetContext c)
```
brthor: What is this doing?

scripts/dotnet-cli-build/CompileTargets.cs
schellap: We don't but I reverted to preserve what you had earlier... I can give the Keys a shot and see if everything is still good.

scripts/dotnet-cli-build/CompileTargets.cs
schellap: Locking down corehost (i.e., restores and publishes a known version of corehost) into `artifacts/osx.10.10-x64/corehost/locked`

scripts/dotnet-cli-build/CompileTargets.cs
brthor: My concern is that this empty package may not get overridden by the *right* package, so we probably shouldn't create it at all (to save headache)

scripts/dotnet-cli-build/CompileTargets.cs
schellap: It is used above in the lambda `HostPackagePlatformRid` which is used in more places.

scripts/dotnet-cli-build/CompileTargets.cs
brthor: ping @Sridhar-MS

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -85,45 +84,78 @@ public static BuildTargetResult Compile(BuildTargetContext c)
...
+                .WorkingDirectory(tempPjDirectory)
+                .Execute()
+                .EnsureSuccessful();
+
+            // Clean out before publishing locked binaries
+            FS.Rmdir(Dirs.CorehostLocked);
+
+            // Use specific RIDS for non-backward compatible platforms.
+            (CurrentPlatform.IsWindows
+                ? DotNetCli.Stage0.Publish("--output", Dirs.CorehostLocked, "--no-build")
```
brthor: It's not too clear why you need a special case for windows here?

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -36,31 +36,30 @@ public class CompileTargets
...
+        public static readonly Dictionary<string, string> HostPackageSupportedRids = new Dictionary<string, string>()
         {
-            "Microsoft.NETCore.DotNetHost",
-            "Microsoft.NETCore.DotNetHostPolicy",
-            "Microsoft.NETCore.DotNetHostResolver"
+            // Key: Current platform RID. Value: The actual publishable (non-dummy) package name produced by the build system for this RID.
+            { "win7-x64", "win7-x64" },
+            { "win7-x86", "win7-x86" },
+            { "osx.10.10-x64", "osx.10.10-x64" },
+            { "osx.10.11-x64", "osx.10.10-x64" },
```
brthor: I am confused by what this is doing, on osx.10.11-x64, this creates a package for osx.10.10-x64?

schellap: Yes. It is the same as win7-x64 produced for win10-x64 and win81-x64.

PR #2654, opened 2016-04-22
dotnet run doesn't accept '--' arg separator anymore

### LINKED ISSUE #2643: `dotnet run` regressed in supporting multiple operands
## Steps to reproduce
Create a command-line app that accepts more than one argument (eg. "Hello" -p) try to a pass the arguments as operands the `run` command, ie your command-line app. 

## Expected  behavior
Previously, `dotnet run` syntax was:
```usage: dotnet [-f <arg>] [-c <arg>] [-p <arg>] [-h] [--] <args>...

    -f, --framework <arg>        Compile a specific framework
    -c, --configuration <arg>    Configuration under which to build
    -p, --project <arg>          The path to the project to run
                                 (defaults to the current directory).
                                 Can be a path to a project.json or a
                                 project directory
    -h, --help                   Help for compile native.
    <args>...                    Arguments to pass to the executable or
                                 script```
so the `--` served to separate the args to `run` from the operands to pass to the app.

## Actual behavior
```
dotnet run "Hello" -p
Specify --help for a list of available options and commands.
Missing value for option 'project'
```
## Environment data
`dotnet --info` output:
.NET Command Line Tools (1.0.0-rc2-002476)

Product Information:
 Version:     1.0.0-rc2-002476
 Commit Sha:  1a8d393b9b

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.11
 OS Platform: Darwin
 RID:         osx.10.11-x64

### PR DESCRIPTION
**Pull request by eerhardt**
When moving from System.CommandLine to the DNX arg parser, we regressed in parsing arguments that should be passed to the underlying process.

Fix #2643 

Questions:

1. Should we just use the existing option `throwOnUnexpectedArg = false` to mean that the args can handle the `--` separator?  Or should I continue with two separate flags on `CommandLineApplication`, like I did?
2. Should the other commands that specify `throwOnUnexpectedArg = false` get this new behavior?
 - dotnet test
 - dotnet restore
 - dotnet repl
 - dotnet repl-csi

`dotnet restore` is the main one that seems weird to have a `--` separator.

@piotrpMSFT @anurse @Sridhar-MS

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2654)

### CODE REVIEW

**Comment by eerhardt** [comment]
This PR is ready for review.

File: src/dotnet/CommandLine/CommandLineApplication.cs [hunk 1]
```diff
@@ -129,10 +130,18 @@ public int Execute(params string[] args)
                     if (longOption != null)
                     {
                         processed = true;
-                        option = command.Options.SingleOrDefault(opt => string.Equals(opt.LongName, longOption[0], StringComparison.Ordinal));
+                        string longOptionName = longOption[0];
```
troydai: `var`

src/dotnet/CommandLine/CommandLineApplication.cs
```diff
@@ -129,10 +130,18 @@ public int Execute(params string[] args)
...
+                        option = command.Options.SingleOrDefault(opt => string.Equals(opt.LongName, longOptionName, StringComparison.Ordinal));
 
                         if (option == null)
                         {
+                            if (string.IsNullOrEmpty(longOptionName) && !command._throwOnUnexpectedArg && AllowArgumentSeparator)
```
troydai: `longOptionName` won't be `null` right?

troydai:
🚢

src/dotnet/CommandLine/CommandLineApplication.cs
eerhardt: Correct, but I felt it more correct to use `string.IsNullOrEmpty` instead of longOptionName.Length == 0, in case that assumption ever changes.

src/dotnet/CommandLine/CommandLineApplication.cs
eerhardt: I follow the [CoreFX coding guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md).  (See rule #10) And I think all C# code under the `dotnet` user should follow the same coding guidelines so people contributing to both don't have to switch their coding style in between repos.

troydai: 👍

davidfowl: You guys should use var more

eerhardt:

eerhardt:

eerhardt:

eerhardt:

PR #2661, opened 2016-04-22
Revert "Revert "Updating CoreFX to rc2-24022""

Petermarcu:
Reverts dotnet/cli#2659

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2661)


Petermarcu:
Unfortunate coincidence that the networking updates in corefx happened at the same time as our network timeouts in the build. Networking issue still repro's without a new corefx. Putting the new build back as we look deeper into the root cause.

Issue #2623, opened 2016-04-21
Create Breadcrumb Location when it is absent


gkhanna79:
If the required .NET FX is not installed on a machine (e.g. clean Win7), then BreadCrumb location will not be present. This needs to be addressed for two scenarios:

1) Create it, if it does not exist, when installing the SharedFX Redist
2) Have a mechanism to create it when standalone apps are executed on a machine that does not contain SharedFX redist.

CC @leecow @schellap @mellinoe @ericstj @ellismg

ellismg:
What is the exact location and and set of ACLs that should be put on the breadcrumb location?

ericstj:
(programdata folder)Microsoft\NetFramework\BreadcrumbStore
sddl="D:(A;OICI;CCDCSWWPLORC;;;WD)(A;OICI;FA;;;BA)(A;OICI;FA;;;SY)"

gkhanna79:
@leecow and I were chatting. It maybe a better idea to have this ACLed BredCrumb location creation (after the existence check) be in its own installer so that both the above scenarios can be addressed. This installer can be used by itself (e.g. for standalone apps) and chained by SharedFX Redist installer to create on machines where it does not exist (e.g. Win7). 

Thoughts?

A question that comes to mind is that once this location is created by the Breadcrumb installer, and later .NET FX is installed, will the latter setup keep the location intact? And if the .NET FX were to get uninstalled, will it still keep the location intact?

@ericstj Do you have insights into the above?

schellap:
@gkhanna79 I was thinking about this and it may help alleviate your NET FX conflict concerns. Now that we have an opportunity to create one, can we create one for ourselves: `CoreBreadcrumbs` inside `NetFramework`?

> It maybe a better idea to have this ACLed BredCrumb location creation (after the existence check) be in its own installer so that both the above scenarios can be addressed. 

It is easier to say if you install `dotnet` you get servicing irrespective of what app you are, than say, go get this special installer to get servicing if you are a standalone app (i.e., it is not obvious to me that I have to get this installer if I am standalone only.)

piotrpMSFT:
#2546

piotrpMSFT:
#2662

piotrpMSFT:
Fixed by #2662

PR #2662, opened 2016-04-22
Create breadcrumb store folder in the sharedframework MSI

mellinoe:
@gkhanna79 , @ericstj 

I tested this by:

1. temporarily changing the folder name and ensuring it was created/deleted as expected.
2. Installing it over my existing Breadcrumb folder. The folder remained, but the permissions on the folder were re-set to what they were initially (i.e. if I give myself access to open the folder and then run the installer, I have to give myself access again). This seems like correct behavior to me. Uninstalling left the folder there, as well.

I'm not very experienced at WiX authoring so I can move this or clean this up if necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2662)

### DIFF
--- a/packaging/windows/sharedframework/sharedframework.wxs
+++ b/packaging/windows/sharedframework/sharedframework.wxs
@@ -11,6 +11,27 @@
         <Feature Id="MainFeature" Title="Main Feature" Level="1">
             <ComponentGroupRef Id="InstallFiles" />
             <ComponentGroupRef Id="AuthoredRegistryKeys"/>
+            <Component Id="BreadcrumbStoreFolder" Directory="BreadcrumbStore" Guid="{DA957490-EA59-42B9-B9D6-3E80F6FB0ED0}">
+                <CreateFolder Directory="BreadcrumbStore">
+                    <Permission
+                        User="Everyone"
+                        CreateChild="no"
+                        CreateFile="yes"
+                        Delete="no"
+                        DeleteChild="no"
+                        Read="yes"
+                        ReadAttributes="yes"
+                        ReadExtendedAttributes="yes"
+                        ReadPermission="yes"
+                        Traverse="yes" />
+                    <Permission
+                        User="SYSTEM"
+                        GenericAll="yes" />
+                    <Permission
+                        User="Administrators"
+                        GenericAll="yes" />
+                </CreateFolder>
+            </Component>
         </Feature>
         <Feature Id="Provider" Absent="disallow" AllowAdvertise="no" Description="Used for Ref Counting" Display="hidden" Level="1" InstallDefault="local" Title="RefCounting" TypicalDefault="install">
             <ComponentRef Id="DotNet_CLI_SharedFramework_$(var.FrameworkName)_$(var.FrameworkComponentVersion)" />
@@ -29,6 +50,13 @@
         <Directory Id="TARGETDIR" Name="SourceDir">
             <Directory Id="$(var.Program_Files)">
                 <Directory Id="DOTNETHOME" Name="dotnet" />
+            </Directory>
+            <Directory Id="CommonAppDataFolder">
+              <Directory Id="MicrosoftCommonAppData" Name="Microsoft">
+                <Directory Id="NetFrameworkCommonAppData" Name="NetFramework">
+                  <Directory Id="BreadcrumbStore" Name="BreadcrumbStore" />
+                </Directory>
+              </Directory>
             </Directory>
         </Directory>
     </Fragment>


ellismg:
Would it be better to have this in the muxer MSI?  I know we were discussing some cases where the muxer would be installed via MSI but the shared frameworks would come from zips?

mellinoe:
> Would it be better to have this in the muxer MSI?

@gkhanna79  ?

gkhanna79:
Ideally, we have a seperate installer for breadcrumb store creation. Installing/updating the muxer should not install/reset/create the breadcrumb folder - the actions should be independent of each other. 

> I know we were discussing some cases where the muxer would be installed via MSI but the shared frameworks would come from zips?

For these cases, it was agreed that we will give guidelines on how to create/prepare the breadcrumb folder. @leecow would be including the steps in his guidelines. Given that, I am fine this to be in SharedFX redist MSI.

CC @schellap

robmen:
Wherever you choose to put the code, @mellinoe the code looks good for what I understand you want to accomplish.

FYI: If you want the folder to be removed on uninstall (not clear that you do) but there are left over files preventing that, you could use a `RemoveFile` element to delete the files during uninstall and the folder would get cleaned up (may need to add a `RemoveFolder` element as well).

gkhanna79:
> If you want the folder to be removed on uninstall (not clear that you do) 

We do not want to remove the folder on uninstall.

schellap:
LGTM. I think any of our installers should create this, but given muxer gets installed by both the fx msi and the muxer msi it is a better place. Avoids the prep step for @ellismg scenario.

gkhanna79:
My concern with adding it to the muxer MSI is that any changes to the layout would require a muxer MSI update. And since that would carry the muxer, wouldnt that create a reboot issue due to muxer overwrite?

ericstj:
> any changes to the layout would require a muxer MSI update

This component is the least likely to change of any component in your installers.  Even if there was a change this particular component has non uninstall characteristics so you don't even need to make the change in this installer.  The proper place to put this component is in whatever MSI contains the piece of code that is aware of the location, eg: the thing that is writing the breadcrumbs.

gkhanna79:
> The proper place to put this component is in whatever MSI contains the piece of code that is aware of the location, eg: the thing that is writing the breadcrumbs

This would be the SharedFX redist MSI

gkhanna79:
@mellinoe @ellismg Are we ready to commit this change? Can you confirm that the ACLs set here are the same as the ones set by the Desktop .NET FX installer?

piotrpMSFT:
This will fix #2623

mellinoe:
@ericstj These are the same permissions as the desktop installer, right?

ericstj:
Yes, this is the same as the desktop MSI.  We should test E2E on a machine without Net 4.x to make sure the host is doing the right thing.

ericstj:
I wouldn't block this PR on that testing.  LGTM.

piotrpMSFT:
@schellap and @Petermarcu just confirmed the Shared FX MSI should be creating the location. :shipit:

packaging/windows/sharedframework/sharedframework.wxs
```diff
@@ -11,6 +11,27 @@
         <Feature Id="MainFeature" Title="Main Feature" Level="1">
             <ComponentGroupRef Id="InstallFiles" />
             <ComponentGroupRef Id="AuthoredRegistryKeys"/>
+            <Component Id="BreadcrumbStoreFolder" Directory="BreadcrumbStore" Guid="{DA957490-EA59-42B9-B9D6-3E80F6FB0ED0}">
```
ericstj: Can we change this to use a `*` GUID instead?

robmen: Unfortunately, no. This Component doesn't meet the requirements.

ericstj: Why not?  It has a well defined machine-wide directory as keypath.  That should be a sufficient seed.  Then again its not really a component where you care about having a stable GUID since it is fire-and-forget.

robmen: Keypath is a folder. Tons of Components can root to the same folder and get colliding keys very easily. For example, you could choose to split these permissions into separate Components and easily get collisions.

Stable and unique. Unique is the hard part.

ericstj: I see, it's trying to enforce uniqueness.  That's why you force folks to have single file components.  I suppose it still isn't foolproof.  You can still have two components with different registry value keypaths but overlapping values since it doesn't enforce the "single registry value" rule.

robmen: Correct. Not perfect but from what I've seen, for the cases WiX handles it handles better than humans. :smile:

ericstj: True, indeed better than humans.  Thanks for explaining the reasoning.  I vaguely remember these discussions happening at WiX night years ago.  :smile:

robmen: Components and their GUIDs are the thing I dislike *most* about the Windows Installer. Ever since I was an intern in 1998... now I just feel old.

PR #2663, opened 2016-04-23
Add performance tests

pakrym:
@davidfowl

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2663)


src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
```diff
@@ -189,6 +190,7 @@ public override bool Compile(ProjectContext context, BuildCommandApp args)
             Reporter errorReporter = Reporter.Error;
             Reporter outputReporter = Reporter.Output;
 
+            Thread.Sleep(1000);
```
eerhardt: ???

test/dotnet-performance.Tests/IncrementalTestBase.cs
```diff
@@ -0,0 +1,190 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Tools.Test.Utilities;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.DotNet.Tools.Builder.Tests
+{
+    public class BuildPerformanceTest : PerformanceTestBase
```
eerhardt: The file name should match the class name.

test/dotnet-performance.Tests/PerformanceTestBase.cs
```diff
@@ -0,0 +1,63 @@
...
+        protected TestInstance CreateTestInstance(string testProjectName, string variation = "", [CallerMemberName] string callingMethod = "")
+        {
+            return TestAssetsManager.CreateTestInstance(Path.Combine("PerformanceTestProjects", testProjectName), callingMethod + variation)
+                .WithLockFiles();
+        }
+
+        public void Iterate(Action<PerformanceIterationContext> action, string variation = "", int iterations = 1, [CallerMemberName] string callingMethod = "")
+        {
+            var fullname = callingMethod + (variation != "" ? "/" + variation : "");
+            var context = new PerformanceIterationContext();
```
eerhardt: This is newed up inside the loop. No reason to do it here, right?

test/dotnet-performance.Tests/project.json
```diff
@@ -0,0 +1,28 @@
+{
+  "version": "1.0.0-*",
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0-rc2-3002439"
```
eerhardt: Should this be *? So it picks up the pkg built from the repo?

If not, what is the strategy for keeping this version updated?

pakrym:
@eerhardt sorry, I should've put `work in progress` when creating PR, it's not merge ready yet, just to show the concept.

src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
davidfowl: This was just to make sure the tests were working 😄

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/BuildCommand.cs
```diff
@@ -199,6 +200,16 @@ private string NoDependencies
             }
         }
 
+        private string Verbose
+        {
+            get
+            {
+                return _verbose ?
+                    "--verbose" :
+                    "";
```
troydai: i think this is short enough to be placed in one line.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/BuildCommand.cs
```diff
@@ -217,7 +228,8 @@ private string NoDependencies
             string cppCompilerFlags="",
             bool buildProfile=true,
             bool noIncremental=false,
-            bool noDependencies=false)
+            bool noDependencies=false,
+            bool verbose=true)
```
troydai: format, add spaces in front of and behind the assignment.

test/dotnet-performance.Tests/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,191 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.Tools.Test.Utilities;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
```
troydai: order using statements

test/dotnet-performance.Tests/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,191 @@
...
+
+namespace Microsoft.DotNet.Tools.Builder.Tests
+{
+    public class BuildPerformanceTest : PerformanceTestBase
+    {
+
```
troydai: format, remove the empty line

test/dotnet-performance.Tests/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                yield return new [] { "TwoTargetApp"};
+                yield return new [] { "SingleTargetApp" };
+            }
+        }
+        public static IEnumerable<object> GraphProjects
+        {
+            get
+            {
+                yield return new object[] {
```
troydai: format, open brace on its own line.

test/dotnet-performance.Tests/PerformanceTestBase.cs
```diff
@@ -0,0 +1,63 @@
...
+            Directory.Delete(Path.Combine(project, "bin"), true);
+        }
+
+        protected TestInstance CreateTestInstance(string testProjectName, string variation = "", [CallerMemberName] string callingMethod = "")
+        {
+            return TestAssetsManager.CreateTestInstance(Path.Combine("PerformanceTestProjects", testProjectName), callingMethod + variation)
+                .WithLockFiles();
+        }
+
+        public void Iterate(Action<PerformanceIterationContext> action, string variation = "", int iterations = 1, [CallerMemberName] string callingMethod = "")
```
troydai: Do we plan to set up any benchmark for the perf tests here?

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/BuildCommand.cs
pakrym: It's consistent with all other similar properties in the file.

test/dotnet-performance.Tests/PerformanceTestBase.cs
pakrym: This `new` is for heat up iteration which is done outside the loop.

test/dotnet-performance.Tests/project.json
pakrym: Removed.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/BuildCommand.cs
pakrym: same as above

test/dotnet-performance.Tests/PerformanceTestBase.cs
pakrym: What do you mean by setup benchmarks?

test/dotnet-performance.Tests/PerformanceTestBase.cs
eerhardt: it may make sense to add this as a comment so someone doesn't remove it.

eerhardt: Also, it looks like the "heat up iteration" is commented out.

troydai:
@pakrym is it still wip?

troydai:
@pakrym I'm wondering the outcome of the tests. It seems to me the just print the elapsed time to the console. It is valuable records the performance for each build however it is also easily ignored. Do we consider to set up a benchmark? That is when performance regresses below the benchmark the test will fail.

test/dotnet-performance.Tests/PerformanceTestBase.cs
troydai: See my comments below

pakrym:
Yes, I'm changing output code to write `csv` file to. For now it only outputs values for developers to compare, setting up failure conditions is out of scope of this PR.

troydai:
@pakrym ok. do we cache the `csv` file over every CI build?

pakrym:
I don't think this should even be ran during CI build, it takes too much time and all CI machines are different so we won't get consistent results anyway.

troydai:
If CI time is an issue, we should make other plan to ensure the performance tests are run regularly. Perf tests run sporadically (actually any automation test) is fairly useless and more expensive in some way:

1. Test won't be well maintained as product evolve.
2. You don't have history data, lost the track of regression.

If this is a onetime thing (it shouldn't), I doubt why should we expand cli test code base. I think we need to run it regularly. Maybe not on the CI so as not to slow down the build turnaround, but leave test code behind is not a good strategy.

Petermarcu:
@piotrpMSFT , how does this overlap with the other effort that is happening to add perf tests to the repo?

test/dotnet-performance.Tests/PerformanceTestBase.cs
```diff
@@ -0,0 +1,63 @@
...
+            Directory.Delete(Path.Combine(project, "bin"), true);
+        }
+
+        protected TestInstance CreateTestInstance(string testProjectName, string variation = "", [CallerMemberName] string callingMethod = "")
+        {
+            return TestAssetsManager.CreateTestInstance(Path.Combine("PerformanceTestProjects", testProjectName), callingMethod + variation)
+                .WithLockFiles();
+        }
+
+        public void Iterate(Action<PerformanceIterationContext> action, string variation = "", int iterations = 3, [CallerMemberName] string callingMethod = "")
```
Sridhar-MS: We must be following a single way to write perf tests. @brianrob is using `Benchmark` attribute to run perf tests - https://github.com/dotnet/cli/pull/2681/files#diff-6876439b00d6c2806174918892c08f48R45. Can we try to do the same here?

pakrym: I don't want to measure some parts of iteration (deleting files for example). I don't think `Benchmark` attribute allows to do this.

pakrym:
@dotnet-bot test Debian8.2 x64 Debug Build  please

test/dotnet-performance.Tests/PerformanceTestBase.cs
pakrym: Also my pattern allows for heat up iteration and having average of multiple runs.

pakrym: Okay, I had couple of wrong assumptions about `Benchmark` attribute, it is able to measure only parts of iterations, I couldn't find any automatic way to do heat up run also is there a `BenchmarkTheoryAttribute`?

Sridhar-MS: From here - https://github.com/Microsoft/xunit-performance/blob/master/README.md#authoring-benchmarks
> The first iteration is the "warmup" iteration; all performance metrics are discarded by the result analyzer. Subsequent iterations are measured.

So I think the first iteration is the warmup iteration.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/BuildCommand.cs
piotrpMSFT: We should make the operators first on the line of the operand instead of last on the preceding line.

test/dotnet-performance.Tests/PerformanceTestBase.cs
piotrpMSFT: The test cases here are more comprehensive than the end-to-end test. We should converge on the BenchmarkAttribute infrastructure however. That one is already fairly developed and has broad ongoing investment.

TestAssets/TestProjects/PerformanceTestProjects/SingleTargetGraph/SingleTargetP0/Program.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
```
brianrob: It looks like SingleTargetP0, P1, and P2 are all the same as SingleTargetApp - both the cs and project.json files.  Is there a reason that you need all of them?

TestAssets/TestProjects/PerformanceTestProjects/TwoTargetGraph/TwoTargetP1/Program.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
```
brianrob: Same for the two target apps.  Do you need all of them?

TestAssets/TestProjects/PerformanceTestProjects/TwoTargetGraphLarge/TwoTargetLargeP4/project.json
```diff
@@ -0,0 +1,22 @@
+{
```
brianrob: For the multi-project apps, is it intentional not to add actual code dependencies?  I would recommend actually calling code in the dependencies so that you get representative build time.

src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
```diff
@@ -8,6 +8,7 @@
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Threading;
```
brianrob: Do you need this?

TestAssets/TestProjects/PerformanceTestProjects/SingleTargetGraph/SingleTargetP0/Program.cs
pakrym: They have dependency on each other.

TestAssets/TestProjects/PerformanceTestProjects/TwoTargetGraph/TwoTargetP1/Program.cs
pakrym: Same, project to project dependencies.

TestAssets/TestProjects/PerformanceTestProjects/SingleTargetGraph/SingleTargetP0/Program.cs
brianrob: Yes, you're right.  I missed that.  They only have a project-to-project dependency though.  No actual code dependency.  Shouldn't we be testing a real scenario with actual code dependency too?  These are just a bunch of empty mains with no dependencies.

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        public void BuildSingleProject(TestInstance instance)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    Build(instance.TestRoot);
+                }
+                RemoveBin(instance.TestRoot);
```
brianrob: Do you need to remove obj dir as well?  I'm thinking you might be better off creating new test instances for each iteration so that you ensure that as people add new persisted artifacts, the tests remain effective.

TestAssets/TestProjects/PerformanceTestProjects/SingleTargetGraph/SingleTargetP0/Program.cs
pakrym: I'll add some method calls to make CSC emit references, but dotnet build itself does not care about code dependencies

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+        [Benchmark]
+        public void IncrementalRebuildWithRootChangedInGraph_TwoTargetGraph() =>
+            IncrementalRebuildWithRootChangedInGraph(CreateTestInstances(TwoTargetGraph));
+
+        [Benchmark]
+        public void IncrementalRebuildWithRootChangedInGraph_TwoTargetGraphLarge() =>
+            IncrementalRebuildWithRootChangedInGraph(CreateTestInstances(TwoTargetGraphLarge));
+
+        public void IncrementalRebuildWithRootChangedInGraph(TestInstance[] instances)
+        {
```
brianrob: This test doesn't actually change the root, but instead just deletes the bin directory.  Is this the scenario you're trying to test, or do you want to actually test changing some code in the root project?  I would think that the performance characteristics might be different because the bin directory will exist in one case but not the other.

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            Build(instance.TestRoot);
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    Build(instance.TestRoot);
+                }
+                RemoveBin(instances.Last().TestRoot);
```
brianrob: Same here.

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        [Benchmark]
+        public void IncrementalSkipAllNoDependenciesInGraph_TwoTargetGraphLarge() =>
+            IncrementalSkipAllNoDependenciesInGraph(CreateTestInstances(TwoTargetGraphLarge));
+
+        public void IncrementalSkipAllNoDependenciesInGraph(TestInstance[] instances)
+        {
+            var instance = instances[0];
+
+            Build(instance.TestRoot);
```
brianrob: Do you need to compile all of the projects before the first iteration?  Otherwise I'd expect all but the first to be compiled the first time around.

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    foreach (var i in instances)
+                    {
+                        Run(new BuildCommand(i.TestRoot,
```
brianrob: Recommend that you produce the build commands outside of the measurement.  This ensures that we keep the overhead of the harness as low a we can.

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+                            framework: DefaultFramework,
+                            noDependencies: true,
+                            buildProfile: false));
+                    }
+                }
+            }
+        }
+        [Benchmark]
+        public void BuildAllNoDependenciesInGraphh_SingleTargetGraph() =>
```
brianrob: Typo: Graphh

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        public void BuildAllNoDependenciesInGraph(TestInstance[] instances)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    foreach (var i in instances.Reverse())
+                    {
+                        Run(new BuildCommand(i.TestRoot,
```
brianrob: Same here.  Can you produce the build commands outside of the measurement?

test/Performance/BuildPerformanceTest.cs
brianrob: Did you mean to Run the app here here, or did you want to build?

test/Performance/BuildPerformanceTest.cs
brianrob: Same question here.  Run vs. Build?

test/Performance/PerformanceTestBase.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Linq;
+using System.Runtime.CompilerServices;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using Microsoft.DotNet.TestFramework;
+
+namespace Microsoft.DotNet.Tools.Builder.Tests
+{
+    public class PerformanceTestBase : TestBase
+    {
+        protected void Build(string project)
```
brianrob: I think it makes more sense to inline the bodies of build and run into the individual tests.  This does two things: (1) allows you to minimize the amount of code inside of the measurement block.  (2) makes it clear what code is being executed in the measurement block.

I also considered the idea of passing the iteration object to these methods, but that means that you have much less control in the test itself of when measurement starts, and it's error prone because it's standard practice to have the call to StartMeasurement in the test method itself.

If we do that, then I think the base class probably isn't helping you much, and I'd just recommend moving the TestInstance methods into the test class itself (or if they belong in TestBase you could consider putting them there as well).

TestAssets/TestProjects/PerformanceTestProjects/SingleTargetGraph/SingleTargetP0/Program.cs
brianrob: Sounds good.

test/Performance/BuildPerformanceTest.cs
pakrym: I wanted enough change to trip incremental check, but you are right modifying source file would be closer to use case.

test/Performance/BuildPerformanceTest.cs
pakrym: Building first project builds all because of P2P dependencies.

test/Performance/PerformanceTestBase.cs
pakrym: Agreed, I'll inline command executions, just wanted to type less arguments and drop base class too.

TestAssets/TestProjects/PerformanceTestProjects/SingleTargetGraph/SingleTargetP0/Program.cs
pakrym: Done.

pakrym:
Added P2P code references, inlined command calls, removed base class, replaced bin removal with touching source file.

pakrym:
@dotnet-bot test Debian8.2 x64 Debug Build  please

pakrym:
@brianrob Do you have any other concerns with this PR?

pakrym:
:up::date:

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,285 @@
...
+        [Benchmark]
+        public void BuildSingleProject_TwoTargetApp() => BuildSingleProject(CreateTestInstance(TwoTargetApp));
+
+        public void BuildSingleProject(TestInstance instance)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    new BuildCommand(instance.TestRoot, buildProfile: false)
```
brianrob: You should create the BuildCommand outside the measurement block to minimize the overhead of setup during the measurement.  I would also recommend not calling Should().Pass() either inside the measurement block.  If you want to call it, you could do so outside of the measurement block as well.  Same applies for the other tests below as well.

pakrym: Overhead is minimal, we are measuring in seconds, object allocation is quick event compared to process start overhead.

test/Performance/BuildPerformanceTest.cs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        protected void TouchSource(string project)
+        {
+            var sourceFile = Directory.GetFiles(project, "*.cs").FirstOrDefault();
+            if (sourceFile == null)
+            {
+                throw new InvalidOperationException($"'.cs' files not found in {project}");
+            }
+            File.SetLastWriteTime(sourceFile, DateTime.Now);
```
brianrob: I've not tried updating the last write time.  However, if this is how both the compiler and the CLI decide whether to compile something, then I think this is fine - you don't need to actually mutate the source.

test/Performance/BuildPerformanceTest.cs
brianrob: Yes, an object allocation doesn't cost very much, but this is about measuring what you want to measure and as little of what you don't want to measure as possible.  BuildCommand could just assign a few fields, or it could choose to do a bunch more.  All you'd see if such a change was made was an increase in time spent in the test.

brianrob:
Before you merge, can you please try running with the script from #2746?  Once this is merged, the script will run your tests as well, so it would be good to make sure that there's no issues there.

pakrym:
I tried:

```
D:\github\dotnet\cli\test\Performance [pakrym/perf-tests +1 ~0 -0 !]> C:\Users\pakrym\AppData\Local\Programs\Python\Python35-32\python.exe .\test.py --name "runid" --xunit-perf-path d:\temp\xperf D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe
# Clone the xunit-performance repo...
  $ git clone https://github.com/microsoft/xunit-performance.git d:\temp\xperf > "D:\github\dotnet\cli\test\Performance\logs\run-perftests\clone-the-xunit-performance-repo.log"
# Build xunit-performance...
  $ cd "d:\temp\xperf"
  $ CiBuild.cmd /release > "D:\github\dotnet\cli\test\Performance\logs\run-perftests\build-xunit-performance.log"
# Microsoft.DotNet.xunit.performance.runner.cli found already published. Use --rebuild to publish again.
# Prepending D:\github\dotnet\cli\artifacts\win10-x64\stage2 to PATH
# Run runid.test...
  $ cd "D:\github\dotnet\cli\test\Performance"
  $ d:\temp\xperf\tools\bin\dotnet run -p d:\temp\xperf\src\cli\Microsoft.DotNet.xunit.performance.runner.cli -c Release -- -runner D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe -runid runid.test -runnerargs "test D:\github\dotnet\cli\test\Performance\project.json -c Release" > "D:\github\dotnet\cli\test\Performance\logs\run-perftests\run-runid-test.log"
# Reverting PATH
! ERROR: Command `d:\temp\xperf\tools\bin\dotnet run -p d:\temp\xperf\src\cli\Microsoft.DotNet.xunit.performance.runner.cli -c Release -- -runner D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe -runid runid.test -runnerargs "test D:\github\dotnet\cli\test\Performance\project.json -c Release" > "D:\github\dotnet\cli\test\Performance\logs\run-perftests\run-runid-test.log"` returned with error code 2147516547. See 'D:\github\dotnet\cli\test\Performance\logs\run-perftests\run-runid-test.log' for details.
```
run-runid-test.log
```
Project Microsoft.DotNet.xunit.performance.runner.cli (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
The targeted framework { "Microsoft.NETCore.App": "1.0.0-rc2-3002545" } was not found.
  - Check application dependencies and target a framework version installed at:
      d:\temp\xperf\tools\bin\shared\Microsoft.NETCore.App
  - The following versions are installed:
      1.0.0-rc2-3002543

```

`dotnet test` works fine

brianrob:
@dpodder, can you please take a look at @pakrym's test result?

dpodder:
@pakrym Do you get the same error doing `dotnet run` manually?
```batch
d:\temp\xperf\tools\bin\dotnet run -p d:\temp\xperf\src\cli\Microsoft.DotNet.xunit.performance.runner.cli -c Release -- -runner D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe -runid runid.test -runnerargs "test D:\github\dotnet\cli\test\Performance\project.json -c Release"
```

pakrym:
Yes

brianrob:
@pakrym, were you and @dpodder able to resolve this?  Want to make sure that this test runs under the perf harness before merging.

pakrym:
No, but I don't think this PR is blocked on script, I added Performance project to list of CLI test projects to make sure it run's fine with `dotnet test`. Script issues should be resolved in script PR.

brianrob:
This really should not have been merged until you were able to see the tests run under the harness.  It's not about the script - it's about whether the test runs under the performance harness, which is a C# app that the script runs.  Can you please work with @dpodder to make sure this runs?

pakrym:
It's not the case of "Tests doesn't run with harness" more of "Harness doesn't run at all". If some issues would be discovered when harness is fixed I'm happy to help and fix them.

dpodder:
@pakrym @brianrob I found a workaround that I can do in the script: running the test using the newest CLI (the one under test) instead of the one from xunit-performance. Even the stage0 dotnet.exe doesn't work. Is this expected/acceptable?

```batch
d:\temp\xperf\tools\bin\dotnet run -p d:\temp\xperf\src\cli\Microsoft.DotNet.xunit.performance.runner.cli -c Release -- -runner D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe -runid runid.test -runnerargs "test D:\github\dotnet\cli\test\Performance\project.json -c Release"
```
becomes:
```batch
D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe run -p d:\temp\xperf\src\cli\Microsoft.DotNet.xunit.performance.runner.cli -c Release -- -runner D:\github\dotnet\cli\artifacts\win10-x64\stage2\dotnet.exe -runid runid.test -runnerargs "test D:\github\dotnet\cli\test\Performance\project.json -c Release"
```

brianrob:
@dpodder, that sound ok, but do you know why stage0 doesn't work?  I would expect there to be a process boundary here so that running with the same version of the CLI wouldn't matter.  Under the covers, CLI process 1 runs the harness and invokes a new dotnet.exe process (2) to run the test.  I would not expect them to have to be the same version of the CLI/runtime/etc.

dpodder:
@brianrob I believe this is resolved with dotnet/cli commit [a677777](https://github.com/dotnet/cli/pull/2746/commits/a677777ecc392ed9780dc03fdaf6020bd1880356).

@pakrym Please test this with the updated script once #2746 is merged, and if you still run into problems let me know.

PR #2679, opened 2016-04-25
align code in dotnet-run

anurse:
we used to use different code when --framework was specified than when it was not specified, this synchronizes them to use the same code path which removes a hidden NullRef

also adds tests to cover both cases

/cc @pakrym @davidfowl

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2679)

### DIFF
--- a/src/dotnet/commands/dotnet-run/RunCommand.cs
+++ b/src/dotnet/commands/dotnet-run/RunCommand.cs
@@ -20,6 +20,13 @@ namespace Microsoft.DotNet.Tools.Run
         public string Project = null;
         public IReadOnlyList<string> Args = null;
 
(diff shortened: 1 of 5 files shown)


pakrym:
:shipit:

src/dotnet/commands/dotnet-run/RunCommand.cs
```diff
@@ -20,6 +20,13 @@ public partial class RunCommand
         public string Project = null;
         public IReadOnlyList<string> Args = null;
 
+        public static readonly string[] DefaultFrameworks = new[]
```
eerhardt: Does this need to be public?

PR #2673, opened 2016-04-25
Encode wide char to CP_ACP for CLR usage

schellap:
Fixes https://github.com/dotnet/cli/issues/2649

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2673)

PR #2680, opened 2016-04-25
dotnet build fails if the directory contains unicode characters

### LINKED ISSUE #2486: dotnet build fails if the directory contains unicode characters
## Steps to reproduce
    cd /d %USERPROFILE%
    mkdir Tést
    cd Tést
    dotnet new
    dotnet restore
    dotnet build

## Expected  behavior
Build succeeds

## Actual behavior

    Response file 'C:\Users\greggm\T├â┬⌐st\obj\Debug\netcoreapp1.0\dotnet-compile.rsp' doesn't exist.
    C:\Program Files\dotnet\dotnet.exe compile-csc @C:\Users\greggm\T├⌐st\obj\Debug\netcoreapp1.0\dotnet-compile.rsp returned Exit Code 1
    
    Compilation failed.

## Environment data
`dotnet --info` output:
.NET Command Line Tools (1.0.0-rc2-002330)

Product Information:
 Version:     1.0.0-rc2-002330
 Commit Sha:  3af38bf0fe

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64

eerhardt:
This fix contains @schellap's fix for https://github.com/dotnet/cli/pull/2673, which is needed to fix this bug and will be rebased out once @schellap merges his change.

Fix #2486 

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2680)

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.ProjectModel.Server;
 using Microsoft.DotNet.Tools.Build;
@@ -40,6 +41,8 @@ namespace Microsoft.DotNet.Cli
         public static int Main(string[] args)
         {
             DebugHelper.HandleDebugSwitch(ref args);
+
+            InitializeProcess();
 
             try
             {
@@ -136,6 +139,13 @@ namespace Microsoft.DotNet.Cli
 
             return exitCode;
 
+        }
+
+        private static void InitializeProcess()
+        {
+            // by default, .NET Core doesn't have all code pages needed for Console apps.
+            // see the .NET Core Notes in https://msdn.microsoft.com/en-us/library/system.diagnostics.process(v=vs.110).aspx
+            Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
         }
 
         internal static bool TryGetBuiltInCommand(string commandName, out Func<string[], int> builtInCommand)
--- a/test/ArgumentForwardingTests/ArgumentForwardingTests.cs
+++ b/test/ArgumentForwardingTests/ArgumentForwardingTests.cs
@@ -51,6 +51,7 @@ namespace Microsoft.DotNet.Tests.ArgumentForwarding
         /// <param name="testUserArgument"></param>
         [Theory]
         [InlineData(@"""abc"" d e")]
+        [InlineData(@"""ábc"" d é")]
         [InlineData(@"""abc""      d e")]
         [InlineData("\"abc\"\t\td\te")]
         [InlineData(@"a\\b d""e f""g h")]
--- a/TestAssets/TestProjects/TestAppWithUnicodéPath/Program.cs
+++ b/TestAssets/TestProjects/TestAppWithUnicodéPath/Program.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace ConsoleApplication
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
+        }
+    }
+}
--- a/TestAssets/TestProjects/TestAppWithUnicodéPath/Program.cs
+++ b/TestAssets/TestProjects/TestAppWithUnicodéPath/Program.cs
@@ -6,7 +6,7 @@ namespace ConsoleApplication
     {
         public static void Main(string[] args)
         {
-            Console.WriteLine("Hello World!");
+            Console.WriteLine("Hélló Wórld!");
         }
     }
 }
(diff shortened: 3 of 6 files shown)


piotrpMSFT:
@brthor review this please

src/dotnet/Program.cs
```diff
@@ -138,6 +141,16 @@ internal static int ProcessArgs(string[] args, ITelemetry telemetryClient)
 
         }
 
+        /// <summary>
+        /// Performs initialization that must be done before anything else in the process.
```
brthor: nit: I don't think this comment is necessary, I would infer this from the method name.

src/dotnet/Program.cs
```diff
@@ -138,6 +141,16 @@ internal static int ProcessArgs(string[] args, ITelemetry telemetryClient)
...
+        /// </summary>
+        private static void InitializeProcess()
+        {
+            // by default, .NET Core doesn't have all code pages needed for Console apps.
```
brthor: 👍 for this comment though

test/ArgumentForwardingTests/ArgumentForwardingTests.cs
```diff
@@ -51,6 +51,7 @@ private void FindAndEnsureReflectorPresent()
         /// <param name="testUserArgument"></param>
         [Theory]
         [InlineData(@"""abc"" d e")]
+        [InlineData(@"""ábc"" d é")]
```
brthor: Although I added this for quick and easy coverage, I think it'd be best to add an explicit unicode fact instead of putting it in this theory. Don't think the PR should block on this though.

brthor:
Didn't review host changes, wasn't sure what was going on there.

Managed changes, couple small comments but :shipit: 

Thanks for fixing this!

test/ArgumentForwardingTests/ArgumentForwardingTests.cs
eerhardt: I'm not sure I follow this fully.  Can you explain a little more what you'd like to see?  What would the fact do that this doesn't?

Also, the `dotnet run` tests exercise this in a more end-to-end manner since dotnet run will spawn the app process, and forward the output back to the CLI.

src/dotnet/Program.cs
eerhardt: Removed.

eerhardt:
> Didn't review host changes, wasn't sure what was going on there.

Those were @schellap's fix for #2673.  I needed them for the tests to pass.  Waiting until his change goes in, then I'll rebase and just merge our 3 commits on the managed side.

src/dotnet/Program.cs
```diff
@@ -138,6 +141,16 @@ internal static int ProcessArgs(string[] args, ITelemetry telemetryClient)
...
+            // see the .NET Core Notes in https://msdn.microsoft.com/en-us/library/system.diagnostics.process(v=vs.110).aspx
+            Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
```
weshaggard: As an FYI System.Text.Encoding.CodePages library is currently only in the shared framework because Roslyn pulls it in and it might not be there in the future which would cause it to be pulled app-local with the CLI because of this (perhaps there are already other dependencies on it as well). That assembly is ~750MB and is intentionally supposed to be pay-for-play, I'm not entirely sure why it is actually needed here as it should be needed to support Unicode encodings.

weshaggard: Reading through some of the other issues I see it appears to be the default Console encoding that is causing the dependency. That is kind of unfortunate that the Console doesn't use a standard algorithm encoding by default.

eerhardt: >  I'm not entirely sure why it is actually needed here as it should be needed to support Unicode encodings.

On my Windows machine, the default code page is:

```
C:\cli>chcp
Active code page: 437
```

Which is not in the default .NET Core code page list.  Similarly, users can set their console code page to whatever they want, and we should be able to work with it, if possible.

Also, another thing we do in the CLI is launch separate processes, and then forward the process's standard error and out to our process, and interpret it. One example of this is when we run `csc.exe`, we Regex the output looking for compiler warnings and errors so we can count them up at the end. Another is when we launch the user's app, we forward the output to our output during `dotnet run`. The separate process can use whatever code page it wants to write to its standard out. If we don't have the correct code page, we won't be reading the results correctly.

Looking at my machine, the System.Text.Encoding.CodePages.dll is 852 **K**B.

/cc @tarekgh

eerhardt:
test OSX x64 Release Build

test/ArgumentForwardingTests/ArgumentForwardingTests.cs
brthor: The fact wouldn't do anything the theory doesn't, so from that perspective, there's not much value. 

The value I see in the separation is better readability of the expectations. This essentially tests that we round trip arguments properly when doing process jumps. Having the fact would just make it clear that we are testing unicode is round tripped right. 

Ultimately it's a fit/finish thing, and I don't feel too strongly that it needs to move.

eerhardt:
test OSX x64 Release Build

PR #2653, opened 2016-04-22
Piotrp msft/issue2584/multicore jit

piotrpMSFT:
Enables Multicore JIT for CLI
Adds off switch for multicore JIT
Refactors some test infrastructure to allow for non-asset-based tests

Fixes #2584 

/cc @brianrob @Sridhar-MS @sokket

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2653)

PR #2681, opened 2016-04-25
Build-Time Hello World Performance Test

brianrob:
Adds a new performance test targeted at measuring the latency of dotnet build on the Hello World app.  The new test is heavily based off of EndToEnd\EndToEndTest.cs.

Next step after merging this PR is to build a script that will run this test under harness and generate a measurement result.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2681)

### DIFF
--- a/test/Performance/HelloWorld.cs
+++ b/test/Performance/HelloWorld.cs
@@ -1,0 +1,108 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using Microsoft.Extensions.PlatformAbstractions;
+using Microsoft.Xunit.Performance;
+using Xunit;
+using System.Diagnostics;
+
+namespace Microsoft.DotNet.Tests.Performance
+{
+    public class HelloWorld : TestBase
+    {
+        private static readonly string NetCoreAppTfm = "netcoreapp1.0";
+        private static readonly string s_expectedOutput = "Hello World!" + Environment.NewLine;
+        private static readonly string s_testdirName = "helloworldtestroot";
+        private static readonly string s_outputdirName = "test space/bin";
+
+        private static string RestoredTestProjectDirectory { get; set; }
+
+        private string TestDirectory { get; set; }
+        private string TestProject { get; set; }
+        private string OutputDirectory { get; set; }
+
+        static HelloWorld()
+        {
+            HelloWorld.SetupStaticTestProject();
+        }
+
+        public HelloWorld()
+        {
+        }
+
+        [Benchmark]
+        public void MeasureDotNetBuild()
+        {
+            foreach (var iter in Benchmark.Iterations)
+            {
+                // Setup a new instance of the test project.
+                TestInstanceSetup();
+
+                // Setup the build command.
+                var buildCommand = new BuildCommand(TestProject, output: OutputDirectory, framework: NetCoreAppTfm);
+                using (iter.StartMeasurement())
+                {
+                    // Execute the build command.
+                    buildCommand.Execute();
+                }
+            }
+        }
+
+        private void TestInstanceSetup()
+        {
+            var root = Temp.CreateDirectory();
+
+            var testInstanceDir = root.CopyDirectory(RestoredTestProjectDirectory);
+
+            TestDirectory = testInstanceDir.Path;
+            TestProject = Path.Combine(TestDirectory, "project.json");
+            OutputDirectory = Path.Combine(TestDirectory, s_outputdirName);
+        }
+
+        private static void SetupStaticTestProject()
+        {
+            RestoredTestProjectDirectory = Path.Combine(AppContext.BaseDirectory, "bin", s_testdirName);
+
+            // Ignore Delete Failure
+            try
+            {
+                Directory.Delete(RestoredTestProjectDirectory, true);
+            }
+            catch (Exception) { }
+
+            Directory.CreateDirectory(RestoredTestProjectDirectory);
+            WriteNuGetConfig(RestoredTestProjectDirectory);
+
+            var currentDirectory = Directory.GetCurrentDirectory();
+            Directory.SetCurrentDirectory(RestoredTestProjectDirectory);
+
+            new NewCommand().Execute().Should().Pass();
+            new RestoreCommand().Execute("--quiet").Should().Pass();
+
+            Directory.SetCurrentDirectory(currentDirectory);
+        }
+
+        // Todo: this is a hack until corefx is on nuget.org remove this After RC 2 Release
+        private static void WriteNuGetConfig(string directory)
+        {
+            var contents = @"<?xml version=""1.0"" encoding=""utf-8""?>
+<configuration>
+<packageSources>
+<!--To inherit the global NuGet package sources remove the <clear/> line below -->
+<clear />
+<add key=""dotnet-core"" value=""https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"" />
+<add key=""api.nuget.org"" value=""https://api.nuget.org/v3/index.json"" />
+</packageSources>
+</configuration>";
+
+            var path = Path.Combine(directory, "NuGet.config");
+
+            File.WriteAllText(path, contents);
+        }
+    }
+}
--- a/test/Performance/HelloWorld.cs
+++ b/test/Performance/HelloWorld.cs
@@ -54,10 +54,11 @@ namespace Microsoft.DotNet.Tests.Performance
 
         private void TestInstanceSetup()
         {
-            var assetsManager = new TestAssetsManager(AssetsRoot);
-            var testInstance = assetsManager.CreateTestInstance(s_testdirName, string.Empty, Guid.NewGuid().ToString()).WithLockFiles();
+            var root = Temp.CreateDirectory();
 
-            TestDirectory = testInstance.Path;
+            var testInstanceDir = root.CopyDirectory(RestoredTestProjectDirectory);
+
+            TestDirectory = testInstanceDir.Path;
             TestProject = Path.Combine(TestDirectory, "project.json");
             OutputDirectory = Path.Combine(TestDirectory, s_outputdirName);
         }
--- a/test/Performance/HelloWorld.cs
+++ b/test/Performance/HelloWorld.cs
@@ -15,11 +15,10 @@ namespace Microsoft.DotNet.Tests.Performance
 {
     public class HelloWorld : TestBase
     {
-        private static readonly string NetCoreAppTfm = "netcoreapp1.0";
-        private static readonly string s_expectedOutput = "Hello World!" + Environment.NewLine;
         private static readonly string s_testdirName = "helloworldtestroot";
         private static readonly string s_outputdirName = "test space/bin";
 
+        private static string AssetsRoot { get; set; }
         private static string RestoredTestProjectDirectory { get; set; }
 
         private string TestDirectory { get; set; }
@@ -44,7 +43,7 @@ namespace Microsoft.DotNet.Tests.Performance
                 TestInstanceSetup();
 
                 // Setup the build command.
-                var buildCommand = new BuildCommand(TestProject, output: OutputDirectory, framework: NetCoreAppTfm);
+                var buildCommand = new BuildCommand(TestProject, output: OutputDirectory, framework: DefaultFramework);
                 using (iter.StartMeasurement())
                 {
                     // Execute the build command.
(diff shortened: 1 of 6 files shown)


brianrob:
@piotrpMSFT, PTAL
@lt72, @cmckinsey FYI

test/Performance/Performance.xproj
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
eerhardt: Can you add this to the .sln file?

test/Performance/project.json
```diff
@@ -0,0 +1,36 @@
...
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0-rc2-*"
+    },
+    "System.Runtime.Serialization.Primitives": "4.1.1-rc2-24015",
```
eerhardt: This is out of date with the version used in the rest of the repo.

test/Performance/project.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    "Microsoft.DotNet.Cli.Utils": {
+      "target": "project"
+    },
+    "Microsoft.DotNet.Tools.Tests.Utilities": {
+      "target": "project"
+    },
+    "xunit": "2.1.0",
+    "xunit.netcore.extensions": "1.0.0-prerelease-00206",
+    "dotnet-test-xunit": "1.0.0-dev-140469-38",
```
eerhardt: This is out of date with the version used in the rest of the repo.

test/Performance/project.json
```diff
@@ -0,0 +1,36 @@
...
+    "Microsoft.DotNet.xunit.performance": "1.0.0-alpha-build0028"
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": [
+        "netstandardapp1.5",
```
eerhardt: What needs `netstandardapp1.5`.  Did you copy this from somewhere?  I removed this line from the rest of the test projects.  It shouldn't be needed

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,117 @@
...
+            var currentDirectory = Directory.GetCurrentDirectory();
+            Directory.SetCurrentDirectory(RestoredTestProjectDirectory);
+
+            new NewCommand().Execute().Should().Pass();
+            new RestoreCommand().Execute("--quiet").Should().Pass();
+
+            Directory.SetCurrentDirectory(currentDirectory);
+        }
+
+        // Todo: this is a hack until corefx is on nuget.org remove this After RC 2 Release
```
eerhardt: Can we share this code with EndToEnd somehow?  Maybe moving it into one of our "test utils/framework" assemblies?

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,117 @@
...
+                    buildCommand.Execute();
+                }
+            }
+        }
+
+        private void TestInstanceSetup()
+        {
+            var root = Temp.CreateDirectory();
+
+            var testInstanceDir = root.CopyDirectory(RestoredTestProjectDirectory);
```
eerhardt: We have a `TestAssetManager` for things like this.  Can we use it here?

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,117 @@
...
+        private string TestDirectory { get; set; }
+        private string TestProject { get; set; }
+        private string OutputDirectory { get; set; }
+
+        static HelloWorld()
+        {
+            HelloWorld.SetupStaticTestProject();
+        }
+
+        public static void Main()
```
eerhardt: ?? is this still necessary? For example, `dotnet-run.Tests` doesn't have this.

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            Console.WriteLine("Dummy Entrypoint.");
+        }
+
+        public HelloWorld()
+        {
+            TestInstanceSetup();
```
eerhardt: Do we need to do this here, if we are doing it everytime through the loop?

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            var root = Temp.CreateDirectory();
+
+            var testInstanceDir = root.CopyDirectory(RestoredTestProjectDirectory);
+
+            TestDirectory = testInstanceDir.Path;
+            TestProject = Path.Combine(TestDirectory, "project.json");
+            OutputDirectory = Path.Combine(TestDirectory, s_outputdirName);
+
+            Rid = PlatformServices.Default.Runtime.GetLegacyRestoreRuntimeIdentifier();
```
eerhardt: Rid doesn't appear to be read anywhere.

test/Performance/HelloWorld.cs
pakrym: Can we use projects from `TestAssets` folder which are restored automatically during build process? I don't think this test needs to do `dotnet new` and `dotnet restore`

eerhardt: If the intent is to measure the build time of a `dotnet new` empty project, then I think actually using `dotnet new` to generate the project is best. The reasoning is because as soon as you check something into `TestAssets` it will start diverging from what `dotnet new` lays down.

If the intent is to measure any old `Hello, World!` app, then I agree with @pakrym, we can just use something in `TestAssets`.

pakrym: https://github.com/dotnet/cli/pull/2681/files#diff-6876439b00d6c2806174918892c08f48R58 only measures build execution

eerhardt: It's not a question of "which operation is it measuring?".  It's a question of "which project is it trying to measure?".

I'm saying if the intent of the test is to measure the build time of a freshly generated **dotnet new** project, then actually using `dotnet new` to create the project is correct. We change the `dotnet new` template all the time without changing a corresponding `TestAsset` project.

test/Performance/Performance.xproj
brianrob: Fixed.

test/Performance/project.json
brianrob: Fixed.

test/Performance/project.json
brianrob: Fixed.

test/Performance/project.json
brianrob: Yes, this came from EndToEnd - I've updated it.

test/Performance/HelloWorld.cs
brianrob: The goal of this test is to test the build time of "dotnet hello world", so running dotnet new is desired in this case.

test/Performance/HelloWorld.cs
brianrob: Looks like this is no longer necessary.  I've removed it.

test/Performance/HelloWorld.cs
brianrob: Nope, fixed.

test/Performance/HelloWorld.cs
brianrob: You're right.  Removed.

brianrob:
I believe I've addressed all of the feedback, with this latest commit.

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                Directory.Delete(RestoredTestProjectDirectory, true);
+            }
+            catch (Exception) { }
+
+            Directory.CreateDirectory(RestoredTestProjectDirectory);
+            WriteNuGetConfig(RestoredTestProjectDirectory);
+
+            var currentDirectory = Directory.GetCurrentDirectory();
+            Directory.SetCurrentDirectory(RestoredTestProjectDirectory);
```
pakrym: Usually we pass project path to commands instead of changing current directory. I think E2E test is written this way to check that user can do it without paths but in case of this test it would be good to avoid changing global state.

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.Extensions.PlatformAbstractions;
+using Microsoft.Xunit.Performance;
+using Xunit;
+using System.Diagnostics;
+
+namespace Microsoft.DotNet.Tests.Performance
+{
+    public class HelloWorld : TestBase
+    {
+        private static readonly string NetCoreAppTfm = "netcoreapp1.0";
```
pakrym: `DefaultFramework` property of `TestBase` class

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,108 @@
...
+        private static readonly string s_expectedOutput = "Hello World!" + Environment.NewLine;
```
pakrym: Not used

test/Performance/HelloWorld.cs
piotrpMSFT: See TestCommand.cs@20: public string WorkingDirectory. Setting this process-wide means we cannot do parallel test execution.

test/Performance/HelloWorld.cs
piotrpMSFT: This should use the TestAssetManager. Unfortunately, it's not currently building new directories that are empty. Fortunately, that fix is in https://github.com/dotnet/cli/pull/2653.

This PR, amongst other things, refactors the TestInstance into TestInstance:TestDirectory where the TestDirectory is just an empty directory placed in "the right path" for this test's usage. 

@brianrob could you either wait for that PR or just cherry pick https://github.com/dotnet/cli/pull/2653/commits/1feed70df57558f0777f850e44904b1ca2eb4835 and move the test onto the common infra?

test/Performance/HelloWorld.cs
eerhardt: So back to my original feedback: Can we share this code with EndToEnd?

test/Performance/project.json
eerhardt: Doesn't appear fixed.

test/Performance/project.json
eerhardt: Doesn't appear fixed.

test/Performance/project.json
eerhardt: Doesn't appear fixed.

test/Performance/HelloWorld.cs
brianrob: Fixed.

test/Performance/HelloWorld.cs
brianrob: Fixed.

test/Performance/HelloWorld.cs
brianrob: Fixed.  FYI, this still exists for EndToEnd.

test/Performance/HelloWorld.cs
brianrob: Are you just referring to the Nuget.config method?  If so, I can move it if you can make a recommendation as to where.

eerhardt: Yes I am referring to this method:

```C#
 private static void WriteNuGetConfig(string directory)
        {
            var contents = @"<?xml version=""1.0"" encoding=""utf-8""?>
<configuration>
<packageSources>
<clear />
<add key=""dotnet-core"" value=""https://dotnet.myget.org/F/dotnet-core/api/v3/index.json"" />
<add key=""api.nuget.org"" value=""https://api.nuget.org/v3/index.json"" />
</packageSources>
</configuration>";

            var path = Path.Combine(directory, "NuGet.config");

            File.WriteAllText(path, contents);
        }
```

I think it can be moved to the `Microsoft.DotNet.Tools.Tests.Utilities` project.  Both EndToEnd and this test reference that assembly.

pakrym:
How do we collect data from this test? Whats the point of merging it without having at leas some way to get performance measurements?

brianrob:
@pakrym, This can be run using the xunit-performance harness at http://github.com/Microsoft/xunit-performance.  Generally, each repo has a script to run these tests.  There are some other folks from my team who will be creating the script to run it.  This has been the pattern in CoreFX and CoreCLR as well - test creation was done in parallel with the work to allow the tests to be run in the repo.

pakrym:
Is there a task tracking harness work, whats the ETA? I'm asking because performance work is happening for a week already and there is no way to measure changes and this PR doesn't solve that issue either.

brianrob:
https://github.com/dotnet/cli/issues/2522 tracks the ability to measure performance of the CLI build scenario.

This PR addresses part of that.  I don't have an ETA yet on when the runner will be scripted nicely, but I'd like to believe we're a few days away.  If you want to run it today, I can provide instructions.

There's been lots of investment in performance across the board, and while having a measuring infrastructure for devs is good, there have been competing priorities around measurement, profiling, and building infra for the team to measure.  Thus, this work had to fit into that prioritization exercise, thus it didn't start immediately - it really only started last Friday.

cc: @lt72

pakrym:
Do you think you can add simple "run performance tests using dotnet currently on `PATH` and  dump result to console" type of script to this PR?

lt72:
Valentin is taking care of providing exactly such script. There are a few build/packaging issues around XUnit.Performance we need to iron out, so it is not that simple and will need a separate PR. It should take just another 24hrs after this PR is merged.

brianrob:

test/Performance/HelloWorld.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
brthor: Would it make sense to change the filename here to represent what this test is measuring?

Hello World doesn't tell me much about what this file tests, but `dotnet-build.Performance.Tests.cs` might (as an example).

brianrob: We could.  However, I was envisioning this test eventually doing more.  Today, we are most concerned with build, but I could imagine us wanting to expand the number of operations, at which point the name of the file would change too.

brthor: Would it then make sense to break the different operations into different files, or is the focus of this file purely how performant different operations are against a hello world app?

brianrob: The way I'm envisioning this is that this file represents "perf tests against hello world".  So, basically the latter option in your comment.

piotrpMSFT:
we can :shipit: without TestAssetManager change. Once that infrastructure is in place I'll go clean up tests to take advantage.

brianrob:
Thanks much @piotrpMSFT.

brianrob:

brianrob:

piotrpMSFT:
@brianrob what were the failure causes you saw?

brianrob:
There were a couple of test failures that I don't believe were related to my changes in Windows Release, and a timeout on Debug.  I've lost the link, but they seem to have passed on re-run.

PR #2696, opened 2016-04-26
abort build when an unresolved dependency is found

### LINKED ISSUE #2695: Errors are unclear when a project has an unresolved dependency
## Steps to reproduce
1. `dotnet new`
2. Add a dependency on a project that does not exist, for example:

```json
{
  "version": "1.0.0-*",
  "compilationOptions": {
    "emitEntryPoint": true
  },
  "dependencies": {
    "Microsoft.NETCore.App": {
      "type": "platform",
      "version": "1.0.0-rc2-3002498"
    },
    "ThisIsNotARealDependencyAndIfSomeoneGoesAndAddsAProjectWithThisNameIWillFindThemAndPunishThem": { "target": "project" }
  },
  "frameworks": {
    "netcoreapp1.0": {
      "imports": "dnxcore50"
    }
  }
}
```
3. `dotnet restore` (this should fail but write a lock file)
4. `dotnet build`

## Expected  behavior
A helpful error message

## Actual behavior
Outputs

```
Unable to cast object of type 'Microsoft.DotNet.ProjectModel.LibraryDescription' to type 'Microsoft.DotNet.ProjectModel.ProjectDescription'.
```

Also, if this error is bypassed, you get even worse behavior. It emits an error for the unresolved dependency, but continues the build and as long as you don't use the code it actually produces a binary. This leads to weird behavior where you try again and it says it doesn't have to compile because the outputs are up-to-date.

## Environment data
`dotnet --info` output:

```
ps1 # dotnet --info

Product Information:
 Version:     1.0.0-rc2-002498
 Commit Sha:  55e561bfb3

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14329
 OS Platform: Windows
 RID:         win10-x64
```

anurse:
Fixes #2695 

/cc @pakrym @davidfowl

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2696)

### DIFF
--- a/src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
+++ b/src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
@@ -45,7 +45,7 @@ namespace Microsoft.DotNet.Tools.Build
                     LibraryDescription libraryDescription;
                     if (lookup.TryGetValue(dependency.Name, out libraryDescription))
                     {
-                        if (libraryDescription.Identity.Type.Equals(LibraryType.Project))
+                        if (libraryDescription.Resolved && libraryDescription.Identity.Type.Equals(LibraryType.Project))
                         {
                             deps.Add(TraverseProject((ProjectDescription)libraryDescription, lookup));
                         }
--- a/src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
+++ b/src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
@@ -63,11 +63,10 @@ namespace Microsoft.DotNet.Tools.Compiler
                 diagnostics.Add(diag);
             }
 
-            if (missingFrameworkDiagnostics.Count > 0)
+            if(diagnostics.Any(d => d.Severity == DiagnosticMessageSeverity.Error))
             {
-                // The framework isn't installed so we should short circuit the rest of the compilation
-                // so we don't get flooded with errors
-                PrintSummary(missingFrameworkDiagnostics, sw);
+                // We got an unresolved dependency or missing framework. Don't continue the compilation.
+                PrintSummary(diagnostics, sw);
                 return false;
             }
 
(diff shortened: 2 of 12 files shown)


pakrym:
:shipit:

anurse:
@dotnet-bot test this please

src/dotnet/commands/dotnet-compile/ManagedCompiler.cs
```diff
@@ -63,11 +63,10 @@ public override bool Compile(ProjectContext context, BuildCommandApp args)
                 diagnostics.Add(diag);
             }
 
-            if (missingFrameworkDiagnostics.Count > 0)
+            if(diagnostics.Any(d => d.Severity == DiagnosticMessageSeverity.Error))
```
davidfowl: nit: spacing

davidfowl:
Good change :shipit:

src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
```diff
@@ -44,7 +44,7 @@ private ProjectGraphNode TraverseProject(ProjectDescription project, IDictionary
                     LibraryDescription libraryDescription;
                     if (lookup.TryGetValue(dependency.Name, out libraryDescription))
                     {
-                        if (libraryDescription.Identity.Type.Equals(LibraryType.Project))
+                        if (libraryDescription.Resolved && libraryDescription.Identity.Type.Equals(LibraryType.Project))
```
cdmihai: Shouldn't it be an error if the project is not resolved?

Right now, if a project is not resolved the else branch is taken which implicitly assumes that it receives a non ProjectDescription. This might lead to a wrong graph.

anurse: It is an error and it will be caught later. This was causing an InvalidCastException.

Before compiling, we do check if all libraries are resolved.

anurse: And, an unresolved project library will not have any dependencies (since we couldn't find the project file) so even the else clause will no-op.

Issue #2698, opened 2016-04-26
Unix builds are failing to publish the new sharedfx+cli zip


Sridhar-MS:
All the unix builds are failing to publish the new sharedfx+cli zip with the below exception

>
2016-04-26T18:38:56.107Z: Build failed: System.AggregateException: One or more errors occurred. (Could not find file '/opt/code/artifacts/ubuntu.14.04-x64/packages/dotnet-sharedframework-sdk-ubuntu-x64.1.0.0-rc2-002500.tar.gz'.) ---> System.IO.FileNotFoundException: Could not find file '/opt/code/artifacts/ubuntu.14.04-x64/packages/dotnet-sharedframework-sdk-ubuntu-x64.1.0.0-rc2-002500.tar.gz'.
2016-04-26T18:38:56.107Z:    at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
2016-04-26T18:38:56.108Z: afeHandle](TSafeHandle handle, String path, Boolean isDirectory, Func`2 errorRewriter)
2016-04-26T18:38:56.108Z:    at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
2016-04-26T18:38:56.108Z:    at System.IO.UnixFileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
2016-04-26T18:38:56.108Z:    at System.IO.UnixFileSystem.Open(String fullPath, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, FileStream parent)
2016-04-26T18:38:56.108Z:    at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
2016-04-26T18:38:56.108Z:    at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob.<>c__DisplayClass23_0.<<UploadFromFileAsync>b__0>d.MoveNext()
2016-04-26T18:38:56.108Z:    --- End of inner exception stack trace ---
2016-04-26T18:38:56.108Z:    at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
[...]

PR #2699, opened 2016-04-26
Only publish SharedFramework + SDK for Windows

ellismg:
We don't build the combined framework + SDK tarball for any Unix
platforms, but we were incorrectly trying to publish it. Make the
publish step for this artifact specific to Windows as well.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2699)

### DIFF
--- a/scripts/dotnet-cli-build/PublishTargets.cs
+++ b/scripts/dotnet-cli-build/PublishTargets.cs
@@ -174,6 +174,7 @@ namespace Microsoft.DotNet.Cli.Build
         }
 
         [Target]
+        [BuildPlatforms(BuildPlatform.Windows)]
         public static BuildTargetResult PublishCombinedFrameworkSDKArchiveToAzure(BuildTargetContext c)
         {
             var version = CliNuGetVersion;


ellismg:
Fixes #2698

ellismg:
@Sridhar-MS @piotrpMSFT 

Sorry about that, What is the correct way to validate these publishing changes in the future to ensure I don't break the internal builds (CI was clean since the publishing step is skipped).

Sridhar-MS:
Why only for windows?

ellismg:
> Why only for windows? I mean why dont we generate fx_cli tarballs for Unix too.

My understanding is we only need these artifacts for scenarios where the host is distributed via MSI but we want to bring in the framework and SDK via a zip file.  I don't think we have this "host from an installer", SDK and SF from a zip/tarball scenario on Unix platforms.  Is that correct @piotrpMSFT?

Sridhar-MS:
Got it. Thanks!

Sridhar-MS:
:shipit:

PR #2703, opened 2016-04-27
Fixing publish scripts to only push to Latest when all bits are there

### LINKED ISSUE #2613: CLI Builds should publish when all builds are finished
## Steps to reproduce
go to https://github.com/dotnet/cli

## Expected  behavior
All versions are the same

## Actual behavior
Builds get out of sync, so you will have fixes on one OS but not on others.

sokket:
Fixes #2613  by waiting until all the current build number bits are available before pushing to the Latest blob storage

/cc @piotrpMSFT

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2703)

### DIFF
--- a/scripts/dotnet-cli-build/PublishTargets.cs
+++ b/scripts/dotnet-cli-build/PublishTargets.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net.Http;
 using System.Text;
 using Microsoft.DotNet.Cli.Build.Framework;
@@ -24,6 +26,8 @@ namespace Microsoft.DotNet.Cli.Build
 
         private static string SharedFrameworkNugetVersion { get; set; }
 
+        private static List<string> _latestBlobs;
+
         [Target]
         public static BuildTargetResult InitPublish(BuildTargetContext c)
         {
@@ -41,19 +45,119 @@ namespace Microsoft.DotNet.Cli.Build
         [Target(nameof(PrepareTargets.Init),
         nameof(PublishTargets.InitPublish),
         nameof(PublishTargets.PublishArtifacts),
-        nameof(PublishTargets.TriggerDockerHubBuilds))]
+        nameof(PublishTargets.TriggerDockerHubBuilds),
+        nameof(PublishTargets.FinalizeBuild))]
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
             return c.Success();
         }
 
+        [Target]
+        public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
+        {
+            if (CheckIfAllBuildsHavePublished())
+            {
+                string targetContainer = $"{Channel}/Binaries/Latest/";
+                string targetVersionFile = $"{targetContainer}{CliNuGetVersion}";
+                string leaseId = AzurePublisherTool.AcquireLeaseOnBlob(targetContainer);
+
+                // Prevent race conditions by dropping a version hint of what version this is. If we see this file
+                // and it is the same as our version then we know that a race happened where two+ builds finished 
+                // at the same time and someone already took care of publishing and we have no work to do.
+                if (AzurePublisherTool.IsLatestSpecifiedVersion(targetVersionFile))
+                {
+                    AzurePublisherTool.ReleaseLeaseOnBlob(targetContainer, leaseId);
+                    return c.Success();
+                }
+                else
+                {
+                    // This is an old drop of latest so remove all old files to ensure a clean state
+                    AzurePublisherTool.ListBlobs($"{targetContainer}").ToList().ForEach(f => AzurePublisherTool.DeleteBlob(f));
+
+                    // Drop the version file signaling such for any race-condition builds (see above comment).
+                    AzurePublisherTool.DropLatestSpecifiedVersion(targetVersionFile);
+                }
+
+                try
+                {
+                    // Iterate over every blob in the latest version folder and copy it to the 'latest' folder
+                    foreach (string blob in _latestBlobs)
+                    {
+                        string targetUrl = $"{targetContainer}{Path.GetFileName(blob)}"
+                                        .Replace(CliNuGetVersion, "latest")
+                                        .Replace(SharedFrameworkNugetVersion, "latest");
+                        AzurePublisherTool.CopyBlob(blob.Replace("/dotnet/", ""), targetUrl);
+                    }
+
+                    // Generate the CLI and SDK Version text files
+                    List<string> versionFiles = new List<string>() { "win.x86.version", "win.x64.version", "ubuntu.x64.version", "rhel.x64.version", "osx.x64.version", "debian.x64.version", "centos.x64.version" };
+                    string cliVersion = Utils.GetCliVersionFileContent(c);
+                    string sfxVersion = Utils.GetSharedFrameworkVersionFileContent(c);
+                    foreach (string version in versionFiles)
+                    {
+                        AzurePublisherTool.PublishStringToBlob($"{Channel}/dnvm/latest.{version}", cliVersion);
+                        AzurePublisherTool.PublishStringToBlob($"{Channel}/dnvm/latest.sharedfx.{version}", sfxVersion);
+                    }
+                }
+                finally
+                {
+                    AzurePublisherTool.ReleaseLeaseOnBlob(targetContainer, leaseId);
+                }
+            }
+
+            return c.Success();
+        }
+
+        private static bool CheckIfAllBuildsHavePublished()
+        {
+             Dictionary<string, bool> badges = new Dictionary<string, bool>()
+             {
+                 { "Windows_x86", false },
+                 { "Windows_x64", false },
+                 { "Ubuntu_x64", false },
+                 { "RHEL_x64", false },
+                 { "OSX_x64", false },
+                 { "Debian_x64", false },
+                 { "CentOS_x64", false }
+             };
+
+            _latestBlobs = new List<string>(AzurePublisherTool.ListBlobs($"{Channel}/Binaries/{CliNuGetVersion}/"));
+
+            var config = Environment.GetEnvironmentVariable("CONFIGURATION");
+            var versionBadgeName = $"{CurrentPlatform.Current}_{CurrentArchitecture.Current}";
+            if (badges.ContainsKey(versionBadgeName) == false)
+            {
+                throw new ArgumentException("A new OS build was added without adding the moniker to the {nameof(badges)} lookup");
+            }
+
+            foreach (string file in _latestBlobs)
+            {
+                string name = Path.GetFileName(file);
+                string key = string.Empty;
+
+                foreach (string img in badges.Keys)
+                {
+                    if ((name.StartsWith($"{img}")) && (name.EndsWith(".svg")))
+                    {
+                        key = img;
+                        break;
+                    }
+                }
+
+                if (string.IsNullOrEmpty(key) == false)
+                {
+                    badges[key] = true;
+                }
+            }
+
+            return badges.Keys.All(key => badges[key]);
+        }
+
         [Target(
             nameof(PublishTargets.PublishInstallerFilesToAzure),
             nameof(PublishTargets.PublishArchivesToAzure),
             nameof(PublishTargets.PublishDebFilesToDebianRepo),
-            nameof(PublishTargets.PublishLatestCliVersionTextFile),
-            nameof(PublishTargets.PublishLatestSharedFrameworkVersionTextFile),
             nameof(PublishTargets.PublishCoreHostPackages),
             nameof(PublishTargets.PublishCliVersionBadge))]
         public static BuildTargetResult PublishArtifacts(BuildTargetContext c) => c.Success();
@@ -87,11 +191,8 @@ namespace Microsoft.DotNet.Cli.Build
         public static BuildTargetResult PublishCliVersionBadge(BuildTargetContext c)
         {
             var versionBadge = c.BuildContext.Get<string>("VersionBadge");
-            var latestVersionBadgeBlob = $"{Channel}/Binaries/Latest/{Path.GetFileName(versionBadge)}";
             var versionBadgeBlob = $"{Channel}/Binaries/{CliNuGetVersion}/{Path.GetFileName(versionBadge)}";
-
             AzurePublisherTool.PublishFile(versionBadgeBlob, versionBadge);
-            AzurePublisherTool.PublishFile(latestVersionBadgeBlob, versionBadge);
             return c.Success();
         }
 
@@ -120,7 +221,7 @@ namespace Microsoft.DotNet.Cli.Build
                 installerFile = Path.ChangeExtension(installerFile, "msi");
             }
 
-            AzurePublisherTool.PublishInstallerFileAndLatest(installerFile, Channel, version);
+            AzurePublisherTool.PublishInstallerFile(installerFile, Channel, version);
 
             return c.Success();
         }
@@ -132,7 +233,7 @@ namespace Microsoft.DotNet.Cli.Build
             var version = SharedFrameworkNugetVersion;
             var installerFile = c.BuildContext.Get<string>("SharedFrameworkInstallerFile");
 
-            AzurePublisherTool.PublishInstallerFileAndLatest(installerFile, Channel, version);
+            AzurePublisherTool.PublishInstallerFile(installerFile, Channel, version);
 
             return c.Success();
         }
@@ -144,7 +245,7 @@ namespace Microsoft.DotNet.Cli.Build
             var version = CliNuGetVersion;
             var installerFile = c.BuildContext.Get<string>("SdkInstallerFile");
 
-            AzurePublisherTool.PublishInstallerFileAndLatest(installerFile, Channel, version);
+            AzurePublisherTool.PublishInstallerFile(installerFile, Channel, version);
 
             return c.Success();
         }
@@ -156,7 +257,7 @@ namespace Microsoft.DotNet.Cli.Build
             var version = SharedFrameworkNugetVersion;
             var installerFile = c.BuildContext.Get<string>("CombinedFrameworkHostInstallerFile");
 
-            AzurePublisherTool.PublishInstallerFileAndLatest(installerFile, Channel, version);
+            AzurePublisherTool.PublishInstallerFile(installerFile, Channel, version);
 
             return c.Success();
         }
@@ -168,7 +269,7 @@ namespace Microsoft.DotNet.Cli.Build
             var version = CliNuGetVersion;
             var installerFile = c.BuildContext.Get<string>("CombinedFrameworkSDKHostInstallerFile");
 
-            AzurePublisherTool.PublishInstallerFileAndLatest(installerFile, Channel, version);
+            AzurePublisherTool.PublishInstallerFile(installerFile, Channel, version);
 
             return c.Success();
         }
@@ -179,7 +280,7 @@ namespace Microsoft.DotNet.Cli.Build
             var version = CliNuGetVersion;
             var archiveFile = c.BuildContext.Get<string>("CombinedFrameworkSDKCompressedFile");
 
-            AzurePublisherTool.PublishArchiveAndLatest(archiveFile, Channel, version);
+            AzurePublisherTool.PublishArchive(archiveFile, Channel, version);
 
             return c.Success();
         }
@@ -190,7 +291,7 @@ namespace Microsoft.DotNet.Cli.Build
             var version = CliNuGetVersion;
             var archiveFile = c.BuildContext.Get<string>("CombinedFrameworkSDKHostCompressedFile");
 
-            AzurePublisherTool.PublishArchiveAndLatest(archiveFile, Channel, version);
+            AzurePublisherTool.PublishArchive(archiveFile, Channel, version);
 
             return c.Success();
         }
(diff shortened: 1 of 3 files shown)


scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
+            if (ShouldPublishToLatest())
+            {
+                string targetContainer = $"{Channel}/Binaries/Latest/";
+
+                foreach (string blob in _latestBlobs)
```
piotrpMSFT: {}

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+                    AzurePublisherTool.CopyBlob(blob.Replace("/dotnet/", ""), (targetContainer + Path.GetFileName(blob)).Replace(CliNuGetVersion, "latest").Replace(SharedFrameworkNugetVersion, "latest"));
```
brthor: make this more readable?

piotrpMSFT: this is a bit hard to read. perhaps:

````
var blobUrl = blob.Replace(...)
                            .Replace(...)
                            .Replace(...)...
````

File: scripts/dotnet-cli-build/PublishTargets.cs [hunk 3]
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+
+                PublishLatestCliVersionTextFile(c);
+                PublishLatestSharedFrameworkVersionTextFile(c);
+            }
+
             return c.Success();
         }
 
+        private static bool ShouldPublishToLatest()
```
piotrpMSFT: @brthor is there a conditionaltargetattribute we could use with this?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+        {
+             Dictionary<string, bool> runtimes = new Dictionary<string, bool>()
```
brthor: It's not too clear that this code is checking that we have runtime packages for each of these rids, maybe make the function name more descriptive?

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+             {
+                 {"win7-x64", false },
+                 {"win7-x86", false },
+                 {"osx.10.10", false },
+                 {"rhel.7", false },
+                 {"ubuntu.14.04", false },
```
piotrpMSFT: CentOS and Debian are missing here.

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
```diff
@@ -36,23 +37,13 @@ private CloudBlobContainer GetDotnetBlobContainer(string connectionString)
...
             var installerFileBlob = CalculateInstallerBlob(installerFile, channel, version);
-
-            var latestInstallerFileName = installerFile.Replace(version, "latest");
-            var latestInstallerFileBlob = CalculateInstallerBlob(latestInstallerFileName, channel, "Latest");
-
             PublishFile(installerFileBlob, installerFile);
-            PublishFile(latestInstallerFileBlob, installerFile);
         }
 
         public void PublishArchiveAndLatest(string archiveFile, string channel, string version)
```
brthor: Change to PublishArchive?

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
```diff
@@ -36,23 +37,13 @@ private CloudBlobContainer GetDotnetBlobContainer(string connectionString)
         public void PublishInstallerFileAndLatest(string installerFile, string channel, string version)
```
brthor: Change to PublishInstaller?

scripts/dotnet-cli-build/PublishTargets.cs
sokket: How's ```CheckIfAllRidsHavePublished```?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
+            if (ShouldPublishToLatest())
```
brthor: If this fails, what action should we take? Is the idea that for the last build running this will be true?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+            _latestBlobs = new List<string>(AzurePublisherTool.ListBlobs($"{Channel}/Binaries/{CliNuGetVersion}/"));
+
+            foreach (string file in _latestBlobs)
+            {
+                string name = Path.GetFileName(file);
+                string key = string.Empty;
+
+                foreach (string runtime in runtimes.Keys)
+                {
+                    if (name.StartsWith($"runtime.{runtime}"))
```
piotrpMSFT: This says that `any` artifact for that runtime will do, right? Can we be more specific? I believe that the version badges are the last artifacts we produce. Can we ensure these are version badge files?

brthor:
Have some stylistic/readability comments but overall looks good, nice!

scripts/dotnet-cli-build/PublishTargets.cs
brthor: Unfortunately you can't pass delegates in the constructor of the attribute. 

So you'd have to create a custom attribute subclassing TargetConditionAttribute and basically calling this function for `EvaluateCondition`

scripts/dotnet-cli-build/PublishTargets.cs
brthor: I like it

piotrpMSFT: We need a method for folks to know to add to this dictionary. Can you add a target that will break the build IIF this is an official build AND the RID of the current build is not in this list?

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
piotrpMSFT: Last build running will be true.

scripts/dotnet-cli-build/PublishTargets.cs
eerhardt: is debian needed here?

sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+                    {
+                        key = runtime;
+                        break;
+                    }
+                }
+
+                runtimes[key] = true;
```
eerhardt: seems like we should check for !string.IsNullOrEmpty(key)

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+            }
+
+            foreach (string key in runtimes.Keys)
```
eerhardt: can be simplified:

return runtimes.Keys.All(key => runtimes[key]);

eerhardt:
:-1:

scripts/dotnet-cli-build/PublishTargets.cs
piotrpMSFT: per offline conversation, let's add a "FinalizeBuildPublishing" target that in turn calls the "PublishVersionBadge" Target. This will show in code that PublishVersion is a Finalization step.

piotrpMSFT: Once that is done, the check here should validate that the version badge exists.

Sridhar-MS: yup. Version badge is the right artifact to use here

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
```
Sridhar-MS: Lets not piggy back on the `Publish` target. This should be its own target `PublishLatestArtifacts`.

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -45,15 +49,63 @@ public static BuildTargetResult InitPublish(BuildTargetContext c)
...
+            if (ShouldPublishToLatest())
+            {
+                string targetContainer = $"{Channel}/Binaries/Latest/";
+
+                foreach (string blob in _latestBlobs)
+                    AzurePublisherTool.CopyBlob(blob.Replace("/dotnet/", ""), (targetContainer + Path.GetFileName(blob)).Replace(CliNuGetVersion, "latest").Replace(SharedFrameworkNugetVersion, "latest"));
+
+                PublishLatestCliVersionTextFile(c);
+                PublishLatestSharedFrameworkVersionTextFile(c);
```
Sridhar-MS: Does this mean we are going to publish the latest version text file only for the current RID?

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
brthor: If the last two builds run the publish step at the same time, seems like this will break down.

scripts/dotnet-cli-build/PublishTargets.cs
brthor: This is a good point, shouldn't we need it for all the supported rids?

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed to look for the badge

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Good catch, fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Nice! Thanks, fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: The ```CheckIfAllBuildsHavePublished``` function will do this check since that's where the dictionary lives; it makes sense to have the exception thrown from as-close-as-possible to where the error is.

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Added taking a lease on the blob container to prevent this

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -48,12 +52,111 @@ public static BuildTargetResult Publish(BuildTargetContext c)
             return c.Success();
         }
 
+        [Target]
+        public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
```
brthor: Might be missing it, but how does this get executed, what target has a dependency on it?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -48,12 +52,111 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+                    {
+                        string targetUrl = $"{targetContainer}{Path.GetFileName(blob)}"
+                                        .Replace(CliNuGetVersion, "latest")
+                                        .Replace(SharedFrameworkNugetVersion, "latest");
+                        AzurePublisherTool.CopyBlob(blob.Replace("/dotnet/", ""), targetUrl);
+                    }
+
+                    // Copy the DNVM latest version files and delete the temp files
+                    IEnumerable<string> dnvmBlobs = AzurePublisherTool.ListBlobs($"{Channel}/dnvm/temp/");
+                    foreach (string blob in dnvmBlobs)
```
Sridhar-MS: Instead we can just publish the latest version text files for all OSes from the current OS. Makes it much simpler.

piotrpMSFT:
@sokket did you send approval mail?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -48,12 +52,111 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+                 { "RHEL_x64", false },
+                 { "OSX_x64", false },
+                 { "Debian_x64", false },
+                 { "CentOS_x64", false }
+             };
+
+            _latestBlobs = new List<string>(AzurePublisherTool.ListBlobs($"{Channel}/Binaries/{CliNuGetVersion}/"));
+
+            var config = Environment.GetEnvironmentVariable("CONFIGURATION");
+            var versionBadgeName = $"{CurrentPlatform.Current}_{CurrentArchitecture.Current}";
```
piotrpMSFT: Are we sure we don't want the Current.CurrentPlatform.Current.Current? haha, we need to fix that up at some point.

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -48,12 +52,111 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+            if (badges.ContainsKey(versionBadgeName) == false)
+                throw new ArgumentException("A new OS build was added without adding the moniker to the {nameof(badges)} lookup");
```
piotrpMSFT: {

}

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -48,12 +52,111 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+                 { "OSX_x64", false },
+                 { "Debian_x64", false },
+                 { "CentOS_x64", false }
+             };
+
+            _latestBlobs = new List<string>(AzurePublisherTool.ListBlobs($"{Channel}/Binaries/{CliNuGetVersion}/"));
+
+            var config = Environment.GetEnvironmentVariable("CONFIGURATION");
+            var versionBadgeName = $"{CurrentPlatform.Current}_{CurrentArchitecture.Current}";
+            if (badges.ContainsKey(versionBadgeName) == false)
```
piotrpMSFT: {

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -48,12 +52,111 @@ public static BuildTargetResult Publish(BuildTargetContext c)
...
+                string targetContainer = $"{Channel}/Binaries/Latest/";
+                string targetVersionFile = $"{targetContainer}{CliNuGetVersion}";
+                string leaseId = AzurePublisherTool.AcquireLeaseOnContainer();
+
+                // Prevent race conditions by dropping a version hint of what version this is. If we see this file
+                // and it is the same as our version then we know that a race happened where two+ builds finished 
+                // at the same time and someone already took care of publishing and we have no work to do.
+                if (AzurePublisherTool.IsLatestSpecifiedVersion(targetVersionFile))
+                {
+                    AzurePublisherTool.ReleaseLeaseOnContainer(leaseId);
```
piotrpMSFT: I guess the scenario here is that:
build 1> grabbed lease
build 1> wrote version file
azure> expired lease
build 2> got lease

?

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Close, the flow is as such (changed slightly between version of my PR):

1) try acquire a lease on the ```latest``` folder
2) Write version and do work
3) Release lease

If there is contention, the second build will wait and then see the version file once the lease expires and exit quickly. The only coordination object is the latest folder itself

scripts/dotnet-cli-build/PublishTargets.cs
sokket: 👍

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -645,11 +640,6 @@ private static BuildTargetResult CompileCliSdk(BuildTargetContext c, DotNetCli d
 
             CrossgenUtil.CrossgenDirectory(c, outputDir);
 
-            // Generate .version file
-            var version = buildVersion.NuGetVersion;
-            var content = $@"{c.BuildContext["CommitHash"]}{Environment.NewLine}{version}{Environment.NewLine}";
-            File.WriteAllText(Path.Combine(outputDir, ".version"), content);
```
eerhardt: Does this mean that the SDK directory won't contain a .version file any more?  We have code that depends on that - specifically: https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/DotnetFiles.cs#L16

sokket: No, the /dnvm directory will still contain version files; they are just populated all at once at the end now

sokket:
@dotnet-bot test this please

scripts/dotnet-cli-build/PublishTargets.cs
brthor: @Sridhar-MS and I thought about this before and came away blank 😄 agreed the name is weird

RuntimePlatform.Current maybe

scripts/dotnet-cli-build/CompileTargets.cs
eerhardt: I mean the .version file being laid down on disk by the installer/tarball/.zip.  Not a version file that is being uploaded to azure.

sokket: Fixed

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -24,6 +26,8 @@ public static class PublishTargets
 
         private static string SharedFrameworkNugetVersion { get; set; }
 
+        private static List<string> _latestBlobs;
```
eerhardt: I think the general approach is to put properties into the BuildTargetContext instead of having static variables.

But looking deeper - it looks like you can just return the List<string> as an out parameter of `CheckIfAllBuildsHavePublished` and remove this static variable.

sokket:
Any further comments / feedback? :)

scripts/dotnet-cli-build/PublishTargets.cs
sokket: Peter and I talked about that offline and we ended up with the static variable being a little more clear than an out var. I'm open to re-opening that discussion if you want :)

eerhardt: No, no reason to re-open.

Looking again, it should be "s_" prefixed to follow the coding style here https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

eerhardt:
:shipit:

Sridhar-MS:
:shipit: Lets keep an eye on the builds after merging.

scripts/dotnet-cli-build/PublishTargets.cs
brthor: Happy that we are moving away from `BuildTargetContext` 👍 It's Hard to keep track of variables as strings

weshaggard:
Is there any reason you aren't also publishing the nuget packages for the host and meta-package as part of this finalize step? The only interesting aspect I can see is that you are essentially doing this Azure publish to Latest form whichever leg finishes last and if that leg is non-windows I don't think we have any great tools to publish the packages. 

My concern here though is that we will still be in a broken state between the time the Latest in blob storage gets updated and when the meta-package gets published.

eerhardt:
NuGet xplat does support Push:

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PushCommand.cs

We could use that from our build scripts by referencing that package.

PR #2705, opened 2016-04-27
Convert UTF8 using MultiByteToWideChar

schellap:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2705)

### DIFF
--- a/src/corehost/common/pal.windows.cpp
+++ b/src/corehost/common/pal.windows.cpp
@@ -10,8 +10,6 @@
 #include <codecvt>
 #include <ShlObj.h>
 
-static thread_local std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> t_converter;
-
 pal::string_t pal::to_lower(const pal::string_t& in)
 {
     pal::string_t ret = in;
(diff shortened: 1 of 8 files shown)


schellap:
@gkhanna79 PTAL. Removes the thread_local usage, I tried to use this to keep code changes minimal in my earlier PR (but this has been bothering me.)

I believe the `thread_local` causes to take critical section which internally depends on some other global class variables to be initialized I believe. This is all unnecessary code either way, as MultiByteToWideChar would do the job for us, so I did not delve too much into this.

schellap:
Fixes #2708

schellap:
@gkhanna79 a look please...

src/corehost/common/pal.windows.cpp
```diff
@@ -253,17 +251,26 @@ bool pal::get_own_executable_path(string_t* recv)
...
 }
 
-pal::string_t pal::to_palstring(const std::string& str)
+pal::string_t pal::utf8_palstring(const std::string& str)
 {
-    return t_converter.from_bytes(str);
+    // No need of explicit null termination, so pass in the actual length.
+    pal::string_t retval;
+    size_t size = ::MultiByteToWideChar(CP_UTF8, 0, str.c_str(), str.size(), nullptr, 0);
+    retval.resize(size, '\0');
```
gkhanna79: What would resize do if size is zero?

gkhanna79: Given that MBTOWC returns zero for failure, I dont think the resize should be done based upon the size. If your goal is to initialize the variable, then please do so explicitly, independent of the call.

Also, if the call failed, we should propagate that failure up instead of returning an empty string!

schellap: It is equivalent of `clear()` setting it back to null terminated empty string `""`

src/corehost/common/pal.windows.cpp
```diff
@@ -253,17 +251,26 @@ bool pal::get_own_executable_path(string_t* recv)
...
+    if (size == 0)
+    {
+        return retval;
+    }
+    ::MultiByteToWideChar(CP_UTF8, 0, str.c_str(), str.size(), &retval[0], retval.size());
```
gkhanna79: We should do error check here (and at other places we use the API).

schellap: I am mostly following what the CLR does in such cases, I took your feedback because it is the right thing to do. However, I am leaving the ignoring the error code to the callers of this library like util function.

https://github.com/dotnet/coreclr/blob/775003a4c72f0acc37eab84628fcef541533ba4e/src/dlls/mscoree/unixinterface.cpp#L59

schellap:

Issue #2664, opened 2016-04-23
Terrible error message: "Invalid .deps file"


BruceForstall:
## Steps to reproduce

Hard to know what details to give. Probably, on Windows 8.1:

1. git clone https://github.com/russellhadley/mcgdiff
2. cd mcgdiff
3. build.cmd -p -f
4. bin\mcgdiff\mcgdiff.exe
5. get: "Invalid .deps file"

unless Russell fixes the project.json, or something else, and makes this repo not reproduce the problem.

This is an app built/published with "dotnet publish". I presume there's some dependency problem, but I don't know how to debug it, especially from this error message.

## Expected  behavior

no error

## Actual behavior

"Invalid .deps file"

General comment:

**This is an absolutely terrible and useless error message.** Which ".deps" file is it referring to? Why is it invalid? What can I do to make it valid?

I notice there are several other cases in the dotnet/cli issue list referring to this error, and in all cases people were complaining about how useless it was and how difficult it was to determine what the actual problem was.

## Environment data
`dotnet --info` output:

.NET Command Line Tools (1.0.0-rc2-002476)

Product Information:
 Version:     1.0.0-rc2-002476
 Commit Sha:  1a8d393b9b

Runtime Environment:
 OS Name:     Windows
 OS Version:  6.3.9600
 OS Platform: Windows
 RID:         win81-x64

piotrpMSFT:
@schellap can you take a look at this? Can we make the message a bit more verbose? Perhaps: `Failed to launch <path to app>: <path to deps> is invalid` ?

schellap:
@BruceForstall hi!

I tried the steps but I couldn't repro. I got this:

```
g:\mcgdiff>bin\mcgdiff\mcgdiff.exe
error: Specify --base and/or --diff.
```

1. Could you please zip up your repo (as is on your machine) and the dotnet root folder (with the FX) and share these zips?
2. Rename dotnet.exe relative `shared\Microsoft.NETCore.App\1.0.0-rc2-23931` folder to `foo`
3. Next, can you target "Microsoft.NETCore.App": "1.0.0-rc2-002476" in your project.json and see if this goes away?

@piotrpMSFT The places where we mark deps as invalid we emit an error message. I'll add the paths to the error message. This looks like some corruption (`1.0.0-rc2-23931` seems suspect which has the non-compatible interface in `hostpolicy.dll` for the newer `hostfxr.dll`)

russellhadley:
[...]

PR #2709, opened 2016-04-27
Improve error messages

schellap:
@gkhanna79 PTAL for error message string changes. Also, silently ignore rid fallback errors @anurse

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2709)

### DIFF
--- a/src/corehost/cli/args.cpp
+++ b/src/corehost/cli/args.cpp
@@ -28,7 +28,7 @@ bool parse_arguments(
     // Get the full name of the application
     if (!pal::get_own_executable_path(&args.own_path) || !pal::realpath(&args.own_path))
     {
-        trace::error(_X("Failed to locate current executable"));
+        trace::error(_X("Failed to resolve full path of the current executable [%s]"), args.own_path.c_str());
         return false;
     }
 
@@ -45,7 +45,7 @@ bool parse_arguments(
         args.managed_application = pal::string_t(argv[1]);
         if (!pal::realpath(&args.managed_application))
         {
-            trace::error(_X("Failed to locate managed application: %s"), args.managed_application.c_str());
+            trace::error(_X("Failed to locate managed application [%s]"), args.managed_application.c_str());
             return false;
         }
         args.app_dir = get_directory(args.managed_application);
@@ -62,7 +62,7 @@ bool parse_arguments(
         args.managed_application = managed_app;
         if (!pal::realpath(&args.managed_application))
         {
-            trace::error(_X("Failed to locate managed application: %s"), args.managed_application.c_str());
+            trace::error(_X("Failed to locate managed application [%s]"), args.managed_application.c_str());
             return false;
         }
         args.app_dir = own_dir;
--- a/src/corehost/cli/deps_format.cpp
+++ b/src/corehost/cli/deps_format.cpp
@@ -93,7 +93,7 @@ void deps_json_t::reconcile_libraries_with_targets(
                     entry.asset_name == LIBCORECLR_FILENAME)
                 {
                     m_coreclr_index = m_deps_entries[i].size() - 1;
-                    trace::verbose(_X("Found coreclr from deps %d [%s, %s, %s]"),
+                    trace::verbose(_X("Found CoreCLR from deps %d [%s, %s, %s]"),
                         m_coreclr_index,
                         entry.library_name.c_str(),
                         entry.library_version.c_str(),
@@ -134,16 +134,18 @@ bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, c
         {
             if (rid_fallback_graph.count(host_rid) == 0)
             {
-                trace::error(_X("Did not find fallback rids for package %s for the host rid %s"), package.first.c_str(), host_rid.c_str());
-                return false;
+                trace::warning(_X("The targeted framework does not support the runtime '%s'. Some native libraries from [%s] may fail to load on this platform."), host_rid.c_str(), package.first.c_str());
             }
-            const auto& fallback_rids = rid_fallback_graph.find(host_rid)->second;
-            auto iter = std::find_if(fallback_rids.begin(), fallback_rids.end(), [&package](const pal::string_t& rid) {
-                return package.second.rid_assets.count(rid);
-            });
-            if (iter != fallback_rids.end())
+            else
             {
-                matched_rid = *iter;
+                const auto& fallback_rids = rid_fallback_graph.find(host_rid)->second;
+                auto iter = std::find_if(fallback_rids.begin(), fallback_rids.end(), [&package](const pal::string_t& rid) {
+                    return package.second.rid_assets.count(rid);
+                });
+                if (iter != fallback_rids.end())
+                {
+                    matched_rid = *iter;
+                }
             }
         }
 
@@ -350,7 +352,7 @@ bool deps_json_t::load(bool portable, const pal::string_t& deps_path, const rid_
     // If file doesn't exist, then assume parsed.
     if (!pal::file_exists(deps_path))
     {
-        trace::verbose(_X("Deps file does not exist [%s]"), deps_path.c_str());
+        trace::verbose(_X("Could not locate the dependencies manifest file [%s]. Some libraries may fail to resolve."), deps_path.c_str());
         return true;
     }
 
@@ -358,7 +360,7 @@ bool deps_json_t::load(bool portable, const pal::string_t& deps_path, const rid_
     pal::ifstream_t file(deps_path);
     if (!file.good())
     {
-        trace::error(_X("Could not open file stream on deps file [%s]"), deps_path.c_str());
+        trace::error(_X("Could not open dependencies manifest file [%s]"), deps_path.c_str());
         return false;
     }
 
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -285,7 +285,11 @@ int fx_muxer_t::parse_args_and_execute(
     std::unordered_map<pal::string_t, std::vector<pal::string_t>> opts;
     if (!parse_known_args(argc - argoff, &argv[argoff], known_opts, &opts, &num_parsed))
     {
-        trace::error(_X("Failed to parse supported arguments."));
+        trace::error(_X("Failed to parse supported options or their values:"));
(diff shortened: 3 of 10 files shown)


schellap:
Fixes #2664

src/corehost/cli/deps_format.cpp
```diff
@@ -134,16 +134,18 @@ bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, c
...
             }
-            const auto& fallback_rids = rid_fallback_graph.find(host_rid)->second;
-            auto iter = std::find_if(fallback_rids.begin(), fallback_rids.end(), [&package](const pal::string_t& rid) {
-                return package.second.rid_assets.count(rid);
-            });
-            if (iter != fallback_rids.end())
+            else
             {
-                matched_rid = *iter;
+                const auto& fallback_rids = rid_fallback_graph.find(host_rid)->second;
```
gkhanna79: Why do we need this to be refactored?

src/corehost/cli/args.cpp
```diff
@@ -28,7 +28,7 @@ bool parse_arguments(
     // Get the full name of the application
     if (!pal::get_own_executable_path(&args.own_path) || !pal::realpath(&args.own_path))
     {
-        trace::error(_X("Failed to locate current executable"));
+        trace::error(_X("Failed to obtain full path to the executable using [%s]"), args.own_path.c_str());
```
anurse: "using" doesn't really indicate why that path was used. How about `Failed to resolve the full path to the executable [%s]`

src/corehost/cli/deps_format.cpp
```diff
@@ -134,16 +134,18 @@ bool deps_json_t::perform_rid_fallback(rid_specific_assets_t* portable_assets, c
         {
             if (rid_fallback_graph.count(host_rid) == 0)
             {
-                trace::error(_X("Did not find fallback rids for package %s for the host rid %s"), package.first.c_str(), host_rid.c_str());
-                return false;
+                trace::warning(_X("Did not find fallback rids for package %s for the host rid %s"), package.first.c_str(), host_rid.c_str());
```
anurse: "rid" is a pretty internal term. How about `The framework 'Microsoft.NETCore.App 1.0.0-abc123' does not support the runtime 'win7-x64'. Some native libraries may fail to load on this platform.`

(where `Microsoft.NETCore.App 1.0.0-abc123` is the shared framework name and version, and `win7-x64` is the host RID)

src/corehost/cli/deps_format.cpp
```diff
@@ -350,15 +352,15 @@ bool deps_json_t::load(bool portable, const pal::string_t& deps_path, const rid_
     // If file doesn't exist, then assume parsed.
     if (!pal::file_exists(deps_path))
     {
-        trace::verbose(_X("Deps file does not exist [%s]"), deps_path.c_str());
+        trace::verbose(_X("Deps.json file [%s] does not exist"), deps_path.c_str());
```
anurse: Let's use full names to be a little more user-friendly. How about `Could not locate dependencies list '[deps file]'. Some libraries may fail to resolve.`

In general, I think it's good to inform users what kinds of failures may occur when these messages are printed (even if they have to turn on tracing to see them). That way we get clearer bug reports and questions :)

src/corehost/cli/deps_format.cpp
```diff
@@ -350,15 +352,15 @@ bool deps_json_t::load(bool portable, const pal::string_t& deps_path, const rid_
...
         return true;
     }
 
     // Somehow the file stream could not be opened. This is an error.
     pal::ifstream_t file(deps_path);
     if (!file.good())
     {
-        trace::error(_X("Could not open file stream on deps file [%s]"), deps_path.c_str());
+        trace::error(_X("Could not open file stream on deps.json file [%s]"), deps_path.c_str());
```
anurse: Since this is an unknown I/O failure, lets say `Unknown error loading [deps_path].` (If there's any way to get an error code or something from the OS, it'd be good to print that out too.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -447,7 +451,11 @@ int fx_muxer_t::read_config_and_execute(
             }
             if (!found)
             {
-                trace::error(_X("Policy library either not found in deps [%s] or not found in %d probe paths."), deps_file.c_str(), probe_realpaths.size());
+                trace::error(_X("Expected hostpolicy lib either not found in deps.json [%s] or not found in %d probe paths: "), deps_file.c_str(), probe_realpaths.size());
```
anurse: The "hostpolicy" library is an internal thing, let's be clearer about exactly what we're looking for and where. How about `Could not find required library 'libhostpolicy.so' in the dependencies list [deps_file], or the probing path [probing_path].` (replacing `libhostpolicy.so` with whatever the platform-specific library name is of course)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -465,7 +473,7 @@ int fx_muxer_t::execute(const int argc, const pal::char_t* argv[])
     // Get the full name of the application
     if (!pal::get_own_executable_path(&own_path) || !pal::realpath(&own_path))
     {
-        trace::error(_X("Failed to locate current executable"));
+        trace::error(_X("Failed to obtain full path to the executable using [%s]"), own_path.c_str());
```
anurse: As above: `Failed to resolve the full path to the current executable [%s]`

anurse: As above: `Failed to resolve the full path to the current executable [%s]`

GuardRex:
Yes, please make exceptions somewhat verbose and explicit and **helpful**. This has been a running problem for Microsoft since ..... uh .... well ...... **1975!!!!**

src/corehost/cli/hostpolicy.cpp
```diff
@@ -20,16 +20,17 @@ int run(const arguments_t& args)
...
-        trace::error(_X("Invalid .deps file"));
+        trace::error(_X("Error initializing the dependency resolver: %s"), resolver_errors.c_str());
         return StatusCode::ResolverInitFailure;
     }
 
     pal::string_t clr_path = resolver.resolve_coreclr_dir();
     if (clr_path.empty() || !pal::realpath(&clr_path))
     {
-        trace::error(_X("Could not resolve coreclr path"));
+        trace::error(_X("Could not resolve coreclr path, turn on logging (set or export COREHOST_TRACE=1) to look for 'CoreCLR' probed paths."));
```
anurse: `Unable to resolve CoreCLR path. For more details, enable tracing by setting the COREHOST_TRACE environment variable to '1'` (users will know how to set environment variables on their platform).

If all calls to `trace::error` result in the termination of the process, perhaps it should just call `exit` itself, and maybe have it emit the line about `COREHOST_TRACE`? For example

```
trace::error(_X("Blah blah something went wrong"), StatusCode::CoreClrResolveFailure)
```

would emit the below message and call `exit(StatusCode::CoreClrResolveFailure)` (if the `exit` refactoring would be complicated, we don't need to do it, just throwing out ideas ;)). I do think that in almost all error cases, the user is going to want to know how to get more details, so we may as well tell them about `COREHOST_TRACE` right there.

```
Blah blah something went wrong
An error occurred initializing the application. For more details, enable tracing by setting the COREHOST_TRACE environment variable to '1'`
```

src/corehost/cli/hostpolicy.cpp
```diff
@@ -112,7 +113,7 @@ int run(const arguments_t& args)
     // Bind CoreCLR
     if (!coreclr::bind(clr_path))
     {
-        trace::error(_X("Failed to bind to coreclr"));
+        trace::error(_X("Failed to bind to coreclr at [%s]"), clr_path.c_str());
```
anurse: We use `CoreCLR` and `coreclr` a little inconsistently in messages. It looks like `CoreCLR` is more common so let's align to that.

src/corehost/common/pal.unix.cpp
```diff
@@ -39,7 +39,7 @@ bool pal::touch_file(const pal::string_t& path)
     int fd = open(path.c_str(), (O_CREAT | O_EXCL), (S_IRUSR | S_IRGRP | S_IROTH));
     if (fd == -1)
     {
-        trace::warning(_X("Failed to open() file descriptor in %s(%s)"), __FUNCTION__, path.c_str());
+        trace::warning(_X("Failed to open() file descriptor in %s(%s)"), _STRINGIFY(__FUNCTION__), path.c_str());
```
anurse: Should the details should be verbose and have a less scary looking warning? Perhaps more like this:

```
trace::warning(_X("Failed to open file: %s"), path.c_str());
trace::verbose(_X("open(%s) failed in %s"), path.c_str(), _STRINGIFY(__FUNCTION__));
```

src/corehost/corehost.cpp
```diff
@@ -63,7 +63,7 @@ int run(const int argc, const pal::char_t* argv[])
     pal::string_t own_path;
     if (!pal::get_own_executable_path(&own_path) || !pal::realpath(&own_path))
     {
-        trace::error(_X("Failed to locate current executable"));
+        trace::error(_X("Failed to obtain full path to the executable using [%s]"), own_path.c_str());
```
anurse: And again :) `Failed to resolve the full path to the current executable [%s]`

anurse:
Really glad to see some nice new error messages! I had some comments to clarify some of them. We can iterate a little on the wording if necessary but in general: :shipit:

src/corehost/cli/deps_format.cpp
schellap: @gkhanna79, this error can occur only if rid graph is totally absent or someone patched hostpolicy built for a higher rid into a lower FX.

This means we cannot find the host rid in the netcore.app fallback graph in the FX deps json. Since it is not an error to skip a package (for loop not matching), it is also not an error to start the fallback for the package.

GuardRex:
@schellap @anurse @gkhanna79 Is this one ...
```
The given key was not present in the dictionary.
```
... part of the updates? It was just mentioned on JabbR due to a bad package name by @Aequoris.

schellap:
@GuardRex that sounds like a managed exception from dictionary lookup. Please file a separate issue for it. This PR fixes native side of things.

PR #2711, opened 2016-04-27
Use only project in incremental checks

pakrym:
@cdmihai PR is work-in-progress but would like you to take a look.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2711)

### DIFF
--- a/src/dotnet/commands/dotnet-build/CompilerIOManager.cs
+++ b/src/dotnet/commands/dotnet-build/CompilerIOManager.cs
@@ -10,6 +10,7 @@ using Microsoft.DotNet.Tools.Compiler;
 using Microsoft.DotNet.ProjectModel;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.Cli.Compiler.Common;
+using Microsoft.DotNet.ProjectModel.Graph;
 using NuGet.Frameworks;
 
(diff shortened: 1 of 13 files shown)


src/dotnet/commands/dotnet-build/CompilerIOManager.cs
```diff
@@ -92,27 +93,28 @@ private CompilerIO ComputeIO(ProjectGraphNode graphNode)
...
+            //if (project.LockFile == null)
+            //{
+            //    var errorMessage = $"Project {project.ProjectName()} does not have a lock file.";
+            //    Reporter.Error.WriteLine(errorMessage);
+            //    throw new InvalidOperationException(errorMessage);
+            //}
+
+            //inputs.Add(project.LockFile.LockFilePath);
+
+            //if (project.LockFile.ExportFile != null)
```
pakrym: I don't know how to fix `ExportFile` how does it get generated? Can we try and hardcode a name somehow?

cdmihai: the name of the file is ExportFile.ExportFileName. By convention, it stays near the lock file. When it exists.

---
In reply to: [61284193](https://github.com/dotnet/cli/pull/2711#discussion_r61284193) [](ancestors = 61284193)

PR #2721, opened 2016-04-27
Fix --no-dependencies incremental check regression

pakrym:
@anurse  @cdmihai

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2721)


src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -12,6 +12,13 @@ namespace Microsoft.DotNet.Tools.Build
 {
     internal abstract class ProjectBuilder
     {
+        private readonly bool _skipDependencies;
+
+        public ProjectBuilder(bool skipDependencies)
```
eerhardt: nit: protected constructors on abstract classes.

anurse:
:shipit:

PR #2724, opened 2016-04-28
Improving Telemetry performance

adamgorMSFT:
Improved telemetry performance to minimize impact on CLI.

Addressing issue #2595
Overall run time now minimally effected by telemetry.

Moved TrackEvent up earlier and run async with the rest.

The Async task will be dropped if not completed by end of normal run time. 
Dropped exitcode from telemetry.
Added flag for Test Machines(e.g CI and others)

Edit:
Dropped collection of all arguments.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2724)

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -132,6 +132,8 @@ namespace Microsoft.DotNet.Cli
                 command = "help";
             }
 
+            telemetryClient.TrackEvent(command, null, null);
+
             int exitCode;
             Func<string[], int> builtIn;
             if (s_builtIns.TryGetValue(command, out builtIn))
@@ -147,13 +149,6 @@ namespace Microsoft.DotNet.Cli
                 exitCode = result.ExitCode;
             }
 
-            telemetryClient.TrackEvent(
-                command,
-                null,
-                new Dictionary<string, double>
-                {
-                    ["ExitCode"] = exitCode
-                });
 
             return exitCode;
 
--- a/src/dotnet/Telemetry.cs
+++ b/src/dotnet/Telemetry.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
--- a/src/dotnet/Telemetry.cs
+++ b/src/dotnet/Telemetry.cs
@@ -58,41 +58,6 @@ namespace Microsoft.DotNet.Cli
                 return;
             }
 
-            _isCollectingTelemetry = true;
-            try
-            {
-                //initialize in task to offload to parallel thread
-                _trackEventTask = Task.Factory.StartNew(() => InitializeTelemetry());
-            }
-            catch(Exception)
-            {
-                Debug.Fail("Exception during telemetry task initialization");
-            }
-        }
-
-        public void TrackEvent(string eventName, IList<string> properties, IDictionary<string, double> measurements)
-        {
-            if (!_isCollectingTelemetry)
-            {
-                return;
-            }
-
-            try
-            {
-                _trackEventTask = _trackEventTask.ContinueWith(
-                    x => TrackEventTask(eventName,
-                        properties,
-                        measurements)
-                );
-            }
-            catch(Exception)
-            {
-                Debug.Fail("Exception during telemetry task continuation");
-            }
-        }
-
-        private void InitializeTelemetry()
-        {
             try
             {
                 using (PerfTrace.Current.CaptureTiming())
@@ -101,11 +66,9 @@ namespace Microsoft.DotNet.Cli
                     _trackEventTask = Task.Factory.StartNew(() => InitializeTelemetry());
                 }
             }
-            catch (Exception)
+            catch(Exception)
             {
-                // we dont want to fail the tool if telemetry fais. We should be able to detect abnormalities from data
-                // at the server end
-                Debug.Fail("Exception during telemetry initialization");
+                Debug.Fail("Exception during telemetry task initialization");
             }
         }
 
--- a/src/dotnet/Telemetry.cs
+++ b/src/dotnet/Telemetry.cs
@@ -52,7 +52,7 @@ namespace Microsoft.DotNet.Cli
                 _sampleRate = ReciprocalSampleRateValueForTest;
             }
 
-            _isCollectingTelemetry = (Environment.TickCount % _sampleRate == 0);    
+            _isCollectingTelemetry = (Environment.TickCount % _sampleRate == 0);
             if(!_isCollectingTelemetry)
             {
                 return;
@@ -69,19 +69,19 @@ namespace Microsoft.DotNet.Cli
                 Debug.Fail("Exception during telemetry task initialization");
             }
         }
-        
+
         public void TrackEvent(string eventName, IList<string> properties, IDictionary<string, double> measurements)
         {
             if (!_isCollectingTelemetry)
             {
                 return;
             }
-            
+
             try
             {
                 _trackEventTask = _trackEventTask.ContinueWith(
-                    x => TrackEventTask(eventName, 
-                        properties, 
+                    x => TrackEventTask(eventName,
+                        properties,
                         measurements)
                 );
             }
@@ -90,7 +90,7 @@ namespace Microsoft.DotNet.Cli
                 Debug.Fail("Exception during telemetry task continuation");
             }
         }
-        
+
         private void InitializeTelemetry()
         {
             try
@@ -108,7 +108,7 @@ namespace Microsoft.DotNet.Cli
                 Debug.Fail("Exception during telemetry initialization");
             }
         }
-        
+
         public void TrackEvent(string eventName, IDictionary<string, string> properties, IDictionary<string, double> measurements)
         {
             if (!Enabled)
@@ -118,7 +118,7 @@ namespace Microsoft.DotNet.Cli
             if (!_isCollectingTelemetry)
             {
                 return;
-            }                
+            }
             try
             {
                 using (PerfTrace.Current.CaptureTiming())
@@ -223,9 +223,9 @@ namespace Microsoft.DotNet.Cli
                 }
                 return eventProperties;
             }
-            else 
+            else
             {
-                return _commonProperties;    
+                return _commonProperties;
             }
         }
     }
--- a/src/dotnet/Telemetry.cs
+++ b/src/dotnet/Telemetry.cs
@@ -14,27 +14,22 @@ namespace Microsoft.DotNet.Cli
     public class Telemetry : ITelemetry
     {
         private bool _isInitialized = false;
-        private bool _isCollectingTelemetry = false;
         private TelemetryClient _client = null;
 
         private Dictionary<string, string> _commonProperties = null;
         private Dictionary<string, double> _commonMeasurements = null;
         private Task _trackEventTask = null;
 
-        private int _sampleRate = 1;
-        private bool _isTestMachine = false;
+        private string _telemetryProfile;
 
-        private const int ReciprocalSampleRateValue = 1;
-        private const int ReciprocalSampleRateValueForTest = 1;
         private const string InstrumentationKey = "74cc1c9e-3e6e-4d05-b3fc-dde9101d0254";
         private const string TelemetryOptout = "DOTNET_CLI_TELEMETRY_OPTOUT";
-        private const string TestMachineFlag = "TEST_MACHINE";
-        private const string TestMachine = "Test Machine";
+        private const string TelemetryProfileEnvironmentVariable = "DOTNET_CLI_TELEMETRY_PROFILE";
         private const string OSVersion = "OS Version";
         private const string OSPlatform = "OS Platform";
         private const string RuntimeId = "Runtime Id";
         private const string ProductVersion = "Product Version";
-        private const string ReciprocalSampleRate = "Reciprocal SampleRate";
+        private const string TelemetryProfile = "Telemetry Profile";
 
         public bool Enabled { get; }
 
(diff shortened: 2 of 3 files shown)


PinpointTownes:
Logging the command name and the exit code for statistical purposes is one thing but logging the entire arguments list is a totally different story... you're clearly opening Pandora's Box here.

Don't forget that one may want to flow secret parameters when using `dotnet run`... logging them - even for correlation - is absolutely evil. It'sspecially true with ASP.NET Core apps, where command line args might be used to configure an OAuth2 `client_secret` or an app key.

Of course, hashing the arguments doesn't offer any real protection here, specially since you're logging the arguments individually and without using any salt (rainbow tables are so fun...).

Please consider reverting this change.

test/dotnet.Tests/TelemetryCommandTest.cs
```diff
@@ -19,6 +19,7 @@ public void TrackEvent(string eventName, IDictionary<string, string> properties,
         {
             EventName = eventName;
         }
+		public void Finish();
```
pakrym: nit: reformat

src/Microsoft.DotNet.Cli.Utils/HashHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Microsoft.DotNet.Cli.Utils
+{
+    public static class HashHelper
+    {
+        public static string sha256_hash(string value)
```
pakrym: CamelCase

test/dotnet.Tests/TelemetryCommandTest.cs
PinpointTownes: Well, actually, it will never compile without a body... so it's not just about code style :trollface:

src/Microsoft.DotNet.Cli.Utils/HashHelper.cs
```diff
@@ -0,0 +1,29 @@
+using System;
```
eerhardt: Copyright header

src/Microsoft.DotNet.Cli.Utils/HashHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Microsoft.DotNet.Cli.Utils
+{
+    public static class HashHelper
+    {
+        public static string sha256_hash(string value)
+        {
+            StringBuilder Sb = new StringBuilder();
```
eerhardt: If this method took a string[], we could hash all the args with 1 stringBuilder and 1 SHA256 instance being created. Instead of needing to create new instances for each arg.

src/dotnet/Program.cs
```diff
@@ -129,13 +139,7 @@ internal static int ProcessArgs(string[] args, ITelemetry telemetryClient)
                 exitCode = result.ExitCode;
             }
 
-            telemetryClient.TrackEvent(
-                command,
-                null,
-                new Dictionary<string, double>
-                {
-                    ["ExitCode"] = exitCode
```
eerhardt: Do we not care about the exit code anymore?

src/dotnet/Telemetry.cs
```diff
@@ -31,12 +42,26 @@ public Telemetry()
                 return;
             }
 
+            int sampleRate = ReciprocalSampleRateValue;
+            bool isCITestMachine = Env.GetEnvironmentVariableAsBool(ContinousIntegrationFlag);
+
+            if(isCITestMachine)
+            {
+                sampleRate = ReciprocalSampleRateValueForCI;
+                _shouldLogTelemetry = (Environment.TickCount % ReciprocalSampleRateValue == 0);
```
eerhardt: Seems like this should use 'sampleRate' variable

src/dotnet/Telemetry.cs
```diff
@@ -31,12 +42,26 @@ public Telemetry()
                 return;
             }
 
+            int sampleRate = ReciprocalSampleRateValue;
+            bool isCITestMachine = Env.GetEnvironmentVariableAsBool(ContinousIntegrationFlag);
```
eerhardt: Who sets this environment variable?

eerhardt: And do we EVER need telemetry from our CI builds?  They do the same thing over and over.

src/dotnet/Telemetry.cs
```diff
@@ -63,7 +90,18 @@ public void TrackEvent(string eventName, IDictionary<string, string> properties,
             {
                 return;
             }
+            if(!_shouldLogTelemetry)
+            {
+                return;
+            }                
 
+            _trackEventTask = Task.Factory.StartNew(
```
eerhardt: The way this is implemented it means an instance of a TelemetryClient can only be used to track a single event.  Is that intended?

src/dotnet/Program.cs
```diff
@@ -114,6 +114,16 @@ internal static int ProcessArgs(string[] args, ITelemetry telemetryClient)
                 command = "help";
             }
 
+            var hashedargs = string.Join(",", args.Select(x => HashHelper.sha256_hash(x)));
```
eerhardt: I think you want "appArgs" here, right?  Or is it ok to have the command inside the args again?

eerhardt:
>                     return 0;

Looks like as it is implemented, you won't get --version, help, info commands tracked with telemetry.  Is that OK?

---
Refers to: src/dotnet/Program.cs:75 in 9c36fd9. [](commit_id = 9c36fd91a702b63a6688bfb08ad239d89a57c079, deletion_comment = False)

src/dotnet/Telemetry.cs
```diff
@@ -9,18 +10,28 @@ namespace Microsoft.DotNet.Cli
...
+        public const double TimeoutInMilliseconds = 100;
+
         private bool _isInitialized = false;
         private TelemetryClient _client = null;
 
         private Dictionary<string, string> _commonProperties = null;
         private Dictionary<string, double> _commonMeasurements = null;
+        private Task _trackEventTask = null;
+
+        private const int ReciprocalSampleRateValue = 5;
```
eerhardt: A question I have about sampling:

When we did Telemetry in the VS features I worked on, everyone used the scenario of "first the user did X, then the user did Y".  If we sample, there is no way we can tie two commands together any more.  Is that OK?  Do we care if a user first did restore and then build?  or build and then publish and then test?

eerhardt: I guess it really comes down to: "What are we going to use this telemetry information for?"

---
In reply to: [61365252](https://github.com/dotnet/cli/pull/2724#discussion_r61365252) [](ancestors = 61365252)

src/dotnet/Telemetry.cs
```diff
@@ -63,7 +90,18 @@ public void TrackEvent(string eventName, IDictionary<string, string> properties,
             {
                 return;
             }
+            if(!_shouldLogTelemetry)
+            {
+                return;
+            }
```
MichaelSimons: <nit> extra space.

src/dotnet/Telemetry.cs
```diff
@@ -45,6 +70,8 @@ public Telemetry()
                 _commonProperties.Add(OSPlatform, runtimeEnvironment.OperatingSystemPlatform.ToString());
                 _commonProperties.Add(RuntimeId, runtimeEnvironment.GetRuntimeIdentifier());
                 _commonProperties.Add(ProductVersion, Product.Version);
+                _commonProperties.Add(CITestMachine, isCITestMachine.ToString());
```
MichaelSimons: This seems wasteful given the number of times it would occur in the grand scheme.  Can telemetry just be turned off for CI or filtered on the back end (e.g. via IP)?

MichaelSimons:
>         public void TrackEvent(string eventName, IDictionary<string, string> properties, IDictionary<string, double> measurements)

One thing that seems a little heavy handed is the number of Dictionaries that are allocated by the callers of this method that will never get used because either telemetry is disabled or isn't being sampled.

---
Refers to: src/dotnet/Telemetry.cs:87 in 9c36fd9. [](commit_id = 9c36fd91a702b63a6688bfb08ad239d89a57c079, deletion_comment = False)

src/dotnet/Program.cs
adamgorMSFT: If we wait until the end to start sending telemetry, there will always be a 200-500 millisecond delay needed to complete the track event. If we wish to keep exitCode, we'd have to delay exiting the program until it was completed. 

There is no way around this, so we have to start sending earlier, which means we have to drop the exitCode.   Its not for a lack of wanting to know about exitCode, its about the constraint of minimizing impact on the runtime.

src/dotnet/Telemetry.cs
adamgorMSFT: Fixed

src/Microsoft.DotNet.Cli.Utils/HashHelper.cs
adamgorMSFT: Fixed

src/Microsoft.DotNet.Cli.Utils/HashHelper.cs
adamgorMSFT: Fixed, changed to Sha256Hash

test/dotnet.Tests/TelemetryCommandTest.cs
adamgorMSFT: fixed tabs/spaces and added empty body

src/dotnet/Telemetry.cs
adamgorMSFT: At this time there should only be the single trackEvent.   However, the only reason we even need to capture the Task to begin with is to add a little wait 100 milliseconds. So we only care about waiting for the latest sent task.  As the earlier ones should have completed by the time the last one finishes.

src/dotnet/Program.cs
anurse: I'd rather drop exit code than incur the delay at the end. Post-RC2 we could look at re-introducing it by having it shell out a new process that won't block the foreground process to capture the exit code and other data.

src/dotnet/Telemetry.cs
```diff
@@ -78,6 +116,13 @@ public void TrackEvent(string eventName, IDictionary<string, string> properties,
             }
         }
 
+        public void Finish()
+        {
+            if(_trackEventTask != null)
+            {
+                _trackEventTask.Wait(TimeSpan.FromMilliseconds(TimeoutInMilliseconds));
```
anurse: This could throw if the task gets cancelled. Can you wrap this in a try/catch just in case?

src/dotnet/Program.cs
eerhardt: I asked this in a separate comment, but I think it is worth repeating here:

"What are we going to use this telemetry information for?"

I don't think anyone has answered that question, and that's why we are having questions like this pop up in the implementation.

Is our goal to use telemetry to understand if customers are successful using our product?

src/Microsoft.DotNet.Cli.Utils/HashHelper.cs
adamgorMSFT: Hashed arguments are still being debated.  However, this change will be made if it stays.

src/dotnet/Telemetry.cs
adamgorMSFT: The EnvVar would be set on the CI machines. 

This isn't finalized yet, but was to help further the conversation regarding CI machines.

src/dotnet/Program.cs
adamgorMSFT: There is still an offline discussion regarding what arguments to capture and how to do it.

src/dotnet/Telemetry.cs
adamgorMSFT: You make a great point regarding the sampling potential gaps.  This was originally requested because of original performance concerns, but after having a recent meeting, the issue might be resolved and sampling might not be needed.

src/dotnet/Telemetry.cs
adamgorMSFT: fixed

src/dotnet/Telemetry.cs
eerhardt: On who's CI machines?  Our customers?  Or the CI machines that build https://github.com/dotnet/cli ?

src/dotnet/Telemetry.cs
adamgorMSFT: The entire Finish and Wait is being dropped.

src/dotnet/Telemetry.cs
adamgorMSFT: I will investigate, but I believe it was for the CI machines that you mentioned.
it has been refactored for now though to be just testMachines.

src/dotnet/Telemetry.cs
adamgorMSFT: Also to address 
"Looks like as it is implemented, you won't get --version, help, info commands tracked with telemetry. Is that OK?"

This is being investigated as to how to better handle this scenario.

src/dotnet/Telemetry.cs
adamgorMSFT: This has been refactored to just TestMachine for now.

Telemetry can be turned off by anyone at anytime for any reason by simply setting the env variable DOTNET_CLI_TELEMETRY_OPTOUT=1

piotrpMSFT:

piotrpMSFT:

adamgorMSFT:

piotrpMSFT:
https://github.com/dotnet/cli/issues/2825

piotrpMSFT:
@adamgorMSFT this looks like a real issue...

Petermarcu:
@adamgorMSFT , whats the status of this change? Is this able to be completed this morning?

piotrpMSFT:
@eerhardt will also take a look at this. @adamgorMSFT please coordinate with Eric.

piotrpMSFT:
:shipit:

piotrpMSFT:
Once this is green, the merge is approved.

src/dotnet/Telemetry.cs
```diff
@@ -33,60 +42,73 @@ public Telemetry()
                 return;
             }
 
-            try
+            _telemetryProfile = Environment.GetEnvironmentVariable(TelemetryProfileEnvironmentVariable);
```
MichaelSimons: Is it necessary to do this in the ctor and store the result as a private member versus reading it in InitializeTelemetry and not storing it?

MichaelSimons:
>         private TelemetryClient _client = null;

It may just be a preference but it doesn't seem necessary to track isInitialized.  Couldn't _client be checked for null instead.  You would need to ensure _client is set to null in the catch of the InitializeTelemetry method.

---
Refers to: src/dotnet/Telemetry.cs:17 in 6d69818. [](commit_id = 6d698181a0c3373adb3e5e7a85ebdca4ca073768, deletion_comment = False)

piotrpMSFT:
This was approved offline. Merging.

piotrpMSFT:
@eerhardt @MichaelSimons had some good feedback on implementation. It's not blocking, so I merged, but do consider once you're back online.

eerhardt:
I addressed the PR feedback with #2858

PinpointTownes:
Really nice to see you've changed your mind about the arguments (well, at least for this milestone :smile:)

:clap:

PR #2726, opened 2016-04-28
[WIP] Resolve hostpolicy out of package graph for servicing

schellap:
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2726)

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -14,6 +14,37 @@
 #include "error_codes.h"
 #include "deps_format.h"
 
+
+void handle_missing_framework_error(const pal::string_t& fx_name, const pal::string_t& fx_version, const pal::string_t& fx_dir)
+{
+    pal::string_t fx_ver_dirs = get_directory(fx_dir);
+
+    trace::error(_X("The targeted framework { \"%s\": \"%s\" } was not found."), fx_name.c_str(), fx_version.c_str());
+    trace::error(_X("  - Check application dependencies and target a framework version installed at:"));
+    trace::error(_X("      %s"), fx_ver_dirs.c_str());
+
+    bool header = true;
+    std::vector<pal::string_t> versions;
+    pal::readdir(fx_ver_dirs, &versions);
+    for (const auto& ver : versions)
+    {
+        fx_ver_t parsed(-1, -1, -1);
+        if (fx_ver_t::parse(ver, &parsed, false))
+        {
+            if (header)
+            {
+                trace::error(_X("  - The following versions are installed:"));
+                header = false;
+            }
+            trace::error(_X("      %s"), ver.c_str());
+        }
+    }
+    if (header)
+    {
+        trace::error(_X("  - Or install the framework version that is being targeted."));
+    }
+}
+
 pal::string_t resolve_impl_version_from_deps(const pal::string_t& deps_json)
 {
     trace::verbose(_X("--- Resolving %s version from deps json [%s]"), LIBHOSTPOLICY_NAME, deps_json.c_str());
@@ -21,14 +52,14 @@ pal::string_t resolve_impl_version_from_deps(const pal::string_t& deps_json)
     pal::string_t retval;
     if (!pal::file_exists(deps_json))
     {
-        trace::verbose(_X("[%s] does not exist"), deps_json.c_str());
+        trace::verbose(_X("Dependency manifest [%s] does not exist"), deps_json.c_str());
         return retval;
     }
 
     pal::ifstream_t file(deps_json);
     if (!file.good())
     {
-        trace::verbose(_X("[%s] could not be opened"), deps_json.c_str());
+        trace::verbose(_X("Dependency manifest [%s] could not be opened"), deps_json.c_str());
         return retval;
     }
 
@@ -59,7 +90,7 @@ pal::string_t resolve_impl_version_from_deps(const pal::string_t& deps_json)
         (void)pal::utf8_palstring(je.what(), &jes);
         trace::error(_X("A JSON parsing exception occurred in [%s]: %s"), deps_json.c_str(), jes.c_str());
     }
-    trace::verbose(_X("%s version is [%s] in deps json file [%s]"), retval.c_str(), deps_json.c_str());
+    trace::verbose(_X("Resolved version %s from dependency manifest file [%s]"), retval.c_str(), deps_json.c_str());
     return retval;
 }
 
@@ -80,10 +111,13 @@ bool to_hostpolicy_package_path(const pal::string_t& dir, const pal::string_t& v
 
     if (!pal::file_exists(path))
     {
+        trace::verbose(_X("Did not find %s in path %s"), LIBHOSTPOLICY_NAME, path.c_str());
         return false;
     }
 
     *candidate = path;
+
+    trace::verbose(_X("Found %s in path %s"), LIBHOSTPOLICY_NAME, path.c_str());
     return true;
 }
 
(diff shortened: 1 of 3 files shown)


brthor:
😄

brthor:
This should let us use the muxer to call xunit after this goes in yeah?

schellap:
@brthor, that works as is today without this PR. This is about servicing the hostpolicy the same way instead of using baked-in versions.

piotrpMSFT:
@gkhanna79 @sokket can one of you give this a shipit after reviewing?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
 #include "deps_format.h"
 
-pal::string_t fx_muxer_t::resolve_fx_dir(const pal::string_t& muxer_dir, runtime_config_t* runtime)
+pal::string_t resolve_impl_version_from_deps(const pal::string_t& deps_json)
 {
-    trace::verbose(_X("--- Resolving FX directory from muxer dir [%s]"), muxer_dir.c_str());
-    const auto fx_name = runtime->get_fx_name();
-    const auto fx_ver = runtime->get_fx_version();
-    const auto patch_roll_fwd = runtime->get_patch_roll_fwd();
+    trace::verbose(_X("--- Resolving %s version from deps json [%s]"), LIBHOSTPOLICY_NAME, deps_json.c_str());
```
sokket: if pal::string is a wstring (on Windows), then the correct printf is %ls not %s (deps_json)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+
+    pal::string_t retval;
+    if (!pal::file_exists(deps_json))
+    {
+        trace::verbose(_X("[%s] does not exist"), deps_json.c_str());
```
sokket: Same

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+        (void)pal::utf8_palstring(je.what(), &jes);
+        trace::error(_X("A JSON parsing exception occurred in [%s]: %s"), deps_json.c_str(), jes.c_str());
+    }
+    trace::verbose(_X("%s version is [%s] in deps json file [%s]"), retval.c_str(), deps_json.c_str());
+    return retval;
+}
+
+bool to_hostpolicy_package_path(const pal::string_t& dir, const pal::string_t& version, pal::string_t* candidate)
+{
+    candidate->clear();
```
sokket: De-referencing a pointer passed into a function without asserting it is non-null; please add an assert != nullptr

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+
+    pal::string_t rel_dir = _STRINGIFY(HOST_POLICY_PKG_REL_DIR);
```
sokket: Why are we using _X some places with _STRINGIFY in others? :)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+    append_path(&path, _STRINGIFY(HOST_POLICY_PKG_NAME));
+    append_path(&path, version.c_str());
+    append_path(&path, rel_dir.c_str());
+
+    if (!pal::file_exists(path))
+    {
+        return false;
+    }
+
+    *candidate = path;
```
sokket: NIT: rather than rely on copy-constructor semantics, perhaps use ```assign```? That will make it more clear

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+    return true;
+}
+
+bool hostpolicy_exists_in_svc(const pal::string_t& version, pal::string_t* resolved_dir)
```
sokket: Add an assert that your pointer is not nullptr

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+}
+
+bool fx_muxer_t::resolve_impl_dir(host_mode_t mode,
+    const pal::string_t& own_dir,
+    const pal::string_t& fx_dir,
+    const pal::string_t& app_candidate,
+    const pal::string_t& deps_file,
+    const std::vector<pal::string_t>& probe_realpaths,
+    const runtime_config_t& config,
+    pal::string_t* impl_dir)
```
sokket: Same about asserting

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+    if (config.get_portable())
+    {
+        pal::string_t fx_deps = fx_dir;
+        pal::string_t fx_deps_name = config.get_fx_name() + _X(".deps.json");
+        append_path(&fx_deps, fx_deps_name.c_str());
+
+        pal::string_t version = resolve_impl_version_from_deps(fx_deps);
+        if (!hostpolicy_exists_in_svc(version, impl_dir))
+        {
+            *impl_dir = fx_dir;
```
sokket: Same as above, use assign so the reader doesn't need to understand std::string copy constructors

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+bool fx_muxer_t::resolve_impl_dir(host_mode_t mode,
+    const pal::string_t& own_dir,
+    const pal::string_t& fx_dir,
+    const pal::string_t& app_candidate,
+    const pal::string_t& deps_file,
+    const std::vector<pal::string_t>& probe_realpaths,
+    const runtime_config_t& config,
+    pal::string_t* impl_dir)
+{
+    if (config.get_portable())
```
sokket: Seems like these two if statements should be broken out into helper functions; one for portable and one for not

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+            for (const auto& probe_path : probe_realpaths)
+            {
+                trace::verbose(_X("Considering %s for hostpolicy library"), probe_path.c_str());
+                if (to_hostpolicy_package_path(probe_path, version, &candidate))
+                {
+                    found = true; // candidate contains the right path.
+                    break;
+                }
+            }
+            if (!found)
```
sokket: NIT: newline

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+    if (skip_utf8_bom(&file))
+    {
+        trace::verbose(_X("UTF-8 BOM skipped while reading [%s]"), deps_json.c_str());
+    }
+
+    try
+    {
+        const auto root = json_value::parse(file);
+        const auto& json = root.as_object();
+        const auto& libraries = json.at(_X("libraries")).as_object();
```
sokket: What types are these ```auto``` types? I think we should be explicit here since it is not immediately obvious by looking (per code styling)

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+    if (mode == host_mode_t::split_fx)
+    {
+        return own_dir;
+    }
+    assert(mode == host_mode_t::muxer);
+
+    trace::verbose(_X("--- Resolving FX directory from muxer dir [%s]"), own_dir.c_str());
+    const auto fx_name = config.get_fx_name();
+    const auto fx_ver = config.get_fx_version();
+    const auto patch_roll_fwd = config.get_patch_roll_fwd();
```
sokket: Same here IRT auto

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -28,7 +204,7 @@ pal::string_t fx_muxer_t::resolve_fx_dir(const pal::string_t& muxer_dir, runtime
         return pal::string_t();
     }
 
-    auto fx_dir = muxer_dir;
+    auto fx_dir = own_dir;
```
sokket: And here

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,188 @@
...
+    trace::verbose(_X("[%s] doesn't exist in servicing [%s]"), LIBHOSTPOLICY_NAME, path.c_str());
+    return false;
+}
+
+bool fx_muxer_t::resolve_impl_dir(host_mode_t mode,
+    const pal::string_t& own_dir,
+    const pal::string_t& fx_dir,
+    const pal::string_t& app_candidate,
+    const pal::string_t& deps_file,
+    const std::vector<pal::string_t>& probe_realpaths,
```
sokket: Since the pal::string objects aren't changing either, should this be ```std::vector<const pal::string_t>``` to be sure nobody changes the underlying strings?

PR #2732, opened 2016-04-28
Remove references to Microsoft.Extensions.Logging

pranavkm:
cc @davidfowl \ @anurse \ @pakrym

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2732)

### DIFF
--- a/test/dotnet-projectmodel-server.Tests/DthTestClient.cs
+++ b/test/dotnet-projectmodel-server.Tests/DthTestClient.cs
@@ -9,7 +9,6 @@ using System.Net;
 using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Xunit;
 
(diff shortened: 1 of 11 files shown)


test/dotnet-projectmodel-server.Tests/DthTestClient.cs
```diff
@@ -238,7 +234,7 @@ public void Dispose()
                 // Swallow this error for now.
                 // This is a temporary fix for a random failure on CI. The issue happens on Windowx x86
                 // only.
-                _logger.LogError($"Exception thrown durning socket shutting down: {ex.SocketErrorCode}.");
+                Console.Error.WriteLine($"Exception thrown durning socket shutting down: {ex.SocketErrorCode}.");
```
pakrym: Reporter?

pakrym:
What about actual reference to package?

pranavkm:
There's a cyclic dependency between coreclr.xunit and CLI. We have to compile CoreCLR.xunit against a CLI where it does not reference `ILogger` at which point we can remove the package reference from `Testing.Abstractions`.

test/dotnet-projectmodel-server.Tests/DthTestClient.cs
pranavkm: Just a test.

pakrym:
:shipit:

PR #2733, opened 2016-04-28
Add performance tracing to dotnet cli

pakrym:
@anurse

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2733)

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -42,16 +42,32 @@ namespace Microsoft.DotNet.Cli
         {
             DebugHelper.HandleDebugSwitch(ref args);
 
+            if (Env.GetEnvironmentVariableAsBool("DOTNET_CLI_CAPTURE_TIMING", false))
+            {
+                PerfTrace.Enabled = true;
+            }
+
             InitializeProcess();
 
             try
             {
-                return Program.ProcessArgs(args, new Telemetry());
+                using (PerfTrace.Current.CaptureTiming())
+                {
+                    return ProcessArgs(args, new Telemetry());
+                }
             }
             catch (GracefulException e)
             {
                 Console.WriteLine(e.Message.Red().Bold());
                 return 1;
+            }
+            finally
+            {
+                if (PerfTrace.Enabled)
+                {
+                    Reporter.Output.WriteLine("Performance Summary:");
+                    PerfTraceOutput.Print(Reporter.Output, PerfTrace.GetEvents());
+                }
             }
         }
 
@@ -107,6 +123,7 @@ namespace Microsoft.DotNet.Cli
             if (verbose.HasValue)
             {
                 Environment.SetEnvironmentVariable(CommandContext.Variables.Verbose, verbose.ToString());
+                Console.WriteLine($"Telemetry is: {(telemetryClient.Enabled ? "Enabled" : "Disabled")}");
             }
 
             if (string.IsNullOrEmpty(command))
@@ -140,6 +157,7 @@ namespace Microsoft.DotNet.Cli
             return exitCode;
 
         }
+
 
         private static void InitializeProcess()
         {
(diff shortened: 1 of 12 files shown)


pakrym:
@davidfowl

test/dotnet.Tests/TelemetryCommandTest.cs
```diff
@@ -14,7 +14,10 @@ namespace Microsoft.DotNet.Tests
 {
     public class MockTelemetry : ITelemetry
     {
-        public string EventName{get;set;}
+        public bool Enabled { get; set; }
+
+        public string EventName{ get; set; }
```
anurse: spacing

anurse:
:shipit:

pakrym:

src/dotnet/Program.cs
```diff
@@ -42,17 +42,33 @@ public static int Main(string[] args)
         {
             DebugHelper.HandleDebugSwitch(ref args);
 
+            if (string.Equals(Environment.GetEnvironmentVariable("DOTNET_CLI_CAPTURE_TIMING"), "1"))
```
eerhardt: We have an Env class that has a "get Boolean value" method. Can you use that instead so we get a consistent handling of Boolean Env vars?

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -68,18 +73,33 @@ private CompilationResult CompileWithDependencies(ProjectGraphNode projectNode)
             }
 
             var context = projectNode.ProjectContext;
-            if (!HasSourceFiles(context))
+            using (PerfTrace.Current.CaptureTiming($"{projectNode.ProjectContext.ProjectName()}", nameof(HasSourceFiles)))
```
eerhardt: Would it make more sense to put this tracing inside each of these methods?

src/dotnet/commands/dotnet-build/ProjectGraphNode.cs
```diff
@@ -20,7 +21,16 @@ public ProjectGraphNode(Task<ProjectContext> projectContext, IEnumerable<Project
             IsRoot = isRoot;
         }
 
-        public ProjectContext ProjectContext { get { return _projectContextCreator.GetAwaiter().GetResult(); } }
+        public ProjectContext ProjectContext
+        {
+            get
+            {
+                using (_projectContextCreator.IsCompleted ? null : PerfTrace.Current.CaptureTiming("", "Blocking ProjectContext wait"))
```
eerhardt: This doesn't seem like a very good use of a property if it requires trace logging inside of it. Property accesses are supposed to be fast.

test/dotnet.Tests/TelemetryCommandTest.cs
```diff
@@ -14,7 +14,10 @@ namespace Microsoft.DotNet.Tests
 {
     public class MockTelemetry : ITelemetry
     {
-        public string EventName{get;set;}
+        public bool Enabled { get; set; }
+
+        public string EventName { get; set; }
```
eerhardt: Needed?  I didn't see this on the interface.

src/dotnet/commands/dotnet-build/ProjectGraphNode.cs
pakrym: Most of the time task finishes before someone accesses this property, tracing is here to check for cases when it's not fast enough.

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
pakrym: I tried, wrapping whole method body in using looked worse then doing it here.

test/dotnet.Tests/TelemetryCommandTest.cs
pakrym: Nice catch!

pakrym: It's used later in the test to store method argument.

eerhardt: Oh right, I see what happened.  Since you reformatted the line it showed up looking like you added a new property. I missed that this property already existed.  Sorry for the distraction.

pakrym:

Issue #2734, opened 2016-04-28
Windows build failure: System.IO.DirectoryNotFoundException: Could not find a part of the path 'F:\Agent\_work\2\s\artifacts\win10-x86\corehost\locked\dotnet.exe'


Sridhar-MS:
Windows build failing in VSO

>
[TARGET     <] [FAIL] [00:01:41.808] CompileStage1                                         (Microsoft.DotNet.Cli.Build.CompileTargets.CompileStage1)
2016-04-28T17:27:54.8758491Z Build failed: System.IO.DirectoryNotFoundException: Could not find a part of the path 'F:\Agent\_work\2\s\artifacts\win10-x86\corehost\locked\dotnet.exe'.
2016-04-28T17:27:54.8758491Z    at System.IO.Win32FileSystem.CopyFile(String sourceFullPath, String destFullPath, Boolean overwrite)
2016-04-28T17:27:54.8758491Z    at System.IO.File.InternalCopy(String sourceFileName, String destFileName, Boolean overwrite)
2016-04-28T17:27:54.8758491Z    at System.IO.File.Copy(String sourceFileName, String destFileName, Boolean overwrite)
2016-04-28T17:27:54.8758491Z    at Microsoft.DotNet.Cli.Build.CompileTargets.CopySharedHost(String outputDir) in F:\agent\_work\2\s\scripts\dotnet-cli-build\CompileTargets.cs:line 463
2016-04-28T17:27:54.8758491Z    at Microsoft.DotNet.Cli.Build.CompileTargets.CompileStage1(BuildTargetContext c) in F:\agent\_work\2\s\scripts\dotnet-cli-build\CompileTargets.cs:line 324
2016-04-28T17:27:54.8758491Z    at Microsoft.DotNet.Cli.Build.Framework.BuildContext.ExecTarget(BuildTarget target) in F:\agent\_work\2\s\scripts\Microsoft.DotNet.Cli.Build.Framework\BuildContext.cs:line 144
2

@schellap I think this is related to your changes.

cc @piotrpMSFT

Petermarcu:
@Sridhar-MS should we pull out #2636 to unblock the builds assuming its related?

schellap:
Flip variabled Dirs.CoreHostLocked to be same as Dirs.CoreHost.

Sridhar-MS:
[...]

PR #2735, opened 2016-04-28
Workaround for 2734

schellap:
I am on my mobile right now, I'll have to debug what the issue is on the VSO machines once I get in. This should solve it in the short term.

@Sridhar-MS @petermarcu @piotrpMSFT.

The restore and publish of hostfxr package is failing.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2735)


schellap:
Works around #2734

Issue #2731, opened 2016-04-28
EULA link in README.md points to an RTF file in GitHub, can't read it


Eilon:
The default readme shown at https://github.com/dotnet/cli/ has a link a the bottom that points at the EULA. The link is to https://aka.ms/dotnet-cli-eula, which redirects to https://github.com/dotnet/cli/blob/master/packaging/osx/sharedframework/resources/en.lproj/eula.rtf.

Unfortunately GitHub doesn't have a viewer for RTF files, so all you see is garbled curly braces and backslashes with some text scattered around.

Is there a TXT or HTML version of the EULA that this could point to instead?

piotrpMSFT:
@blackdwarf was looking at adding EULAs to https://dotnet.github.io. Those are supposed to handle our web-readable EULA needs.

blackdwarf:
This is now fixed.

Eilon:
Looks good, thanks!

PR #2738, opened 2016-04-28
Add markdown verison of en-EULA

jasonwilliams200OK:
Contributes to #2731.

cc: @Elion

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2738)

### DIFF
--- a/packaging/osx/sharedframework/resources/en.lproj/eula.md
+++ b/packaging/osx/sharedframework/resources/en.lproj/eula.md
@@ -1,0 +1,94 @@
+**MICROSOFT SOFTWARE LICENSE TERMS**
+
+**MICROSOFT .NET LIBRARY**
+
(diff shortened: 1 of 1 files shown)


blowdart:
This appears to be for the ".NET Library". It's frankly not clear what that covers, is the cli? the framework? something else entirely?

jasonwilliams200OK:
@blowdart, this is just a markdown variant of https://github.com/dotnet/cli/blob/13817c7/packaging/osx/sharedframework/resources/en.lproj/eula.rtf
See https://github.com/dotnet/cli/tree/13817c7/packaging/osx/sharedframework/resources/en.lproj

blowdart:
Yes .. so I'd posit that's wrong there too.

blackdwarf:
@jasonwilliams200OK let's hold off on this one for a while, there are talks about whether the EULA is the right one or not for the CLI we should figure it out this week.

Issue #2660, opened 2016-04-22
dotnet build reporting "PE image does not have metadata" with no details.


dantraMSFT:
## Steps to reproduce
I'm hitting this in a private repro building an assembly building against the full CLR.

## Expected  behavior
Builds clean or reports the errant image not containing metadatea

## Actual behavior
Compilation Succeeds with 0 errors and warnings. Elapsed time is reported. The text "PE image does not have metadata." is reported and dotnet build exits with 1.
The DLL and PDB files are produced and ildasm is able to report metadata for the assembly.

## Environment data
`dotnet --info` output:
.NET Command Line Tools (1.0.0-rc2-002444)

Product Information:
 Version:     1.0.0-rc2-002444
 Commit Sha:  ada4e363c0

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14329
 OS Platform: Windows
 RID:         win10-x64

piotrpMSFT:
@jaredpar @cdmihai does the error message "PE image does not have metadata" mean something to either of you? Is CSC producing this?

@dantraMSFT can you provide a minimal repro?

dantraMSFT:
@piotrpMSFT - I haven't been able to make it too minimal but I did managed to isolate it down to the single project and as set of packages.  How should I get it to you?

cdmihai:
@piotrpMSFT never seen it. Seems like a resource that gets written in the obj? https://github.com/search?q=PEImageDoesNotHaveMetadata&type=Code&utf8=%E2%9C%93

@dantraMSFT a zip with command line repro steps would be nice.

piotrpMSFT:
@dantraMSFT can you share the repro with @cdmihai?

jaredpar:
This error is actually coming from System.Reflection.MetadataReader:

> https://github.com/dotnet/corefx/blob/d0dc5fc099946adc1035b34a8b1f6042eddb0c75/src/System.Reflection.Metadata/src/System/Reflection/PortableExecutable/PEReader.cs#L337

@tmat has more context but I'm guessing the compiler is being asked to reference a file that isn't actually a valid PE.

piotrpMSFT:
@cdmihai @tmat, I trust you guys will figure out whose fix this is :)

piotrpMSFT:
any updates here?

cdmihai:
Haven't had the time to look into this yet. I'll try to repro it tomorrow.
[...]

PR #2744, opened 2016-04-28
Add exception handling when makerunable fails

cdmihai:
Better error handling for #2660

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2744)

### DIFF
--- a/src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
+++ b/src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
@@ -179,33 +179,40 @@ namespace Microsoft.DotNet.Cli.Compiler.Common
 
         private static AssemblyReferenceInfo GetAssemblyInfo(LibraryAsset arg)
         {
-            using (var peReader = new PEReader(File.OpenRead(arg.ResolvedPath)))
+            try
             {
-                var metadataReader = peReader.GetMetadataReader();
-
-                var definition = metadataReader.GetAssemblyDefinition();
-
-                var identity = new AssemblyIdentity(
-                    metadataReader.GetString(definition.Name),
-                    definition.Version,
-                    metadataReader.GetString(definition.Culture),
-                    GetPublicKeyToken(metadataReader.GetBlobBytes(definition.PublicKey))
-                );
-
-                var references = new List<AssemblyIdentity>(metadataReader.AssemblyReferences.Count);
-
-                foreach (var assemblyReferenceHandle in metadataReader.AssemblyReferences)
+                using (var peReader = new PEReader(File.OpenRead(arg.ResolvedPath)))
                 {
-                    var assemblyReference = metadataReader.GetAssemblyReference(assemblyReferenceHandle);
-                    references.Add(new AssemblyIdentity(
-                        metadataReader.GetString(assemblyReference.Name),
-                        assemblyReference.Version,
-                        metadataReader.GetString(assemblyReference.Culture),
-                        GetPublicKeyToken(metadataReader.GetBlobBytes(assemblyReference.PublicKeyOrToken))
-                    ));
-                }
+                    var metadataReader = peReader.GetMetadataReader();
 
-                return new AssemblyReferenceInfo(identity, references.ToArray());
+                    var definition = metadataReader.GetAssemblyDefinition();
+
+                    var identity = new AssemblyIdentity(
+                        metadataReader.GetString(definition.Name),
+                        definition.Version,
+                        metadataReader.GetString(definition.Culture),
+                        GetPublicKeyToken(metadataReader.GetBlobBytes(definition.PublicKey))
+                        );
+
+                    var references = new List<AssemblyIdentity>(metadataReader.AssemblyReferences.Count);
+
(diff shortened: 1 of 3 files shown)


TestAssets/TestProjects/AppWithDependencyOnMalformedPackage/NuGet.Config
```diff
@@ -0,0 +1,12 @@
...
+  <config>
+  </config>
+  <packageSources>
+    <clear />
+    <add key="local" value="nuget-feed" />
+    <add key="dotnet-core" value="https://www.myget.org/F/dotnet-core/api/v3/index.json" />
+    <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+</configuration>
```
piotrpMSFT: Why do we need a copy of NuGet.config?

TestAssets/TestProjects/AppWithDependencyOnMalformedPackage/project.json
```diff
@@ -0,0 +1,14 @@
+{
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "BadPackage": "1.0.0"
```
piotrpMSFT: Where do we generate this package?

src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
```diff
@@ -179,33 +179,40 @@ internal static AssemblyRedirect[] CollectRedirects(IEnumerable<AssemblyReferenc
...
+                            metadataReader.GetString(assemblyReference.Culture),
+                            GetPublicKeyToken(metadataReader.GetBlobBytes(assemblyReference.PublicKeyOrToken))
+                            ));
+                    }
 
-                return new AssemblyReferenceInfo(identity, references.ToArray());
+                    return new AssemblyReferenceInfo(identity, references.ToArray());
+                }
+            }
+            catch (Exception e)
```
piotrpMSFT: What kind of exception is this?

src/dotnet/Program.cs
```diff
@@ -50,7 +50,8 @@ public static int Main(string[] args)
             }
             catch (GracefulException e)
             {
-                Console.WriteLine(e.Message.Red().Bold());
+                Console.WriteLine(CommandContext.IsVerbose() ? e.ToString() : e.Message.Red().Bold());
```
piotrpMSFT: Why do we turn off color coding + bolding for Verbose?

TestAssets/TestProjects/AppWithDependencyOnMalformedPackage/NuGet.Config
brthor: We should avoid checking this in if at all possible.

TestAssets/TestProjects/AppWithDependencyOnMalformedPackage/project.json
brthor: I see this package checked in above, is there a way to generate a bad package so we can avoid checking it in? When things like tfms change, it can be tough to maintain these checked in nupkgs since the code to re-generate them isn't here.

TestAssets/TestProjects/AppWithDependencyOnMalformedPackage/NuGet.Config
cdmihai: It's because the test involves using a malformed Nuget package that has a dll without any metadata.

Does `dotnet restore --source "./nuget-feed"` achieve the same thing?

brthor: `dotnet restore --fallbacksource ./nuget-feed` is probably what you are looking for.

`--source` excludes any other sources
`--fallbacksource` adds sources

TestAssets/TestProjects/AppWithDependencyOnMalformedPackage/project.json
cdmihai: It's hard to generate, because the bad dll should be present in lib but not in ref.

Right now I'd vote for deleting the test and its test data, and just keeping the exception handling. It's more trouble than it's worth :)

brthor: I think if you must check it in, at least put in the nuspec or some way to regenerate the package if need be

src/Microsoft.DotNet.Compiler.Common/BindingRedirectGenerator.cs
cdmihai: No idea what PEReader might throw. I made it general to indicate: If anything crashes while reading the assembly info, we'll wrap over the arbitrary exception with a more clear message: "Could not read assembly metadata ... "

Petermarcu:
@piotrpMSFT , Can you get this merged today?

piotrpMSFT:
:shipit:

PR #2747, opened 2016-04-29
Replaces references to Microsoft.Extensions.PlatformAbstractions

pranavkm:
… Microsoft.Extensions.PlatformAbstractions.Internal

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2747)


krwq:
Is this refactoring needed for RC2?

pranavkm:
@krwq - we were making some breaking changes to PlatformAbstractions (https://github.com/aspnet/PlatformAbstractions/pull/39) but doing so causes tests to fail with type load exceptions. The plan was to move the relevant code in to CLI to break this cycle.

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/NativeMethods.Unix.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Extensions.PlatformAbstractions.Internal.Native
+{
+    internal static partial class NativeMethods
+    {
+        public static class Unix
+        {
+            public unsafe static string GetUname()
```
eerhardt: I do believe this is exposed by `System.Runtime.InteropServices.RuntimeInformation.OSDescription`.  Maybe we could use that instead?

https://github.com/dotnet/corefx/blob/master/src/System.Runtime.InteropServices.RuntimeInformation/src/System/Runtime/InteropServices/RuntimeInformation/RuntimeInformation.Unix.cs#L25

calls

https://github.com/dotnet/corefx/blob/master/src/Native/System.Native/pal_runtimeinformation.cpp#L16

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/PlatformApis.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            if (isWindows)
+            {
+                return Platform.Windows;
+            }
+            else
+            {
+                try
+                {
+                    var uname = NativeMethods.Unix.GetUname();
```
eerhardt: Do we support non-windows on NET451?  Is this for Mono?  Do we support Mono?

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/NativeMethods.Unix.cs
```diff
@@ -0,0 +1,33 @@
...
+                    }
+                }
+                catch (Exception ex)
+                {
+                    throw new PlatformNotSupportedException("Error reading Unix name", ex);
+                }
+                throw new PlatformNotSupportedException("Unknown error reading Unix name");
+            }
+
+            [DllImport("libc")]
```
eerhardt: We spent a lot of time in CoreFX not directly DllImport'ing native functions whose signatures could change from platform to platform (i.e. C++ 'int' could return 64-bit int on some platforms).  IMO - we shouldn't be DllImporting native functions here for the same reasons.

@sokket @nguerrera @stephentoub

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/PlatformApis.cs
```diff
@@ -0,0 +1,169 @@
...
+                case Platform.Linux:
+                    return GetDistroId() ?? "Linux";
+                case Platform.Darwin:
+                    return "Mac OS X";
+                default:
+                    return "Unknown";
+            }
+        }
+
+        public static string GetOSVersion()
```
eerhardt: I think a lot of these method could be replaced with `System.Runtime.InteropServices.RuntimeInformation`

src/Microsoft.Extensions.PlatformAbstractions.Internal/project.json
```diff
@@ -0,0 +1,24 @@
+{
```
eerhardt: I don't think we should be adding a new assembly.  Instead can't this code all go inside the existing `Microsoft.DotNet.InternalAbstractions` project?

That way we don't have to ship a new NuGet package (which it doesn't look like you are publishing here), sign another assembly, etc.

eerhardt:
I really like having less dependencies.  :+1: 

I have some questions on the implementation, and if it is necessary to create a completely new assembly.  If we are going to be creating this new assembly, we should probably be publishing it as a NuGet package since our current NuGet packages depend on it.

src/Microsoft.Extensions.PlatformAbstractions.Internal/Platform.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.PlatformAbstractions.Internal
+{
+    public enum Platform
```
eerhardt: Can this be replaced with `System.Runtime.InteropServices.OSPlatform`?

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/NativeMethods.Unix.cs
pranavkm: There's an issue with the InteropServices package in net4x where the package fails to work at runtime unless the application explicitly includes the runtime package. https://github.com/dotnet/corefx/issues/7863 fixes it for RTM.

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/PlatformApis.cs
pranavkm: No idea - I copied the code over from https://github.com/aspnet/PlatformAbstractions/blob/release/src/Microsoft.Extensions.PlatformAbstractions/DefaultRuntimeEnvironment.cs as is.

src/Microsoft.Extensions.PlatformAbstractions.Internal/project.json
pranavkm: Sure

src/Microsoft.Extensions.PlatformAbstractions.Internal/Native/NativeMethods.Unix.cs
davidfowl: /cc @ericstj If we had to port it to RC2 would it be a huge task?

src/Microsoft.DotNet.InternalAbstractions/Native/NativeMethods.Darwin.cs
```diff
@@ -0,0 +1,52 @@
...
+                    throw new PlatformNotSupportedException("Error reading Darwin Kernel Version", ex);
+                }
+                throw new PlatformNotSupportedException("Unknown error reading Darwin Kernel Version");
+            }
+
+            [DllImport("libc")]
+            private unsafe static extern int sysctl(
+                int* name,
+                uint namelen,
+                byte* oldp,
```
stephentoub: size_t is not guaranteed to equal the size of a pointer, only that it's at least that big.

src/Microsoft.DotNet.InternalAbstractions/Native/NativeMethods.Unix.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Microsoft.DotNet.InternalAbstractions.Native
+{
+    internal static partial class NativeMethods
+    {
+        public static class Unix
+        {
+            public unsafe static string GetUname()
+            {
+                // Utsname shouldn't be larger than 2K
+                var buf = stackalloc byte[2048];
```
stephentoub: How do you know this is sufficient space for any platform we might be targeting?  How do you know that padding or header magic or such things doesn't influence where the target field is aligned in the buffer?  This might be fine on the platforms we're currently targeting, but it's risky.

src/Microsoft.DotNet.InternalAbstractions/Native/NativeMethods.Windows.cs
```diff
@@ -0,0 +1,40 @@
...
+                internal uint dwMinorVersion;
+                internal uint dwBuildNumber;
+                internal uint dwPlatformId;
+                [MarshalAs(UnmanagedType.ByValTStr, SizeConst = 128)]
+                internal string szCSDVersion;
+            }
+
+            // This call avoids the shimming Windows does to report old versions
+            [DllImport("ntdll")]
+            private static extern int RtlGetVersion(out RTL_OSVERSIONINFOEX lpVersionInformation);
```
stephentoub: Can we just use RuntimeInformation instead of all of these DllImports?  If there's a problem with RuntimeInformation, we should fix that so it can be usable in these situations.

cc: @Priya91, @weshaggard

src/Microsoft.DotNet.InternalAbstractions/ApplicationEnvironment.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
eerhardt: Copyright headers are needed on all the new files.

src/Microsoft.DotNet.InternalAbstractions/Native/NativeMethods.Unix.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.DotNet.InternalAbstractions.Native
+{
+    internal static partial class NativeMethods
+    {
+        public static class Unix
+        {
+            public unsafe static string GetUname()
```
eerhardt: This method only appears to be called from #if NET451.  If we need to keep it, can it just be compiled for NET451?

src/Microsoft.DotNet.InternalAbstractions/Native/NativeMethods.Windows.cs
pranavkm: Per https://github.com/dotnet/cli/pull/2747#discussion_r61533348, there's an issue with the package in RC2 where it doesn't pull in the runtime dependencies for net45x. Since a fair bit of code is meant to run in desktop (e.g. Testing.Abstractions in xunit runner, DependencyContext in web applications targeting desktop, tools calling Cli.Utils) - I'd rather not change what currently works. We should replace it for RTM where the issue with the RuntimeInformation package has been addressed.

eerhardt: @pranavkm Can you open an RTM issue for this?

src/Microsoft.Extensions.Testing.Abstractions/project.json
```diff
@@ -14,6 +14,7 @@
     "Newtonsoft.Json": "7.0.1",
     "Microsoft.DotNet.ProjectModel": "1.0.0-*",
     "Microsoft.Extensions.Logging.Abstractions": "1.0.0-rc2-20581",
+    "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc2-20581",
```
pranavkm: We need to recompile `dotnet-test-xunit against this build of Testing.Abstractions before we can remove this. Not having this dependency causes loader exceptions. I have a similar change for Logging.Abstractions https://github.com/dotnet/cli/pull/2732/files and I'll address those both once these changes go in.

src/Microsoft.DotNet.InternalAbstractions/Native/NativeMethods.Windows.cs
pranavkm: https://github.com/dotnet/corefx/issues/7863

eerhardt: I meant a CLI issue to take all this out and use RuntimeInformation instead.

pranavkm: Oops, pasted in the wrong link - https://github.com/dotnet/cli/issues/2754

pranavkm:
@stephentoub \ @eerhardt  Are we ok with taking in the native code as is for now? None of it is new and https://github.com/dotnet/cli/issues/2754 should address this for RTM by switching to RuntimeInformation. (although we would need to figure out how to get the OSVersion since that's not exposed by `RuntimeInformation`)

stephentoub:
> Are we ok with taking in the native code as is for now?

Yes, that's fine by me.  It's not new, we know it works on the platforms we're currently targeting, and we plan to replace it.  Thanks for checking.

eerhardt:
I'm fine with it for RC2.  We've been living with it this long and are just moving the code.

pranavkm:

pranavkm:

Petermarcu:
Lets do what we can to get this in today!

pranavkm:
Yup, waiting for the Windows build to pass

pranavkm:

pranavkm:

pranavkm:

src/Microsoft.DotNet.InternalAbstractions/Platform.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
eerhardt: This should be MIT license

src/Microsoft.DotNet.InternalAbstractions/RuntimeEnvironment.cs
```diff
@@ -0,0 +1,104 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
eerhardt: This should be MIT license.

eerhardt:
:shipit: once fixing the copyright header and CI passes.

PR #2749, opened 2016-04-29
Add commit hash to dotnet

schellap:
Helps in debugging and finding symbols etc. quickly.

@gkhanna79 PTAL.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2749)


gkhanna79:
LGTM

anurse:
Could we embed this and the version in the same way that CoreCLR is now doing (i.e. an SCCS ID: https://en.wikipedia.org/wiki/Source_Code_Control_System)? Then I can run `what dotnet` and it will tell me the version and commit hash :)

schellap:
@anurse, for RTM: https://github.com/dotnet/cli/issues/2761

schellap:
That would be super useful, btw!

PR #2769, opened 2016-04-29
Fixing BuiltInCommandTests so it doesn't fail sporatically.

### LINKED ISSUE #2768: BuiltInCommandTests is flaky because it changes Console.Out and expects no one else to write to it
## Steps to reproduce

1. Run Microsoft.DotNet.Cli.Utils.Tests on a Linux box

## Expected  behavior

The test should pass always

## Actual behavior

It fails sometimes with an error like:

```
2016-04-29T18:47:26.384Z:     Microsoft.DotNet.Cli.Utils.BuiltInCommandTests.TestOnOutputLines [FAIL]
2016-04-29T18:47:26.384Z:       System.AggregateException : One or more errors occurred. (Assert.Equal() Failure
2016-04-29T18:47:26.384Z:                 ↓ (pos 0)
2016-04-29T18:47:26.385Z:       Expected: firstsecond
2016-04-29T18:47:26.386Z:       Actual:       Microsoft.DotNet.Cli.Utils.Tests.Give···
2016-04-29T18:47:26.386Z:                 ↑ (pos 0))
```

## Environment data
`dotnet --info` output:

N/A

## Notes

Probably the right thing to do is to mock out the Console during the tests.  For now, I am using XUnit Collections so the tests don't fail.

eerhardt:
The BuiltInCommandTests sets the current Console.Out and Console.Error, which causes the test to fail if some other test is running and writes to the console at the same time.

Fix #2768

@Sridhar-MS @pakrym

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2769)


pakrym:
Would it really help? What if xunit logs something?

eerhardt:
Do you think I should just put the assembly level attribute for now?

[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]

pakrym:
I think we would need to.

eerhardt:
OK - I had time to fixed this for real today.  PTAL.

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
```diff
@@ -100,6 +108,29 @@ public ICommand OnErrorLine(Action<string> handler)
             return this;
         }
 
+        private class BuiltInCommandEnvironment : IBuiltInCommandEnvironment
```
pakrym: `IBuiltInCommandConsole`? Doesn't have anything except console yet.

eerhardt:
@brthor - can you take a look?

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
eerhardt: It is going to have Get/Set CurrentDirectory once https://github.com/dotnet/cli/pull/2792 is merged.

brthor: Would it make sense to move this over to our InternalAbstractions?

I like the console abstraction, and think it could be useful in other places.

test/Microsoft.DotNet.Cli.Utils.Tests/project.json
```diff
@@ -1,7 +1,8 @@
 {
   "version": "1.0.0-*",
   "compilationOptions": {
-    "emitEntryPoint": true
+    "emitEntryPoint": true,
+    "keyFile": "../../tools/Key.snk"
```
brthor: I'm curious why the test needs this?

eerhardt: In order for InternalsVisibleTo to work.  The product assembly is strong named.  In order for the product to have InternalsVisibleTo a different assembly, that assembly must be strong named as well.

brthor:
Small Factoring comment, but this is pretty nice. 

As an aside @krwq has been telling me for a little while that we shouldn't use console in our tests, and that xunit has some abstractions. This may be the best long term fix.

https://xunit.github.io/docs/capturing-output.html

:shipit:

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
eerhardt: I think we could move it lower in the future, if we needed to.  But for now, I think it being internal and staying close to where it is used makes sense.

PR #2774, opened 2016-04-29
Fixing error where lock could not be taken on build folder

sokket:
cc @piotrpMSFT 

Ran this and it was able to take the lock, this should fix the 404 blob not found problem

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2774)

### DIFF
--- a/scripts/dotnet-cli-build/PublishTargets.cs
+++ b/scripts/dotnet-cli-build/PublishTargets.cs
@@ -60,7 +60,9 @@ namespace Microsoft.DotNet.Cli.Build
             {
                 string targetContainer = $"{Channel}/Binaries/Latest/";
                 string targetVersionFile = $"{targetContainer}{CliNuGetVersion}";
-                string leaseId = AzurePublisherTool.AcquireLeaseOnBlob(targetContainer);
+                string semaphoreBlob = $"{Channel}/Binaries/publishSemaphore";
+                AzurePublisherTool.CreateBlobIfNotExists(semaphoreBlob);
+                string leaseId = AzurePublisherTool.AcquireLeaseOnBlob(semaphoreBlob);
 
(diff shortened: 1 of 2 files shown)


sokket:
@dotnet-bot test this please

sokket:
@dotnet-bot test  OSX x64 Release Build

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -60,7 +60,9 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
             {
                 string targetContainer = $"{Channel}/Binaries/Latest/";
                 string targetVersionFile = $"{targetContainer}{CliNuGetVersion}";
-                string leaseId = AzurePublisherTool.AcquireLeaseOnBlob(targetContainer);
+                string semaphoreBlob = $"{Channel}/Binaries/publishSemaphore";
```
Sridhar-MS: Should we have something similar for `{Channel}/Installers/publishSemaphore`

sokket: That is published in a different code-path, is it not?

Sridhar-MS: Are we not 'Finalizing' installer artifacts too..

sokket: I don't believe in this code path; we should only be touching the binaries/Latest folder here. Adding a semaphore for the installer binaries would be out-of-scope for this PR

sokket:
@dotnet-bot test this please

sokket:

sokket:
Any comments on this or is it good to go when green?

piotrpMSFT:
:shipit:

sokket:
Thanks! Merging

PR #2781, opened 2016-04-29
fixes #2760

### LINKED ISSUE #2760: Can't run console app built for win10-x64 (probably other rids as well)
dotnet version: 1.0.0-rc2-002520

## Steps to reproduce
1. New console app
2. Remove type: platform from dependency
3. Add runtimes: {"win10-x64" : {}}
4. dotnet build
5 Output should be in bin\debug\netcoreapp10\win10-x64
6. run the built executable

See the following error
Could not resolve coreclr path

brthor:
Fixes #2760 

Running through the repro steps there you'll now see: 

```
Expected coreclr library not found in package graph. Please try running restore again.
```
cc @krwq @Sridhar-MS @eerhardt

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2781)


eerhardt:
i think this looks ok.  Can you describe the issue here and why this fixes it?

brthor:
@eerhardt This is not yet done. There are some deeper issues here this doesn't fix all the way.

brthor:
Please review this @eerhardt @Sridhar-MS

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -53,6 +53,23 @@ public void MakeCompilationOutputRunnable()
             ExportRuntimeAssets();
         }
 
+        private void VerifyCoreclrPresenceInPackageGraph()
```
eerhardt: nit: the capitalization of Coreclr doesn't match the way you have it below in LibCoreClrBinaryNames.

brthor: I feel that Coreclr is the right thing here, but don't actually know, what do you think?

eerhardt: Looking at the Constants.cs file, it already uses CoreClr.

My opinion is since CLR is an acronym, it is separate from "core", thus in PascalCasing would require a new captial letter.

---
In reply to: [61924689](https://github.com/dotnet/cli/pull/2781#discussion_r61924689) [](ancestors = 61924689)

brthor: Will make the change 👍

TestAssets/TestProjects/DependencyChangeTest/StandaloneApp_Portable/project.json
```diff
@@ -0,0 +1,22 @@
+{
```
eerhardt: This test asset doesn't appear to be used in a test.

brthor: good catch, will remove. I had a test but it ended up being unnecessary.

brthor:
PR feedback applied

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -56,6 +56,23 @@ public void MakeCompilationOutputRunnable()
...
+                .SelectMany(e => e.NativeLibraryGroups)
+                .SelectMany(g => g.Assets)
+                .Select(a => a.FileName)
+                .Where(f => Constants.LibCoreClrBinaryNames.Contains(f))
+                .Any();
+
+            // coreclr should be present for standalone apps
+            if (! isCoreClrPresent)
+            {
+                throw new InvalidOperationException("Expected coreclr library not found in package graph. Please try running restore again.");
```
eerhardt: This could be mis-leading for *nix since the library is named libcoreclr.dylib/so.  I'm not sure how to make it better though...

eerhardt: "running `dotnet restore`" ??  Or do you think `running restore` is explicit enough?

We say `dotnet restore` in our other error messages.

brthor: Don't necessarily agree with this. The error doesn't attempt to point out an individual file, only notify the user that something is going wrong with coreclr, and they need to re-run restore.

eerhardt:
:shipit:

src/Microsoft.DotNet.Compiler.Common/Executable.cs
brthor: yeah `dotnet restore` is better. A newbie might be confused here

brthor: done.

piotrpMSFT:
This is approved once it turns green.

brthor:
Sweet, latest code added was a simple string change so not expecting any issues there.

Sridhar-MS:
:shipit:

piotrpMSFT:
Approved offline. Merging.

Issue #2408, opened 2016-04-11
Stop using corehost to activate apps


schellap:
## Steps to reproduce
`dotnet test` uses `corehost` to activate test xunit.dll

## Expected  behavior
`dotnet test` always uses the muxer to activate xunit.dll

## Actual behavior
corehost in `sdk` is being used.

## Environment data
`dotnet --info` output:

brthor:
I thought we decided to leave this in for RC2?

The issue was resolving hostpolicy.dll from the nuget cache for standalone apps when we decided.

Do we still need to do that or no longer required?

schellap:
I've added this for RTM. @gkhanna79

brthor:
Going to work on this tomorrow, @schellap do you expect that the host should handle this scenario fine?

schellap:
@brthor, we are RTM-ing. Let me know if you find anything busted. It should work as is.

brthor:
Fixed in #3346

PR #2790, opened 2016-04-30
Resolve hostpolicy out of package graph for servicing

schellap:
Resolve hostpolicy version out of the FX or App deps files and look up servicing folders for that version.

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2790)

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -161,7 +161,7 @@ bool hostpolicy_exists_in_svc(const pal::string_t& version, pal::string_t* resol
 }
 
 /**
- * Given path to app binary, say app.dll, retrieve the app.deps.json.
+ * Given path to app binary, say app.dll or app.exe, retrieve the app.deps.json.
  */
 pal::string_t get_deps_from_app_binary(const pal::string_t& app)
 {
@@ -212,12 +212,12 @@ bool resolve_hostpolicy_dir_from_probe_paths(const pal::string_t& version, const
 }
 
 /**
- * Given FX location, app dll and specified --depsfile, return deps that contains hostpolicy.dll
+ * Given FX location, app binary and specified --depsfile, return deps that contains hostpolicy.dll
  */
 pal::string_t get_deps_file(
     const pal::string_t& fx_dir,
     const pal::string_t& app_candidate,
-    const pal::string_t& deps_file,
+    const pal::string_t& specified_deps_file,
     const runtime_config_t& config)
 {
     if (config.get_portable())
@@ -228,25 +228,25 @@ pal::string_t get_deps_file(
     else
     {
         // Standalone app's hostpolicy is from specified deps or from app deps.
-        return !deps_file.empty() ? deps_file : get_deps_from_app_binary(app_candidate);
+        return !specified_deps_file.empty() ? specified_deps_file : get_deps_from_app_binary(app_candidate);
     }
 }
 
 /**
- * Given own location, FX location, app dll and specified --depsfile and probe paths
+ * Given own location, FX location, app binary and specified --depsfile and probe paths
  *     return location that is expected to contain hostpolicy
  */
 bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
     const pal::string_t& own_dir,
     const pal::string_t& fx_dir,
     const pal::string_t& app_candidate,
-    const pal::string_t& deps_file,
+    const pal::string_t& specified_deps_file,
     const std::vector<pal::string_t>& probe_realpaths,
     const runtime_config_t& config,
     pal::string_t* impl_dir)
 {
     // Obtain deps file for the given configuration.
-    pal::string_t resolved_deps = get_deps_file(fx_dir, app_candidate, deps_file, config);
+    pal::string_t resolved_deps = get_deps_file(fx_dir, app_candidate, specified_deps_file, config);
 
     // Resolve hostpolicy version out of the deps file.
     pal::string_t version = resolve_hostpolicy_version_from_deps(resolved_deps);
@@ -276,10 +276,13 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
     else
     {
         // Standalone apps can be activated by muxer or by standalone host or "corehost"
+        // 1. When activated with dotnet.exe or corehost.exe, check for hostpolicy in the deps dir or
+        //    app dir.
+        // 2. When activated with app.exe, the standalone host, check own directory.
         assert(mode == host_mode_t::muxer || mode == host_mode_t::standalone || mode == host_mode_t::split_fx);
-        expected = (mode == host_mode_t::muxer)
-            ? get_directory(deps_file.empty() ? app_candidate : deps_file)
-            : own_dir;
+        expected = (mode == host_mode_t::standalone)
+            ? own_dir
+            : get_directory(specified_deps_file.empty() ? app_candidate : specified_deps_file);
     }
 
     // Check if hostpolicy exists in "expected" directory.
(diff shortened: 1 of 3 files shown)


schellap:
@gkhanna79 PTAL.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,310 @@
...
+    return to_hostpolicy_package_dir(svc_dir, version, resolved_dir);
+}
+
+/**
+ * Given path to app binary, say app.dll, retrieve the app.deps.json.
+ */
+pal::string_t get_deps_from_app_binary(const pal::string_t& app)
+{
+    assert(app.find(DIR_SEPARATOR) != pal::string_t::npos);
+    assert(ends_with(app, _X(".dll"), false) || ends_with(app, _X(".exe"), false));
```
gkhanna79: Why do we have the EXE check if the path will be for the DLL?

schellap: Fsc, csc are .exe cc @brthor

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,310 @@
...
+{
+    if (config.get_portable())
+    {
+        // Portable app's hostpolicy is resolved from FX deps
+        return fx_dir + DIR_SEPARATOR + config.get_fx_name() + _X(".deps.json");
+    }
+    else
+    {
+        // Standalone app's hostpolicy is from specified deps or from app deps.
+        return !deps_file.empty() ? deps_file : get_deps_from_app_binary(app_candidate);
```
gkhanna79: Please rename "deps_file" to "specified_deps_file" to indicate that it was passed at the command line by the user.

src/corehost/cli/fxr/fx_muxer.cpp
gkhanna79: In that case, please update the comment.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,310 @@
...
+            handle_missing_framework_error(config.get_fx_name(), config.get_fx_version(), fx_dir);
+            return false;
+        }
+        
+        expected = fx_dir;
+    }
+    else
+    {
+        // Standalone apps can be activated by muxer or by standalone host or "corehost"
+        assert(mode == host_mode_t::muxer || mode == host_mode_t::standalone || mode == host_mode_t::split_fx);
```
gkhanna79: Why would a standalone app be having mode of muxer even if "dotnet.exe" (global host) was used to launch its .DLL?

gkhanna79: Can you elaborate on the "corehost"/split_fx mode?

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,310 @@
...
+    return false;
+}
+
+pal::string_t fx_muxer_t::resolve_fx_dir(host_mode_t mode, const pal::string_t& own_dir, const runtime_config_t& config)
 {
-    trace::verbose(_X("--- Resolving FX directory from muxer dir [%s]"), muxer_dir.c_str());
-    const auto fx_name = runtime->get_fx_name();
-    const auto fx_ver = runtime->get_fx_version();
-    const auto patch_roll_fwd = runtime->get_patch_roll_fwd();
+    if (mode == host_mode_t::split_fx)
```
gkhanna79: No standalone mode check?

src/corehost/cli/fxr/fx_muxer.cpp
schellap: (Runtime) Config is standalone or  portable, but activation mode is using the "muxer".

Earlier, the `corehost.exe` was being used to activate xunit and the standalone app's build output did not carry a hostpolicy.dll. The expectation for not carrying `hostpolicy.dll` post build was for the host to resolve `hostpolicy` out of the additional probing paths. I tried this out now, and is no longer the case.

Ideally, this should read as:
`pal::string_t expected = (mode == standalone) ? own_dir : get_directory(deps_file.empty() ? app_candidate : deps_file)`

irrespective of whether the activation was using corehost or dotnet.exe from FX location or the muxer.

I'll make the change, if you are on the same page.

schellap: Note however, we do probe additional probing paths if we can't find in the expected directory.

schellap: @brthor, for xunit activation of a standalone app, we expect `dotnet build` to copy `hostpolicy.dll` to the standalone directory's build output similar to how it copies `hostfxr.dll` there.

Is this how you expect this to work as well for RTM?

src/corehost/cli/fxr/fx_muxer.cpp
brthor: Yes, fsc.exe was the original impetus here.

src/corehost/cli/fxr/fx_muxer.cpp
brthor: I'm no longer certain how exactly this will work in RTM 😄

We should have some discussion around this.

src/corehost/cli/fxr/fx_muxer.cpp
schellap: Will add an assert here.

src/corehost/cli/fxr/fx_muxer.cpp
gkhanna79: @brthor and @schellap Can you please syncup on this?

gkhanna79: > we do probe additional probing paths if we can't find in the expected directory.

I would expect that the probe order for this is no different from probe order of any other nuget package. Correct?

schellap: > I would expect that the probe order for this is no different from probe order of any other nuget package. Correct?

Yes. Using the version obtained from the deps file.

schellap: @piotrpMSFT, Could you please help clarify the RTM solution for testing standalone apps? i.e., XUnit is activated with the Muxer like:

`dotnet exec --additionalprobingpath C:\ --depsfile a.deps xunit.dll standalone.dll`

and the muxer figures out `hostpolicy.dll` from additional probing path.

But for RC2, `dotnet build` is copying it next to the `standalone.dll`?

src/corehost/cli/fxr/fx_muxer.cpp
schellap: Fixed.

src/corehost/cli/fxr/fx_muxer.cpp
schellap: Fixed.

schellap:
@gkhanna79, let me know if you are good on the changes post PR so I can merge.

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -14,12 +14,317 @@
...
+        {
+            handle_missing_framework_error(config.get_fx_name(), config.get_fx_version(), fx_dir);
+            return false;
+        }
+        
+        expected = fx_dir;
+    }
+    else
+    {
+        // Standalone apps can be activated by muxer or by standalone host or "corehost"
```
gkhanna79: Is there a bug to remove the corehost support?

Otherwise, LGTM.

gkhanna79: Did you try out the servicing scenario for hostpolicy?

schellap: Yes. Of course, we're shipping soon right? Tried portable, standalone build/run, xunit, published and unpublished with muxer, corehost and standalone where applicable.

schellap: Oh, music store as well on windows and on osx (for Unix)

To answer the other question: yes, here: https://github.com/dotnet/cli/issues/2408

PR #2792, opened 2016-04-30
Execute 'csc' with working directory set to the project directory.

### LINKED ISSUE #2710: Relative paths in compilationOptions/additionalArguments don't work as expected when running build from a different folder
## Steps to reproduce
Given the following folder structure
```
- global.ruleset
- src
| - project
  | - project.json
```

With `project.json` including:
```
"compilationOptions": {
    "emitEntryPoint": true,
    "debugType": "portable",
    "additionalArguments": [ "/ruleset:../../global.ruleset" ]
  },
```

## Expected  behavior
Running `dotnet build` from the `\src\project` folder compiles the project.
Running `dotnet build .\src\project` from root folder compiles the project.

## Actual behavior
Running `dotnet build` from the `\src\project` folder compiles the project.
Running `dotnet build .\src\project` from root folder fails with error:
> C:\some\path\src\project\error CS8035: Error reading ruleset file C:\some\path\src\\..\\..\global.ruleset - File not found.

## Environment data
`dotnet --info` output:

```

Product Information:
 Version:     1.0.0-rc2-002498
 Commit Sha:  55e561bfb3

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

eerhardt:
When using a ruleset with a relative path in buildOptions, csc can't
find the file because it is not working in the same directory as the
project.

Fix #2710

@brthor @cdmihai

---
This change is [Reviewable](https://reviewable.io/reviews/dotnet/cli/2792)

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
@@ -20,6 +20,7 @@ namespace Microsoft.DotNet.Cli.Utils
         private readonly Func<string[], int> _builtInCommand;
         private readonly StreamForwarder _stdOut;
         private readonly StreamForwarder _stdErr;
+        private string _workingDirectory;
 
         public string CommandName { get; }
         public string CommandArgs => string.Join(" ", _commandArgs);
@@ -38,6 +39,7 @@ namespace Microsoft.DotNet.Cli.Utils
         {
             TextWriter originalConsoleOut = Console.Out;
             TextWriter originalConsoleError = Console.Error;
+            string originalWorkingDirectory = Directory.GetCurrentDirectory();
 
             try
             {
@@ -51,6 +53,11 @@ namespace Microsoft.DotNet.Cli.Utils
 
                     // Reset the Reporters to the new Console Out and Error.
                     Reporter.Reset();
+
+                    if (!string.IsNullOrEmpty(_workingDirectory))
+                    {
+                        Directory.SetCurrentDirectory(_workingDirectory);
+                    }
 
                     var taskOut = _stdOut.BeginRead(new StreamReader(outStream));
                     var taskErr = _stdErr.BeginRead(new StreamReader(errorStream));
@@ -71,6 +78,7 @@ namespace Microsoft.DotNet.Cli.Utils
             {
                 Console.SetOut(originalConsoleOut);
                 Console.SetError(originalConsoleError);
+                Directory.SetCurrentDirectory(originalWorkingDirectory);
 
                 Reporter.Reset();
             }
@@ -96,6 +104,13 @@ namespace Microsoft.DotNet.Cli.Utils
             }
 
             _stdErr.ForwardTo(writeLine: handler);
+
+            return this;
+        }
+
+        public ICommand WorkingDirectory(string workingDirectory)
+        {
+            _workingDirectory = workingDirectory;
 
             return this;
         }
@@ -129,11 +144,6 @@ namespace Microsoft.DotNet.Cli.Utils
         }
 
         public ICommand ForwardStdOut(TextWriter to = null, bool onlyIfVerbose = false, bool ansiPassThrough = true)
-        {
-            throw new NotImplementedException();
-        }
-
-        public ICommand WorkingDirectory(string projectDirectory)
         {
             throw new NotImplementedException();
         }
--- a/src/dotnet/commands/dotnet-compile-csc/Program.cs
+++ b/src/dotnet/commands/dotnet-compile-csc/Program.cs
@@ -84,6 +84,7 @@ namespace Microsoft.DotNet.Tools.Compiler.Csc
 
                 // Execute CSC!
                 var result = RunCsc(new string[] { $"-noconfig", "@" + $"{rsp}" })
+                    .WorkingDirectory(Directory.GetCurrentDirectory())
                     .ForwardStdErr()
                     .ForwardStdOut()
                     .Execute();
--- a/TestAssets/TestProjects/TestRuleSet/TestLibraryWithRuleSet/Violations.cs
+++ b/TestAssets/TestProjects/TestRuleSet/TestLibraryWithRuleSet/Violations.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+
+namespace TestLibrary
+{
+    public class AttributeWithoutUsage : Attribute
+    {
+    }
+
+    public class ClassWithUndisposedStream
+    {
+        private Stream _nonDisposedStream = new MemoryStream();
+
+        public ClassWithUndisposedStream()
+        {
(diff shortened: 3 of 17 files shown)


pakrym:
Are you sure running `csc` in the right directory is a fix, we could just generate full path of `rullset` file and pass it.

eerhardt:
> we could just generate full path of rullset file and pass it.

@agocke says in the issue:

>Also, I should note that additionalArguments are, by design, applied verbatim. Parsing is explicitly not applied to the compiler arguments, since one of the things it's supposed to support is arguments and argument formats that dotnet doesn't necessarily understand.

pakrym:
I think if analyzer support is first level `ruleset` should be too. But in case if we want to leave it as additional argument your fix is good.

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
```diff
@@ -52,6 +54,11 @@ public CommandResult Execute()
                     // Reset the Reporters to the new Console Out and Error.
                     Reporter.Reset();
 
+                    if (!string.IsNullOrEmpty(_workingDirectory))
+                    {
+                        Directory.SetCurrentDirectory(_workingDirectory);
```
pakrym: Why? We are executing csc out of proc anyway, why mutate environment inside `dotnet.dll`?

src/dotnet/commands/dotnet-compile-csc/Program.cs
```diff
@@ -84,6 +84,7 @@ public static int Run(string[] args)
 
                 // Execute CSC!
                 var result = RunCsc(new string[] { $"-noconfig", "@" + $"{rsp}" })
+                    .WorkingDirectory(Directory.GetCurrentDirectory())
```
cdmihai: Currently I only see BuiltInCommand implementing WorkingDirectory, but RunCsc returns a Command. Shouldn't you add WorkingDirectory to ICommand, and then have Command and BuiltInCommand implement it? 
Maybe have the common code in an AbstractCommand

eerhardt: There is an existing Command class that only knows how to spin up processes:

https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/Command.cs

It and BuiltInCommand have very little common implementation.

WorkingDirectory is already on ICommand:  https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/ICommand.cs#L13.  

The two classes only share the interface.

cdmihai: Aha, did not notice ICommand already has that :) My bad.

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
eerhardt: This is needed because the only information passed into `compile-csc` is:  a string[] args and the current working directory.  `compile-csc` currently has no knowledge of the project.json or the project context information at all.

In order for `compile-csc` to launch `csc` in the directory that contains the project.json file, it needs to know which directory the project.json file is in.  This was the way I decided to pass that info into `compile-csc`.  Is there a better way?

agocke: Personally, I would have created an additional "known" argument to `compile-csc`, like `temp-out`, that takes the working directory for `csc` as the argument value. This would avoid manipulating the current exe's working directory.

agocke:
I think this is a fix for the broader issue of supporting relative paths in `additionalArguments`.

However, I think for `ruleset` in particular, we could add that as a known argument in the `analyzerOptions` section of `compilerOptions`.

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
eerhardt: I can totally do that.  What is the general approach for designing these arguments?  Should the name match? "--build-base-path"?  Also, how do you communicate these changes to the other language teams? i.e. F#.

@enricosada @KevinRansom

eerhardt:
> However, I think for ruleset in particular, we could add that as a known argument in the analyzerOptions section of compilerOptions.

How far do you go with things like this?  What if a new language comes along that doesn't support analyzers and rulesets?  For example, does Python?  It feels like `dotnet build` shouldn't know about every languages special parameters/options.  Instead, each language should have their own options.  If some languages happen to have common options, so be it. But that shouldn't mean `dotnet build` needs to know about them.

agocke:
@eerhardt I agree. My bar has been that I would only accept things that are language neutral and generally useful.

A rule set file, for Roslyn and presumably for other analyzer systems, is basically just a text file that lists configuration options for the analyzers. Since we don't actually check anything about the file except its path, it seems like a reasonable addition as a language-neutral analyzer hook point.

I could imagine every language making use of this, possibly in different ways, and simply exposing this path as the primary integration point.

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
eerhardt: Actually thinking more on this, I don't think this approach would work.  Why would a variable like `temp-out` mean that should be the working directory for `csc`?  Shouldn't it be named something else?  Like `project-directory` or `compiler-working-directory`?

eerhardt:
@agocke - I think moving the "ruleset" to a different location in project.json is a different issue. If you feel strongly about this, can you open a new issue?

The issue I'm fixing with this change (#2710) is explicitly for any relative paths inside `additionalArguments` aren't working. `ruleset` just happens to be one of them.  (OK, it might be the only one right now, but that doesn't mean some other language wouldn't want relative paths in `additionalArguments`.)

agocke:
@eerhardt Sure. I don't disagree that this is a good fix.

src/Microsoft.DotNet.Cli.Utils/BuiltInCommand.cs
eerhardt: I see my mistake.  You were just saying "like `temp-out`".  "It could be a new argument like this existing one."

I can go either way on this one.  But a part of me kind of thinks this is the right fix.  We are invoking `dotnet-compile-{language}` and instead of passing in a `this-is-your-working-directory` argument, we actually set the working directory. It's kind of what WorkingDirectory is there for.

Thoughts?

agocke: Up to you. My original design was to allow practically any piece of the command tree to be easily swapped out, so someone could call "dotnet-compile-csc" from bash and they would just have to pass the right arguments to link it up. Basically, there would be no friction in switching languages aside from finding the right arguments to be passed, which could be just copy-pasted from a `dotnet -v` call.

Adding more "types" of options or more mechanisms for delivering those options seems to make things more complicated, IMHO, but we've significantly diverged in overall command structure from where we originally were, so it's possible this approach is more consistent with the overall tool's current design.

eerhardt:
Approved in email.

PR #2808, opened 2016-05-02
TestPlatform ObjectModel convergence

singhsarab:
Changing the dotnet test verb to use TestPlatform ObjectModel for the test objects (For e.g. TestCase, TestResult and TestOutcome)

Moving forward, "dotnet test" verb will be based on TestPlatform (v2.0) designed by Unit Testing team. Hence, making this change to prevent breaking changes in RTM.


davidfowl:
Where is that package **Microsoft.TestPlatform.ObjectModel** coming from?

piotrpMSFT:
Let's get this work into a decent place in a feature branch before merging it into rel/1.0.0. There are a lot of moving pieces here.

ManishJayaswal:
+1 on what @piotrpMSFT said. 
Additionally, do we want "visualstudio" in the testplatform namespace?

PR #2813, opened 2016-05-02
Fixing semaphore locking during build finalization

sokket:
Fixing semaphore locking during build finalization

cc @piotrpMSFT @Sridhar-MS


piotrpMSFT:
:shipit:

This code won't affect any Jenkins-hit codepaths. Please merge once we see that the build script itself is building in Jenkins.

Sridhar-MS:
:shipit:

piotrpMSFT:
verified this builds on Windows as well. Merging as this will impact VSO only and we need to know if there are issues.

Issue #2819, opened 2016-05-02
Adding buildOptions\embed causes resources to be skipped from being embedded


pranavkm:
## Steps to reproduce
1. Clone https://github.com/aspnet/Microsoft.Data.Sqlite/ and checkout e00935cd02226b112b49a5221106e0650c53bf4a
2. Build src/Microsft.Data.Sqlite. Notice that the generated binary has the resource file.
3. Modify the project.json to change `resource` node to `buildOptions/embed/include`. Build it again

## Expected  behavior
The generated binary should include the resx file.

## Actual behavior
The generated binary is missing the resx file.

## Environment data
`dotnet --info` output:
1.0.0-rc2-002543

Specifying the resx file in the embed\include array works.

pranavkm:
cc @ajaybhargavb

pranavkm:
Possibly related: Adding items to `buildOptions\embed\include` does not exclude them. I have to explicitly add items to `buildOptions/compile/exclude`.

ajaybhargavb:
> Adding items to buildOptions\embed\include does not exclude them. I have to explicitly add items to buildOptions/compile/exclude.

I investigated this and it is not related to this issue. Could you please file a new issue?

piotrpMSFT:
@ajaybhargavb I assume you want to get this into RC2. Please move it out to RTM if that's not the case.

pranavkm:
It would be nice to fix this in RC2. Adding a compile `exclude` causes `dotnet build` to succeed, but no binary is produced without this fix. It's not entirely intuitive what is required to fix it.

mellinoe:
I'm hitting this as well. I get no warning or anything, but the resource mentioned in the "embed" item is not included in the compiled binary.

mellinoe:
I see this announcement now: https://github.com/aspnet/Announcements/issues/175. It seems that my syntax is wrong.

ajaybhargavb:
@piotrpMSFT, is there any place in cli repo or the dotnet org that this announcement can be linked to?

piotrpMSFT:
@blackdwarf can you help with communication here?

piotrpMSFT:
Fixed by #2822

PR #2822, opened 2016-05-02
Fix: builtIn defaults are not used when builtIn is not specified

ajaybhargavb:
Issue #2819 

Currently the default BuiltIn values are being ignored if there is no `builtIns` specified in project.json.

@troydai @pranavkm


pranavkm:
:shipit: once @troydai has had a look.

troydai:
:shipit:

piotrpMSFT:
@petermarcu approved offline. Merging.

Issue #2588, opened 2016-04-19
[MSI]: CLI and SharedFx bundle must set up the right registry keys for install detection


Sridhar-MS:
For example VC++ redist bundle sets up the correct registry keys for detection - https://github.com/dotnet/cli/blob/rel/1.0.0/packaging/windows/clisdk/bundle.wxs#L34

The CLI bundle and SharedFx bundle must do something similar. 

Note: CLI and SharedFx are installed SxS. So these registry keys must also be SxS and not overwrite each other.

cc @Petermarcu @joeloff

Sridhar-MS:
@Petermarcu @robmen @joeloff @piotrpMSFT I like to get your feedback on this

Below is the registry I like to setup for CLI bundle detection

![image](https://cloud.githubusercontent.com/assets/9423246/14830226/7085255c-0ba4-11e6-8a25-98352ac9586c.png)

Since the CLI is installed SxS I am planning to set the registry keys in the following way
`HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\dotnet\Setup\$(platform)\CLI\$(cli_nugetversion) Value Install = 1`
Where
`$platform = x64 or x86`
`$cli_nugetversion = nuget version of the CLI (example - 1.0.0-rc2-002501)`

Some points to consider
- Both x64 and x86 installer will set the reg keys in the wow subsystem. This is to enable 32 bit installer to detect x64 or x86 installation.
- Since the CLI nuget version is part of the key any SxS installation will create a new key as shown in the above pic.
- CLI bundle detection logic for a give version example - `1.0.0-rc2-002501`
    - Check for the below reg key 
```xml
    <util:RegistrySearch
       Id="CLI_KeyExists"
       Variable="CLI_KeyExists"
       Root="HKLM"
       Key="SOFTWARE\dotnet\Setup\$(platform)\CLI\1.0.0-rc2-002501"
       Win64="no"
       Result="exists" />
```    
   
One disadvantage I see is that it becomes difficult to do a `greater than` detection. Similar to [here](https://github.com/dotnet/cli/blob/rel/1.0.0/packaging/windows/clisdk/bundle.wxs#L34). I am wondering that since CLI is SxS it does not make sense to do a `greater than` detection.

joeloff:
Have you considered using the OptionalUpdateRegistration in the bundle? See http://wixtoolset.org/documentation/manual/v3/xsd/wix/optionalupdateregistration.html

[...]

PR #2828, opened 2016-05-02
Setup the registry keys for CLI bundle detection.

Sridhar-MS:
For 1.0.0-rc2 release of the CLI/SharedFx only set the current version in the registry. The future releases will also set any backward compatible versions along with the current version.

The spec - https://github.com/dotnet/cli/issues/2588#issuecomment-215912323

Fixes - #2588

cc @piotrpMSFT @krwq @robmen @joeloff

### DIFF
--- a/packaging/windows/clisdk/registrykeys.wxs
+++ b/packaging/windows/clisdk/registrykeys.wxs
@@ -3,23 +3,9 @@
   <?include "Variables.wxi" ?>
   <Fragment>
     <ComponentGroup Id="AuthoredRegistryKeys">
-      <!-- Include the 64-bit registry keys only in the 64-bit installer -->
-      <?if $(var.Platform) = x64?>
-      <Component Id="SetupRegistry_x64" Directory="TARGETDIR" Win64="yes">
-        <RegistryKey Root="HKLM" Key="SOFTWARE\dotnet\Setup">
-          <RegistryValue Action="write" Name="Install" Type="integer" Value="1" KeyPath="yes"/>
-          <RegistryValue Action="write" Name="InstallDir" Type="string" Value="[DOTNETHOME]" />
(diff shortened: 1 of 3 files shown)


Sridhar-MS:
@BillHiebert  FYI we are changing reg keys set by the CLI/SharedFx installers.

packaging/windows/clisdk/registrykeys.wxs
```diff
@@ -3,23 +3,9 @@
...
-      </Component>
-      <?endif?>
-
-      <!-- Always install the 32-bit registry keys and env vars -->
       <Component Id="SetupRegistry_x86" Directory="TARGETDIR" Win64="no">
-        <RegistryKey Root="HKLM" Key="SOFTWARE\dotnet\Setup">
-          <RegistryValue Action="write" Name="Install" Type="integer" Value="1" KeyPath="yes"/>
-          <RegistryValue Action="write" Name="InstallDir" Type="string" Value="[DOTNETHOME]" />
-          <RegistryValue Action="write" Name="Version" Type="string" Value="$(var.Dotnet_ProductVersion)" />
+        <RegistryKey Root="HKLM" Key="SOFTWARE\dotnet\Setup\InstalledVersions\$(var.Platform)\sdk">
```
krwq: check the guids, add a comment to have specific hard-coded guids when supporting older versions so we don't forget it when doing the work

krwq:
LGTM :+1:

joeloff:
:shipit:

packaging/windows/clisdk/registrykeys.wxs
Sridhar-MS: The guids are automatically generated based on the registry value that is set, the documentation is [here](http://wixtoolset.org/documentation/manual/v3/xsd/wix/component.html). And same guid is generated for the same registry value. So we should not be worrying about it.

piotrpMSFT:
@petermarcu approved offline. merging.

Issue #2824, opened 2016-05-02
ASP.NET core (Kestrel) apps blocked from running on .NET 4.6 (S.N.V contains _._ in a non-empty ref folder)


joshfree:
Moving issue from https://github.com/dotnet/corefx/issues/8189 per offline discussion

System.Numerics.Vectors 4.1.1-rc2-24018 contains ``_._`` in a folder that also contains dlls. 

From ``ref/net46``:
![image](https://cloud.githubusercontent.com/assets/6381138/14932324/32c4319a-0e29-11e6-886d-604d93446be2.png)

Normally ``_._`` is added only to allow an empty folder in nupkgs (OPC does not allow completely empty folders)

Having the extra placeholder here causes problems: https://github.com/aspnet/KestrelHttpServer/issues/773

ericstj:
CLI is inconsistent with MSBuild here.  This package works fine in desktop project using MSBuild.

Issue is caused by https://github.com/dotnet/cli/pull/2597, specifically https://github.com/dotnet/cli/blob/4af795b9e8bcd5974d6087b9b4e631eb3e42886d/src/Microsoft.DotNet.ProjectModel/PackageDescription.cs#L39.

Seems like that should be using "All" as opposed to "Any".

In general, we should avoid any special casing of the `_._` to trigger behavior, instead we should just be filtering it out and treating the absence of assets as a trigger.  We don't want that file to appear as any sort of *magic* other than an ignored asset similar to any other random file that might show up in the lib/ref folder.

We're still fixing this in CoreFx since we didn't need the file.

davidfowl:
> In general, we should avoid any special casing of the _._ to trigger behavior, instead we should just be filtering it out and treating the absence of assets as a trigger. We don't want that file to appear as any sort of magic other than an ignored asset similar to any other random file that might show up in the lib/ref folder.

That won't work. We need to know if there is an empty folder at the very least. The check could be something like, if there's a folder and no compilation assemblies but it's not worth that complexity for rc2.

Also, MSBuild already traverses the assembly dependency graph so it's no concerned with packages at all it doesn't need to care about this.

The CLI is trying to fill a gap where the **System.* packages with placeholders on certain frameworks** = use the framework assembly instead since the packages can't declare the framework assembly (as I'm sure you know).

Petermarcu:
Fixed by https://github.com/dotnet/cli/pull/2834

PR #2834, opened 2016-05-03
Fix placeholder resolution logic

davidfowl:
System.Numerics.Vector has a placeholder next to a .dll in the lib folder. The scenario isn't really valid but take it into account and make the placeholder resolution logic a little more robust. A package is considered to have a compile time placeholder if it has files and *all* files are placeholders.

#2824

/cc @ericstj @anurse @pakrym


piotrpMSFT:
:shipit:

davidfowl:

pakrym:
:shipit:

ericstj:
:shipit:

Issue #2258, opened 2016-04-04
Update non-ENU EULA's for telemetry policies


danquirk:
In https://github.com/dotnet/cli/pull/2149 we're updating the EULA text to explain the new data collection policy. At the moment we do not have non-ENU EULA text, so all locales' copies have been replaced with the ENU copy. This item tracks getting the correct LOC'd copies back where they belong instead of the ENU copies.

blackdwarf:
/cc @richlander 

@richlander we also have to do this for the release.

PR #2838, opened 2016-05-03
Fix EULA privacy links

GuardRex:
Current link results in a 404 that redirects to the MS homepage.

capture

I don't think this conflicts with @jasonwilliams200OK PR for a markdown version. https://github.com/dotnet/cli/issues/2731

Also cross referencing for telemetry policy updates: https://github.com/dotnet/cli/issues/2258

Note: The EULA at https://github.com/dotnet/cli/blob/rel/1.0.0/pkg/projects/dotnet_library_license.txt does not match the text in these files and was committed fairly recently.


vcsjones:
While you're fixing the links it might be nice to move them to HTTPS, too, since `go.microsoft.com` supports HTTPS.

GuardRex:
I thought this was being fixed with recent updates, but it's still there.

capture

jasonwilliams200OK:
cc @blackdwarf (a minor but important readability fix in the preface text)

eerhardt:
@piotrpMSFT @blackdwarf @brthor - are we still looking to take this change?  If so, it will need to be split and half of the changes will need to go to `dotnet/core-setup`.

blackdwarf:
@eerhardt we should take it. 

@jasonwilliams200OK I'm not sure we can change the EULA text like this. @richlander owns the text here.

GuardRex:
I'm going to close now. It looks like you're good. I'll leave the branch up for a bit. Since those files are already modded, that might save someone some time.

PR #2849, opened 2016-05-03
add PR template

krwq:
- If this PR should not run tests please add text "skip-ci-please" (replace dashes with spaces, no quotes).


krwq:
The text should not automatically disable running tests

PR #2858, opened 2016-05-04
Address telemetry PR feedback

eerhardt:
- Track telemetry isInitialized by checking for TelemetryClient != null
- Don't store the telemetry profile string

@piotrpMSFT @MichaelSimons


MichaelSimons:
:shipit:

PR #2860, opened 2016-05-04
Update NuGet to 3.5.0-beta-1246

alpaix:
Updating NuGet packages version to 3.5.0-beta-1246.

//cc @emgarten @Petermarcu


weshaggard:
Is this a downgrade? At least from a semantic versioning prospective isn't `beta-1246` less then `rc-1233`?

eerhardt:
From semantic versioning, yes it is a downgrade.  But changing it from beta to rc was a mistake in the first place.  This is correcting that mistake.

@yishaigalatzer

yishaigalatzer:
@eerhardt is right

weshaggard:
Sounds good I just wanted to make sure it was intentional and that we don't expect it to impact any scenarios.

PR #2875, opened 2016-05-04
Fix casing of Sha

jasonwilliams200OK:
`Commit Sha: xyz` -> `Commit SHA-1 hash: xyz`  for `dotnet --info` output.

Skip ci please.

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -182,8 +182,8 @@ namespace Microsoft.DotNet.Cli
             var commitSha = GetCommitSha() ?? "N/A";
             Reporter.Output.WriteLine();
             Reporter.Output.WriteLine("Product Information:");
-            Reporter.Output.WriteLine($" Version:     {Product.Version}");
-            Reporter.Output.WriteLine($" Commit Sha:  {commitSha}");
+            Reporter.Output.WriteLine($" Version:            {Product.Version}");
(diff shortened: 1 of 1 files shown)


jasonwilliams200OK:
@piotrpMSFT, can you take a look please.

eerhardt:
@jasonwilliams200OK - thanks for the contribution.  This change LGTM.

In general, "skip ci please" should only be used for non-product changes, like .md files. Even changing a single string in the product could cause test failures if there were tests that called `dotnet --info` and tried verifying the output.

src/dotnet/Program.cs
```diff
@@ -183,7 +183,7 @@ private static void PrintInfo()
             Reporter.Output.WriteLine();
             Reporter.Output.WriteLine("Product Information:");
             Reporter.Output.WriteLine($" Version:     {Product.Version}");
```
eerhardt: Should we add more whitespace before `{Product.Version}` so the values stay lined up vertically?

jasonwilliams200OK:
@eerhardt, duly noted. Thanks! I will keep it in mind. :)

src/dotnet/Program.cs
jasonwilliams200OK: Fixed and rebased.

eerhardt:
:shipit:

Issue #2873, opened 2016-05-04
Install script issues


blackdwarf:
Here are some things that we need to do with the install script to make it better. Some of these are bugs, some are deviations from the spec and some are good to have.

I'm using the following command, as seen in the installation spec to get and run the script:

`curl -sSL https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.sh | bash /dev/stdin --version 1.0.0-rc2-002659`

**Issues:**
- [ ] If no installation directory is provided, the script tries to install into a global location. This fails with "permission denied" even when using `sudo`. This is also not the behavior that we specified in the spec. 
- [ ] There is no progress for the download.
- [ ] The $PATH is not modified. 

/cc @piotrpMSFT @krwq

krwq:
I agree that installation directory is a bug. It seemed to be fine on Ubuntu where I tested it.

Progress for the download: I have mixed feelings about that, in interactive mode I agree it would be useful but since the script is mainly meant to be used in non-interactive mode it is not interesting to see a progress bar on the log. If the main intent is to be in interactive mode then should we ask for download path and add -qq option?

PATH is not modified - The default should not modify global settings on the machine but modify it on the current session - this is not doable unless you source the script which currently have side effects. We should change one liner to something like:
```
old one liner && export PATH=your/install/dir;$PATH
```

we can alternatively also set some global variable with dotnet installation path i.e.: DOTNET_INSTALL_DIR

blackdwarf:
@krwq the progress and the $PATH are not RC2 important, however, the install dir is since it will block people from installing properly. So you should fix that and we will leave the rest for RTM.

piotrpMSFT:
Fixed by https://github.com/dotnet/cli/issues/2873

techiechap:
dotnet script installation successful unfortunately dotnet command not found error

PR #2876, opened 2016-05-04
fix default INSTALL_DIR & --dry-run issue in dotnet-install.sh

krwq:
https://github.com/dotnet/cli/issues/2873

- Fixes default install dir
- --dry-run is returning 0 when script is finished


piotrpMSFT:
:shipit:

ellismg:
I believe the documentation in the script on how install dir is picked (https://github.com/krwq/cli/blob/b18df1c51cfbfcbe4a27deb48b577737c9800751/scripts/obtain/dotnet-install.sh#L605) is now wrong.

PR #2898, opened 2016-05-05
Prefer servicing for native dll search order

schellap:
@gkhanna79 PTAL. Because we process native DLL directories in the order of the assets' directories, the directories of non-serviced assets gets appended first to the `NATIVE_DLL_SEARCH_DIRECTORIES`. However, servicing locations should get preference first. Note that this is not a problem for TPA because they are files and not directories.


schellap:
This can be made a bit more efficient in

schellap:
This can be more efficient in space and time if two lists were built and then merge one with other.

gkhanna79:
LGTM

PR #2930, opened 2016-05-06
dotnet build raises errors MVC apps target net46.

### LINKED ISSUE #2906: Can't build for net46 (net451, net452 works fine)
## Steps to reproduce
Use the https://github.com/aspnet/cli-samples/tree/master/HelloMvc sample. 
Add net46 as the target framework
Restore (restores fine)
Build (`dotnet build -f net46`)
## Expected  behavior
Project builds successfully

## Actual behavior
```
PS C:\Source\cli-samples\HelloMvc> dotnet build -f net46
Project HelloMvc (.NETFramework,Version=v4.6) will be compiled because expected outputs are missing
Compiling HelloMvc for .NETFramework,Version=v4.6
C:\Source\cli-samples\HelloMvc\project.json(15,42): error NU1001: The dependency System.Diagnostics.FileVersionInfo >= 4.0.0-rc2-24027 could not be resolved.
C:\Source\cli-samples\HelloMvc\project.json(15,42): error NU1001: The dependency System.Diagnostics.StackTrace >= 4.0.1-rc2-24027 could not be resolved.
C:\Source\cli-samples\HelloMvc\project.json(15,42): error NU1001: The dependency System.Text.Encoding.CodePages >= 4.0.1-rc2-24027 could not be resolved.
C:\Source\cli-samples\HelloMvc\project.json(15,42): error NU1001: The dependency System.Threading.Thread >= 4.0.0-rc2-24027 could not be resolved.
C:\Source\cli-samples\HelloMvc\project.json(15,42): error NU1001: The dependency System.Xml.XmlDocument >= 4.0.1-rc2-24027 could not be resolved.
C:\Source\cli-samples\HelloMvc\project.json(15,42): error NU1001: The dependency System.Xml.XPath.XDocument >= 4.0.1-rc2-24027 could not be resolved.

Compilation failed.
    0 Warning(s)
    6 Error(s)

Time elapsed 00:00:00.0945403
```

Note when targeting net451 or net452 the project builds just fine. 

## Environment data
`dotnet --info` output:

```
PS C:\Source\cli-samples\HelloMvc> dotnet --info
.NET Command Line Tools (1.0.0-rc2-002670)

Product Information:
 Version:     1.0.0-rc2-002670
 Commit Sha:  f0018642a9

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

eerhardt:
The issue is when the ProjectContextBuilder sees a CompileTimePlaceholder `_._` file on a full framework, it assumes that dependency has to come from the "Reference Assemblies" directory.  If it can't be found there, an error is raised.  However, there are other reasons `_._` placeholders are created (when a NuGet package doesn't want its dependencies to be exposed in the Compile dependencies of its consumers). And these placeholders can exist for assemblies that aren't in the full framework - in this case System.Diagnostics.FileVersionInfo and others.

To fix this, if the reference can't be resolved from the "Reference Assemblies" folder, it is just skipped. If the compiler really needs that assembly, it will raise an error to the user.  Dotnet build shouldn't raise the error.

Fix #2906 

@piotrpMSFT @davidfowl @ericstj

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
@@ -440,15 +440,21 @@ namespace Microsoft.DotNet.ProjectModel
                     package.HasCompileTimePlaceholder &&
                     !TargetFramework.IsPackageBased)
                 {
+                    // requiresFrameworkAssemblies is true whenever we find a CompileTimePlaceholder in a non-package based framework, even if
+                    // the reference is unresolved. This ensures the best error experience when someone is building on a machine without
+                    // the target framework installed.
+                    requiresFrameworkAssemblies = true;
+
                     var newKey = new LibraryKey(library.Identity.Name, LibraryType.ReferenceAssembly);
                     var dependency = new LibraryRange(library.Identity.Name, LibraryType.ReferenceAssembly);
 
-                    // If the framework assembly can't be resolved then mark it as unresolved but still replace the package
-                    // dependency
-                    var replacement = referenceAssemblyDependencyResolver.GetDescription(dependency, TargetFramework) ??
-                                        UnresolvedDependencyProvider.GetDescription(dependency, TargetFramework);
-
-                    requiresFrameworkAssemblies = true;
+                    var replacement = referenceAssemblyDependencyResolver.GetDescription(dependency, TargetFramework);
+                    
+                    // If the reference is unresolved, just skip it.  Don't replace the package dependency
+                    if (replacement == null)
+                    {
+                        continue;
+                    }
 
                     // Remove the original package reference
                     libraries.Remove(pair.Key);
--- a/TestAssets/TestProjects/AppWithNet46AndRoslyn/.noautobuild
+++ b/TestAssets/TestProjects/AppWithNet46AndRoslyn/.noautobuild
@@ -1,0 +1,1 @@
+noautobuild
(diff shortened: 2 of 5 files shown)


src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -440,15 +440,21 @@ private static string ComposeMessageFromInnerExceptions(Exception exception)
                     package.HasCompileTimePlaceholder &&
                     !TargetFramework.IsPackageBased)
                 {
+                    // requiresFrameworkAssemblies is true whenever we find a CompileTimePlaceholder in a non-package based framework, even if
+                    // the reference is unresolved. This ensures the best error experience when someone is building on a machine without
+                    // the target framework installed.
+                    requiresFrameworkAssemblies = true;
```
ericstj: Should this be called "mayRequireFrameworkAssemblies"?  This will occur when folks use `exclude="compile"` which we commonly do in CoreFx packages (as does Roslyn which caused this issue).  Probably not a serious issue now since these are all likely to require a targeting pack, but this might need to be revisited when we look at replacing targeting packs with packages.

ericstj:
LGTM

eerhardt:
Approved in email.  Merging.

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
eerhardt: Setting this to true enables the check to make sure the target framework is installed on the machine.

@davidfowl and I discussed maybe this rule changes to just be `!TargetFramework.IsPackageBased`, instead of also needing a `HasCompileTimePlaceholder` package.

PR #2936, opened 2016-05-06
Make GracefulExceptions print with color on Windows 7 and 8.

AustinWise:
When an invalid dotnet subcommand name is used, the resulting error message prints ASNI escape codes to the console on Windows 7 and 8:
looks bad

The root cause is the use of System.Console instead of Reporter to print the text of GracefulExceptions.

Windows 10 1511 and Unix support ANSI escape codes, so that's why this only affects Windows 7 and 8.

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -61,7 +61,7 @@ namespace Microsoft.DotNet.Cli
             }
             catch (GracefulException e)
             {
-                Console.WriteLine(CommandContext.IsVerbose() ? e.ToString().Red().Bold() : e.Message.Red().Bold());
+                Reporter.Output.WriteLine(CommandContext.IsVerbose() ? e.ToString().Red().Bold() : e.Message.Red().Bold());
 
                 return 1;
             }


jasonwilliams200OK:
Should we have a same thing on [line 130](https://github.com/AustinWise/cli/blob/e1a601a/src/dotnet/Program.cs#L130) ?
There are some others instances: https://github.com/dotnet/cli/search?utf8=%E2%9C%93&q=Console+WriteLine+path%3Asrc+path%3Atools&type=Code (except the Reporter.cs)

eerhardt:
@AustinWise - I had to make a different change to the exact same line of code in #2980.  When a GracefulException occurs, it should be writing to standard error, not standard out. That is consistent with other places where we write errors:

https://github.com/dotnet/cli/blob/8013398defcb53e1807f027726ed4a587d655f1a/src/dotnet/Program.cs#L109
https://github.com/dotnet/cli/blob/7c504b2456de30c715de30feba9386b3d9e6ac3f/src/dotnet/commands/dotnet-run/Program.cs#L47
https://github.com/dotnet/cli/blob/9a85205781bc21610111cff9042ee279aa0343ff/src/dotnet/commands/dotnet-publish/Program.cs#L72

I think this PR can be closed now, right?

AustinWise:
@eerhardt Yeah, it looks like you already got this one. Thanks!

Issue #2572, opened 2016-04-18
`Could not load host policy...` when dll does not exist


moozzyk:
## Steps to reproduce
Misspell dll name when doing `dotnet my.dll`

## Expected  behavior
An error message saying that the dll can't be found. This is the dll that does not exist in the first place and not the host policy. 

## Actual behavior
```
PS > dotnet .\bin\Debug\netcoreapp1.0\dotnet-scratch.dll
dotnet-scratch: 1.2.3.0

PS > dotnet .\bin\Debug\netcoreapp1.0\dotnet-scratch1.dll
Could not load host policy library from [C:\Users\\Documents\Visual Studio 2015\Projects\dotnet-scratch\bin\Debug\netcoreapp1.0]
```

## Environment data
`dotnet --info` output:

moozzyk:
Note that the message changed to: 

```
PS C:\Temp\HelloMvc> dotnet .\HelloMvcaaa.dll
Expect required library hostpolicy.dll to be present in [C:\Temp\HelloMvc]
  - This may be because of an invalid .NET Core FX configuration in the directory.
```

Not sure if the old message was just rephrased or if I am hitting a different code path now but the  usefulness and correctness of the new message is about the same as of the original one (i.e. very low).

PR #2946, opened 2016-05-07
Make error message better

schellap:
**This is an RTM PR.**

Fixes https://github.com/dotnet/cli/issues/2572

### DIFF
--- a/src/corehost/cli/fxr/fx_muxer.cpp
+++ b/src/corehost/cli/fxr/fx_muxer.cpp
@@ -303,9 +303,8 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
         return true;
     }
 
-    // If it still couldn't be found, flag an error for the "expected" location.
-    trace::error(_X("Expect required library %s to be present in [%s]"), LIBHOSTPOLICY_NAME, expected.c_str());
-    trace::error(_X("  - This may be because of an invalid .NET Core FX configuration in the directory."));
+    // If it still couldn't be found, somebody upstack messed up. Flag an error for the "expected" location.
+    trace::error(_X("A fatal error was encountered. The library '%s' required to execute the application was not found in '%s'."), LIBHOSTPOLICY_NAME, expected.c_str());
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -303,9 +303,9 @@ bool fx_muxer_t::resolve_hostpolicy_dir(host_mode_t mode,
         return true;
     }
 
-    // If it still couldn't be found, flag an error for the "expected" location.
-    trace::error(_X("Expect required library %s to be present in [%s]"), LIBHOSTPOLICY_NAME, expected.c_str());
-    trace::error(_X("  - This may be because of an invalid .NET Core FX configuration in the directory."));
+    // If it still couldn't be found, somebody upstack messed up. Flag an error for the "expected" location.
+    trace::error(_X("The library '%s' required to execute the application was not found in '%s'"), LIBHOSTPOLICY_NAME, expected.c_str());
+    trace::error(_X("  - This is a fatal error, and an invalid application configuration in the above directory."));
```
eerhardt: I don't know exactly what `an invalid application configuration in the above directory.` means.  Do we need this line at all? Is the first line enough information?

schellap: This is just a fatal error with some files missing, so somebody messed up the configuration. Do you have any suggestions for rewording it? I'm inclined to making it "Fatal error: The library %s is required..."

eerhardt: Yes, I think that would be perfect.  Just move "Fatal error:" to the beginning of the first line, and drop the second line.

schellap:
@gkhanna79 PTAL

src/corehost/cli/fxr/fx_muxer.cpp
```diff
@@ -612,30 +612,31 @@ int fx_muxer_t::parse_args_and_execute(
         }
 
         app_candidate = argv[cur_i];
-        bool is_app_runnable = (ends_with(app_candidate, _X(".dll"), false) || ends_with(app_candidate, _X(".exe"), false)) && pal::realpath(&app_candidate);
-        trace::verbose(_X("App %s runnable=[%d]"), app_candidate.c_str(), is_app_runnable);
-        // If exec mode is on, then check we have a dll at this point
-        if (exec_mode)
+        bool is_app_managed = (ends_with(app_candidate, _X(".dll"), false) || ends_with(app_candidate, _X(".exe"), false)) && pal::realpath(&app_candidate);
+        trace::verbose(_X("App candidate '%s' has %d managed executable extension"), app_candidate.c_str(), is_app_managed);
```
eerhardt: I know this is only a verbose message, but I think it could be worded better if we switched on `is_app_managed`.

```
trace::verbose(
  is_app_managed ? 
         _X("App candidate '%s' has a managed executable extension") :
         _X("App candidate '%s' does not have a managed executable extension"),
  app_candidate.c_str());
```

schellap: Thanks, refactored so this pattern doesn't occur.

eerhardt:
:shipit: when CI is green.

eerhardt:
test OSX x64 Release Build

eerhardt:

HarelM:
I'm looking at the following issues:
https://developercommunity.visualstudio.com/content/problem/7876/iis-express-fails-to-start-with-a-migrated-aspnet.html
Which was linked from a feedback with the name:
ASP.NET CORE App fails to run after updating to latest Nuget Packages
I'm not sure what's the current status of this issues but my user experience was as follows (did it today):
1. Install VS 2017 RC.
2. Create new wep api .net core application.
3. Update Nuget packages
4. Run

**Actual results**:
Crash (with non informative error massage):
```the program '[8648] dotnet.exe' has exited with code -2147450749 (0x80008083).```
**Expected results**:
Tell me there's a problem with my dotnet core version
- or -
Do not allow to upgrade a project which targeting .net core 1.0 with packages that targets 1.1
- or -
Install with VS 2017 RC the latest dotnet core

Installing .net core 1.1 solved the above issue but caused me to waste time and have less faith in the stability of this project (After the mess with xproj you guys have to step it up with the quality...).

piotrpMSFT:
@HarelM I'm sorry you hit this issue, and I'm glad to say it's resolved. CLI already carries .net core 1.0 and 1.1, so recent builds would not have reproduced what you saw below. I think you hit this issue because the ASP.NET packages, available via NuGet, were available to you sooner than the latest VS [and therefore CLI]. I think that once 1.0 is released we will need to invest more in features that help keep versions aligned for scenarios like this one.

ionut2:
HI,
I am having the same issue with a dotnet core console application, but only when I try to debug it from VS code. If I run it from command line it works fine. Also yesterday it worked fine. I believe a dotnet restore  caused the issue but now I do not know how to fix ti in order to be able to debug again from VS code

Issue #2892, opened 2016-05-04
dotnet run app.dll throws null reference exception


javiercn:
## Steps to reproduce
1. build a console app.
2. run dotnet publish
3. navigate to the publish folder
4. run dotnet run app.dll
## Expected  behavior

An error indicating that to use dotnet on a dll you shoud do dotne app.dll
## Actual behavior

Object reference not set to an instance of an object.
## Environment data

`dotnet --info` output:
.NET Command Line Tools (1.0.0-rc2-002611)

Product Information:
 Version:     1.0.0-rc2-002611
 Commit Sha:  bf8f0edd89

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  14.04
 OS Platform: Linux
 RID:         ubuntu.14.04-x64

Other info
  "dependencies": {
    "Microsoft.NETCore.App": {
      "version": "1.0.0-rc2-24008",
      "type": "platform"
    },
    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.0-rc2-20581",
    "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-rc2-20581"
  },

  "frameworks": {
    "netcoreapp1.0": {
      "imports": [
        "dotnet5.6",
        "dnxcore50",
        "portable-net45+win8"
      ]
    }
  }

moozzyk:
Hit this today as well and reopened https://github.com/dotnet/cli/issues/941

vcsjones:
This is related to #2965. CLI is looking for a project.json in the current path and wants to pass "app.dll" as an argument.

eerhardt:
I think this was fixed with #2890.  Can you try again with the latest Preview2 build and close if this has been fixed?

BTW - to run your published app, you don't say `dotnet run`.  Instead you call, `dotnet myapp.dll`.  `dotnet run` is a "development time" command, thus it needs a project.json file, and your project to be restored, etc.

Sebazzz:
Perhaps this can be clarified in documentation. I was trying to run my published app using `dotnet run`, only to find out now it should be run using `dotnet filename.dll`. Never mind, it shouldn't crash with such a vague error message. I haven't tried the preview2 bits yet though.

toddams:
> dotnet run is a "development time" command, thus it needs a project.json file, and your project to be restored, etc.

@eehardt
But when you actually use `dotnet run app.dll` on self-contained app - it works. That is an expected behaviour?

304NotModified:
[...]

Issue #941, opened 2016-01-21
dotnet run NREs on project whose dependencies were not restored


moozzyk:
I cloned https://github.com/davidfowl/dotnetcli-aspnetcore and forgot to run restore:
```
PS C:\source\dotnetcli-aspnetcore\HelloMvc> dotnet run
Object reference not set to an instance of an object.
```

(Not sure if it is relevant - I did not have `DOTNET_HOME` set)

blackdwarf:
@moozzyk this doesn't repro for me on the latest bits. Will close this, plese reopen if you find it still repros.

moozzyk:
When I run `dotnet run` in a folder an application was published to:

```
PS C:\temp\publishout> dotnet -v run
Telemetry is: Enabled
Object reference not set to an instance of an object.
```

```
PS C:\temp\publishout> dotnet --info
.NET Command Line Tools (1.0.0-rc2-002655)

Product Information:
 Version:     1.0.0-rc2-002655
 Commit Sha:  3e96a05f9c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

eerhardt:
I think this was resolved with #2980.  Can you retry with the latest Preview2 bits and close if this is fixed?

piotrpMSFT:
I verified this is fixed in preview2.

PR #2969, opened 2016-05-10
Fix null reference output on missing project for dotnet run

vcsjones:
This pull request attempts to address #2965 when `dotnet run` cannot find a project to run.

This change simply guards against the null value, and if null, gives a more helpful error message.

It may be worthwhile to pull out the logic from `Workspace.NormalizeProjectPath` to show the possible paths that it is looking for a project file, but I'm not sure where that new logic should belong or if it is worth doing.

### DIFF
--- a/src/dotnet/commands/dotnet-run/RunCommand.cs
+++ b/src/dotnet/commands/dotnet-run/RunCommand.cs
@@ -72,8 +72,8 @@ namespace Microsoft.DotNet.Tools.Run
             if (projectContextCollection == null)
             {
                 throw new InvalidOperationException("Couldn't find a project to run. Possible causes:" + Environment.NewLine +
-                    "1. There is no project.json file in the current working directory." + Environment.NewLine +
-                    "2. The path specified with `--project` does not exist or does not contain a project.json file.");
+                    "1. There is no project in the current working directory - use `dotnet run --project` to specify a path." + Environment.NewLine +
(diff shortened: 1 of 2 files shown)


vcsjones:
Windows failure looks unrelated.

vcsjones:
This PR also probably addresses #2892 and #941.

eerhardt:
Thanks @vcsjones for the contribution.

I was in the middle of a separate change (#2980) that was modifying the exact same line of code when I first saw your change. My scenario was when the project.lock.json file didn't exist (since the project wasn't restored). My fix checks for any diagnostic errors and throws a GracefulException if there are any.

I incorporated this scenario into my change and took the test you added to ensure the scenario works correctly going forward. Adding this scenario to my change fixed this same issue across multiple commands: `test`, `publish`, `run`.

Since we were touching the same line of code in RunCommand.cs, your PR now has conflicts with my change.

With your permission, I think this PR can be closed now, as the behavior is covered in #2980.

If I was to do this over, I think I should have just continued my change without incorporating your scenario into it. And then you could have added your scenario to my change once the conflicts were resolved. My thinking at the time was combining them got to the end result faster.

Thanks for reporting the issue and creating this PR.  It even has a test - which is great!

vcsjones:
>With your permission, I think this PR can be closed now, as the behavior is covered in #2980.

Awesome. Your change is obviously much more complete.

PR #2990, opened 2016-05-11
Corrected path to built dotnet

svick:
There is no `artifacts\rid\stage2\bin`, `dotnet` is in `artifacts\rid\stage2`.


eerhardt:
:shipit:

PR #2998, opened 2016-05-11
Fixed broken Equals (2)

bkoelman:
Supersedes https://github.com/dotnet/cli/pull/2601. I lost my clone so recreated the PR.

@eerhardt I believe this is more correct. Overloaded operators defer to `.Equals` and proper null checks are in place. For `GetHashCode` to be reliable, it must use immutable fields. So I have made the `LanguageId` property readonly and added a constructor.

It's easiest to review the `AnalyzerOptions` file in full mode.


eerhardt:
:shipit:

eerhardt:
test OSX x64 Release Build

PR #3001, opened 2016-05-11
dotnet crashes with an unresolved 'platform' dependency.

### LINKED ISSUE #2693: dotnet crashes on invalid command
## Steps to reproduce

1. `dotnet new`
2. Edit `project.json`: depend on a version of `Microsoft.NETCore.App` that won't resolve, e.g. `42.42.42`.
3. `dotnet restore`
4. `dotnet crash`

## Expected  behavior
```
λ dotnet crash
No executable found matching command "dotnet-crash"
```

## Actual behavior
```
λ dotnet crash

Unhandled Exception: System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Microsoft.DotNet.ProjectModel.ProjectContextBuilder.Build()
   at Microsoft.DotNet.ProjectModel.ProjectContextBuilder.<BuildAllTargets>d__66.MoveNext()
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
   at Microsoft.DotNet.Cli.Utils.ProjectToolsCommandResolver.GetProjectContextFromDirectoryForFirstTarget(String projectRootPath)
   at Microsoft.DotNet.Cli.Utils.ProjectToolsCommandResolver.ResolveFromProjectTools(String commandName, IEnumerable`1 args, String projectDirectory)
   at Microsoft.DotNet.Cli.Utils.ProjectToolsCommandResolver.Resolve(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.CompositeCommandResolver.Resolve(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.CommandResolver.TryResolveCommandSpec(String commandName, IEnumerable`1 args, NuGetFramework framework, String configuration, String outputPath)
   at Microsoft.DotNet.Cli.Utils.Command.Create(String commandName, IEnumerable`1 args, NuGetFramework framework, String configuration, String outputPath)
```

## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-rc2-002490)

Product Information:
 Version:     1.0.0-rc2-002490
 Commit Sha:  06f57c5c81

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
```

eerhardt:
Fixed by calling TryGetValue instead of a dictionary indexer when looking up platform dependencies in ProjectContextBuilder.

Fix #2693

@brthor

### DIFF
--- a/TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedPlatformDependency/project.json
+++ b/TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedPlatformDependency/project.json
@@ -1,0 +1,17 @@
+{
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "ThisIsNotARealDependencyAndIfSomeoneGoesAndAddsAProjectWithThisNameIWillFindThemAndPunishThem": {
+      "type": "platform",
+      "version": "1.0.0"
+    }
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": "dnxcore50"
+    }
+  }
+}
--- a/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
+++ b/src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
@@ -277,7 +277,7 @@ namespace Microsoft.DotNet.ProjectModel
 
                     if (platformDependency != null)
                     {
-                        platformLibrary = libraries[new LibraryKey(platformDependency.Value.Name)];
+                        libraries.TryGetValue(new LibraryKey(platformDependency.Value.Name), out platformLibrary);
                     }
                 }
             }
--- a/TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedPlatformDependency/Program.cs
+++ b/TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedPlatformDependency/Program.cs
@@ -1,0 +1,12 @@
+﻿using System;
+
+namespace ConsoleApplication
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.WriteLine("Hello World!");
+        }
+    }
+}
--- a/test/dotnet.Tests/GivenThatDotNetRunsCommands.cs
+++ b/test/dotnet.Tests/GivenThatDotNetRunsCommands.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved. 
+// Licensed under the MIT license. See LICENSE file in the project root for full license information. 
+
+using System.IO;
+using FluentAssertions;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using Xunit;
+
+namespace Microsoft.DotNet.Tests
+{
+    public class GivenThatDotNetRunsCommands : TestBase
+    {
+        [Fact]
+        public void UnresolvedPlatformReferencesFailAsExpected()
+        {
+            var testAssetsManager = GetTestGroupTestAssetsManager("NonRestoredTestProjects");
+            var testInstance = testAssetsManager.CreateTestInstance("TestProjectWithUnresolvedPlatformDependency");
+
+            new RestoreCommand()
+                .WithWorkingDirectory(testInstance.TestRoot)
+                .Execute()
+                .Should()
+                .Fail();
+            new DirectoryInfo(testInstance.TestRoot)
+                .Should()
+                .HaveFile("project.lock.json");
+
+            new DotnetCommand()
+                .ExecuteWithCapturedOutput("crash")
+                .Should()
+                .Fail()
+                .And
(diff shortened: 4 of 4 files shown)


TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedPlatformDependency/project.json
```diff
@@ -0,0 +1,17 @@
+{
+  "version": "1.0.0-*",
+  "compilationOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "ThisIsNotARealDependencyAndIfSomeoneGoesAndAddsAProjectWithThisNameIWillFindThemAndPunishThem": {
```
brthor: I love this name, but realistically will we hit any long path issues?

eerhardt: This name was taken from an existing test asset testing a "project" dependency: https://github.com/dotnet/cli/blob/rel/1.0.0/TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedDependency/project.json#L11

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
```diff
@@ -277,7 +277,7 @@ public ProjectContext Build()
 
                     if (platformDependency != null)
                     {
-                        platformLibrary = libraries[new LibraryKey(platformDependency.Value.Name)];
+                        libraries.TryGetValue(new LibraryKey(platformDependency.Value.Name), out platformLibrary);
```
brthor: When this happens, isn't this an error state? If this comes back null, should we throw here, print a warning, catch it in the command resolver and then return null from that resolver?

TestAssets/NonRestoredTestProjects/TestProjectWithUnresolvedPlatformDependency/project.json
brthor: as long as it is working, cool

src/Microsoft.DotNet.ProjectModel/ProjectContextBuilder.cs
eerhardt: Just returning null here provides the correct behavior.  I'm not certain why we would try to add more complexity.

1. `dotnet restore` has already failed, telling the user the dependency doesn't exist.
2. If I try running `dotnet build` on this failed-to-restore project it prints out:

```
C:\cli\TestAssets\NonRestoredTestProjects\TestProjectWithUnresolvedPlatformDependency\project.json(7,103): error NU1001: The dependency ThisIsNotARealDependencyAndIfSomeoneGoesAndAddsAProjectWithThisNameIWillFindThemAndPunishThem >= 1.0.0 could not be resolved.
```

brthor: Sure, but assuming the unresolved dependency is not the tool in a project, a ProjectToolsCommandResolver may still try to use the projectcontext returned by this to find the tool. 

What do you think the correct behavior should be there?

eerhardt: I think ALL things that try to use a ProjectContext should verify that the ProjectContext is valid before using it, like so:

https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/Extensions/ProjectContextCollectionExtensions.cs#L14

If a ProjectContext has diagnostic errors, it should fail and those errors should be written to the screen.

brthor: You are saying *everywhere* we create a projectcontext we should do validation? 

Why wouldn't we handle the validation in the common piece (project context creation), and use built in error mechanisms to communicate the validation failures?

eerhardt: That could work too.  I was assuming there are places that need to be able to work with an invalid ProjectContext.  If there aren't any, then we could put the validation into the builder.

brthor: I'm thinking in terms of usage of the projectmodel api using exceptions would make error cases more visible and bug fixes faster. 

If we need invalid ProjectContext support you are completely right though and I don't know the answer to that question.

eerhardt: I think that is a larger change than fixing this one bug.  Maybe log an issue for it?

brthor: Yup you're right. Not sure it'd take precedence over the more pressing issues for preview2 as well. 

Sounds good

brthor:
:shipit:

eerhardt:
test OSX x64 Release Build

PR #3011, opened 2016-05-12
Fix install scripts to deal with content type change.

eerhardt:
@piotrpMSFT @brthor


eerhardt:
Merging to get the build unblocked.

PR #3020, opened 2016-05-13
Strict deps resolution else fail fast

schellap:
When assets in the deps are not found, fail fast.

Only when deps file is missing add the assemblies in the app base.

### DIFF
--- a/build_projects/dotnet-cli-build/CompileTargets.cs
+++ b/build_projects/dotnet-cli-build/CompileTargets.cs
@@ -396,7 +396,8 @@ namespace Microsoft.DotNet.Cli.Build
             File.Copy(Path.Combine(Dirs.CorehostLocked, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), Path.Combine(outputDir, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), overwrite: true);
             File.Copy(Path.Combine(Dirs.CorehostLatest, $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}"), Path.Combine(outputDir, $"{Constants.DynamicLibPrefix}hostpolicy{Constants.DynamicLibSuffix}"), overwrite: true);
 
-            var binaryToCorehostifyOutDir = Path.Combine(outputDir, "runtimes", "any", "native");
+            var binaryToCorehostifyRelDir = Path.Combine("runtimes", "any", "native");
+            var binaryToCorehostifyOutDir = Path.Combine(outputDir, binaryToCorehostifyRelDir);
             // Corehostify binaries
             foreach (var binaryToCorehostify in BinariesForCoreHost)
             {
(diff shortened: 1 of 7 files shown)


scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -329,7 +329,6 @@ public static BuildTargetResult CompileStage1(BuildTargetContext c)
                 dotnet: DotNetCli.Stage0,
                 outputDir: Dirs.Stage1);
 
-            CleanOutputDir(Path.Combine(Dirs.Stage1, "sdk"));
```
eerhardt: Looks like the method `CleanOutputDir` can now be removed as well. And also the `FilesToClean` static array.

scripts/dotnet-cli-build/CompileTargets.cs
```diff
@@ -657,7 +655,6 @@ private static BuildTargetResult CompileCliSdk(BuildTargetContext c, DotNetCli d
                 {
                     // Yes, it is .exe even on Linux. This is the managed exe we're working with
                     File.Copy(Path.Combine(binaryToCorehostifyOutDir, $"{binaryToCorehostify}.exe"), Path.Combine(outputDir, $"{binaryToCorehostify}.dll"));
-                    File.Delete(Path.Combine(binaryToCorehostifyOutDir, $"{binaryToCorehostify}.exe"));
```
eerhardt: ?? I think we should continue deleting the `csc.exe` file, since it isn't being used - we don't want it in our SDK folder.

schellap: If an assembly specified by the deps file is not found, I am flagging an error.

So I had to remove this deletion. The other option is to remove this entry from the deps file as well. At that point, it is editing a project.lock.json. Both of these are not pretty, do you have other ideas or opinions? @piotrpMSFT.

eerhardt: Maybe we shouldn't be renaming `csc.exe` to `csc.dll` at all?  Maybe we should just leave it `csc.exe` - just like we are for `vbc.exe` now.

/cc @pakrym who added this code.

pakrym: I've added renaming code because `csc.exe` could not be ran directly with portable cli setup. Having two csc's does not make sense, but you can change `ChangeEntryPointLibraryName` second argument to null and it will delete deps entry completely.

schellap:
Adding a note: test sqlclient, musicstore.

schellap:
@gkhanna79 PTAL. In commit 2, I am retaining mscorlib.dll in the Shared FX. NI images can be incomplete right and may need the IL? If not, I can undo commit 2 and not flag an error for mscorlib, because coreclr anyways would error out if it can't find it.

@eerhardt I manually removed the entries from the `runtimeTargets` section.

gkhanna79:
> Only when deps file is missing add the assemblies in the app base.

Shouldn't this say "only when app is standalone..."?

build_projects/dotnet-cli-build/CompileTargets.cs
```diff
@@ -318,13 +318,6 @@ public static void PublishSharedFramework(BuildTargetContext c, string outputDir
                 Path.Combine(Dirs.CorehostLatest, HostPolicyBaseName),
                 Path.Combine(SharedFrameworkNameAndVersionRoot, HostPolicyBaseName), true);
 
-            if (File.Exists(Path.Combine(SharedFrameworkNameAndVersionRoot, "mscorlib.ni.dll")))
```
gkhanna79: NI contains the IL. We should undo this change.

schellap: > Shouldn't this say "only when app is standalone..."?

No, this is for portable too without rid assets for app dependency assemblies.

schellap: > NI contains the IL

I thought only R2R contained them. Thanks for clarifying. I'm leaving the error handling for mscorlib to CoreCLR since apps can choose to delete one or the other.

schellap:

schellap:
@dotnet-bot test this please

schellap:
@dotnet-bot test this please

schellap:
@dotnet-bot test this please

schellap:

schellap:

schellap:

schellap:

schellap:

schellap:
@dotnet-bot test this please

PR #3027, opened 2016-05-13
debian package uninstallation script

brthor:
cc @piotrpMSFT @blackdwarf @Sridhar-MS 

tested this out, and this will remove any dotnet installations post-shared framework.

### DIFF
--- a/scripts/obtain/uninstall/debian-packages.sh
+++ b/scripts/obtain/uninstall/debian-packages.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+current_user=$(whoami)
+if [ $current_user != "root" ]; then
+    echo "debian-packages.sh uninstallation script requires superuser privileges to run"
+    exit 1
+fi
+
+host_package_name="dotnet-host"
+
+remove_all(){
+    apt-get purge -y $host_package_name
+}
+
+is_dotnet_host_installed(){
+    local out="$(dpkg -l | grep $host_package_name)"
+    [ -z "$out" ]
+}
+
+is_dotnet_host_installed
+[ "$?" -eq 0 ] && echo "Unable to find dotnet installation to remove." \
+    && exit 0
+
+remove_all
+[ "$?" -ne 0 ] && echo "Failed to remove dotnet packages." && exit 1
+
+is_dotnet_host_installed
+[ "$?" -ne 0 ] && \
+    echo "dotnet package removal succeeded but appear to still be installed. Please file an issue at https://github.com/dotnet/cli" && \
+    exit 1
+
+echo "dotnet package removal succeeded."
+exit 0
--- a/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
+++ b/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
@@ -17,7 +17,7 @@ dotnet_pkg_name_suffix="com.microsoft.dotnet"
 dotnet_install_root="/usr/local/share/dotnet"
 dotnet_path_file="/etc/paths.d/dotnet"
 
-remove_all(){
+remove_dotnet_pkgs(){
     installed_pkgs=($(pkgutil --pkgs | grep $dotnet_pkg_name_suffix))
     
     for i in "${installed_pkgs[@]}"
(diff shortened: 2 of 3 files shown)


scripts/obtain/uninstall/debian-packages.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
Sridhar-MS: Can we rename the file to `dotnet-uninstall-debian.sh` ?

Sridhar-MS: Or we can have a single file `dotnet-uninstall.sh` which works for `OSX` and `Ubuntu`.

brthor: I don't think we should have one script for OSX and Ubuntu. 

The processes are meaningfully different and having that detection logic seems unnecessary when a user could very easily choose a different script. 

the name change sounds find though.

Sridhar-MS:
I have also pushed my commit to remove OSX pkgs to the same branch.

scripts/obtain/uninstall/debian-packages.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+current_user=$(whoami)
+if [ $current_user != "root" ]; then
+    echo "debian-packages.sh uninstallation script requires superuser privileges to run"
```
Sridhar-MS: Use `basename "$0"`

scripts/obtain/uninstall/debian-packages.sh
Sridhar-MS: Does the script also remove the dotnet path from $PATH?

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
Sridhar-MS: @blackdwarf can you take a look? This will remove all dotnet installation in `/usr/local/share/dotnet`

scripts/obtain/uninstall/debian-packages.sh
brthor: yes the host package carries a symlink and that will be removed.

scripts/obtain/uninstall/debian-packages.sh
brthor: done

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
```diff
@@ -0,0 +1,38 @@
...
+        pkgutil --force --forget "$i"
+    done
+    
+    echo "Deleting install root - $dotnet_install_root"
+    rm -rf $dotnet_install_location
+    rm -f $dotnet_path_file
+}
+
+remove_all
+[ "$?" -ne 0 ] && echo "Failed to remove dotnet packages." && exit 1
```
brthor: Have you tested a case where an error occurs?

Without `set -e` or some test at the end of the function `ie. [ -z "$out" ]` I believe the exit code of the function will only be that of the last command run `rm -f` which I wouldn't ever expect to return non-zero.

blackdwarf:
Wait...what are we doing here? I get the PKG ones, but what are the DEB uninstall ones?

brthor:
> Wait...what are we doing here? I get the PKG ones, but what are the DEB uninstall ones?

It uninstalls all the dotnet packages on your machine, what is the concern?

blackdwarf:
@brthor I thought there is an OS-level mechanism to do this?

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
```diff
@@ -0,0 +1,38 @@
...
+dotnet_pkg_name_suffix="com.microsoft.dotnet"
+dotnet_install_root="/usr/local/share/dotnet"
+dotnet_path_file="/etc/paths.d/dotnet"
+
+remove_all(){
+    installed_pkgs=($(pkgutil --pkgs | grep $dotnet_pkg_name_suffix))
+    
+    for i in "${installed_pkgs[@]}"
+    do
+        echo "Removing dotnet component - \"$i\""
```
blackdwarf: You should say "Removing recipes" or similar.

blackdwarf:
@Sridhar-MS the PKG uninstall one LGTM. :shipit:

piotrpMSFT:

blackdwarf:
LGTM all up :shipit:

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
brthor: ping on this

PR #3028, opened 2016-05-13
Fix publish CoreHost NuGet packages

eerhardt:
Merging the PullNupkgFilesFromBlob target into rel/1.0.0 so it picks up the correct channel.

Fixing the publish script to push the CoreHost nupkgs to the correct version folder, not the CLI version folder.

This needs to be fixed ASAP so consumers can start consuming our RC3 packages.  I consciously didn't refactor the publishing code into the 'FinalizeBuild' target because that would have taken more time. We can do that separately, if we feel it is needed.

@weshaggard @Sridhar-MS @brthor @livarcocc

### DIFF
--- a/scripts/dotnet-cli-build/PublishTargets.cs
+++ b/scripts/dotnet-cli-build/PublishTargets.cs
@@ -458,7 +458,7 @@ namespace Microsoft.DotNet.Cli.Build
 
             List<string> buildVersions = new List<string>();
 
-            Regex buildVersionRegex = new Regex(@"Binaries/(?<version>\d+\.\d+\.\d+-[^-]+-\d{6})/$");
+            Regex buildVersionRegex = new Regex(@"Binaries/(?<version>\d+\.\d+\.\d+(?<prerelease>-[^-]+-\d{6})?)/$");
 
             foreach (string file in AzurePublisherTool.ListBlobs(hostBlob))
             {
@@ -535,7 +535,9 @@ namespace Microsoft.DotNet.Cli.Build
             }
 
             if (missingRuntime)
+            {
                 Console.WriteLine($"Build version {buildVersion} is missing some runtime packages so not pushing this drop.");
+            }
 
             return !missingRuntime;
         }
--- a/scripts/dotnet-cli-build/PublishTargets.cs
+++ b/scripts/dotnet-cli-build/PublishTargets.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Net.Http;
 using System.Text;
+using System.Text.RegularExpressions;
 using Microsoft.DotNet.Cli.Build.Framework;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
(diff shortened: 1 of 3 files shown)


brthor:
I may be missing some context here, but what is the purpose of this? Why do we need to pull these packages from blob?

eerhardt:
> I may be missing some context here, but what is the purpose of this? Why do we need to pull these packages from blob?

This is the way it is currently working.  We have a separate VSO build (on Windows) that runs every so often and looks to see if any builds have completed across all platforms. If so, it downloads the .nupkgs from the Azure blob, and then a 2nd build step publishes those .nupkgs to MyGet.

We currently lack the infrastructure to publish .nupkgs from non-Windows machines. Our "CLI" packages are only published from our Windows build today.  These "CoreHost" packages contain platform-specific assets, which means all packages from all platforms need to be published.

Today it works by pushing all the platforms packages to Azure, and then a Windows "clean up" build comes along and picks them up and pushes them to MyGet.

We can change this in the future, but currently all "rc3" packages aren't getting published.  So this is to get "rc3" unblocked.

brthor:
Thanks @eerhardt that explanation makes sense. 

Whilst I understand the need for urgency here, NuGet xplat appears to have some support for push, and in the future we should consider taking advantage of it to avoid the complexity of needing a clean up build.

eerhardt:
> Whilst I understand the need for urgency here, NuGet xplat appears to have some support for push, and in the future we should consider taking advantage of it to avoid the complexity of needing a clean up build.

I think there needs to be a separate issue for this, and it should be prioritized accordingly.

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -436,5 +437,116 @@ public static BuildTargetResult TriggerDockerHubBuilds(BuildTargetContext c)
...
+            }
+
+            return c.Success();
+        }
+
+        private static bool ShouldDownloadAndPush(string hostBlob, string buildVersion)
+        {
+            // Set of runtime ids to look for to act as the signal that the build
+            // as finished each of these legs of the build.
+            Dictionary<string, bool> runtimes = new Dictionary<string, bool>()
```
brthor: This is the list of rids that we run builds for? Could we move this to a more central location?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -436,5 +437,116 @@ public static BuildTargetResult TriggerDockerHubBuilds(BuildTargetContext c)
...
+            {
+                if (!runtime.Value)
+                {
+                    missingRuntime = true;
+                    Console.WriteLine($"Version {buildVersion} missing packages for runtime {runtime.Key}");
+                }
+            }
+
+            if (missingRuntime)
+                Console.WriteLine($"Build version {buildVersion} is missing some runtime packages so not pushing this drop.");
```
brthor: nit: open/close brackets

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -436,5 +437,116 @@ public static BuildTargetResult TriggerDockerHubBuilds(BuildTargetContext c)
...
+
+            return !missingRuntime;
+        }
+
+        private static void DownloadPackagesForPush(string pathToDownload)
```
brthor: I'm slightly confused of the scope of this function. What role does the PackagePushedSemaphore play here?

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
```diff
@@ -187,5 +187,23 @@ public string CalculateArchiveBlob(string archiveFile, string channel, string ve
         {
             return $"{channel}/Binaries/{version}/{Path.GetFileName(archiveFile)}";
         }
+
+        public void DownloadFiles(string blobVirtualDirectory, string fileExtension, string downloadPath)
```
brthor: `DownloadFilesOfExtension`?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -227,7 +228,7 @@ public static BuildTargetResult PublishCoreHostPackages(BuildTargetContext c)
         {
             foreach (var file in Directory.GetFiles(Dirs.CorehostLocalPackages, "*.nupkg"))
             {
-                var hostBlob = $"{Channel}/Binaries/{CliNuGetVersion}/{Path.GetFileName(file)}";
+                var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}/{Path.GetFileName(file)}";
```
brthor: Shouldn't this go to the `SharedHostNugetVersion `?

scripts/dotnet-cli-build/PublishTargets.cs
```diff
@@ -436,5 +437,116 @@ public static BuildTargetResult TriggerDockerHubBuilds(BuildTargetContext c)
...
+            if (!string.IsNullOrEmpty(forcePushBuild))
+            {
+                Console.WriteLine($"Forcing all nupkg packages for build version {forcePushBuild}.");
+                DownloadPackagesForPush(hostBlob + forcePushBuild);
+                return c.Success();
+            }
+
+            List<string> buildVersions = new List<string>();
+
+            Regex buildVersionRegex = new Regex(@"Binaries/(?<version>\d+\.\d+\.\d+-[^-]+-\d{6})/$");
```
brthor: After the shared framework ships in june this regex may break. 

Recommend making the prerelease portion an optional group:
i.e `Binaries/(?<version>\d+\.\d+\.\d+([^-]+-\d{6})?)/$`

brthor:
> I think there needs to be a separate issue for this, and it should be prioritized accordingly.

sure. 

https://github.com/dotnet/cli/issues/3031

scripts/dotnet-cli-build/PublishTargets.cs
weshaggard: I don't mind if this is optional but I suspect we will never be producing packages that don't have a prerelease id in our dev builds. We may for release builds though so it might make since for those cases.

scripts/dotnet-cli-build/PublishTargets.cs
weshaggard: Do we maintain this list in any other location?

scripts/dotnet-cli-build/PublishTargets.cs
brthor: Those were my thoughts as well

scripts/dotnet-cli-build/PublishTargets.cs
weshaggard: The goal of the semaphore is mainly to signal that we no longer need to push the nupkg's for this blob any more. If you see how we are polling the last 10 builds above we use this to make sure we don't repush the same blob multiple times.

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
weshaggard: I think DownloadFiles is fine here. Think of Directory.GetFiles(...) it takes a wildcard as well.

weshaggard:
Thanks @eerhardt for merging this change! The changes LGTM

scripts/dotnet-cli-build/PublishTargets.cs
brthor: I understand the semaphore purpose, my concern was more over the naming of the function not signifying the semaphore side effects.

Also what if there was a failure between here and when they are actually pushed, doesn't that leave us in a broken state?

brthor: I realize the potential broken state may be unavoidable for the moment, I thought it was worth calling out though.

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
brthor: But this doesn't appear to take a wildcard. It explicitly searches for files with the provided extension no?

scripts/dotnet-cli-build/PublishTargets.cs
brthor: Unfortunately only here:
https://github.com/dotnet/cli/blob/rel/1.0.0/netci.groovy#L12

Parsing that may not be a terrible idea, but I don't see it in the scope of this PR. I'm okay with leaving this here for now.

scripts/dotnet-cli-build/PublishTargets.cs
eerhardt: SharedHostNuGetVersion is defined as:

```
SharedHostNugetVersion = c.BuildContext.Get<HostVersion>("HostVersion").LockedHostVersion;
```

Where LockedHostVersion is:

```
        public string LockedHostVersion => IsLocked ? "1.0.1-rc2-002468-00" : LatestHostVersion;
```

So, if IsLocked is turned on, it will use a hard-coded version, and not the current version.

brthor: Hmm this is not good, perhaps LatestHostVersion would be more appropriate.

What I'm worried about here is that the host versions and sharedframework versions are different.

eerhardt: > What I'm worried about here is that the host versions and sharedframework versions are different.

Yup, they are different, but that doesn't matter so much.  The SharedFramework references the DotNetHost packages.  So SharedFX v1.0.0-rc3-002799 references DotNetHost v1.0.1-rc3-002799 (or in some cases it might reference the "locked" version).

So the idea here is that we will upload all the corresponding packages (SharedFX and DotNetHost) to the most consumable package's version folder, namely the SharedFX:  `1.0.0-rc3-002799`.  The DotNetHost packages that correspond to SharedFX `1.0.0-rc3-002799` will be in that folder, even though they have a different version number.

The CLI works the same way.  We add these DotNetHost packages into our SDK installer, and even though our installer has a different version, we still upload the installer to a folder with the installer's version number.  Since it is actually the thing people consume.

I don't think it would make sense to split the SharedFX and the DotNetHost packages into two different Azure folders corresponding to their versions.  We built them together - we put them in the same folder together.

scripts/dotnet-cli-build/PublishTargets.cs
eerhardt: Fixed.

scripts/dotnet-cli-build/PublishTargets.cs
brthor: That sounds reasonable, and I think I misunderstood the purpose here. 

My first take was that we were trying to put artifacts in a versioned folder that matches, but I think what I'm understanding now is just that we're switching the grouping from being under the CLI version to being under the sharedframework version. That sounds fine to me.

scripts/dotnet-cli-build/PublishTargets.cs
eerhardt: Fixed.

scripts/dotnet-cli-build/Publishing/AzurePublisher.cs
eerhardt: Since the parameter is named "fileExtension" calling this DownloadFilesWithExtension seems like it would be redundant.  Also, this is just a utility function that is only used in our build scripts and can easily be renamed/refactored later.

Leaving as is, for now.

scripts/dotnet-cli-build/PublishTargets.cs
eerhardt: This will all go away when #3031 is fixed.

brthor:
:shipit:

PR #2864, opened 2016-05-04
Maintain folder structure in mappings

ajaybhargavb:
Fixes #2829 and #2861 

@troydai @pranavkm @piotrpMSFT

PR #3039, opened 2016-05-14
Fix warnings in project.json

### LINKED ISSUE #3038: Project.json warnings in build
After the schema change at #2493, there are several hundred warnings in the tree.

## Steps to reproduce
build the CLI tree

## Expected  behavior
Warnings only related to tests

## Actual behavior
Several hundred warnings 

## Environment data
`dotnet --info` output:

billwert:
PR #2493 introduced the new project.json schema. The tree has 124 files
with the old schema, which added several hundred warnings.

This change can't go in until PR #2864 does - it relies on those bug
fixes.

This fixes #3038


billwert:
The test failures are due to missing #2864; not sure how y'all typically handle things like this.

eerhardt:
@billwert - #2864 is now in.  Can you rebase and we can get this merged?

billwert:
@eerhardt all good.

test/Microsoft.DotNet.Cli.Utils.Tests/project.json
```diff
@@ -1,8 +1,17 @@
 {
   "version": "1.0.0-*",
-  "compilationOptions": {
+  "buildOptions": {
     "emitEntryPoint": true,
-    "keyFile": "../../tools/Key.snk"
+    "keyFile": "../../tools/Key.snk",
+    "copyToOutput": {
+      "include": [
+        "../../TestAssets/TestProjects/OutputStandardOutputAndError/*",
```
eerhardt: I know you didn't design the format - but IMO this points out a design flaw.  How often do you want the "copyToOutput" files copied both on build and on publish?  Most of the time, I would think.  I would also want them in my "pack" as well ...

eerhardt:
:shipit:

test/Microsoft.DotNet.Cli.Utils.Tests/project.json
billwert: agreed. and you're probably right that I should have added the pack option as well, to match the previous semantic.. @ajaybhargavb - thoughts on this overall? perhaps with msbuild coming it doesn't matter much.

PR #3042, opened 2016-05-15
Update uninstall script to correctly target install root

richlander:
- The .NET Core uninstall script did not work for me. 
- I made this change (which made sense to me upon visual inspection of the script) and then the script worked.

### DIFF
--- a/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
+++ b/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
@@ -31,7 +31,7 @@ remove_dotnet_pkgs
 [ "$?" -ne 0 ] && echo "Failed to remove dotnet packages." && exit 1
 
 echo "Deleting install root - $dotnet_install_root"
-rm -rf $dotnet_install_location
+rm -rf $dotnet_install_root
 rm -f $dotnet_path_file
 
 echo "dotnet packages removal succeeded."
--- a/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
+++ b/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
@@ -17,6 +17,9 @@ dotnet_pkg_name_suffix="com.microsoft.dotnet"
 dotnet_install_root="/usr/local/share/dotnet"
 dotnet_path_file="/etc/paths.d/dotnet"
 
+# nuget paths
+nuget_root=~/.nuget
+
 remove_dotnet_pkgs(){
     installed_pkgs=($(pkgutil --pkgs | grep $dotnet_pkg_name_suffix))
     
@@ -33,6 +36,9 @@ remove_dotnet_pkgs
 echo "Deleting install root - $dotnet_install_root"
 rm -rf $dotnet_install_root
 rm -f $dotnet_path_file
+
+echo "Deleting NuGet package location"
+rm -rf $nuget_root
 
 echo "dotnet packages removal succeeded."
 exit 0
--- a/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
+++ b/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
@@ -8,7 +8,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 current_user=$(whoami)
 if [ $current_user != "root" ]; then
-    echo "`basename "$0"` uninstallation script requires superuser privileges to run"
+    echo "$(basename "$0") uninstallation script requires superuser privileges to run"
     exit 1
 fi
 
@@ -16,9 +16,6 @@ fi
 dotnet_pkg_name_suffix="com.microsoft.dotnet"
 dotnet_install_root="/usr/local/share/dotnet"
 dotnet_path_file="/etc/paths.d/dotnet"
-
-# nuget paths
-nuget_root=~/.nuget
 
 remove_dotnet_pkgs(){
     installed_pkgs=($(pkgutil --pkgs | grep $dotnet_pkg_name_suffix))
@@ -34,11 +31,8 @@ remove_dotnet_pkgs
 [ "$?" -ne 0 ] && echo "Failed to remove dotnet packages." && exit 1
 
 echo "Deleting install root - $dotnet_install_root"
-rm -rf $dotnet_install_root
-rm -f $dotnet_path_file
-
-echo "Deleting NuGet package location"
-rm -rf $nuget_root
+rm -r "$dotnet_install_root"
+rm "$dotnet_path_file"
 
 echo "dotnet packages removal succeeded."
 exit 0
--- a/scripts/obtain/uninstall/dotnet-uninstall-debian-packages.sh
+++ b/scripts/obtain/uninstall/dotnet-uninstall-debian-packages.sh
@@ -8,7 +8,7 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 current_user=$(whoami)
 if [ $current_user != "root" ]; then
-    echo "$(basename $0) uninstallation script requires superuser privileges to run"
+    echo "$(basename "$0") uninstallation script requires superuser privileges to run"
     exit 1
 fi
 


danroth27:
:shipit:

Petermarcu:
@brthor , can you bless this?

richlander:
Added logic to remove NuGet package location. Seems like the correct behavior.

blackdwarf:

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
```diff
@@ -17,6 +17,9 @@ dotnet_pkg_name_suffix="com.microsoft.dotnet"
 dotnet_install_root="/usr/local/share/dotnet"
 dotnet_path_file="/etc/paths.d/dotnet"
 
+# nuget paths
+nuget_root=~/.nuget
```
blackdwarf: Please quote this path as well.

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
```diff
@@ -31,8 +34,11 @@ remove_dotnet_pkgs
 [ "$?" -ne 0 ] && echo "Failed to remove dotnet packages." && exit 1
 
 echo "Deleting install root - $dotnet_install_root"
-rm -rf $dotnet_install_location
+rm -rf $dotnet_install_root
 rm -f $dotnet_path_file
 
+echo "Deleting NuGet package location"
```
blackdwarf: You are not deleting the package location, you are blowing away the entire .nuget dir. :)

blackdwarf:
LGTM modulo two comments.

blackdwarf:
@richlander can you also add, while you're at it, removing the NuGet package cache to the Ubuntu uninstall script? Should be the same logic, actually.

julielerman:
I tried this version of the uninstaller after having some problems, then reinstalling the new CLI bits. Now the dotnet command is not even recognized. Seems like a far shot but is there ANY chance that this removed something that the installer is not giving me? thanks

blackdwarf:
@julielerman more info on what are you using? Are you using `zsh` by any chance? Also, did you try the version of the script in the repo or the one in this PR?

julielerman:
@blackdwarf yes -> zsh! I wouldn't have even dreamt that was related. Is it? And yes, I used the file in the PR.  What else can I tell you.

blackdwarf:
@julielerman told me enough. :) There is an issue in the way `path_helper` is working with `zsh` on OSX. The easiest way for you to get unblocked is to simply symlink the `dotnet` binary to `/usr/local/bin` using the following:

```console
ln -s /usr/local/share/dotnet/dotnet /usr/local/bin
```

Just check the first path in the command pls, I'm not near my Mac so doing this from the top of my head. :smile:

julielerman:
@blackdwarf thanks. And thanks for showing my *how* to symlink not just saying that I need to. I'm fairly new to osx and still following the google/copy/paste method of making my way through most tasks. Will report back shortly.

julielerman:
still trying to find the dotnet libraries.All I've got in local share is nuget and dnu .
Also, FWIW, I didn't have this problem earlier. Are you saing this path helper issue affected the results of the uninstall script or of just running dotnet now. I was able to run dotnet commands earlier after runing the original uninstall script and installing the 2702 pkg. I'll keep looking for dotnet to link

blackdwarf:
@julielerman that's weird. OK, can you open up an issue and do the following:

1. Outline the steps you did
2. Paste in `ls /user/local/share/dotnet` ouput
3. /cc me (@blackdwarf)

We will sort out this problem, must be something...weird.

julielerman:
no such directory.

Okay so ... I ran the original uninstall script.
then I got the 2072 package and installed it. 
to test I did dotnet --version (it gave correct version)
then created a test folder. ran dotnet new then dotnet run to verify.
Then I went to start updating a aspnetcore web api sample solution I'd created a few months ago. after updating project.json and nuget.config, I was having a lot of problems with dotnet restore around the "type": "platform" setting on the microsoft.netcore.app dependency. DotNet restore said it could not recognize platform. Removing that resulted in dotnet restore failing right away with something like (I can't try it now) ..can't find a target starting point ..platform...can't remember.  Various conversations around the platform unrecogznied issue pointed to people not having updated to the latest version of the CLI. That's when I thought that maybe the earlier version hadn't been correctly uninstalled. So I modifiied the uninstaller script using the PR, ran it again, re-ran  the 2072 package and then tried dotnet again and got the "not recognized". I reran the 2072 installer a few times. Restarted the machine. Still totally stuck. but not it seems that it's literally not installing if I can't even find the dotnet folder! What do you think?

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
brthor: in case there's a space in the username, or home folder path for some weird reason

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
```diff
@@ -31,8 +34,11 @@ remove_dotnet_pkgs
 [ "$?" -ne 0 ] && echo "Failed to remove dotnet packages." && exit 1
 
 echo "Deleting install root - $dotnet_install_root"
-rm -rf $dotnet_install_location
+rm -rf $dotnet_install_root
```
Sridhar-MS: Thanks for fixing this. I renamed some variables as I made addressed some PR feedback and it did not test it enough :(

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
brthor: Are we sure we want to blow away the nuget packages on uninstall? What of built (`dotnet build`) Standalone apps that resolve from the package cache? They don't rely on this installation but they do rely on the cache.

Sridhar-MS: Actually there are multiple nuget cache folders you may have to delete
•	~/.local/share/NuGet/v3-cache
•	~/.nuget/packages
•	/tmp/NuGetScratch/

cc: @joelverhagen @emgarten

blackdwarf:
Can we merge this?

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
blackdwarf: @Sridhar-MS I thought that there is only `$HOME/.nuget`? Why do we have these? 

@brthor what do you mean? This is impacting also built shared apps, since you can run your built output out of the NuGet cache.

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
blackdwarf: Yeah. Though, it should all expand with "~". But, better safe than 3x sorry. :smile:

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
tn0502: Nuking anything inside a user's home directory sounds really dangerous to me. If we're really doing this, what about other users' home directories?

eerhardt:
> Can we merge this?

It looks like there is review feedback that needs to be responded to. Has all of the feedback been addressed?

tn0502:
> Nuking anything inside a user's home directory sounds really dangerous to me. If we're really doing this, what about other users' home directories?
>
> IMO, adding a note with instructions is better.

Deleting the user's cache sounds like we're doing more than what the script suggests it does, namely uninstalling the previous version of .NET Core.

eerhardt:
Also see the discussion here:  #3056.

scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh
brthor: @blackdwarf Built shared apps don't have a host dropped app local, so they rely on the muxer being present, this is not true for standalone apps.

Also @tn0502 is right in that if we consider the pkg as a machine wide installation, and the nuget cache as a user local cache, the two things don't exist in the same context. This is especially true for debs, although I'm not sure how true it is for PKGs.

brthor:
https://github.com/dotnet/cli/pull/3056

Brings up good points on using `rm -rf` as root.

eerhardt:
@blackdwarf @richlander - can one of you make the call whether the nuget cache should be cleared or not?  I'd like to push this issue along, but I also want us to be doing the right thing.

whoisjake:
I'll repeat the feedback I left on #3056 and just say please don't use `rm -rf` with elevated privilege...

Tighten the scope of the delete (using find)... 

Also, please don't delete things from $HOME either... (which is interesting, because your script isn't doing that, though it's written with that intention)

Lastly... using `~` as root:
`nuget_root=~/.nuget`

While running elevated is actually: `/var/root`

*Not* the users $HOME dir

👎

Sridhar-MS:
@whoisjake Declaring the following variables read only should alleviate your concern
`declare -r dotnet_install_root="/usr/local/share/dotnet"`
`declare -r dotnet_path_file="/etc/paths.d/dotnet"`

Then `rm -rf` will always act on dotnet files.

blackdwarf:
OK, so I think the consensus is that we don't have NuGet package cache cleaning scripts. This will be a part of the documentation. So, this script should just be fixed to actually work and merged. Right @piotrpMSFT?

eerhardt:
@brthor said he would fix this PR as needed and get it merged.

brthor:
After conversations with @piotrpMSFT and @blackdwarf , we're planning to remove the nuget package clearing here. It doesn't fit in the context of removing the uninstallers, and the user is root when these scripts are executed.

brthor:
Pushed changes, tests are meaningless here, so will merge after :shipit:

Sridhar-MS:
:shipit:

PR #3046, opened 2016-05-16
Add slack channel

benaadams:
skipciplease

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # .NET Command Line Interface
 
-[![Join the chat at https://gitter.im/dotnet/cli](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dotnet/cli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # .NET Command Line Interface
 
-[![Join the chat at https://gitter.im/dotnet/cli](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dotnet/cli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![.NET Slack Status](https://aspnetcoreslack.herokuapp.com/badge.svg?2)](http://tattoocoder.com/aspnet-slack-sign-up/) [![Join the chat at https://gitter.im/dotnet/cli](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dotnet/cli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
```
eerhardt: It feels like the site being pointed to here should be a 'slack' page.

@blackdwarf @piotrpMSFT - any thoughts?

benaadams: That takes it to the slack page an has no sign up or signup customisation so looks like:

slack-signup

Where as the link allows new people to get signed in

blackdwarf:
If we do this, then we need to point to the link @benaadams put in, because people will have no idea how to sign up. Now, whether we should do this...

piotrpMSFT:

eerhardt:
Looks like this same change was made in dotnet/corefx.  Merging.

PR #3047, opened 2016-05-16
Update known-issues.md

blackdwarf:
Add instructions on how to symlink `dotnet` for users of zsh until we fix that one.

skip ci please


eerhardt:
:shipit:

PR #3052, opened 2016-05-16
Fix bug preventing MakeRunnable on dependent project when using globbiing build

pakrym:
Fixes: https://github.com/dotnet/cli/issues/3055

/cc @anurse @cdmihai @eerhardt


TestAssets/TestProjects/AppWithAppDependency/TestApp1/project.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilationOptions": {
```
eerhardt: buildOptions.

TestAssets/TestProjects/AppWithAppDependency/TestApp2/project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "compilationOptions": {
```
eerhardt: buildOptions

src/dotnet/commands/dotnet-build/ProjectBuilder.cs
```diff
@@ -29,7 +29,9 @@ public IEnumerable<CompilationResult> Build(IEnumerable<ProjectGraphNode> roots)
             {
                 using (PerfTrace.Current.CaptureTiming($"{projectNode.ProjectContext.ProjectName()}"))
                 {
-                    yield return Build(projectNode);
+                    var result = Build(projectNode);
```
eerhardt: Would it make more sense to just have Build be virtual, and then DotNetProjectBuilder overrides Build, calls base, and then performs the "after" piece?  This way if someone ever needed to perform a "before" piece they could override Build and do something before calling 'base'.

test/dotnet-build.Tests/BuildOutputTests.cs
```diff
@@ -176,6 +176,25 @@ public void SettingVersionSuffixFlag_ShouldStampAssemblyInfoInOutputAssembly()
             informationalVersion.Should().BeEquivalentTo("1.0.0-85");
         }
 
+        [Fact]
+        public void BuildGlobbingMakesAllRunnable()
+        {
+            var testInstance = TestAssetsManager.CreateTestInstance("AppWithAppDependency")
+                                                .WithLockFiles();
+
+            var cmd = new BuildCommand(string.Format("*{0}project.json", Path.DirectorySeparatorChar), skipLoadProject: true)
```
eerhardt: This can all be one fluent expression:

```C#
new BuildCommand(string.Format("*{0}project.json", Path.DirectorySeparatorChar), skipLoadProject: true)
  .WithWorkingDirectory(testInstance.TestRoot)
  .Execute()
  .Should()
  .Pass();
```

Note you don't need to `ExecuteWithCapturedOutput` if you aren't going to do anything with the captured output.

anurse:
:shipit:

PR #3056, opened 2016-05-16
Fixed uninstaller not removing installation folder.

tn0502:


piotrpMSFT:
@sridhar-ms @robmen can you review?

eerhardt:
Looks like a dupe of #3042

tn0502:
Looks like it. Although, the extra lines for deleting things inside one's home directory sounds really dangerous to me. Not something I'd do lightly.

whoisjake:
Since this is a script running in an elevated privilege state... Please don't use `rm -rf`

The script was literally *one* character away from being dangerous (had there been a trailing slash):
`rm -rf $dotnet_install_location/`

I can see how this script is a workaround for uninstalling previous beta versions for now, so maybe:
`find $dotnet_install_root -delete` or `rm -ri $dotnet_install_root`

The idea with using `find` is that you can tighten the scope of what you want to delete using regex.

eerhardt:
I'm closing this as a dupe of #3042.  Let's make #3042 as good as it can be.

PR #3068, opened 2016-05-17
Eliminate vcredist dependency

schellap:
@gkhanna79 PTAL.

### DIFF
--- a/src/corehost/cli/setup.cmake
+++ b/src/corehost/cli/setup.cmake
@@ -11,6 +11,12 @@ if(WIN32)
     add_compile_options($<$<CONFIG:Release>:-DNDEBUG>)
     add_compile_options($<$<CONFIG:RelWithDebInfo>:-DNDEBUG>)
     add_compile_options($<$<CONFIG:Debug>:/Od>)
+    add_compile_options(/guard:cf) 
+    add_compile_options(/d2Zi+) # make optimized builds debugging easier
+    add_compile_options(/Oi) # enable intrinsics
+    add_compile_options(/Oy-) # disable suppressing of the creation of frame pointers on the call stack for quicker function calls
+    add_compile_options(/GF) # enable read-only string pooling
+    add_compile_options(/FC) # use full pathnames in diagnostics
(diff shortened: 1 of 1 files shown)


schellap:

schellap:
@gkhanna79 CRT changes, PTAL.

src/corehost/cli/setup.cmake
```diff
@@ -11,6 +11,12 @@ if(WIN32)
     add_compile_options($<$<CONFIG:Release>:-DNDEBUG>)
     add_compile_options($<$<CONFIG:RelWithDebInfo>:-DNDEBUG>)
     add_compile_options($<$<CONFIG:Debug>:/Od>)
+    add_compile_options(/guard:cf) 
+    add_compile_options(/d2Zi+) # make optimized builds debugging easier
+    add_compile_options(/Oi) # enable intrinsics
+    add_compile_options(/Oy-) # disable suppressing of the creation of frame pointers on the call stack for quicker function calls
+    add_compile_options(/GF) # enable read-only string pooling
+    add_compile_options(/FC) # use full pathnames in diagnostics
```
gkhanna79: You are missing /MT[d]. This will enable CRT to be statically linked. If you dumpbin the host with your current change, you will see a dependency on vcruntime140?.dll, which needs to go away.

schellap: Those are added in the component binaries in their CMakeLists.txt pls. see 

https://github.com/dotnet/cli/blob/rel/1.0.0/src/corehost/cli/CMakeLists.txt#L9

https://github.com/dotnet/cli/blob/rel/1.0.0/src/corehost/cli/fxr/CMakeLists.txt#L9

https://github.com/dotnet/cli/blob/rel/1.0.0/src/corehost/cli/dll/CMakeLists.txt#L9

gkhanna79: Cool! Just make sure that the debug build does not have any imports and release build will only have imports into ucrt. If that is the case, the change will be good.

PR #3074, opened 2016-05-17
Move RestorePackages to Prepare, not Init.

eerhardt:
RestorePackages will fail if the 'artifacts\rid\corehost` folder doesn't exist. So other targets that depend on Init can't run unless the host has been built.

This is blocking the 'PullNupkgFilesFromBlob' target from publishing our nupkgs.

@brthor


brthor:
Good catch. Thanks for fixing this!
:shipit:

eerhardt:
test Ubuntu x64 Release Build

Issue #2999, opened 2016-05-11
Produce a file with the versions for packages produced as part of build


pranavkm:
There's a file at https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/cli/rel/1.0.0-preview1/Latest_Packages.txt that contains the versions for packages produced by the CLI build. @eerhardt created one for preview1 but there's none available for preview2. This is required for AspNet to pick up the right set of CLI packages.

eerhardt:
I'm working on fixing this.

eerhardt:
This issue is now fixed with #3092 and adding the appropriate build steps.

See https://github.com/dotnet/versions/blob/master/build-info/dotnet/cli/rel/1.0.0/Latest_Packages.txt

PR #3092, opened 2016-05-18
Update github.com/dotnet/versions when new CLI builds are produced.

eerhardt:
Adding a build target to the CLI build that will update the dotnet/versions repo of all the newly published NuGet package Ids and versions. This will get invoked after the "NuGetPublish" steps in our VSO build.

The output is will match: https://github.com/dotnet/versions/blob/master/build-info/dotnet/corefx/master/Latest_Packages.txt

This is basically a port of the PS1 code in corefx: https://github.com/dotnet/corefx/blob/master/UpdatePublishedVersions.ps1

This will be used to push both our DotNetHost* and Microsoft.NETCore.App package information, as well as our CLI nuget packages (ProjectModel, DependencyModel, etc.).  I plan on making these 2 separate files in the dotnet/versions repo.

Fix for #2999

@brthor @dagood @pranavkm


build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.DotNet.Cli.Build
+{
+    public class VersionRepoUpdater
+    {
+        private static Regex s_nugetFileRegex = new Regex("^(.*?)\\.(([0-9]+\\.)?[0-9]+\\.[0-9]+(-([A-z0-9-]+))?)\\.nupkg$");
+
+        private HttpClient _client = new HttpClient();
```
pranavkm: Initializing this as part of `UpdatePublishedVersions` would make it a bit easier to dispose this.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -0,0 +1,126 @@
...
+            await UpdateGitHubFile(packageInfoFilePath, packageInfoFileContent, message);
+        }
+
+        private static List<Tuple<string, string>> GetPackageInfo(string nupkgFilePath)
+        {
+            List<Tuple<string, string>> packages = new List<Tuple<string, string>>();
+
+            foreach (string filePath in Directory.GetFiles(nupkgFilePath, "*.nupkg"))
+            {
+                Match match = s_nugetFileRegex.Match(Path.GetFileName(filePath));
```
pranavkm: How tricky is it to use NuGet.Packaging to read the id and version here?

build_projects/shared-build-targets-utils/shared-build-targets-utils.xproj
```diff
@@ -4,16 +4,16 @@
...
+    <RootNamespace>Microsoft.DotNet.Cli.Build</RootNamespace>
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin</OutputPath>
   </PropertyGroup>
 
   <PropertyGroup>
     <SchemaVersion>2.0</SchemaVersion>
   </PropertyGroup>
-  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.targets" Condition="'$(VSToolsPath)' != ''" />
+  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.targets" Condition="'$(VSToolsPath)' != ''" />
```
pranavkm: Accidental?

pranavkm:
:shipit:

build_projects/shared-build-targets-utils/shared-build-targets-utils.xproj
eerhardt: No. This matches our other .xproj files.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
eerhardt: I haven't tried. I was just porting the power shell script. What is the concern - that the regex would be flaky or the format would change?

pranavkm: Just flakiness - in the event the naming convention breaks it in some way.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -0,0 +1,126 @@
...
+            string gitHubUser = null, 
+            string gitHubEmail = null, 
+            string versionRepoOwner = null, 
+            string versionsRepo = null, 
+            string gitHubAuthToken = null)
+        {
+            _gitHubUser = gitHubUser ?? "dotnet-bot";
+            _gitHubEmail = gitHubEmail ?? "dotnet-bot@microsoft.com";
+            _versionsRepoOwner = versionRepoOwner ?? "dotnet";
+            _versionsRepo = versionsRepo ?? "versions";
```
dagood: Why are these here instead of written as the optional parameters? Behavior's a bit different if you explicitly pass a null--is that a reason?

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
pranavkm: @dagood - we've been using the `3.5.0-*` NuGet APIs to work with our packages. They do crack open the nupkg, but we haven't seen any perf issues when working with over 400+ (CoreFx + CLI + NuGet + AspNetCore) packages.

eerhardt: I was thinking about this more last night. I think this code is going to need to be extremely portable to other places/repos that will update github.com/dotnet/versions.

Given that I'd like to keep the dependencies to a minimum. The regex works for CoreFX and there is minimal advantage to taking the dependency.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            string firstVersionWithPrerelease = publishedPackages
+                .FirstOrDefault(t => t.Item2.Contains('-'))
+                ?.Item2;
+
+            string prereleaseVersion = null;
+            if (!string.IsNullOrEmpty(firstVersionWithPrerelease))
+            {
+                prereleaseVersion = firstVersionWithPrerelease.Substring(firstVersionWithPrerelease.IndexOf('-') + 1);
+            }
```
dagood: This logic is quite different from the powershell--but I see that you aren't making a Latest.txt file and this is just for the message. Do you plan on making one?

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
dagood: @pranavkm Good to know, thanks. For reference, the perf issues I'm referring to were in a dir that now has ~49,000 packages. (Especially bad over a network share.)

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
eerhardt: I haven't decided on that yet.... If we do it would be the full version, not just the pre-release version, like we do for CoreFX.

I think moving the towards the model that you update each individual dependency with its "full new version" is better than the model we started with of updating just the pre-release and leaving the beginning unchanged.

Another issue right now is that CLI publishes two sets of NuGet packages at two different times. 

We can definitely add a Latest.txt file in the future when needed.

dagood: Cool, agreed not to do it now. I think it's good to push in the direction of updating using full-versions, since using just the prerelease causes plenty of issues. I'll make an issue on buildtools for changing upgrade/validation for discussion.

About the two sets: I would assume those would be to different paths, so there could be two Latest.txts...

Going back to this code, the logic here is just for the commit message so it's not too important, but it does seem to duplicate stuff already done by the regex... not sure how flowing that forward vs. recalculating here affects the code readability, up to you.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
eerhardt: `null` is definitely not valid for any of these parameters, they all need to be specified.

When I first wrote this, I was going to get these values from environment variables (like gitHubAuthToken is), which I didn't think is possible with optional parameters.

Another reasoning is how optional parameters work.  The optional value actually gets compiled into the calling assembly.  So if you were to update the optional value without recompiling all your callers, the callers still get the old optional value.  In general, to me it seems to be a good idea to make optional parameter values something that won't ever change, `null`, and then provide their defaults in actual code.

dagood: I see, that makes sense, thanks.

dagood:
LGTM! :+1:

eerhardt:
I've addressed the PR feedback:
- Dispose of the HttpClient
- Use the Regex to get the Prerelease version info.

Thanks for the review.

build_projects/dotnet-cli-build/PackageDependencies.cs
```diff
@@ -1,96 +0,0 @@
-﻿using System;
```
brthor: Duplicate file?

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -550,6 +550,29 @@ private static void DownloadPackagesForPush(string pathToDownload)
...
+            string nupkgFilePath = EnsureVariable("NUPKG_FILE_PATH");
+            string versionsRepoPath = EnsureVariable("VERSIONS_REPO_PATH");
+
+            VersionRepoUpdater repoUpdater = new VersionRepoUpdater();
+            repoUpdater.UpdatePublishedVersions(nupkgFilePath, versionsRepoPath).Wait();
+
+            return c.Success();
+        }
+
+        private static string EnsureVariable(string variableName)
```
brthor: This is useful, and we should probably use it all over the place, can we put it in utils?

build_projects/dotnet-cli-build/PackageDependencies.cs
eerhardt: Yes it was in both dotnet-cli-build and shared-build-targets-utils. And was causing build warnings.

I kept the shared-build-targets-utils one.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -0,0 +1,135 @@
...
+            string gitHubAuthToken = null,
+            string versionRepoOwner = null,
+            string versionsRepo = null)
+        {
+            _gitHubUser = gitHubUser ?? "dotnet-bot";
+            _gitHubEmail = gitHubEmail ?? "dotnet-bot@microsoft.com";
+            _versionsRepoOwner = versionRepoOwner ?? "dotnet";
+            _versionsRepo = versionsRepo ?? "versions";
+
+            _gitHubAuthToken = gitHubAuthToken ?? Environment.GetEnvironmentVariable("GITHUB_PASSWORD");
```
brthor: Should we have the fallback here? Why not just let this get passed in?

brthor: Also this would be a good place for the `EnsureVariable`

eerhardt: > Should we have the fallback here? Why not just let this get passed in?

Not sure I follow this question.

> Also this would be a good place for the EnsureVariable

I wouldn't be using it here for 2 reasons:
1. to keep this code self contained with minimal dependencies.
1. Because I want an ArgumentException here.

brthor: The first question was pertaining to the scope of this class. It sets a bunch of default values to strings, but pulls one default value from an environment variable. I'm wondering if it would make the class more intuitive to use if it required the GitHub auth token to be passed in instead of defaulting to the environment variable.

eerhardt: I can change that.  It doesn't matter to me either way.

eerhardt:
test OSX x64 Release Build

PR #3100, opened 2016-05-18
Made path to delete binaries correct in the uninstall script

jchannon:
skipciplease

Unless `$dotnet_install_location` is set somewhere else I think this will fail to delete anything


jchannon:
Aha this is a duplicate of #3042

PR #3104, opened 2016-05-18
Fix build minor and add sccsid

schellap:


schellap:

build_projects/shared-build-targets-utils/Utils/HostVersion.cs
```diff
@@ -24,7 +24,8 @@ public class HostVersion : Version
         public string LatestHostPolicyVersionNoSuffix => $"{Major}.{Minor}.{Patch}";
         public string LatestHostPrerelease => ReleaseSuffix;
         public string LatestHostBuildMajor => $"{CommitCountString}";
-        public string LatestHostSuffix => $"{ReleaseSuffix}-{LatestHostBuildMajor}-00";
+        public string LatestHostBuildMinor => "00";
```
eerhardt: Why do we even have this part?  Why not just have the commit count and remove `-00` from all usages?

brthor: my impression was that it was in case we needed to manually patch a package, but I agree the commit count should be sufficient.

schellap:
This is some weird characteristic of the pkgproj mechanism. I'm just doing it to make it happy. Take a look at all the native pkgs in dotnet-core feed.

PR #3107, opened 2016-05-18
Fix stub package generation to just depend on System.Console.

eerhardt:
NuGet was complaining about some packages having compile-time assemblies, but not having runtime assemblies. Removing a bunch of unnecessary dependencies fixes this.

Currently our x86 VSO build is broken.

@brthor

### DIFF
--- a/build_projects/dotnet-host-build/StubPackageBuilder.cs
+++ b/build_projects/dotnet-host-build/StubPackageBuilder.cs
@@ -42,9 +42,8 @@ namespace Microsoft.DotNet.Host.Build
         {
             var projectJson = new StringBuilder();
             projectJson.Append("{");
-            projectJson.Append("  \"dependencies\": { \"NETStandard.Library\": \"1.5.0-rc2-24027\" },");
-            projectJson.Append("  \"frameworks\": { \"netcoreapp1.0\": { \"imports\": [\"netstandard1.5\", \"dnxcore50\"] } },");
-            projectJson.Append("  \"runtimes\": { \"win7-x64\": { } },");
+            projectJson.Append("  \"dependencies\": { \"System.Console\": \"4.0.0-*\" },");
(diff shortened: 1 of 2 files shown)


brthor:
This time x86 went down? 

Can you try adding Microsoft.NETCore.Platforms? It's possible not having the rid graph could cause this.

build_projects/dotnet-host-build/StubPackageBuilder.cs
```diff
@@ -42,9 +42,8 @@ private void BuildStubBits(DotNetCli dotnet, string intermediateDirectory)
         {
             var projectJson = new StringBuilder();
             projectJson.Append("{");
-            projectJson.Append("  \"dependencies\": { \"NETStandard.Library\": \"1.5.0-rc2-24027\" },");
-            projectJson.Append("  \"frameworks\": { \"netcoreapp1.0\": { \"imports\": [\"netstandard1.5\", \"dnxcore50\"] } },");
-            projectJson.Append("  \"runtimes\": { \"win7-x64\": { } },");
```
brthor: Why take out the rid?

eerhardt:

build_projects/dotnet-host-build/StubPackageBuilder.cs
eerhardt: It isn't necessary, and is wrong on 6/7 of our builds.

brthor: If it works then it doesn't matter, the rid here is meaningless with respect to the actual platform though. The entire intention is to produce a stub package with the current rid in the *name*. The contents of those packages don't matter.

eerhardt: Right - I was even throwing around an idea about making an empty package, with no dependencies, but I kept getting "EmptyPackage" errors.

brthor: I tried too :)

PR #3134, opened 2016-05-19
Change stub packages to no longer contain any dependencies.  

eerhardt:
Instead they just contain a dummy.txt file.

@brthor


brthor:
:shipit:

Issue #3051, opened 2016-05-16
Better Dianostic if you don't do a dotnet restore before a dotnet build


vancem:
When I was dogfooding the RC2 bits  https://go.microsoft.com/fwlink/?LinkId=798481 

I issued the following command 

mkdir app
cd app
dotnet new 
dotnet run

Produces the following diagnostic.  

Project hello does not have a lock file.
Project hello does not have a lock file.

I realize this is now an error, as you are now required to do the dotnet restore, however the current diagnostic does nothing to lead the user in the correct direction.   At the least it should suggest that 
a dotnet restore will solve the problem. 

This is particularly bad because dotnet restore did not need to be done in RC1, so it is likely that it will be a common mistake.

colemickens:
It's also confusing if you have a lock file already but haven't restored packages yet (or cleared your cache, etc). I don't have the exact error, but it's not super user friendly.

blackdwarf:
This is now fixed in #3150 . @vancem can you verify and close if it behaves correctly? This is on the latest.

PR #3150, opened 2016-05-20
Add 'dotnet restore' instruction when lock file is missing (#3051)

ErikSchierboom:
As per #3051, if you run `dotnet run` but forgot to run `dotnet restore` before that, the error message only states what is wrong (missing lock file), not how to fix this. This PR adds an instruction to the error message that states the user should run `dotnet restore`. The exact instruction has been copied from the instruction used in Visual Studio when this error occurs.


eerhardt:
:shipit:

PR #3160, opened 2016-05-20
perf tests: add fallback feed to dotnet restore

dpodder:
The perf tests use `dotnet restore` to download test dependencies. The
perf tests' project.json may sometimes refer to a version of the
framework that is not yet published, which can cause the restore command
to fail. This change addresses the issue by adding a fallback nuget feed
that contains not-yet-published version of the framework.


dpodder:
@lt72 @brianrob PTAL

brthor:
:shipit:

lt72:
good for me

piotrpMSFT:
:shipit:

PR #3161, opened 2016-05-20
remove check for gettext dependency in install.sh

### LINKED ISSUE #3108: Install Script should not look for `gettext` dependency
https://github.com/dotnet/cli/blob/rel/1.0.0/scripts/obtain/dotnet-install.sh#L124

cc @ellismg @krwq

krwq:
Fixes #3108


piotrpMSFT:
:shipit:

brthor:
nice! :shipit:

Issue #2988, opened 2016-05-11
Installing packages appends many times to /etc/paths.d/dotnet


vcsjones:
## Steps to reproduce

Run the .pkg for the SDK installer multiple times. Can be the same version, can be different ones.

## Expected  behavior

The `/etc/paths.d/dotnet` file contains a single line to the path of where the CLI tools are run.

## Actual behavior

Every time the package installer is run, another path line is added:

```
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
/usr/local/share/dotnet
..etc
```

I only discovered this because it looks like at some point, the path was once `/usr/local/share/dotnet/bin`, which was getting added to my path and causing problems.

## Environment data
`dotnet --info` output:

>.NET Command Line Tools (1.0.0-preview1-002700)
>
>Product Information:
> Version:     1.0.0-preview1-002700
> Commit Sha:  0336f6bb34

>Runtime Environment:
> OS Name:     Mac OS X
> OS Version:  10.11
> OS Platform: Darwin
> RID:         osx.10.11-x64

oold:
Apparently, it also doesn't clean old versions. My old dotnet binary still exists in the bin directory.

vcsjones:
Migrated to dotnet/core-setup#18.

Issue #1567, opened 2016-02-24
Issues installing on OSX using ZSH


earlbalai:
The path to dotnet was not added to my zshrc automatically during installation. After manually adding it myself it works like a charm. 

os: 10.11
zsh version: zsh 5.0.8 (x86_64-apple-darwin15.0)
dotnet version: 1.0.0-beta-001517 

@blackdwarf

agocke:
Maybe if we detect a ~/.zshrc we can print a message suggesting they add the os x `path_helper` call to their zshrc? I don't think we should change peoples' `.rc` files without approval.

blackdwarf:
@agocke no, changing people's .rc files is out of question. I'm still confused about why this works on my machine, so repro time. :)

brthor:
This shouldn't beed to be added to the .rc file right, we symlink to /usr/bin?

blackdwarf:
Found this http://unix.stackexchange.com/questions/22979/path-helper-and-zsh. According to this, `path_helper` is not getting called in zsh. This would explain @earlbalai and @barambani issues with not having the tools in the $PATH. 

There are two ways to solve this:

1. We detect and modify the .zshrc file by `echo "export PATH=$PATH:/usr/local/share/dotnet >> .zshrc` or similar. This would solve the problem but is invasive.
2. We print out the warning that if you use zsh you need to add this to your path or invoked the `path_helper` from the profile.

blackdwarf:
Of course, this still doesn't explain why is it working for me :(

blackdwarf:
@earlbalai @barambani can you guys check if you have a file in `/etc/` which is called `zprofile` and does it contain anything?

brthor:
@blackdwarf What if we symlink to /usr/bin again? Are the bugs preventing that still happening?

blackdwarf:
@brthor no, but...seems *wrong* to do it. Also, would like to know if @earlbalai has El Capitan, 'cause it seems that the zprofile is something that El Capitan brings by default, which would explain why the installer "just worked" on my machine.

barambani:
no `zprofile` in `/etc/` for me, but I have a file `zshenv` there that contains the lines

```
# system-wide environment settings for zsh(1)
if [ -x /usr/libexec/path_helper ]; then
	eval `/usr/libexec/path_helper -s`
fi```

I'm on 10.10.5 Yosemite

blackdwarf:
[...]

PR #3167, opened 2016-05-20
Adding the symlink to dotnet binary to /usr/bin

### LINKED ISSUE #3063: dotnet: command not found Mac OS X 10.11.4, dotnet-osx-x64.1.0.0-rc2-3002702.pkg
## Steps to reproduce
1. Run the latest released .NET Core `pkg` installer: `dotnet-osx-x64.1.0.0-rc2-3002702.pkg`.
2. Pull up a fresh instance of your preferred shell.
3. Run `dotnet`

## Expected  behavior
`dotnet` executes and probably prints usage information because no specific command was passed.

## Actual behavior
I'm given this output:
```shell
zsh: command not found: dotnet
```

## Environment data
`dotnet --info` output:

```shell
shawn ~  dotnet --info
zsh: command not found: dotnet
```
¯\\\_(ツ)\_/¯ 

```shell
normandy:~ shawn$ /usr/local/share/dotnet/dotnet --info
.NET Command Line Tools (1.0.0-preview1-002702)

Product Information:
 Version:     1.0.0-preview1-002702
 Commit Sha:  6cde21225e

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.11
 OS Platform: Darwin
 RID:         osx.10.11-x64
```

--

This package was found at `https://go.microsoft.com/fwlink/?LinkID=798388`, and was not the link specified at the actual instruction guide at `https://www.microsoft.com/net/core#macosx`.

Verifying the hash:
```shell
shawn ~ shasum ~/Downloads/dotnet-osx-x64.1.0.0-rc2-3002702.pkg -a 512
5f00fd91e1bbffe55d091c398913847bab0766ab00987d9164826a1c15f2b05eb8b56f801c0625b75161237dd31eb9180b6468b2bbd9c85077caf2fb4cae6e7a  /Users/shawn/Downloads/dotnet-osx-x64.1.0.0-rc2-3002702.pkg
```

It's also worth mentioning that `dotnet` does, in fact, exist at `/usr/local/share/dotnet/dotnet`, ~~but there's definitely no symlink in `/usr/local/bin/`.~~ While there's definitely no symlink, the more I read about this issue, the more it seems like that's intentional. This is most likely actually a duplicate of #1567 and highlights the insufficiency of using `path_helper` for zsh users.

blackdwarf:
This change removes adding the path through /etc/paths.d and
goes to adding a symlink to /usr/bin. The reason is that there is
a lot of users who are running into problems with zsh and oh-my-zsh has
a bug in the installer where they generate a .zshrc file with an
exploded $PATH at the time of install. This overrides the dynamic path
that is created with path_helper.

Fixes #3063, #2988, #1567

/cc @brthor @Sridhar-MS @piotrpMSFT


vcsjones:
I don't think this will work in El Capitan's "rootless" mode.  Even as root this will spit out "Operation not permitted" since this is a protected operating system path.

I would suggest using `/usr/local/bin` instead.

blackdwarf:
@vcsjones ah, good one. OK, fix is easy enough. However, is the `/usr/local/bin` per-user?

vcsjones:
@blackdwarf No. System wide things typically go in there. However, `/usr/local` may not exist on a clean OS X install (Yosemite). You may have to create it and the `bin` directory. It should be in `$PATH` even if the directory doesn't exist yet. If it's useful I can spin up a clean OS 10.11 install and let you know what is there by default and what isn't.

blackdwarf:
@vcsjones that would be useful, thanks! I don't have access to a clean machine now.

vcsjones:
@blackdwarf 

For 10.11:

* `/usr/local` exists. There is nothing at all inside of it.
* `/usr/local/bin` does *not* exist.
* The default `$PATH` is `/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin`.

So, for 10.11 it sounds like you want to make the `bin` directory under `/usr/local` if it doesn't exist.

The default permissions on `/usr/local` is `drwxr-xr-x` which is 755.

vcsjones:
I think this could be your installer could do this then:

```
mkdir -p /usr/local/bin && ln -s $INSTALL_DESTINATION/dotnet/dotnet /usr/local/bin
```

blackdwarf:
@vcsjones nice, thanks! :)

vcsjones:
@blackdwarf if you can point me toward how to make the PKG or a nightly after this is merged in I can try it on the clean OSX instance.

blackdwarf:
@vcsjones you can checkout the branch from my fork and just run build.sh. By default, it will create the PKGs in `artifacts/` directory under `packages`.

brthor:
We'll need to move this over to github.com/dotnet/core-setup as soon as the initial PR goes in

dasMulli:
The uninstall script should also be updated to remove the symlink. It would stay around in the path and would cause an error message since its target is gone. Shouldn't be much of a problem as long as install paths don't change but a hell to investigate. (currently only a problem if you uninstall preview1 and add a custom stage2 to your PATH).

dasMulli:
Also, maybe remove any old symlink before creating a new one.. `ln` will fail if `/usr/local/bin` already exists (and possibly points to sth else rendering the installation unusable)

brthor:
Can we move this to https://github.com/dotnet/core-setup

blackdwarf:
@brthor yes, we can. However, I'm starting to feel that we maybe shouldn't "solve" this. In essence, we are adding a thing to our installer to work around an issue in oh-my-zsh installer. We don't have a bug.

dasMulli:
It's probably enough to document the issue somewhere (known issues, FAQ?). Both `bash` and `zsh` shipping with OS X work fine with `/etc/paths.d` .
Go also decided to not take any action in the installer https://github.com/golang/go/issues/11052
There also is some (slow) discussion and 2 standing pull requests on the installer in the oh-my-zsh repo.

vcsjones:
If `paths.d` stays, which I have no issues with, perhaps I should close #2988 here are re-open it on core-setup.

blackdwarf:
OK, after doing some more thinking, I think that we should not do this. It is clear that this is a bug in a different product. We are using a system-level support that is documented and valid. Oh-my-zsh already has at least one PR to solve this issue. We will document this properly in the known issues document here in the repo and on the getting started site. 

@vcsjones can you open the issue you referenced on dotnet/core-setup and close it here?

PR #3182, opened 2016-05-22
Move PullNupkgFilesFromBlob into FinalizeBuild.

### LINKED ISSUE #3031: Remove build dependency on VSO NuGet Publisher
Currently our builds have a dependency on the VSO NuGet publisher in order to push our packages to myget. 

This dependency is only available on Windows and forces us to workaround with separate build definitions only for handling publishing of our nuget packages. 

NuGet.Xplat has the push command available and we should consider switching to that to remove this dependency and the need to work around it.

cc @Sridhar-MS @piotrpMSFT @eerhardt @sokket

eerhardt:
Removing the need for a separate VSO build definition to push our nuget packages to MyGet by adding a NuGetUtil.PushPackages method that calls into NuGet XPlat CommandLine "push" command.

Also, update dotnet/versions for the DotNetHost and NetCore.App packages.

Fix #3031

@sokket @brthor @livarcocc 

/cc @weshaggard

### DIFF
--- a/build_projects/dotnet-cli-build/NuGetUtil.cs
+++ b/build_projects/dotnet-cli-build/NuGetUtil.cs
@@ -1,0 +1,32 @@
+﻿using System.IO;
+using System.Reflection;
+using Microsoft.DotNet.Cli.Build.Framework;
+using NugetProgram = NuGet.CommandLine.XPlat.Program;
+
+namespace Microsoft.DotNet.Cli.Build
+{
+    public static class NuGetUtil
+    {
+        public static void PushPackages(string packagesPath, string destinationUrl, string apiKey)
+        {
+            int result = RunNuGetCommand(
+                "push",
+                "-s", destinationUrl,
+                "-k", apiKey,
+                Path.Combine(packagesPath, "*.nupkg"));
+
+            if (result != 0)
+            {
+                throw new BuildFailureException($"NuGet Push failed with exit code '{result}'.");
+            }
+        }
+
+        private static int RunNuGetCommand(params string[] nugetArgs)
+        {
+            var nugetAssembly = typeof(NugetProgram).GetTypeInfo().Assembly;
+            var mainMethod = nugetAssembly.EntryPoint;
+            return (int)mainMethod.Invoke(null, new object[] { nugetArgs });
+        }
+    }
+}
+
--- a/build_projects/dotnet-cli-build/PrepareTargets.cs
+++ b/build_projects/dotnet-cli-build/PrepareTargets.cs
@@ -62,17 +62,8 @@ namespace Microsoft.DotNet.Cli.Build
         [Target]
         public static BuildTargetResult GenerateVersions(BuildTargetContext c)
         {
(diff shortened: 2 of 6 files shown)


build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -62,17 +62,8 @@ public static BuildTargetResult Init(BuildTargetContext c)
...
-                .Execute();
-            gitResult.EnsureSuccessful();
-            var commitCount = int.Parse(gitResult.StdOut);
-
-            gitResult = Cmd("git", "rev-parse", "HEAD")
-                .CaptureStdOut()
-                .Execute();
-            gitResult.EnsureSuccessful();
-            var commitHash = gitResult.StdOut.Trim();
+            var commitCount = GitUtils.GetCommitCount();
```
brthor: 👍

brthor:
Small comment on moving a file, if you don't want to move it here, I can move it when I bring this over to core-setup.
:shipit:

build_projects/dotnet-cli-build/NuGetUtil.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.IO;
```
eerhardt: The reason this is in `dotnet-cli-build` is because it needs to reference "NuGet.CommandLine.XPlat.Program", which is targeting `netcoreapp1.0`.

If I put it in the shared project, I would need to move the shared project to target `netcoreapp1.0`. Do you think that is acceptable?

brthor: That's a good point. Moving the shared project to target that will probably work, but if we don't do it we'll need to duplicate this between the host and cli builds, which we'll need to do anyways so It's fine to leave it I think.

eerhardt:
@weshaggard - FYI with this change I have now disabled the `weshaggard-CLI-PackagePublish` build in VSO.

weshaggard:
@eerhardt Sounds good, we can probably just delete it completely once it is confirmed to not be needed.

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -136,6 +138,22 @@ private static void CopyBlobs(string sourceFolder, string destinationFolder)
...
+            var hostBlob = $"{Channel}/Binaries/{SharedFrameworkNugetVersion}";
+
+            Directory.CreateDirectory(Dirs.PackagesNoRID);
+            AzurePublisherTool.DownloadFiles(hostBlob, ".nupkg", Dirs.PackagesNoRID);
+
+            string nugetFeedUrl = EnvVars.EnsureVariable("NUGET_FEED_URL");
+            string apiKey = EnvVars.EnsureVariable("NUGET_API_KEY");
+            NuGetUtil.PushPackages(Dirs.PackagesNoRID, nugetFeedUrl, apiKey);
+
+            string githubAuthToken = EnvVars.EnsureVariable("GITHUB_PASSWORD");
```
weshaggard: This seems kind of scary too me.

PR #3225, opened 2016-05-24
Remove updating dotnet/versions repo with the latest shared framework version.

eerhardt:
This is breaking the build, and it should be done in core-setup instead.

@brthor


brthor:
:shipit:

PR #3227, opened 2016-05-25
Specify KB information in the prerequisites

schellap:
skip ci please

Since we use LoadLibraryExW with flags that require hotfix KB2533623 to be installed on Windows 7 and Windows Server 2008 R2 and Windows Server 2008, we need to update our pre-requisites to reflect this.
See: https://msdn.microsoft.com/en-us/library/windows/desktop/ms684179(v=vs.85).aspx

https://support.microsoft.com/en-us/kb/2533623
This is a 5 year old KB, but there are OSes that may not have updated, yet want to use .NET Core.

@gkhanna79, @blackdwarf

### DIFF
--- a/Documentation/cli-prerequisites.md
+++ b/Documentation/cli-prerequisites.md
@@ -12,7 +12,7 @@ On Windows, the only dependency is the VC++ Redistributable. Depending on the ve
 * Windows 10
     * [Visual C++ Redistributable for Visual Studio 2015](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
 * Pre-Windows 10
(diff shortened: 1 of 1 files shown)


gkhanna79:
@schellap Is there a Windows Update number we can call out here? Otherwise, LGTM.

@blackdwarf Is this the right place to update the pre-req documentation for .NET Core for RTM?

blackdwarf:
@gkhanna79 yes, it is.

schellap:
> Is there a Windows Update number we can call out here?

The KB number is called out and linked in the md file. Is there a separate number for WU?

Issue #1623, opened 2016-02-28
dotnet restore needs a less chatty default verbosity level


blackdwarf:
Currently, the default verbosity level is set to output the entire operation log to STDOUT. For new users this can be daungting. We should switch to a less chatty verbosity level by default. 

/cc @billwert @piotrpMSFT

natemcmaster:
:+1: for quieter default. For big projects like EntityFramework, the default output of dotnet-restore can be literally thousands of lines long. Here is an idea we proposed for dnu restore but never implemented: https://github.com/aspnet/dnx/issues/251#issuecomment-130088659

blackdwarf:
@natemcmaster for RTM we could go a long way if we default to `--verbosity minimal` honestly. It seems to provide success/fail response back to the user.

emgarten:
Minimal should have everything but the HTTP request messages. I am fine with this change since the output can be overwhelming, but to provide some background on why network calls are shown by default today, it is to let the user know which source feed is slowing down the restore. Without this it can look like look like restore is hanging when it's actually being caused by a misbehaving source that the user may not realize they have.

Maybe NuGet could provide a short summary or warning if a source is extremely slow to solve the original problem without writing out every request.

blackdwarf:
@emgarten thanks for the clarification. NuGet showing such errors in general in any verbosity level that displays output would be a good addtion.

billwert:
@emgarten Having a summary of speed per source available would be good. The current thing doesn't print timings anyway, so I'd just be left to "feeling" which of the sources was slower or faster. (Unless one was just timing out but I'd like to think that's rareish.

blackdwarf:
@billwert you may time out on a feed that is not nuget.org/official feed. For instance, you added a myget feed to your global Nuget.config file.

billwert:
@blackdwarf Sure. Is the general case of using myget unstable enough that I need a lot of diagnostic output every time I run the tool? Or is it typically true that once I have a myget feed figured out and am using it in a stable way it will just stop working a lot?

blackdwarf:
@billwert my point was just an explanation that you could be in a situation where the restore is slow/failing because in some config file you have a misbehaving feed. Replace "myget" with "internal corporate feed on a 386SX" and similar applies. :)

emgarten:
@blackdwarf exactly :)

billwert:
[...]

PR #3229, opened 2016-05-25
Reduce nuget output

billwert:
This addresses part of #1623. Unfortunately, because the CLI takes Nuget
as a binary, it is hard to get to where I think we should really be.
This change makes default verbosity "minimal", which is the first level
where you get any status output. Unfortunately, things like package
downgrade warnings and the like still appear there. This does hide all
the "info" and "trace" messages by default.

I also removed the now useless (and previously undocumented)
--quiet.

### DIFF
--- a/src/dotnet/commands/dotnet-restore/NuGet3.cs
+++ b/src/dotnet/commands/dotnet-restore/NuGet3.cs
@@ -10,13 +10,13 @@ namespace Microsoft.DotNet.Tools.Restore
 {
     internal static class NuGet3
     {
-        public static int Restore(IEnumerable<string> args, bool quiet)
+        public static int Restore(IEnumerable<string> args)
         {
             var prefixArgs = new List<string>();
-            if (quiet)
+            if (!args.Any(s => s.Equals("--verbosity", StringComparison.OrdinalIgnoreCase) || s.Equals("-v", StringComparison.OrdinalIgnoreCase)))
             {
                 prefixArgs.Add("--verbosity");
-                prefixArgs.Add("Error");
+                prefixArgs.Add("minimal");
             }
             prefixArgs.Add("restore");
 
--- a/src/dotnet/commands/dotnet-restore/Program.cs
+++ b/src/dotnet/commands/dotnet-restore/Program.cs
@@ -24,16 +24,12 @@ namespace Microsoft.DotNet.Tools.Restore
                 Description = "Restores dependencies listed in project.json"
             };
 
-            // Parse --quiet, because we have to handle that specially since NuGet3 has a different
-            // "--verbosity" switch that goes BEFORE the command
-            var quiet = args.Any(s => s.Equals("--quiet", StringComparison.OrdinalIgnoreCase));
-            args = args.Where(s => !s.Equals("--quiet", StringComparison.OrdinalIgnoreCase)).ToArray();
 
             app.OnExecute(() =>
             {
                 try
                 {
-                    return NuGet3.Restore(args, quiet);
+                    return NuGet3.Restore(args);
                 }
                 catch (InvalidOperationException e)
                 {
--- a/test/EndToEnd/EndToEndTest.cs
+++ b/test/EndToEnd/EndToEndTest.cs
@@ -218,7 +218,7 @@ namespace Microsoft.DotNet.Tests.EndToEnd
             Directory.SetCurrentDirectory(RestoredTestProjectDirectory);
 
             new NewCommand().Execute().Should().Pass();
-            new RestoreCommand().Execute("--quiet").Should().Pass();
+            new RestoreCommand().Execute().Should().Pass();
 
             Directory.SetCurrentDirectory(currentDirectory);
         }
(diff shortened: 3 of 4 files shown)


brianrob:
👍 I like it!

src/dotnet/commands/dotnet-restore/Program.cs
```diff
@@ -24,16 +24,12 @@ public static int Run(string[] args)
                 Description = "Restores dependencies listed in project.json"
             };
 
-            // Parse --quiet, because we have to handle that specially since NuGet3 has a different
-            // "--verbosity" switch that goes BEFORE the command
```
eerhardt: If NuGet uses `--verbosity` BEFORE the "restore" command, does passing in `--verbosity` into `dotnet restore` work?

Ex. `dotnet restore --verbosity Verbose`.  Does that actually use print out verbose messages?

brthor:
:+1: Like it as well!

src/dotnet/commands/dotnet-restore/Program.cs
billwert: Yes, passing --verbosity|-v does work. (In the build log from CI you can see some commands passing --verbosity and getting full output.)

Frankly, I'm not sure what that comment really meant.

blackdwarf:
:shipit:

piotrpMSFT:
:shipit:

src/dotnet/commands/dotnet-restore/NuGet3.cs
```diff
@@ -10,13 +10,13 @@ namespace Microsoft.DotNet.Tools.Restore
...
-        public static int Restore(IEnumerable<string> args, bool quiet)
+        public static int Restore(IEnumerable<string> args)
         {
             var prefixArgs = new List<string>();
-            if (quiet)
+            if (!args.Any(s => s.Equals("--verbosity", StringComparison.OrdinalIgnoreCase) || s.Equals("-v", StringComparison.OrdinalIgnoreCase)))
             {
                 prefixArgs.Add("--verbosity");
-                prefixArgs.Add("Error");
+                prefixArgs.Add("minimal");
```
eerhardt: Do we need prefixArgs at all?  

Also, I don't feel it is the CLI's place to decide what the "default" verbosity should be.  We should just pass through to NuGet, and NuGet gets to decide what the "default" verbosity is.

eerhardt: What if we just said:

```
 var result = Run(Enumerable.Concat(
                     new [] { "restore" },
                     args).ToArray());
```

billwert: The default verbosity seems to be "Information", the level above "verbosity." Thus we need prefixArgs to set it to minimal. 

IMO (and the O of the folks in the issues about this) Information is too spammy. We could push the change all the way down, but I think it is very reasonable that the CLI, as a thing that uses NuGet as a dependency, gets to decide what the CLI shows CLI consomers.

eerhardt: I agree "Information" is too spammy.

I would really like to see NuGet "do the right thing here", and not have this "default" in CLI.

@yishaigalatzer @emgarten @rrelyea  - care to comment?

eerhardt:
:shipit:

billwert:
As long as we're tagging our nuget folks, how is "package downgrade warning" supposed to be actionable? it's not clear to me what I as a developer am supposed to do with this information, and googling with bing's first few hits are for issues in the Nuget repo, not something helpful to customers.

yishaigalatzer:
This sounds wrong, the restore command should be an absolute pass through

eerhardt:
> This sounds wrong, the restore command should be an absolute pass through

I totally agree. But I also think that the NuGet default should be changed.  "Information" is too much.

yishaigalatzer:
I Don't disagree. Another side question - if such flags are set, how would a user override them

billwert:
@yishaigalatzer not sure what you mean. -v|--verbosity are still passed through.. are you talking about something else.

also, is there a doc or explanation of what the package downgrade warning means?

yishaigalatzer:
I mean that these are flags for a user to specify, how does a user override what is pre-spec'd?

Downgrade warnings are pretty simple, they are supposed to capture bad project dependencies, but unfortunately many of the packages are poorly authored and expose them. There is no specific doc for them, but you can go on the nuget docs site and read about resolution rules, that would be the closest

eerhardt:
@yishaigalatzer - if the user specifies a flag, it gets passed through. The CLI code is checking if the user doesn't specify any flags, only then does it put the default verbosity:

```C#
            if (!args.Any(s => s.Equals("--verbosity", StringComparison.OrdinalIgnoreCase) || s.Equals("-v", StringComparison.OrdinalIgnoreCase)))
              {
                  prefixArgs.Add("--verbosity");
                  prefixArgs.Add("minimal");
```

yishaigalatzer:
And what if the user passes --v ?

billwert:
That's not a legal option on the CLI.

PR #3230, opened 2016-05-25
Remove host + sharedfx build from CLI Repo

brthor:
This PR removes the host and shared fx build from this repo and sets up the consumption from the `core-setup` repo.

Currently it is set up to consume from the `core-setup-testing` myget feed and channel. 

After we see this PR go green, I will update the core-setup repo to push to our preview feeds and then update this PR to consume from there.

The reasoning for this is that currently all of the tests for the host and sharedfx live in the cli. We should work to create tests in the `core-setup` repo but for now the cli holds validation.

PTAL @eerhardt @piotrpMSFT @weshaggard @schellap @sokket @livarcocc


scripts/update-dependencies/UpdateFilesTargets.cs
```diff
@@ -190,7 +190,7 @@ private static IEnumerable<JProperty> FindAllDependencyProperties(JObject projec
         [Target]
         public static BuildTargetResult ReplaceCrossGen(BuildTargetContext c)
         {
-            ReplaceFileContents(@"build_projects\dotnet-cli-build\CompileTargets.cs", compileTargetsContent =>
+            ReplaceFileContents(@"build_projects\shared-build-targets-utils\BuildPackageVersions.cs", compileTargetsContent =>
```
eerhardt: Should be "DependecyVersions.cs", right?

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
```diff
@@ -0,0 +1,219 @@
...
+            string corehostLatestDirectory,
+            string corehostPackageSource,
+            string sharedFrameworkNugetVersion)
+        {
+            _repoRoot = repoRoot;
+            _corehostLockedDirectory = corehostLockedDirectory;
+            _corehostLatestDirectory = corehostLatestDirectory;
+            _corehostPackageSource = corehostPackageSource;
+
+            _sharedFrameworkTemplateSourceRoot = Path.Combine(repoRoot, "src", "sharedframework", "framework");
```
eerhardt: We shouldn't need this whole file anymore, right?  This path doesn't exist in the CLI repo anymore.

build_projects/shared-build-targets-utils/Utils/Dirs.cs
```diff
@@ -27,6 +27,7 @@ public static class Dirs
         public static readonly string CorehostLocked = Path.Combine(Output, "corehost", "locked");
         public static readonly string CorehostLocalPackages = Path.Combine(Output, "corehost");
         public static readonly string CorehostDummyPackages = Path.Combine(Output, "corehostdummypackages");
+        public static readonly string SharedFrameworkPublish = Path.Combine(Output, "sharedFrameworkPublish");
```
eerhardt: What do you feel about this being under "intermediate", since it really isn't an "output" of the CLI build.  It is just something on the way of getting to the real output.

eerhardt:
>         public static readonly string CorehostLatest = Path.Combine(Output, "corehost"); // Not using Path.Combine(Output, "corehost", "latest") to keep signing working.

A lot of these "corehost" dirs can be remove now, right?

---
Refers to: build_projects/shared-build-targets-utils/Utils/Dirs.cs:26 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

build_projects/shared-build-targets-utils/Utils/Dirs.cs
```diff
@@ -27,6 +27,7 @@ public static class Dirs
         public static readonly string CorehostLocked = Path.Combine(Output, "corehost", "locked");
         public static readonly string CorehostLocalPackages = Path.Combine(Output, "corehost");
         public static readonly string CorehostDummyPackages = Path.Combine(Output, "corehostdummypackages");
```
eerhardt: Please delete this one, and never let it come back.  :)

build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs
brthor: I am trying to keep this project the same between the two repos to make it easy to propagate changes.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -22,7 +22,7 @@ public class AzurePublisher
 
         public AzurePublisher()
         {
-            _connectionString = Environment.GetEnvironmentVariable("CONNECTION_STRING").Trim('"');
+            _connectionString = Environment.GetEnvironmentVariable("CONNECTION_STRING")?.Trim('"');
```
eerhardt: Does this work later?  Or do you just get an less specific error later in the execution?

Maybe this should use EnvVars.EnsureVariable, so we get a good error message that CONNECTION_STRING is a required env var.

eerhardt:
>         public string CalculateArchiveUploadUrl(string archiveFile, string channel, string version)

Should this be `static` like the other two around it?

---
Refers to: build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs:181 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -183,12 +183,37 @@ public string CalculateArchiveUploadUrl(string archiveFile, string channel, stri
...
         }
 
-        public void DownloadFiles(string blobVirtualDirectory, string fileExtension, string downloadPath)
+        public static void DownloadFile(string blobFilePath, string localDownloadPath)
+        {
+            var blobUrl = $"{s_dotnetBlobRootUrl}{blobFilePath}";
+
+            using (var client = new HttpClient())
+            {
+                client.Timeout = TimeSpan.FromMinutes(1);
```
eerhardt: Really?  I'm in Bismarck, we have slow internets here...

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -183,12 +183,37 @@ public string CalculateArchiveUploadUrl(string archiveFile, string channel, stri
...
+
+                var request = new HttpRequestMessage(HttpMethod.Get, blobUrl);
+                var sendTask = client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead);
+                var response = sendTask.Result.EnsureSuccessStatusCode();
+
+                var httpStreamTask = response.Content.ReadAsStreamAsync();
+                //httpStreamTask.Wait();
```
eerhardt: Remove.

build_projects/shared-build-targets-utils/Utils/Dirs.cs
brthor: Yeah that makes sense.

build_projects/shared-build-targets-utils/Utils/Dirs.cs
brthor: Have to keep it around for the local builds of core-setup unfortunately.

brthor:
> A lot of these "corehost" dirs can be remove now, right?

@eerhardt 
This project is the same (some changes need to be brought over) between the two repos currently so unless we break that they need to stay because they're used in corehost.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: 5 minutes? We don't want this to hang

brthor:
> Should this be static like the other two around it?
@eerhardt 
VS is showing 0 references on that method, I will just remove it.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: I agree we should use EnvVars.EnsureVariable.

I placed the null operator before I made the other methods static. Will fix.

scripts/update-dependencies/UpdateFilesTargets.cs
brthor: Yes indeed.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
eerhardt: 5 would be better than 1.  How big is the shared framework zip file?  40 MB?  Why do we care about hanging now?  We don't care about a timeout when downloading Stage0, do we?

---
In reply to: [64613316](https://github.com/dotnet/cli/pull/3230#discussion_r64613316) [](ancestors = 64613316)

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -183,12 +183,37 @@ public string CalculateArchiveUploadUrl(string archiveFile, string channel, stri
...
+
+                var httpStream = httpStreamTask.Result;
```
eerhardt: In my general opinion, these "utils" should be async as needed.  And then force the callers to deal with the async behavior.

Someday I could envision a Build.Framework that supported async Targets, and then it would all flow nicely.  (Of course it is going away, so that won't happen).

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: I don't know for sure, but we should. Infinitely hanging http request would be terrible on a build machine. I'm fine with making this a larger number say 10m though.

eerhardt: I think 10 should suffice.  We'll learn soon enough.

---
In reply to: [64615875](https://github.com/dotnet/cli/pull/3230#discussion_r64615875) [](ancestors = 64615875)

eerhardt:
>             ReplaceFileContents(@"pkg\dir.props", contents =>

I think this needs to be removed, as this file no longer exists.

---
Refers to: scripts/update-dependencies/UpdateFilesTargets.cs:210 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

eerhardt:
>         private static void PublishCoreHostPackagesToFeed()

This shouldn't be needed in CLI anymore.  It happens in core-setup.

---
Refers to: build_projects/dotnet-cli-build/PublishTargets.cs:141 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

brthor:
>         private static void PublishCoreHostPackagesToFeed()

Will clean this up.

---
In reply to: [221650102](https://github.com/dotnet/cli/pull/3230#issuecomment-221650102) [](ancestors = 221650102)

---
Refers to: build_projects/dotnet-cli-build/PublishTargets.cs:141 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

eerhardt:
>                     CopyBlobs($"{Channel}/Binaries/{SharedFrameworkNugetVersion}/", targetContainer);

Won't the core-setup build do this?  I don't think the CLI build should be responsible for publishing the SharedFramework artifacts.  Especially since we can have 80 builds of CLI using the same SharedFx.

---
Refers to: build_projects/dotnet-cli-build/PublishTargets.cs:95 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -145,17 +140,94 @@ public static BuildTargetResult ExpectedBuildArtifacts(BuildTargetContext c)
...
+        [Target(
+            nameof(ExpectedBuildArtifacts), 
+            nameof(DownloadHostAndSharedFxArchives), 
+            nameof(DownloadHostAndSharedFxInstallers))]
+        public static BuildTargetResult DownloadHostAndSharedFxArtifacts(BuildTargetContext c) => c.Success();
+
+        [Target]
+        public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetContext c)
+        {
+            Mkdirp(Dirs.Packages);
```
eerhardt: Is this directory used?  I think this line can be removed.

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -145,17 +140,94 @@ public static BuildTargetResult ExpectedBuildArtifacts(BuildTargetContext c)
...
+            }
+
+            return c.Success();
+        }
+
+        [Target]
+        [BuildPlatforms(BuildPlatform.Windows, BuildPlatform.OSX, BuildPlatform.Ubuntu)]
+        public static BuildTargetResult DownloadHostAndSharedFxInstallers(BuildTargetContext c)
+        {
+            Mkdirp(Dirs.Packages);
```
eerhardt: Can this be removed?  I don't see it used.

brthor:
>                     CopyBlobs($"{Channel}/Binaries/{SharedFrameworkNugetVersion}/", targetContainer);

Yes the core-setup build does this. I need to take another pass on this file.

---
In reply to: [221657078](https://github.com/dotnet/cli/pull/3230#issuecomment-221657078) [](ancestors = 221657078)

---
Refers to: build_projects/dotnet-cli-build/PublishTargets.cs:95 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

brthor:
All but `dotnet-new.Tests` are passing here, I think this is safe for NETCore.App to publish to our preview feeds.

schellap:
LGTM. I didn't understand the usage of the SharedFrameworkPublisher.cs, other than that, thank you for turning this around super fast!

build_projects/dotnet-cli-build/MsiTargets.cs
```diff
@@ -258,5 +172,11 @@ private static void ReattachEngineToBundleHelper(string bundle, string engine)
 
             File.Delete(engine);
         }
+
+        private static string GetEngineName(string bundle)
```
eerhardt: Does this have to move?  Can it be in the same spot?

build_projects/dotnet-cli-build/CompileTargets.cs
```diff
@@ -423,16 +226,31 @@ private static BuildTargetResult CompileCliSdk(BuildTargetContext c, DotNetCli d
...
+
+            // Crossgen SDK directory
+            var sharedFrameworkNugetVersion = DependencyVersions.SharedFrameworkVersion;
+            var sharedFrameworkNameVersionPath = SharedFrameworkPublisher.GetSharedFrameworkPublishPath(
+                rootOutputDirectory,
+                sharedFrameworkNugetVersion);
+            
+            // Copy Host to SDK Directory
+            File.Copy(Path.Combine(sharedFrameworkNameVersionPath, DotnetHostBaseName), Path.Combine(sdkOutputDirectory, $"corehost{Constants.ExeSuffix}"), overwrite: true);
+            File.Copy(Path.Combine(sharedFrameworkNameVersionPath, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), Path.Combine(sdkOutputDirectory, $"{Constants.DynamicLibPrefix}hostfxr{Constants.DynamicLibSuffix}"), overwrite: true);
```
eerhardt: Can these constants use the values in `HostArtifactNames` instead of duplicating it here?

eerhardt:
:clock1:

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
weshaggard: Do we even need to change from the default timeout? Which is 100seconds (https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpClient.cs#L18). I wouldn't set it explicitly unless we run into an issue which we believe needs a custom timeout.

build_projects/dotnet-cli-build/MsiTargets.cs
brthor: No, it happened during refactoring of this file. It doesn't affect anything though.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: Alright I'm okay with using the default timeout.

---
In reply to: [64637390](https://github.com/dotnet/cli/pull/3230#discussion_r64637390) [](ancestors = 64637390)

build_projects/dotnet-cli-build/CompileTargets.cs
brthor: I copied these from above but will clean up.

---
In reply to: [64636417](https://github.com/dotnet/cli/pull/3230#discussion_r64636417) [](ancestors = 64636417)

brthor: fixed.

---
In reply to: [64651283](https://github.com/dotnet/cli/pull/3230#discussion_r64651283) [](ancestors = 64651283,64636417)

build_projects/dotnet-cli-build/PrepareTargets.cs
brthor: This is implicit in the sharedFrameworkInstallerFile Path below

I guess I should mkdirp on those.

---
In reply to: [64623409](https://github.com/dotnet/cli/pull/3230#discussion_r64623409) [](ancestors = 64623409)

brthor: fixed.

---
In reply to: [64651911](https://github.com/dotnet/cli/pull/3230#discussion_r64651911) [](ancestors = 64651911,64623409)

brthor:
SharedFrameworkPublisher isn't used here, but the idea is for shared-utils project to be the same between repos so we don't need to reason about as many differences. 
There are some slight differences which exist that I will clean up after this PR.

---
In reply to: [221676308](https://github.com/dotnet/cli/pull/3230#issuecomment-221676308) [](ancestors = 221676308)

scripts/update-dependencies/UpdateFilesTargets.cs
brthor: fixed.

---
In reply to: [64609110](https://github.com/dotnet/cli/pull/3230#discussion_r64609110) [](ancestors = 64609110)

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: fixed.

---
In reply to: [64611612](https://github.com/dotnet/cli/pull/3230#discussion_r64611612) [](ancestors = 64611612)

build_projects/shared-build-targets-utils/Utils/Dirs.cs
brthor: fixed.

---
In reply to: [64611103](https://github.com/dotnet/cli/pull/3230#discussion_r64611103) [](ancestors = 64611103)

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: fixed

---
In reply to: [64612599](https://github.com/dotnet/cli/pull/3230#discussion_r64612599) [](ancestors = 64612599)

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: fixed.

---
In reply to: [64612639](https://github.com/dotnet/cli/pull/3230#discussion_r64612639) [](ancestors = 64612639)

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
brthor: :+1: fixed.

---
In reply to: [64614363](https://github.com/dotnet/cli/pull/3230#discussion_r64614363) [](ancestors = 64614363)

build_projects/dotnet-cli-build/PrepareTargets.cs
brthor: fixed as mentioned on other comment.

---
In reply to: [64623376](https://github.com/dotnet/cli/pull/3230#discussion_r64623376) [](ancestors = 64623376)

build_projects/dotnet-cli-build/MsiTargets.cs
brthor: going to leave this as is, it is not breaking anything and reverting this file would require redoing the changes.

---
In reply to: [64635837](https://github.com/dotnet/cli/pull/3230#discussion_r64635837) [](ancestors = 64635837)

brthor:
>         public string CalculateArchiveUploadUrl(string archiveFile, string channel, string version)

removed.

---
In reply to: [221640118](https://github.com/dotnet/cli/pull/3230#issuecomment-221640118) [](ancestors = 221640118)

---
Refers to: build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs:181 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

brthor:
>             ReplaceFileContents(@"pkg\dir.props", contents =>

fixed.

---
In reply to: [221648661](https://github.com/dotnet/cli/pull/3230#issuecomment-221648661) [](ancestors = 221648661)

---
Refers to: scripts/update-dependencies/UpdateFilesTargets.cs:210 in 74f49a8. [](commit_id = 74f49a80ed876f31ee880fec8b9f32dda6afccd3, deletion_comment = False)

brthor:
PR feedback taken into account, fixed the Ubuntu bug. 

Also had to update all the Microsoft.NETCore.App project.json references to no longer use `*`, so this has grown in enormity. 

Test feeds removed and now consuming from `dotnet-core` and preview channel.

brthor:
This is ready to merge pending a shipit

eerhardt:
:shipit:

brthor:
:+1:

PR #3231, opened 2016-05-25
Setup clrjit path as property

schellap:
Will rebase onto dotnet/core-setup.

### DIFF
--- a/build_projects/shared-build-targets-utils/DependencyVersions.cs
+++ b/build_projects/shared-build-targets-utils/DependencyVersions.cs
@@ -9,8 +9,8 @@ namespace Microsoft.DotNet.Cli.Build
     {
         public static readonly string CoreCLRVersion = "1.0.2-rc3-24123-01";
 
-        public static readonly string SharedFrameworkVersion = "1.0.0-rc3-004306";
-        public static readonly string SharedHostVersion = "1.0.1-rc3-004306-00";
+        public static readonly string SharedFrameworkVersion = "1.0.0-rc3-004308";
+        public static readonly string SharedHostVersion = "1.0.1-rc3-004308-00";
 
         public static readonly string SharedFrameworkChannel = "preview";
         public static readonly string SharedHostChannel = "preview";


src/corehost/cli/deps_resolver.cpp
```diff
@@ -354,58 +354,21 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
 }
 
 /**
- * Helper for obtaining CoreCLR from a given deps file and the deps file's directory.
+ * Probe for coreclr relative to the app and then from the framework root.
+ * If deps does not exist, then just use the app base.
  */
```
schellap: Update comments

src/corehost/cli/deps_resolver.cpp
```diff
@@ -354,58 +354,21 @@ bool deps_resolver_t::probe_deps_entry(const deps_entry_t& entry, const pal::str
...
-    return false;
+    return m_coreclr_path;
 }
 
-
 /**
- * Probe for coreclr relative to the app and then from the framework root.
+ * Probe for clrjit relative to the app and then from the framework root.
  * If deps does not exist, then just use the app base.
  */
```
schellap: Update comments

src/corehost/cli/deps_resolver.cpp
```diff
@@ -497,6 +460,25 @@ bool deps_resolver_t::resolve_tpa_list(
     return true;
 }
 
+bool deps_resolver_t::check_if_entry_known(const deps_entry_t& entry, pal::string_t& path)
+{
```
schellap: const qualifier.

src/corehost/cli/hostpolicy.cpp
```diff
@@ -68,14 +73,15 @@ int run(const arguments_t& args)
...
-    std::vector<char> tpa_paths_cstr, app_base_cstr, native_dirs_cstr, resources_dirs_cstr, fx_deps, deps;
+    std::vector<char> tpa_paths_cstr, app_base_cstr, native_dirs_cstr, resources_dirs_cstr, fx_deps, deps, clrjit_path_cstr;
     pal::pal_clrstring(probe_paths.tpa, &tpa_paths_cstr);
     pal::pal_clrstring(args.app_dir, &app_base_cstr);
     pal::pal_clrstring(probe_paths.native, &native_dirs_cstr);
     pal::pal_clrstring(probe_paths.resources, &resources_dirs_cstr);
 
     pal::pal_clrstring(resolver.get_fx_deps_file(), &fx_deps);
     pal::pal_clrstring(resolver.get_deps_file() + _X(";") + resolver.get_fx_deps_file(), &deps);
+    pal::pal_clrstring(clrjit_path, &clrjit_path_cstr);
```
schellap: move this under the if check.

schellap:
Cc @gkhanna79 JIT path changes

src/corehost/cli/deps_resolver.cpp
gkhanna79: NIT: You should add a comment about the nature of this function (checking for well known packages).

src/corehost/cli/hostpolicy.cpp
```diff
@@ -54,6 +42,23 @@ int run(const arguments_t& args)
...
+    if (clr_path.empty() || !pal::realpath(&clr_path))
+    {
+        trace::error(_X("Could not resolve CoreCLR path. For more details, enable tracing by setting COREHOST_TRACE environment variable to 1"));;
+        return StatusCode::CoreClrResolveFailure;
+    }
+
+    pal::string_t clrjit_path = resolver.get_clrjit_path();
+    if (clrjit_path.empty() || !pal::realpath(&clrjit_path))
+    {
+        trace::warning(_X("Could not resolve CLRJit path, but CoreCLR might load the JIT if it is present side-by-side."));;
```
gkhanna79: I think this should be two different blocks:

if (jit path is empty) {warning: No explicit JIT path specified.}
else if (JIT path cannot be resolved via realpath) {warning: Possibly invalid JIT path.}

We should not call out CoreCLr policy (about JIT load) since it can change independent of the host.

src/corehost/cli/deps_resolver.cpp
```diff
@@ -497,6 +460,25 @@ bool deps_resolver_t::resolve_tpa_list(
     return true;
 }
 
+bool deps_resolver_t::check_if_entry_known(const deps_entry_t& entry, pal::string_t& path)
```
gkhanna79: You are not using the return value, so it will be helpful to make the return type as void and also update the method name to reflect the intent (e.g. InitKnownPackageData).

gkhanna79:
LGTM modulo comments.

gkhanna79:
@schellap @brthor If the host has moved to core-setup, shouldnt this change go there? Or, is there going to be a mop-up from CLI->core-setup?

CC @eerhardt

brthor:
Ideally yes this should go to `core-setup` 

Ran into some issues on PR #3230 yesterday but `core-setup` is publishing successfully to it's test feeds.

If this is an extremely urgent issue we can back port it to core-setup fairly easily.

schellap:
Yes, see description in PR. I'm using CLI as testing ground before @brthor merges.

gkhanna79:
It will be good to get this change in CLI since SharedFW and host are still coming from CLI repo. 

We should then port it to core-setup before SharedFW starts to come from there and send AskMode email when SharedFW for RTM comes from Core-setup RTM branch.

CC @eerhardt

brthor:
FYI merging this won't have affect now as builds from `core-setup` are superseding any subsequent packages from cli.

gkhanna79:
In that case, lets move the change to core-setup and drive it from there. @schellap can you please do that?

schellap:

PR #3241, opened 2016-05-25
Enhance update-dependencies

eerhardt:
- Fix update-dependencies to read all package versions from Latest_Packages.txt.
- Add support to update-dependencies to not update the dependencies in a folder

@brthor @livarcocc


scripts/update-dependencies.ps1
```diff
@@ -32,9 +32,15 @@ $env:PATH = "$env:DOTNET_INSTALL_DIR;$env:PATH"
 
 $appPath = "$PSScriptRoot\update-dependencies"
 
-# Restore the build scripts
-Write-Host "Restoring Build Script projects..."
-pushd $PSScriptRoot
+# Restore the build_projects
+Write-Host "Restoring Microsoft.DotNet.Cli.Build.Framework..."
+pushd $PSScriptRoot\..\build_projects\Microsoft.DotNet.Cli.Build.Framework
```
piotrpMSFT: please don't use ..\ to determine paths in scripts. This makes it really hard to refactor later. Use a `$RepoRoot`-relative path instead.

piotrpMSFT:
:shipit:, modulo 1 comment.

scripts/update-dependencies.ps1
eerhardt: But it is OK to use `..\` in establishing `$RepoRoot`, right?

eerhardt: I am leaning towards moving update-dependencies into the `build_projects` folder to fix this kind of problem - it will live next to Microsoft.DotNet.Cli.Build.Framework again.

Any objections?  @brthor

scripts/update-dependencies.ps1
```diff
@@ -32,9 +32,15 @@ $env:PATH = "$env:DOTNET_INSTALL_DIR;$env:PATH"
...
+dotnet restore --infer-runtimes
```
brthor: do we need infer runtimes?

scripts/update-dependencies/Config.cs
```diff
@@ -33,7 +33,7 @@ public class Config
         private Lazy<string> _email = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_EMAIL"));
         private Lazy<string> _password = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_PASSWORD"));
         
-        private Lazy<string> _coreFxVersionUrl = new Lazy<string>(() => GetEnvironmentVariable("COREFX_VERSION_URL", "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/master/Latest.txt"));
+        private Lazy<string> _coreFxVersionUrl = new Lazy<string>(() => GetEnvironmentVariable("COREFX_VERSION_URL", "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/corefx/release/1.0.0/Latest_Packages.txt"));
```
brthor: should this maybe read from some configuration file, branchinfo equivalent?

scripts/update-dependencies.ps1
brthor: Moving it makes sense 👍

scripts/update-dependencies/Config.cs
eerhardt: It is read from an EnvironmentVariable.  If that EnvironmentVariable doesn't exist, the hard-coded string is just a default.  Putting the default in scripts or in a config file doesn't really make a difference to me.  It just seems easier to keep all the defaults in the same code file.

scripts/update-dependencies.ps1
eerhardt: Good point.  At one time I had all the RIDs in a "runtimes" section, but that must have never made it into the repo.

I'll add the RIDs and remove --infer-runtimes.

scripts/update-dependencies/Config.cs
brthor: Ah I didn't notice the environment variable. That seems fine to me.

brthor:
Will this update the version of Microsoft.NETCore.App in our project.jsons?

eerhardt:
> Will this update the version of Microsoft.NETCore.App in our project.jsons? We no longer need to use *

It doesn't right now.  But it will...  We need:

- [x] `core-setup` producing dotnet/versions outputs: https://github.com/dotnet/versions/blob/master/build-info/dotnet/core-setup/master/Latest_Packages.txt
- [ ] To add a new dependency to this code that reads in the `core-setup` _Packages.txt file, and update any versions
- [ ] Add a subscription to https://github.com/dotnet/versions/blob/master/Maestro/subscriptions.json so that udpate-dependencies gets invoked every time `core-setup` produces a build

PR #3208, opened 2016-05-23
change optimization data path

brthor:
Fixes #3083 

cc @piotrpMSFT @troydai

PR #3242, opened 2016-05-25
Remove optimizationdata subdirectory from optimizationdata path

piotrpMSFT:
PR #3208 moved optimizationdata to a new root directory called optimizationdata. The leaf-node directory optimizationdata is therefore redundant.

/cc @brthor


brthor:
:shipit:

PR #3248, opened 2016-05-26
Fix build failure in PublishTargets.

eerhardt:
InitPublish is trying to access a BuildContext variable that is no longer there. It doesn't need that variable, so removing it.

@piotrpMSFT @brthor


piotrpMSFT:
:shipit:

eerhardt:
Merging to get our VSO builds passing again.

brthor:
Thanks Eric

PR #3254, opened 2016-05-26
Fix dotnet-new project.json and more update-dependencies enhancements

eerhardt:
There is no more need to have project.json.pretemplate now that Microsoft.NETCore.App is built in a separate repo.

Also, adding `core-setup` as a dependency to update-dependencies.

@brthor @livarcocc @piotrpMSFT 

### Note
These are grouped together in an effort to get update-dependencies back up and running smoothly.


brthor:
:shipit:

Issue #3252, opened 2016-05-26
Build time dependencies are not filtered from full framework embedded deps file


pakrym:
@JunTaoLuo

PR #3256, opened 2016-05-26
Filter build dependencies from fullclr embedded deps file

pakrym:
@anurse 
Fixes: https://github.com/dotnet/cli/issues/3252


anurse:
:shipit:

PR #3257, opened 2016-05-26
stop nuking latest binaries

brthor:
cc @eerhardt @sokket

### DIFF
--- a/build_projects/dotnet-cli-build/PublishTargets.cs
+++ b/build_projects/dotnet-cli-build/PublishTargets.cs
@@ -73,9 +73,12 @@ namespace Microsoft.DotNet.Cli.Build
                 }
                 else
                 {
-                    // This is an old drop of latest so remove all old files to ensure a clean state
+                    Regex versionFileRegex = new Regex(@"(?<version>\d\.\d\.\d)-(?<release>.*)?");
+
+                    // Delete old version files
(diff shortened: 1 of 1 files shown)


build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -73,9 +73,12 @@ public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
                 }
                 else
                 {
-                    // This is an old drop of latest so remove all old files to ensure a clean state
+                    Regex versionFileRegex = new Regex(@"(?<version>\d\.\d\.\d)-(?<release>.*)?");
```
sokket: Is there a '.' between 'version' and any \d?

sokket:
LGTM

build_projects/dotnet-cli-build/PublishTargets.cs
brthor: I'm not sure I understand.
`<version>` creates a named group, doesn't affect the matching pattern in any way.

sokket: Ah sorry, forgot my Regex for a minute....carry on, nothing to see here...

brthor: :smile:

PR #3258, opened 2016-05-26
Stop downloading core setup dependencies in every build

brthor:
cc @eerhardt

### DIFF
--- a/build_projects/dotnet-cli-build/PrepareTargets.cs
+++ b/build_projects/dotnet-cli-build/PrepareTargets.cs
@@ -148,27 +148,36 @@ namespace Microsoft.DotNet.Cli.Build
             var sharedFrameworkVersion = DependencyVersions.SharedFrameworkVersion;
             var sharedFrameworkChannel = DependencyVersions.SharedFrameworkChannel;
 
-            var combinedSharedHostAndFrameworkArchiveFile = c.BuildContext.Get<string>("CombinedFrameworkHostCompressedFile");
+            var combinedSharedHostAndFrameworkArchiveDownloadFile =
+                Path.Combine(Dirs.CoreSetupDownload, "combinedSharedHostAndFrameworkArchive");
 
-            Mkdirp(Path.GetDirectoryName(combinedSharedHostAndFrameworkArchiveFile));
+            Mkdirp(Path.GetDirectoryName(combinedSharedHostAndFrameworkArchiveDownloadFile));
 
-            AzurePublisher.DownloadFile(
-                AzurePublisher.CalculateArchiveBlob(
-                    combinedSharedHostAndFrameworkArchiveFile, 
-                    sharedFrameworkChannel, 
-                    sharedFrameworkVersion), 
-                combinedSharedHostAndFrameworkArchiveFile).Wait();
-
-            // Unpack the combined archive to shared framework publish directory
-            Rmdir(Dirs.SharedFrameworkPublish);
-            Mkdirp(Dirs.SharedFrameworkPublish);
-            if(CurrentPlatform.IsWindows)
+            if ( ! File.Exists(combinedSharedHostAndFrameworkArchiveDownloadFile))
             {
-                ZipFile.ExtractToDirectory(combinedSharedHostAndFrameworkArchiveFile, Dirs.SharedFrameworkPublish);
-            }
-            else
-            {
-                Exec("tar", "xf", combinedSharedHostAndFrameworkArchiveFile, "-C", Dirs.SharedFrameworkPublish);
+                // Needed for computing the blob path
+                var combinedSharedHostAndFrameworkArchiveBuildContextFile = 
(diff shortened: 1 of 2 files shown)


build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -148,27 +148,38 @@ public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetConte
...
             var sharedFrameworkChannel = DependencyVersions.SharedFrameworkChannel;
 
-            var combinedSharedHostAndFrameworkArchiveFile = c.BuildContext.Get<string>("CombinedFrameworkHostCompressedFile");
+            var combinedSharedHostAndFrameworkArchiveDownloadFile =
+                Path.Combine(Dirs.CoreSetupDownload, "combinedSharedHostAndFrameworkArchive");
+            var combinedSharedHostAndFrameworkArchiveDestinationFile = 
+                c.BuildContext.Get<string>("CombinedFrameworkHostCompressedFile");
 
-            Mkdirp(Path.GetDirectoryName(combinedSharedHostAndFrameworkArchiveFile));
+            Mkdirp(Path.GetDirectoryName(combinedSharedHostAndFrameworkArchiveDownloadFile));
```
eerhardt: This could be moved inside the "if (!File.Exists)"

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -148,27 +148,38 @@ public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetConte
             var sharedFrameworkVersion = DependencyVersions.SharedFrameworkVersion;
             var sharedFrameworkChannel = DependencyVersions.SharedFrameworkChannel;
 
-            var combinedSharedHostAndFrameworkArchiveFile = c.BuildContext.Get<string>("CombinedFrameworkHostCompressedFile");
+            var combinedSharedHostAndFrameworkArchiveDownloadFile =
+                Path.Combine(Dirs.CoreSetupDownload, "combinedSharedHostAndFrameworkArchive");
+            var combinedSharedHostAndFrameworkArchiveDestinationFile =
```
eerhardt: I think it would be easier to reason about the code if this was moved inside the `if (!File.Exists` block.  That way it is really easy to tell that the above file is the one being checked for existence.

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -184,25 +195,41 @@ public static BuildTargetResult DownloadHostAndSharedFxInstallers(BuildTargetCon
...
-                sharedFrameworkInstallerFile).Wait();
-
-            AzurePublisher.DownloadFile(
-                AzurePublisher.CalculateInstallerBlob(
-                    sharedHostInstallerFile,
-                    sharedHostChannel,
-                    hostVersion),
-                sharedHostInstallerFile).Wait();
+            var sharedFrameworkInstallerDownloadFile = Path.Combine(Dirs.CoreSetupDownload, "sharedFrameworkInstaller");
+            var sharedHostInstallerDownloadFile = Path.Combine(Dirs.CoreSetupDownload, "sharedHostInstaller");
```
eerhardt: Same two comments in this method as above.  Putting the Mkdirp calls inside the if checks, and moving the "destinationFile" variables inside as well would make this code easier to read coming from the outside.

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -148,27 +148,38 @@ public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetConte
...
-            {
-                Exec("tar", "xf", combinedSharedHostAndFrameworkArchiveFile, "-C", Dirs.SharedFrameworkPublish);
+                AzurePublisher.DownloadFile(
+                    AzurePublisher.CalculateArchiveBlob(
+                        combinedSharedHostAndFrameworkArchiveDestinationFile,
+                        sharedFrameworkChannel,
+                        sharedFrameworkVersion),
+                    combinedSharedHostAndFrameworkArchiveDownloadFile).Wait();
+
+                File.Copy(combinedSharedHostAndFrameworkArchiveDownloadFile,
```
eerhardt: Why do we copy the file?  Why not just unzip it from the place we downloaded it to?

eerhardt:
Some nits.  Otherwise, LGTM.

Thanks for doing this so quickly.  This will speed up our 'inter-loop' a lot.

build_projects/dotnet-cli-build/PrepareTargets.cs
brthor: Good point.

build_projects/dotnet-cli-build/PrepareTargets.cs
brthor: done.

brthor:
ahhhhhh osx

PR #3270, opened 2016-05-27
Fixing up build publishing to be more organized

### LINKED ISSUE #3120: Refactor the Publishing code to be more organized
# Overview
When we publish built code and installers today, compiled units are compressed into archives and put into the same root folder in one Azure Blob Container (ie ```<root>\<cli_version>``` or ```<root>\<other version>```). This requires consumers of the archives (or installers) to understand the product difference between versions 1.0.0-rc2-3002520 and 1.0.0-rc2-002520.

# Proposal
In order to organize the intermediate and final build outputs (archive files and installers, respectively), I propose the following:

1. Abandon the ```beta``` folder and instead have the folder driven by a branch-specific-channel
2. Create two subfolders in every Channel folder: ```Packages``` for build consumables (nupkgs, MSIs, debian packages) and ```Intermediates``` for build intermediates (direct build outputs packaged as archives). This adds a logical separation between the build stage and the packaging stage; this gives us the freedom to rev these two pipelines independently.
3. Both ```Packages``` and ```Intermediates``` would have children folders for the specific products (Shared Framework, Host, SDK). This removes the complexity around knowing which version number means what (mainly around the '3' prefix for the version number means a different product than when the '3' is not present).
4. Each product and package folder will contain the ship train folders: Latest, LKG, Production, which will contain the relevant bits as well as a version badge and version textual file. This will allow automated systems to watch the specific train and take action when necessary, rev'ing individual products independently.

# Changes
The changes to make this happen would be contained entirely in two files: [AzurePublisher.cs](https://github.com/dotnet/cli/blob/rel/1.0.0-preview1/scripts/dotnet-cli-build/Publishing/AzurePublisher.cs) and [PublishTargets.cs](https://github.com/dotnet/cli/blob/rel/1.0.0-preview1/scripts/dotnet-cli-build/PublishTargets.cs)

# Future
Once the switch to use msbuild comes along, the only changes that should be necessary would be to the code to publish the bits to the blob locations; no external or breaking changes should be necessary.

Thoughts / comments / feedback @piotrpMSFT, @Sridhar-MS, @livarcocc, @brthor, @krwq?

sokket:
Partly fixes #3120 by organizing the output files in blob storage according to the design in the bug

@brthor @eerhardt @piotrpMSFT

### DIFF
--- a/build_projects/dotnet-cli-build/PrepareTargets.cs
+++ b/build_projects/dotnet-cli-build/PrepareTargets.cs
@@ -153,31 +153,39 @@ namespace Microsoft.DotNet.Cli.Build
 
             Mkdirp(Path.GetDirectoryName(combinedSharedHostAndFrameworkArchiveDownloadFile));
 
-            if ( ! File.Exists(combinedSharedHostAndFrameworkArchiveDownloadFile))
+            if (!File.Exists(combinedSharedHostAndFrameworkArchiveDownloadFile))
             {
-                // Needed for computing the blob path
-                var combinedSharedHostAndFrameworkArchiveBuildContextFile = 
-                    c.BuildContext.Get<string>("CombinedFrameworkHostCompressedFile");
-
-                AzurePublisher.DownloadFile(
-                    AzurePublisher.CalculateArchiveBlob(
-                        combinedSharedHostAndFrameworkArchiveBuildContextFile,
-                        sharedFrameworkChannel,
-                        sharedFrameworkVersion),
-                    combinedSharedHostAndFrameworkArchiveDownloadFile).Wait();
-
+                AzurePublisher.DownloadFile(combinedSharedHostAndFrameworkArchiveDownloadFile, 
+                                              AzurePublisher.Product.HostAndFramework, 
+                                              sharedFrameworkVersion, 
+                                              combinedSharedHostAndFrameworkArchiveDownloadFile).Wait();
 
                 // Unpack the combined archive to shared framework publish directory
                 Rmdir(Dirs.SharedFrameworkPublish);
                 Mkdirp(Dirs.SharedFrameworkPublish);
-                if (CurrentPlatform.IsWindows)
-                {
-                    ZipFile.ExtractToDirectory(combinedSharedHostAndFrameworkArchiveDownloadFile, Dirs.SharedFrameworkPublish);
-                }
-                else
-                {
-                    Exec("tar", "xf", combinedSharedHostAndFrameworkArchiveDownloadFile, "-C", Dirs.SharedFrameworkPublish);
-                }
+            }
+
+            // Needed for computing the blob path
+            var combinedSharedHostAndFrameworkArchiveBuildContextFile = 
+            c.BuildContext.Get<string>("CombinedFrameworkHostCompressedFile");
+
+            AzurePublisher.DownloadFile(
+                combinedSharedHostAndFrameworkArchiveBuildContextFile,
+                AzurePublisher.Product.HostAndFramework,
+                sharedFrameworkVersion,
+                combinedSharedHostAndFrameworkArchiveBuildContextFile).Wait();
+
+
+            // Unpack the combined archive to shared framework publish directory
+            Rmdir(Dirs.SharedFrameworkPublish);
+            Mkdirp(Dirs.SharedFrameworkPublish);
+            if (CurrentPlatform.IsWindows)
+            {
+                ZipFile.ExtractToDirectory(combinedSharedHostAndFrameworkArchiveDownloadFile, Dirs.SharedFrameworkPublish);
+            }
+            else
+            {
+                Exec("tar", "xf", combinedSharedHostAndFrameworkArchiveDownloadFile, "-C", Dirs.SharedFrameworkPublish);
             }
 
             return c.Success();
@@ -198,36 +206,29 @@ namespace Microsoft.DotNet.Cli.Build
 
             Mkdirp(Path.GetDirectoryName(sharedFrameworkInstallerDownloadFile));
             Mkdirp(Path.GetDirectoryName(sharedHostInstallerDownloadFile));
-            
-            if ( ! File.Exists(sharedFrameworkInstallerDownloadFile))
+
+            if (!File.Exists(sharedFrameworkInstallerDownloadFile))
             {
                 var sharedFrameworkInstallerDestinationFile = c.BuildContext.Get<string>("SharedFrameworkInstallerFile");
                 Mkdirp(Path.GetDirectoryName(sharedFrameworkInstallerDestinationFile));
-                
                 AzurePublisher.DownloadFile(
-                    AzurePublisher.CalculateInstallerBlob(
-                        sharedFrameworkInstallerDestinationFile,
-                        sharedFrameworkChannel,
-                        sharedFrameworkVersion),
-                    sharedFrameworkInstallerDownloadFile).Wait();
-
+                    sharedFrameworkInstallerDestinationFile, 
+                    AzurePublisher.Product.SharedFramework, 
+                    sharedFrameworkVersion, 
+                    sharedFrameworkInstallerDestinationFile).Wait();
                 File.Copy(sharedFrameworkInstallerDownloadFile, sharedFrameworkInstallerDestinationFile, true);
             }
-            
-            if ( ! File.Exists(sharedHostInstallerDownloadFile))
+
+            if (!File.Exists(sharedHostInstallerDownloadFile))
             {
                 var sharedHostInstallerDestinationFile = c.BuildContext.Get<string>("SharedHostInstallerFile");
                 Mkdirp(Path.GetDirectoryName(sharedHostInstallerDestinationFile));
-
                 AzurePublisher.DownloadFile(
-                   AzurePublisher.CalculateInstallerBlob(
-                       sharedHostInstallerDestinationFile,
-                       sharedHostChannel,
-                       hostVersion),
-                   sharedHostInstallerDownloadFile).Wait();
-
+                    sharedHostInstallerDestinationFile, 
+                    AzurePublisher.Product.Host, 
+                    hostVersion, 
+                    sharedHostInstallerDestinationFile).Wait();
                 File.Copy(sharedHostInstallerDownloadFile, sharedHostInstallerDestinationFile, true);
-
             }
 
             return c.Success();
--- a/build_projects/dotnet-cli-build/PublishTargets.cs
+++ b/build_projects/dotnet-cli-build/PublishTargets.cs
@@ -21,6 +21,8 @@ namespace Microsoft.DotNet.Cli.Build
 
         private static string Channel { get; set; }
 
+        private static string Pointer { get; set; }
+
         private static string CliVersion { get; set; }
 
         private static string CliNuGetVersion { get; set; }
@@ -37,6 +39,7 @@ namespace Microsoft.DotNet.Cli.Build
             CliNuGetVersion = c.BuildContext.Get<BuildVersion>("BuildVersion").NuGetVersion;
             SharedFrameworkNugetVersion = CliDependencyVersions.SharedFrameworkVersion;
             Channel = c.BuildContext.Get<string>("Channel");
+            Pointer = "Latest";
 
             return c.Success();
         }
@@ -44,8 +47,8 @@ namespace Microsoft.DotNet.Cli.Build
         [Target(nameof(PrepareTargets.Init),
         nameof(PublishTargets.InitPublish),
         nameof(PublishTargets.PublishArtifacts),
-        nameof(PublishTargets.TriggerDockerHubBuilds),
-        nameof(PublishTargets.FinalizeBuild))]
+        nameof(PublishTargets.FinalizeBuild),
+        nameof(PublishTargets.TriggerDockerHubBuilds))]
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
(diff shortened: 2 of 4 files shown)


build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -152,12 +152,10 @@ public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetConte
 
             Mkdirp(Path.GetDirectoryName(combinedSharedHostAndFrameworkArchiveFile));
 
-            AzurePublisher.DownloadFile(
-                AzurePublisher.CalculateArchiveBlob(
-                    combinedSharedHostAndFrameworkArchiveFile, 
-                    sharedFrameworkChannel, 
-                    sharedFrameworkVersion), 
-                combinedSharedHostAndFrameworkArchiveFile).Wait();
+            new AzurePublisher().DownloadFile(combinedSharedHostAndFrameworkArchiveFile,
```
brthor: Use the static method here, an instance of AzurePublisher depends on some environment variables which are not present in Jenkins.

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -155,6 +155,18 @@ public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetConte
 
             if ( ! File.Exists(combinedSharedHostAndFrameworkArchiveDownloadFile))
             {
+                new AzurePublisher().DownloadFile(combinedSharedHostAndFrameworkArchiveFile,
```
brthor: This method was static to bypass the environment variable checks in the constructor of AzurePublisher since they're not necessary for htis method to work.

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -155,6 +155,18 @@ public static BuildTargetResult DownloadHostAndSharedFxArchives(BuildTargetConte
...
+                                              AzurePublisher.Products.HostAndFramework, 
+                                              sharedFrameworkVersion, 
+                                              combinedSharedHostAndFrameworkArchiveFile).Wait();
+
+                // Unpack the combined archive to shared framework publish directory
+                Rmdir(Dirs.SharedFrameworkPublish);
+                Mkdirp(Dirs.SharedFrameworkPublish);
+            }
+            
+            if(CurrentPlatform.IsWindows)
```
brthor: ?? This is needed on all platforms

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -203,31 +216,16 @@ public static BuildTargetResult DownloadHostAndSharedFxInstallers(BuildTargetCon
...
                 Mkdirp(Path.GetDirectoryName(sharedFrameworkInstallerDestinationFile));
-                
-                AzurePublisher.DownloadFile(
-                    AzurePublisher.CalculateInstallerBlob(
-                        sharedFrameworkInstallerDestinationFile,
-                        sharedFrameworkChannel,
-                        sharedFrameworkVersion),
-                    sharedFrameworkInstallerDownloadFile).Wait();
-
+                azure.DownloadFile(sharedFrameworkInstallerFile, AzurePublisher.Products.HostAndFramework, sharedFrameworkVersion, sharedFrameworkInstallerFile).Wait();
```
brthor: This is not hte host and framework, only the framework.

build_projects/dotnet-cli-build/PrepareTargets.cs
```diff
@@ -203,31 +216,16 @@ public static BuildTargetResult DownloadHostAndSharedFxInstallers(BuildTargetCon
...
                 Mkdirp(Path.GetDirectoryName(sharedHostInstallerDestinationFile));
-
-                AzurePublisher.DownloadFile(
-                   AzurePublisher.CalculateInstallerBlob(
-                       sharedHostInstallerDestinationFile,
-                       sharedHostChannel,
-                       hostVersion),
-                   sharedHostInstallerDownloadFile).Wait();
-
+                azure.DownloadFile(sharedHostInstallerFile, AzurePublisher.Products.HostAndFramework, hostVersion, sharedHostInstallerFile).Wait();
```
brthor: This is not hte host and framework, only the host.

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -19,160 +19,115 @@ public static class PublishTargets
...
+        nameof(PublishTargets.TriggerDockerHubBuilds))]
         [Environment("PUBLISH_TO_AZURE_BLOB", "1", "true")] // This is set by CI systems
         public static BuildTargetResult Publish(BuildTargetContext c)
         {
             return c.Success();
         }
 
         [Target]
-        public static BuildTargetResult FinalizeBuild(BuildTargetContext c)
+        public static BuildTargetResult PublishBuildFinalizationMoniker(BuildTargetContext c)
```
brthor: What is the finalization moniker?

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -19,160 +19,115 @@ public static class PublishTargets
...
+        
+        public static void PublishCliVersionInformation(BuildTargetContext c)
+        {
+            var versionBadge = c.BuildContext.Get<string>("VersionBadge");
+            VersionUpdater.PublishVersionBadge(versionBadge, Channel, Pointer).Wait();
+            VersionUpdater.PublishVersionFile("build.version", Utils.GetCliVersionFileContent(c), Channel, Pointer).Wait();
+        }
 
-            return badges.Keys.All(key => badges[key]);
+        public static void PublishCliNupkgs(BuildTargetContext c)
```
brthor: This isn't publishing the packages but rather just updating the `dotnet/versions` repo, shouldnt the name reflect that?

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -15,6 +15,16 @@ namespace Microsoft.DotNet.Cli.Build
...
     public class AzurePublisher
     {
+        public enum Products
+        {
+            SharedFramework,
+            Host,
+            SDK,
+            HostAndFramework,
+            FrameworkAndSdk,
+            CombinedPackage,
```
brthor: Package is somewhat ambiguous given the 4 or 5 different kinds of packages we ship out of the CLI.

What do you think of `CombinedHostFrameworkSdk` to make it more explicit?

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -266,15 +207,14 @@ public static BuildTargetResult PublishSDKSymbolsArchiveToAzure(BuildTargetConte
         public static BuildTargetResult PublishSdkDebToDebianRepo(BuildTargetContext c)
         {
             var version = CliNuGetVersion;
-
             var packageName = Monikers.GetSdkDebianPackageName(c);
             var installerFile = c.BuildContext.Get<string>("SdkInstallerFile");
-            var uploadUrl = AzurePublisherTool.CalculateInstallerUploadUrl(installerFile, Channel, version);
+            string url = AzurePublisherTool.UploadFile(installerFile, AzurePublisher.Products.SDK, version);
```
brthor: This file is already uploaded above in  AzurePublisherTool.UploadFile(installerFile, AzurePublisher.Products.SDK, version);

build_projects/dotnet-cli-build/PrepareTargets.cs
sokket: Fixed

build_projects/dotnet-cli-build/PrepareTargets.cs
sokket: Weird merge conflict resolution...removing the Windows check

eerhardt:
>         private static string ToBase64(string value)

Is this used anymore?

---
Refers to: build_projects/shared-build-targets-utils/VersionRepoUpdater.cs:138 in 3dff39d. [](commit_id = 3dff39de49b28ea0ca9fe86cb35394a44bdc469c, deletion_comment = False)

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.DotNet.Cli.Build
     public class VersionRepoUpdater
     {
         private static Regex s_nugetFileRegex = new Regex("^(.*?)\\.(([0-9]+\\.)?[0-9]+\\.[0-9]+(-([A-z0-9-]+))?)\\.nupkg$");
-
+        private static string s_baseGitHubPath = $"/build-info/dotnet/cli";
```
eerhardt: We need to keep "CLI" specific concepts out of this class, as it is duplicated between both the 'cli' and 'core-setup' repo, and @brthor wants to keep the files exactly the same.

Also, I imagine other teams copying this file into their own repos in order to update 'dotnet/versions' with their build info.  It'd be easier to dupe if it is as generic as possible.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -60,6 +60,19 @@ public class VersionRepoUpdater
             await UpdateGitHubFile(packageInfoFilePath, packageInfoFileContent, message);
         }
 
+        public async Task PublishVersionBadge(string badgePath, string channel, string pointer)
```
eerhardt: See comment above about keeping this file as "repo-agnostic" as possible.  Maybe these could be extension methods on VersionRepoUpdater.

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -81,31 +94,33 @@ private static List<NuGetPackageInfo> GetPackageInfo(string nupkgFilePath)
...
 
-                Console.WriteLine($"Getting the 'sha' of the current contents of file '{_versionsRepoOwner}/{_versionsRepo}/{path}'");
+                Console.WriteLine($"Getting the 'sha' of the current contents of file '{_versionsRepoOwner}/{_versionsRepo}/{githubFile}'");
 
                 string currentFile = await client.GetStringAsync(fileUrl);
                 string currentSha = JObject.Parse(currentFile)["sha"].ToString();
 
                 Console.WriteLine($"Got 'sha' value of '{currentSha}'");
 
-                Console.WriteLine($"Request to update file '{_versionsRepoOwner}/{_versionsRepo}/{path}' contents to:");
```
eerhardt: This is actually really helpful when trying to diagnose issues when the script fails.  Can we attempt to keep this logging somehow?  (Maybe blindly UTF8 convert the data, and even if is garbage print it out?)

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
```diff
@@ -60,6 +60,19 @@ public class VersionRepoUpdater
             await UpdateGitHubFile(packageInfoFilePath, packageInfoFileContent, message);
         }
 
+        public async Task PublishVersionBadge(string badgePath, string channel, string pointer)
+        {
+            byte[] data = System.IO.File.ReadAllBytes(badgePath);
+            string githubFileName = $"{s_baseGitHubPath}/{channel}/{pointer}/{Path.GetFileName(badgePath)}";
+            await PushChangesToGitHubFile(githubFileName, data, "Updating CLI version badge");
```
eerhardt: These messages should have the version number in them, if possible.  This makes the commit history a lot easier to consume.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -15,6 +15,16 @@ namespace Microsoft.DotNet.Cli.Build
 {
     public class AzurePublisher
     {
+        public enum Products
```
eerhardt: Making this enum plural seems odd since it is only used for a single value.  [https://msdn.microsoft.com/en-us/library/bb264487.aspx](https://msdn.microsoft.com/en-us/library/bb264487.aspx)

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -19,160 +19,115 @@ public static class PublishTargets
...
         private static string CliNuGetVersion { get; set; }
 
-        private static string SharedFrameworkNugetVersion { get; set; }
-
         [Target]
         public static BuildTargetResult InitPublish(BuildTargetContext c)
         {
             AzurePublisherTool = new AzurePublisher();
             DebRepoPublisherTool = new DebRepoPublisher(Dirs.Packages);
+            VersionUpdater = new VersionRepoUpdater(EnvVars.EnsureVariable("GITHUB_PASSWORD"));
```
eerhardt: This is probably going to fail on Jenkins and everyone's machines since this EnvVar isn't set.  It is only set on VSO.

build_projects/dotnet-cli-build/PublishTargets.cs
```diff
@@ -19,160 +19,115 @@ public static class PublishTargets
...
             return c.Success();
         }
 
         [Target(nameof(PrepareTargets.Init),
         nameof(PublishTargets.InitPublish),
         nameof(PublishTargets.PublishArtifacts),
-        nameof(PublishTargets.TriggerDockerHubBuilds),
-        nameof(PublishTargets.FinalizeBuild))]
+        nameof(PublishTargets.FinalizeBuild),
+        nameof(PublishTargets.TriggerDockerHubBuilds))]
```
eerhardt: I believe @MichaelSimons was going to work on moving this into FinalizeBuild.

Any reason why these are being switched around?

build_projects/dotnet-cli-build/PublishTargets.cs
eerhardt: Currently the Windows x64 VSO build publishes the NuGet packages after the "Publish" step.

We either need to:

1. continue updating dotnet/versions of the new packages AFTER the packages have actually been published
2. move the publishing of the nupkgs into the FinalizeBuild step, and ensure that we ONLY publish the Windows x64 nupkgs.
 - I imagine this working like:  
 - Windows x64 knows that it should publish the .nupkgs to Azure
 - FinalizeBuild downloads those packages and then publishes them to MyGet

We shouldn't be updated dotnet/versions before the packages have actually been published to MyGet (in case they never get published).

---
In reply to: [64946020](https://github.com/dotnet/cli/pull/3270#discussion_r64946020) [](ancestors = 64946020)

build_projects/dotnet-cli-build/PublishTargets.cs
MichaelSimons: I am actually hoping to get TriggerDockerHubBuilds out of the CLI repo all together and use Maestro instead.  This requires moving UpdateVersionsRepo to FinalizeBuild however in order to ensure all platforms are built prior to updating the Versions repo.

eerhardt:
I think we may need to push "latest.version" files to both Azure and dotnet/versions to start.  @piotrMSFT had some concerns about changing the dotnet-install scripts to consuming GitHub because of GitHub's rate limits.  Also, we will definitely need to do both to start because dotnet-install will still be depending on the version file in Azure.

build_projects/dotnet-cli-build/PrepareTargets.cs
sokket: Fixed

build_projects/dotnet-cli-build/PrepareTargets.cs
sokket: Fixed

build_projects/dotnet-cli-build/PublishTargets.cs
sokket: This is a build-{OS}.{version} file that is dropped to indicate that the build for that OS has completed

brthor: Why moniker? Would version file or something indicating this is an artifact be more appropriate?

We have a Monikers.cs and I think that also makes this confusing.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
sokket: Fixed

build_projects/dotnet-cli-build/PublishTargets.cs
sokket: Fixed

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
sokket: If that is the case then we should move this out into it's own package and consume it that way. Keeping a source file sync'd across multiple repos without changes is not a viable long-term solution; there is no way to enforce that and keep it synchronized

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
sokket: Responded above :)

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
sokket: Fixed

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
sokket: Fixed

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
brthor: Shipping it via core-setup would be possible and straightforward but given that, that repo is soon moving to MSBuild in the short-term it didn't seem worth the investment to make it happen. 

If you feel strongly about it, I wouldn't be opposed to a PR enabling this.

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
sokket: Fixed

build_projects/shared-build-targets-utils/VersionRepoUpdater.cs
eerhardt: You didn't respond to the part:

> Also, I imagine other teams copying this file into their own repos in order to update 'dotnet/versions' with their build info. It'd be easier to dupe if it is as generic as possible.

build_projects/dotnet-cli-build/PublishTargets.cs
sokket: Fixed to make it local to the functions that need it

build_projects/dotnet-cli-build/PublishTargets.cs
sokket: A rebase should fix this then

sokket:
Note - updating my PR to get the code over to my Windows box to debug

sokket:
Punting this for now due to Escrow

eerhardt:
We can still do this for rel/1.0.0 right?  Escrow shouldn't block us from getting our publish artifacts in order for the future, AFAIK.

/cc @piotrpMSFT

PR #3295, opened 2016-05-31
Fix Building from source links in README

markrendle:
skipciplease

Links in `Building from source` are full URLs to `master` branch; changed to relative URLs.


eerhardt:
:shipit:   Thanks, @markrendle !

PR #3337, opened 2016-06-02
update nuget to 3.5.0-beta2-1392

rohit21agrawal:
retargeting pull request https://github.com/dotnet/cli/pull/3329 to rel/1.0.0-preview2 .

Update nuget to 3.5.0-beta2-1392 . This build has been upgraded to depend on the following:
• CoreCLR/CoreFX/WCF – rc3-24128-00
• Microsoft.NetCore.App – 1.0.0-rc3-004338
• CLI – 1.0.0-preview2-002911

### DIFF
--- a/build_projects/dotnet-cli-build/project.json
+++ b/build_projects/dotnet-cli-build/project.json
@@ -13,7 +13,7 @@
     "System.Runtime.Serialization.Primitives": "4.1.1-rc3-24131-00",
     "System.Xml.XmlSerializer": "4.0.11-rc3-24131-00",
     "WindowsAzure.Storage": "6.2.2-preview",
-    "NuGet.CommandLine.XPlat": "3.5.0-rc-1285",
+    "NuGet.CommandLine.XPlat": "3.5.0-beta2-1392",
     "Microsoft.DotNet.Cli.Build.Framework": {
       "target": "project"
     },
--- a/build_projects/update-dependencies/project.json
+++ b/build_projects/update-dependencies/project.json
@@ -12,7 +12,7 @@
     "Microsoft.DotNet.Cli.Build.Framework": {
       "target": "project"
     },
-    "NuGet.Versioning": "3.5.0-rc-1285",
+    "NuGet.Versioning": "3.5.0-beta2-1392",
     "Newtonsoft.Json": "7.0.1",
     "Octokit": "0.18.0",
     "Microsoft.Net.Http": "2.2.29"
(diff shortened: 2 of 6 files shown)


schellap:
What I find really bizarre is that, the commit hash printed by the host binary doesn't correspond to the version on the nuget feed. The NuGet cache contains some package that is not in our feeds.

```
--- Invoked dotnet [version: 1.0.1-rc3-004338-00, commit hash: b2e2280386209841b943c851d48014ce0529069c]
```
Note that this commit does not exist in the core-setup repo.

@brthor @eerhardt For repo builds, don't we redirect the nuget package to be repo local in CI builds and hence we shouldn't be picking up this rogue package from the dotnet-bot's nuget cache?

Cc @livarcocc @ellismg

See below for exact details:
```
[dotnet-bot@Azure0523130133 cli]$ cat ~/.nuget/packages/runtime.rhel.7-x64.Microsoft.NETCore.DotNetHostResolver/1.0.1-rc3-004338-00/version.txt 
b2e2280386209841b943c851d48014ce0529069c
Microsoft.DotNet.Cli.Build.HostVersion+VerInfo
```

However, if you download from MyGet:
https://dotnet.myget.org/F/dotnet-core/api/v2/package/runtime.rhel.7-x64.Microsoft.NETCore.DotNetHostResolver/1.0.1-rc3-004338-00

```
dotnet-bot@Azure0523111231:~/cli/tmep$ ls
dotnet-bot@Azure0523111231:~/cli/tmep$ wget https://dotnet.myget.org/F/dotnet-core/api/v2/package/runtime.rhel.7-x64.Microsoft.NETCore.DotNetHostResolver/1.0.1-rc3-004338-00 -O a.zip 2> /dev/null
dotnet-bot@Azure0523111231:~/cli/tmep$ unzip a.zip > /dev/null
dotnet-bot@Azure0523111231:~/cli/tmep$ sudo cat version.txt 
05e286dd641ee67543bcb83110eb5b8610925c57
1.0.1-rc3-004338-00
```

brthor:
@schellap currently core-setup and cli both place the package cache in the parent directory so it's not wiped out with each build. That could explain this issue. @eerhardt and I had some discussion about this and it's going to cause other issues as well. 

@eerhardt What do you think of placing the package cache in the repo workspace so it is build specific?

eerhardt:
> @eerhardt What do you think of placing the package cache in the repo workspace so it is build specific?

In theory it shouldn't be necessary. But giving this more thought, and given the world we live in where the same machine might run both core-setup builds and cli builds, I think it has to happen.  The core-setup builds produce NuGet packages and put them in the cache, and those aren't the "official" version of those NuGet packages, which can cause all sorts of bad issues.

brthor:
@eerhardt 

Even between core-setup builds we have the same issue where builds with the same number of commits but aren't "official" essentially produce a race condition on the cache, especially if they're running at the same time.

PR #3106, opened 2016-05-18
Add --fx-version option to the muxer

schellap:

Issue #2129, opened 2016-03-29
corehost filters out muxer aguments in standalone mode


pakrym:
## Steps to reproduce

```
{
  "version": "1.0.0-*",
  "compilationOptions": {
    "emitEntryPoint": true
  },
  "dependencies": {
    "Microsoft.NETCore.App": {
      "version": "1.0.0-rc2-23928"
    }
  },
  "frameworks": {
    "netstandardapp1.5": {
      "imports": [
        "portable-net45+win8",
        "dnxcore50"
      ]
    }
  }
}
```

```
using System;

namespace ConsoleApplication
{
    public class Program
    {
        public static void Main(string[] args)
        {
            Console.WriteLine(string.Join(",", args));
        }
    }
}
```

Build, run exe directly with `--depsFile` argument
## Expected  behavior

All arguments get forwarded to managed code
## Actual behavior

Muxer known arguments get filtered:

```
D:\temp\1> D:\temp\1\bin\Debug\netstandardapp1.5\win7-x64\1.exe --depsFile C:\Users\pakrym\.nuget\packages

D:\temp\1> D:\temp\1\bin\Debug\netstandardapp1.5\win7-x64\1.exe --depsFile1 C:\Users\pakrym\.nuget\packages
--depsFile1,C:\Users\pakrym\.nuget\packages
```
## Environment data

`dotnet --version` output:

```
1.0.0-beta-002112
```

piotrpMSFT:
What's the scenario here? We can debate what the intention should be, but how is this getting in your way?

I'm not a big fan of introducing behavioral differences between standalone and portable mode. Folks should be easily able to flip the switch. 

i'm also curious why no runtimes are listed in pj. We have said that an app is not standalone until RIDs are listed.

pakrym:
It's not getting in my way, it's just `corehost` bug I noticed. This app was published before we disabled rid inference so it didn't need rids in `project.json`.

schellap:
@pakrym, in one thread you want this and in another it is a bug? Just use the muxer then. We'll kill --additionalprobingpath in standalone @davidfowl

pakrym:
@chellap Where do I wan't this?

schellap:
https://github.com/dotnet/cli/issues/2127#issuecomment-203849540. Here... this bug and that comment don't make sense together.

pakrym:
"Expected" there is running standalone exe part, not passing `additionalprobingpath` part

gkhanna79:
[...]

PR #3346, opened 2016-06-02
fixes #2408 stop using corehost to activate projectdependencycommands

### LINKED ISSUE #2408: Stop using corehost to activate apps

brthor:
cc @schellap @eerhardt 

fixes #2408

### DIFF
--- a/build_projects/dotnet-cli-build/PrepareTargets.cs
+++ b/build_projects/dotnet-cli-build/PrepareTargets.cs
@@ -232,7 +232,6 @@ namespace Microsoft.DotNet.Cli.Build
                    sharedHostInstallerDownloadFile).Wait();
 
                 File.Copy(sharedHostInstallerDownloadFile, sharedHostInstallerDestinationFile, true);
-
             }
 
             return c.Success();
--- a/src/Microsoft.DotNet.Cli.Utils/CommandResolution/PackagedCommandSpecFactory.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/CommandResolution/PackagedCommandSpecFactory.cs
@@ -36,15 +36,12 @@ namespace Microsoft.DotNet.Cli.Utils
                 return null;
             }
 
-            var isPortable = IsPortableApp(commandPath, runtimeConfigPath);
-
-            return CreateCommandSpecWrappingWithCorehostIfDll(
+            return CreateCommandSpecWrappingWithMuxerIfDll(
                 commandPath, 
                 commandArguments, 
                 depsFilePath, 
                 commandResolutionStrategy,
                 nugetPackagesRoot,
-                isPortable,
                 runtimeConfigPath);
         }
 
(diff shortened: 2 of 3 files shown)


schellap:
Can we get rid of `corehost.exe` and friends in the SDK binaries and installers?

eerhardt:
:shipit: 

I assume we can remove `corehost` and `dotnet` from the shared fx after this, right?

brthor:
Corehost definitely, I don't know who has a dependency on the `dotnet` there.

brthor:

eerhardt:
> Corehost definitely, I don't know who has a dependency on the dotnet there.

I think the idea of https://github.com/dotnet/cli/pull/3106 was that you'd always use the muxer, no matter what.

brthor:
@schellap Can you comment on whether we could remove `dotnet` from the sharedfx?

And yes we should be able to remove corehost.exe

eerhardt:
We need to remove the CoreHost class too https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Cli.Utils/CoreHost.cs.

Looks like there is still 1 dependency on it: 

https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.DotNet.Compiler.Common/Executable.cs#L106

brthor:
I was just checking on this, thanks @eerhardt :smile: 

I'm actually not sure what we should do there, that is the host the sdk copies on build output, I guess let's copy the muxer?

eerhardt:
>  I guess let's copy the muxer?

Yes, that was my thought too.  It should be the exact same executable as corehost.exe, just a different name and in a different folder.

schellap:
Dotnet is removed already as part of https://github.com/dotnet/cli/pull/3106/files

schellap:
> I'm actually not sure what we should do there, that is the host the sdk copies on build output, I guess let's copy the muxer?

You should get everything from package graph. Otherwise you will end up with problems with app specified version vs. installed SDK versions. Also, hostfxr.dll won't be sxs in future SharedFX builds.

brthor:
@schellap Okay this has become a bigger change.

I think I will put this PR in as is as only a fix to activating xunit with corehost. 

As for removing corehost we need to deal with https://github.com/dotnet/cli/issues/2444, resolve the host from the package graph during build, and then remove corehost. These will be breaking changes so we'll need to communicate appropriately.

eerhardt:
`corehost.exe` is still in the shared framework:

https://github.com/dotnet/core-setup/blob/master/build_projects/shared-build-targets-utils/Utils/SharedFrameworkPublisher.cs#L182-L184

```
            File.Copy(
               Path.Combine(_corehostLockedDirectory, HostArtifactNames.DotnetHostBaseName),
               Path.Combine(sharedFrameworkNameAndVersionRoot, $"corehost{Constants.ExeSuffix}"), true);
```

brthor:

schellap:
> These will be breaking changes so we'll need to communicate appropriately.

I see these solutions and only 1 below is breaking. Have you decided on its way forward?

1. Activate with muxer: `dotnet.exe --additionalprobingpath c:\ache\ standalone.dll`
2. Build copies `standalone.exe` and `hostfxr.dll` from package graph
  * `standalone.exe --additionalprobingpath c:\ache`
3. `dotnet.exe` and `hostfxr.dll` loaded in the CLI SDK's process is copied into `standalone.exe` base.

schellap:
If we are getting this right and can make breaking changes we should do 1. Standalone apps are not truly standalone in the sense they don't have everything next to them. So they should be activated by the muxer, that is the rationale here.

1 will also address: #2129

brthor:
@schellap 2 is breaking because any app which doesn't already reference HostPolicy will break. I agree it's the ideal solution though

schellap:
> @schellap 2 is breaking because any app which doesn't already reference HostPolicy will break. I agree it's the ideal solution though

I see, 2 would have to resolve hostpolicy as well :)

PR #3370, opened 2016-06-03
Make the nuget cache local to the repo

### LINKED ISSUE #3368: Make CI Build NuGet Cache Independent
Today our CI build nuget cache is placed in the parent directory of the CI workspace to prevent it from being wiped out in each build to save on caching time. 

This presents multiple issues:
1. If two builds are running on the same machine, with the same number of commits in each build, there is a race condition in the cache as to which version of the packages produced in the build will be present in the cache. This will (and likely is) cause test instability, and overall inexplicable failures.

2. If a build has previously run with the same commit count, it's possible the cache will not be re-hydrated with new packages and tests may be run against old packages (from some PR) with unknown bits.

3. This same method is shared by the Core-Setup repo so number 2 also applies to nuget packages in that repo (Microsoft.NETCore.App).

To solve this we should make the nuget cache of each build independent to that build. Cleaning the cache of each package the build produces at the beginning of the build is not sufficient to solve number 1.

cc @eerhardt @livarcocc @piotrpMSFT

brthor:
fixes #3368

@eerhardt @livarcocc

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/Dirs.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Dirs.cs
@@ -45,10 +45,6 @@ namespace Microsoft.DotNet.Cli.Build
 
         private static string GetNuGetPackagesDir()
         {
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                return Path.Combine(Dirs.RepoRoot, ".nuget", "packages");
-            }
             return Path.Combine(Dirs.RepoRoot, ".nuget", "packages");
         }
     }
--- a/build_projects/shared-build-targets-utils/Utils/Dirs.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Dirs.cs
@@ -47,9 +47,9 @@ namespace Microsoft.DotNet.Cli.Build
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                return Path.Combine(Environment.GetEnvironmentVariable("USERPROFILE"), ".nuget", "packages");
+                return Path.Combine(Dirs.RepoRoot, ".nuget", "packages");
             }
-            return Path.Combine(Environment.GetEnvironmentVariable("HOME"), ".nuget", "packages");
+            return Path.Combine(Dirs.RepoRoot, ".nuget", "packages");
         }
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@
 # NuGet keeps dropping
 Library/
 
+# local nuget cache
+.nuget/
+
 # NuGet v3 restore drops these even though we don't use MSBuild :(
 *.nuget.targets
 *.nuget.props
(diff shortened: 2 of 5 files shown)


eerhardt:
I'll wait until this is green to review.

brthor:
@dotnet-bot test this please

eerhardt:
Are we concerned that parts of our build are still using the %userprofile% cache?  For example, when we restore and build the 'dotnet-cli-build' project, we are using the default cache.

brthor:
Are you concerned about it?

My opinion is that it is inconsequential since the build framework is essentially a tool. Leaving it pulling from the machine nuget cache won't cause any of the issues I presented in #3368 . 

However, if we're planning on regularly needing to consume manually patched corefx bits then I agree it's a concern.

eerhardt:
There was a problem last week that CoreFX had bad packages, that our build servers cached the bad packages.  The packages were fixed up on myget, but since our build servers already had them cached, we were unable to build.

This is the same type of problem this PR is addressing, but it wouldn't be fixed by this PR.  Basically all the packages that our build projects depend on can't have the issues described in #3368.

brthor:
This PR was intended to fix race conditions due to generated nuget packages in concurrent builds. Ideally we wouldn't have non-versioned package replacements, but I agree that pragmatically we should grow the scope here and cover that issue. 

I'll make the changes today and in core-setup as well.

brthor:
@eerhardt updated as we discussed

@dotnet-bot test this please

build_projects/shared-build-targets-utils/Utils/Dirs.cs
```diff
@@ -47,9 +47,9 @@ private static string GetNuGetPackagesDir()
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
eerhardt: This "if" can go away, right?  We do the exact same thing on all platforms.

eerhardt:
:shipit: just one minor comment.

build_projects/shared-build-targets-utils/Utils/Dirs.cs
brthor: yes not sure what I was thinking here

piotrpMSFT:
Ready to merge?

eerhardt:
We can address my minor comment in a post-PR PR.

brthor:
I will fix this now, been swamped with the core-setup tests

brthor:

brthor:
Anyone seen this before

```
Expected command to pass but it did not. File Name: /Users/dotnet-bot/j/workspace/dotnet_cli/rel_1.0.0-preview2/release_osx_x64_prtest/artifacts/osx.10.11-x64/stage2/dotnet Arguments: run -p "/Users/dotnet-bot/j/workspace/dotnet_cli/rel_1.0.0-preview2/release_osx_x64_prtest/test/dotnet-run.Tests/bin/Release/netcoreapp1.0/ItPassesArgumentsToTheApp/TestAppWithArgs" one --two -three Exit Code: 139 StdOut: StdErr:
```

brthor:

eerhardt:
> Exit Code: 139

In the past, this has normally been "access violation" or "seg fault" errors happening in native code.

@piotrpMSFT

PR #3373, opened 2016-06-03
Move dependency from Microsoft.NETCore.Runtime to Microsoft.NETCore.Runtime.CoreCLR.

eerhardt:
Making the same change in the CLI as in core-setup.

These 5 projects are all 'self-contained' apps.  The 3 perf ones can't reference NETCore.App because they target 'netstandard1.6', which NETCore.App doesn't support.  (They need to target 2 TFMs - the point of the test.)

@weshaggard @brthor


eerhardt:

Dth tests fail all the time -- @troydai - I am thinking of shutting them off. They are not providing value right now.

weshaggard:
LGTM

eerhardt:
test Fedora23 x64 Debug Build 

all flaky Dth tests

PR #3375, opened 2016-06-03
Adding support for --serviceable option to pack command

toddm:
Adding support for --serviceable option to pack command which puts < serviceable>true< /serviceable> into the output nuspec

This addresses https://github.com/NuGet/Home/issues/2900

@rohit21agrawal @rrelyea

### DIFF
--- a/src/dotnet/commands/dotnet-pack/NuGet/ManifestSchemaUtility.cs
+++ b/src/dotnet/commands/dotnet-pack/NuGet/ManifestSchemaUtility.cs
@@ -44,7 +44,7 @@ namespace NuGet
         /// <summary>
         /// Added serviceble element under metadata.
         /// </summary>
-        internal const string SchemaVersionV7 = "http://schemas.microsoft.com/packaging/2016/06/nuspec.xsd";
+        internal const string SchemaVersionV8 = "http://schemas.microsoft.com/packaging/2016/06/nuspec.xsd";
 
         private static readonly string[] VersionToSchemaMappings = new[] {
             SchemaVersionV1,
@@ -53,7 +53,7 @@ namespace NuGet
             SchemaVersionV4,
             SchemaVersionV5,
             SchemaVersionV6,
-            SchemaVersionV7
+            SchemaVersionV8
         };
 
         public static int GetVersionFromNamespace(string @namespace)
--- a/src/dotnet/commands/dotnet-pack/NuGet/ManifestSchemaUtility.cs
+++ b/src/dotnet/commands/dotnet-pack/NuGet/ManifestSchemaUtility.cs
@@ -41,13 +41,19 @@ namespace NuGet
         /// </summary>
         internal const string SchemaVersionV6 = "http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd";
 
+        /// <summary>
+        /// Added serviceble element under metadata.
+        /// </summary>
+        internal const string SchemaVersionV7 = "http://schemas.microsoft.com/packaging/2016/06/nuspec.xsd";
+
         private static readonly string[] VersionToSchemaMappings = new[] {
             SchemaVersionV1,
             SchemaVersionV2,
             SchemaVersionV3,
             SchemaVersionV4,
             SchemaVersionV5,
-            SchemaVersionV6
+            SchemaVersionV6,
+            SchemaVersionV7
         };
 
         public static int GetVersionFromNamespace(string @namespace)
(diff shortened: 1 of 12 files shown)


src/dotnet/commands/dotnet-pack/NuGet/ManifestSchemaUtility.cs
```diff
@@ -41,13 +41,19 @@ internal static class ManifestSchemaUtility
         /// </summary>
         internal const string SchemaVersionV6 = "http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd";
 
+        /// <summary>
+        /// Added serviceble element under metadata.
+        /// </summary>
+        internal const string SchemaVersionV7 = "http://schemas.microsoft.com/packaging/2016/06/nuspec.xsd";
```
rohit21agrawal: release-3.5.0-beta2 will have serviceable element in the SchemaVersion8 . If Joel's PackageType change is eventually going to be merged in here, should we try to be consistent and make this SchemaVersionV8? And while merging PackageType change, add that to SchemaVersionV7?

rohit21agrawal:
LGTM. Are you planning on adding tests?
I am going to be updating my PR to add tests for the NuGet side of things later today.

src/dotnet/commands/dotnet-pack/NuGet/ManifestSchemaUtility.cs
toddm: I thought about doing but wasn't sure. I'll change it.

rohit21agrawal: @rrelyea @joelverhagen any thoughts on this?

toddm:
I added a test.

src/dotnet/commands/dotnet-pack/NuGet/ManifestVersionUtility.cs
```diff
@@ -15,6 +15,7 @@ internal static class ManifestVersionUtility
         public const int TargetFrameworkSupportForDependencyContentsAndToolsVersion = 4;
         public const int TargetFrameworkSupportForReferencesVersion = 5;
         public const int XdtTransformationVersion = 6;
+        public const int ServiceableVersion = 8;
```
rohit21agrawal: This will throw right now since there are only 7 entries here. We still need to keep this equal to 7 , and add a comment that it needs to be changed to 8 when we integrate PackageType change here.

src/dotnet/commands/dotnet-pack/Program.cs
```diff
@@ -29,6 +29,7 @@ public static int Run(string[] args)
             var buildBasePath = app.Option("-b|--build-base-path <OUTPUT_DIR>", "Directory in which to place temporary build outputs", CommandOptionType.SingleValue);
             var configuration = app.Option("-c|--configuration <CONFIGURATION>", "Configuration under which to build", CommandOptionType.SingleValue);
             var versionSuffix = app.Option("--version-suffix <VERSION_SUFFIX>", "Defines what `*` should be replaced with in version field in project.json", CommandOptionType.SingleValue);
+            var serviceable = app.Option("-s|--serviceable", "Set the serviceable flag in the output nuspec", CommandOptionType.NoValue);
```
eerhardt: "the output nuspec" seems an odd phrase to an outside user.  The user is calling `dotnet pack` and the output is a .nupkg.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/PackCommand.cs
```diff
@@ -64,13 +65,24 @@ private string VersionSuffixOption
...
+
         public PackCommand(
             string projectPath,
             string output = "",
             string buildBasePath = "",
             string tempOutput="", 
             string configuration="", 
-            string versionSuffix="")
+            string versionSuffix="",
+            string serviceable = "")
```
eerhardt: shouldn't this be a boolean?

src/dotnet/commands/dotnet-pack/NuGet/ManifestVersionUtility.cs
toddm: You're right.  I changed it to 7 and added a comment.

src/dotnet/commands/dotnet-pack/Program.cs
toddm: I changed the text.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/PackCommand.cs
toddm: Agreed, a bool would be better.  I changed it.

toddm:
I pushed a commit to address the comments.

rohit21agrawal:
:shipit:

eerhardt:
:shipit: 

I assume you will want this in 'preview2'.  So we will have to port it to the rel/1.0.0-preview2 branch. Let me know if you will take care of the porting, or if you need someone else to port it.

toddm:
Thanks.  I just merged to rel/1.0.0.  Yes, it needs to be ported to preview2.  I'll work on porting now.

PR #3378, opened 2016-06-04
[Fixes #3377] Update Json.NET to 9.0.1-beta1

### LINKED ISSUE #3377: Update JSON.NET to 9.0.1
It has support for NetStandard and allows to remove imports from project.json, we want to update ASP.NET to use it, but it is causing issues with our test runner.

javiercn:
Updates Json.NET to 9.0.1-beta1 which has support for netstandard.

Related issue:
https://github.com/dotnet/cli/issues/3377


eerhardt:
From @joshfree  - 
joshfree [1:38 PM] 
We can't take dependencies on -prerelease a week before escrow.

Let's make sure this doesn't get into preview2.  This is only going into rel/1.0.0.

Eilon:
@eerhardt we need Json.NET 9.0.1 RTM, which @jamesnk will produce for us in time for the CLI Preview2 release. Right now we just need to switch to the beta build, then when escrow happens we will work with @jamesnk to get an RTM build and we'll all switch to that.

So, this needs to stay in the preview2 milestone.

cc @Petermarcu @piotrpMSFT because they are both aware of this.

eerhardt:
Ok. Then this PR should be opened in the rel/1.0.0-preview2 branch if it should go into preview2. Then you don't need to dual checkin.

Eilon:
@eerhardt ah, got it, thanks! @javiercn please do this 😄

javiercn:
Will do, closing this PR as its against the wrong base

PR #3389, opened 2016-06-06
Expand timeout setting for project model server

troydai:
This is an experiment to verify if the timeout setting of current project model server tests causes the random failures.

Here's what we know:
* There aren't changes made to either project model server codes or tests.
* The random failure are all on Linux based CI.
* The random failure begin showing up recently. 
* There is a reconfiguration happened to Linux CI machines. 

In addition, in my investigation of the test log I found a common symptom. Over the course of the test run all failed tests happened at the same moment. The tests ended before the moment and the tests began after the moment all passed. As if the test machine frozen at a moment which causes the timeout.

_Update_

@eerhardt points out a PR in which tests failed on OSX and Windows as well: https://github.com/dotnet/cli/pull/3388.

### DIFF
--- a/test/dotnet-projectmodel-server.Tests/DthTestClient.cs
+++ b/test/dotnet-projectmodel-server.Tests/DthTestClient.cs
@@ -22,6 +22,7 @@ namespace Microsoft.DotNet.ProjectModel.Server.Tests
         private readonly NetworkStream _networkStream;
         private readonly BlockingCollection<DthMessage> _messageQueue;
         private readonly CancellationTokenSource _readCancellationToken;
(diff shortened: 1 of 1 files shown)


troydai:
Note: expand the timeout __won't__ slow down the test. It only affects the case when expected responses are missing, which are failed cases.

troydai:
@eerhardt @piotrpMSFT @brthor

eerhardt:
:shipit:

merging this to get it run on all our other PRs and see if it fixes the random errors.

brthor:
thanks @troydai

PR #3398, opened 2016-06-06
Copy pdb's for msbuildpackage references

pakrym:
https://github.com/dotnet/cli/issues/2961
Side effect of this is when package contains `.pdb` file next to '.dll' it would be copied to the output directory.

@davidfowl @cdmihai @anurse


anurse:

PR #3403, opened 2016-06-06
Porting serviceable change from rel/1.0.0 to 1.0.0-preview2

toddm:
Adding support for --serviceable option to pack command which puts <serviceable>true</serviceable> into the output nuspec

This addresses https://github.com/NuGet/Home/issues/2900

@eerhardt


eerhardt:
test  Windows_NT x64 Release Build

Issue #3385, opened 2016-06-04
StackOverflowException when a project references itself


niemyjski:
Running RC2 on Windows 10 x64 (fully updated) 

![screen shot 2016-06-04 at 4 20 17 pm](https://cloud.githubusercontent.com/assets/1020579/15802220/f9ff2cd4-2a70-11e6-8c3a-6ba01b8089bc.png)

```
Log Name:      Application
Source:        Application Error
Date:          6/4/16 4:23:35 PM
Event ID:      1000
Task Category: (100)
Level:         Error
Keywords:      Classic
User:          N/A
Computer:      BLAKENIEMYJF00D
Description:
Faulting application name: dotnet.exe, version: 0.0.0.0, time stamp: 0x573411d9
Faulting module name: coreclr.dll, version: 4.6.24027.0, time stamp: 0x572035de
Exception code: 0xc00000fd
Fault offset: 0x000000000005dd14
Faulting process id: 0x3ef4
Faulting application start time: 0x01d1bea75a5ca44b
Faulting application path: C:\Program Files\dotnet\dotnet.exe
Faulting module path: C:\Program Files\dotnet\shared\Microsoft.NETCore.App\1.0.0-rc2-3002702\coreclr.dll
Report Id: 85959d99-c027-485f-9645-1395620b0ceb
Faulting package full name: 
Faulting package-relative application ID: 
Event Xml:
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
    <Provider Name="Application Error" />
    <EventID Qualifiers="0">1000</EventID>
    <Level>2</Level>
    <Task>100</Task>
    <Keywords>0x80000000000000</Keywords>
    <TimeCreated SystemTime="2016-06-04T21:23:35.767619700Z" />
    <EventRecordID>30282</EventRecordID>
    <Channel>Application</Channel>
    <Computer>BLAKENIEMYJF00D</Computer>
    <Security />
  </System>
  <EventData>
    <Data>dotnet.exe</Data>
    <Data>0.0.0.0</Data>
    <Data>573411d9</Data>
    <Data>coreclr.dll</Data>
    <Data>4.6.24027.0</Data>
    <Data>572035de</Data>
    <Data>c00000fd</Data>
    <Data>000000000005dd14</Data>
    <Data>3ef4</Data>
    <Data>01d1bea75a5ca44b</Data>
    <Data>C:\Program Files\dotnet\dotnet.exe</Data>
[...]

PR #3413, opened 2016-06-06
Adding check for self-referencing projects

sokket:
Fixing bug #3385 by adding a check to make sure a project does not reference itself in project.json

/cc @piotrpMSFT @niemyjski

### DIFF
--- a/src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
+++ b/src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
@@ -43,7 +43,7 @@ namespace Microsoft.DotNet.Tools.Build
                 foreach (var dependency in project.Dependencies)
                 {
                     LibraryDescription libraryDescription;
-                    if (lookup.TryGetValue(dependency.Name, out libraryDescription))
+                    if ((lookup.TryGetValue(dependency.Name, out libraryDescription)) && (!libraryDescription.Identity.Name.Equals(project.Identity.Name)))
                     {
                         if (libraryDescription.Resolved && libraryDescription.Identity.Type.Equals(LibraryType.Project))
                         {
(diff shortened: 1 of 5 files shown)


src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
```diff
@@ -43,7 +43,7 @@ private ProjectGraphNode TraverseProject(ProjectDescription project, IDictionary
                 foreach (var dependency in project.Dependencies)
                 {
                     LibraryDescription libraryDescription;
-                    if (lookup.TryGetValue(dependency.Name, out libraryDescription))
+                    if ((lookup.TryGetValue(dependency.Name, out libraryDescription)) && (!libraryDescription.Identity.Name.Equals(project.Identity.Name)))
```
brthor: Can you add a test for this as well, so we don't regress the scenario?

sokket: Added...testing locally

sokket:

src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
brthor: Nice!

I'm not too familiar with this part of the code so tell me if I'm way off here. From looking at the `if` statement it seems like this would also fail if you had a package reference (not `target: project`) with the same name as the current project. 

Is that the case? If so should we scope it only to fail for `target: project` references?

sokket: Sure, I'll add the check for that

sokket: Ends up that the nuget restore will still think it is a cyclical reference if the dependency has the same name, regardless of the type. Due to this, I don't think we need to worry about only failing for project types since the same exact thing happens regardless of the type

piotrpMSFT:
:shipit:

src/dotnet/commands/dotnet-build/ProjectGraphCollector.cs
brthor: Sounds good to me. Just wanted to check :+1:

brthor:
:shipit:

PR #3423, opened 2016-06-07
Fixed links in README.md

svick:
There is no `master` branch, so the old links didn't work.

New links are relative, so they should work correctly on any branch.


eerhardt:
:shipit: The Ubuntu failure has been fixed with PR #3429. Since this is only a .md file update, no need to run CI.  Merging.

PR #3435, opened 2016-06-08
Minor improvements to build documentation

svick:
- explain how to speed up building using `Compile` target
- sort both sets of options in `build.sh` usage the same
- remove redundant and confusing `TARGETS` line from `build.sh` usage

### DIFF
--- a/build_projects/dotnet-cli-build/build.sh
+++ b/build_projects/dotnet-cli-build/build.sh
@@ -37,16 +37,15 @@ while [[ $# > 0 ]]; do
             export DOTNET_INSTALL_SKIP_PREREQS=1
             ;;
         --help)
-            echo "Usage: $0 [--configuration <CONFIGURATION>] [--skip-prereqs] [--nopackage] [--docker <IMAGENAME>] [--help] [--targets <TARGETS...>]"
+            echo "Usage: $0 [--configuration <CONFIGURATION>] [--targets <TARGETS...>] [--skip-prereqs] [--nopackage] [--docker <IMAGENAME>] [--help]"
             echo ""
             echo "Options:"
             echo "  --configuration <CONFIGURATION>     Build the specified Configuration (Debug or Release, default: Debug)"
             echo "  --targets <TARGETS...>              Comma separated build targets to run (Init, Compile, Publish, etc.; Default is a full build and publish)"
-            echo "  --nopackage                         Skip packaging targets"
             echo "  --skip-prereqs                      Skip checks for pre-reqs in dotnet_install"
+            echo "  --nopackage                         Skip packaging targets"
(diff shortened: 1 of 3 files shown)


build_projects/dotnet-cli-build/build.sh
```diff
@@ -41,12 +41,11 @@ while [[ $# > 0 ]]; do
             echo ""
             echo "Options:"
             echo "  --configuration <CONFIGURATION>     Build the specified Configuration (Debug or Release, default: Debug)"
-            echo "  --targets <TARGETS...>              Comma separated build targets to run (Init, Compile, Publish, etc.; Default is a full build and publish)"
-            echo "  --nopackage                         Skip packaging targets"
             echo "  --skip-prereqs                      Skip checks for pre-reqs in dotnet_install"
+            echo "  --nopackage                         Skip packaging targets"
```
eerhardt: Any reason you switched this around?

build_projects/dotnet-cli-build/build.sh
```diff
@@ -41,12 +41,11 @@ while [[ $# > 0 ]]; do
...
             echo "  --docker <IMAGENAME>                Build in Docker using the Dockerfile located in scripts/docker/IMAGENAME"
             echo "  --help                              Display this help message"
-            echo "  <TARGETS...>                        The build targets to run (Init, Compile, Publish, etc.; Default is a full build and publish)"
+            echo "  --targets <TARGETS...>              Comma separated build targets to run (Init, Compile, Publish, etc.; Default is a full build and publish)"
```
eerhardt: I think the `--targets` option is important enough, that it should be up at the top where it was.

Removing the "  <TARGETS...> " line is a good change though, since that doesn't work.  Thanks for the sharp eye!

build_projects/dotnet-cli-build/build.sh
svick: To make it consistent with [the Usage line above](https://github.com/dotnet/cli/pull/3435/files/33f61e7dcf02d14e81ea7a2a405ff076d1fe8669#diff-77fc6850a27234595d773b7c80f2280eR40).

build_projects/dotnet-cli-build/build.sh
svick: So what's the right order? Should I keep the order here the same, but reorder Usage?

eerhardt: My opinion:

- `--configuration`
- `--targets`
- `--skip-prereqs`
- `--nopackage`
- rest the same

svick: Done, here and also in build.ps1.

eerhardt:
:shipit: when it's green.

Thanks, @svick, for the contribution!

PR #3470, opened 2016-06-09
Publish symbol.nupkg files to myget.

### LINKED ISSUE #3445: CLI symbols.nupkg packages should be published to myget
We should be publishing our symbols packages to myget. This is how CoreFX and CoreCLR are handling their symbols.

Note: we still have binaries that aren't in nuget packages (for example dotnet.dll). We should either continue what we are doing for it - and publishing the symbols in a zip file on Azure blob, or we should call `dotnet pack` on it, and publish its symbols.nupkg to myget as well.

/cc @piotrpMSFT @brthor @dagood

eerhardt:
Fix #3445

@brthor @dagood


dagood:
LGTM!

PR #3491, opened 2016-06-10
Fix the VersionRepoUpdater to handle symbols.nupkg files correctly.

eerhardt:
When the VersionRepoUpdater pushes a change to dotnet/versions, it is putting a bunch of blank lines at the beginning of the Packages.txt file. This can break anyone trying to parse the file.

The problem is that our Regex to parse package Ids and versions based on the file name doesn't account for .symbols.nupkg files. Fixing this generates the correct file again.

See dotnet/versions@9ab2856 for an example of a bad Packages.txt file.

Porting the change from `core-setup`: https://github.com/dotnet/core-setup/pull/139

@brthor


eerhardt:
test OSX x64 Release Build

PR #3494, opened 2016-06-10
Adding machine ids to telemetry

adamgorMSFT:
Currently collected telemetry has no ability to accurately differentiate users or machines. 

Adding a hashed machine ID to collected telemetry to improve ability to differentiate distinct machines(users)


src/dotnet/project.json
```diff
@@ -54,7 +54,8 @@
     "Microsoft.Win32.Registry": {
       "version": "4.0.0-rc3-24201-00",
       "exclude": "compile"
-    }
+    },
+	"System.Net.NetworkInformation": "4.1.0-rc2-*"
```
eerhardt: We do not want * versions in our project.json files.  Especial "rc2" versions.  Please use the current CoreFX version.

src/dotnet/Telemetry.cs
```diff
@@ -146,5 +155,49 @@ private void TrackEventTask(string eventName, IDictionary<string, string> proper
...
+            }
+            
+            foreach (NetworkInterface adapter in nics)
+            {
+                IPInterfaceProperties properties = adapter.GetIPProperties();
+
+                PhysicalAddress address = adapter.GetPhysicalAddress();
+                byte[] bytes = address.GetAddressBytes();
+                macs.Add(string.Join("-", bytes.Select(x => x.ToString("X2"))));
+                if (macs.Count() >= 10)
```
eerhardt: Shouldn't use `.Count()` here - List<T> has a .Count property, which is much faster.

src/dotnet/Telemetry.cs
```diff
@@ -146,5 +155,49 @@ private void TrackEventTask(string eventName, IDictionary<string, string> proper
...
+                return macs;
+            }
+            
+            foreach (NetworkInterface adapter in nics)
+            {
+                IPInterfaceProperties properties = adapter.GetIPProperties();
+
+                PhysicalAddress address = adapter.GetPhysicalAddress();
+                byte[] bytes = address.GetAddressBytes();
+                macs.Add(string.Join("-", bytes.Select(x => x.ToString("X2"))));
```
eerhardt: Can't we just use UTF8 or some other encoding to turn these bytes into a string?  It would probably be much faster.

eerhardt: Also, looking below, we are just turning byte[]s into strings here. Then later, turning those strings back into byte[]s.  Why not just return List<byte[]> from here, and skip the back and forth?

src/dotnet/project.json
adamgorMSFT: changed.

src/dotnet/Telemetry.cs
adamgorMSFT: changed

src/dotnet/Telemetry.cs
adamgorMSFT: The reason they are getting turned into a string and back again, is an attempt to match and hash a specific format (a string in form of FF-FF-FF-FF-FF-FF) in as most similar way that VsCode hashes their mac address. 

Essentially I am trying to duplicate the following
[VsCode hashing](https://github.com/Microsoft/vscode/blob/224bbb8b453715395ec5e7b7a6dc546f7eb2c08b/src/vs/platform/telemetry/node/commonProperties.ts)
e.g. 
result = crypto.createHash('sha256').update(macAddress, 'utf8').digest('hex');
Which gets the 
[macAddress ](https://github.com/bevry/getmac/blob/master/src/lib/getmac.coffee) from a regex match from the string of getmac or ifconfig


If there is a way I can do this in a better way without the back and forth. I'd love to improve it, I am just not sure how to do this any other way.

eerhardt: Why would we care if the string is in a specific format?  It is hashed, so it should be 100% opaque to us.

IMO - it should look like this:

```
private List<byte[]> GetMacs()
{
    ... same as above only add address.GetAddressBytes() to the 'macs' list
}

       private List<string> HashSha256(List<byte[]> values)
        {
            var sha256 = SHA256.Create();
            var hashedStrings = new List<string>();
            foreach (var value in values)
            {
                byte[] hash = sha256.ComputeHash(value);
                string hashString = string.Empty;
                foreach (byte x in hash)
                {
                    hashString += String.Format("{0:x2}", x);
                }
                hashedStrings.Add(hashString);
            }
            return hashedStrings;
        }
```

adamgorMSFT: For future correlation purposes(e.g. matching the hashed value with VsCode's telemetry). I have added a comment in the code to indicate the need for a specific format.

PinpointTownes:
Am I really the only guy having a problem with all these crazy telemetry-related PRs?

Just like command line arguments, sending machine/environment-specific things like MAC addresses sucks really hard from a privacy perspective, even if they are hashed.

If you really want to be able to correlate collected traces, why not simply generating a unique identifier when installing .NET CLI and storing it somewhere in the user profile? Not perfect, but still much better than using MAC addresses for that.

GuardRex:
@PinpointTownes You're not the only one upset about these developments. It doesn't bother me that the program exists and provides MS with basic usage information, but it does bother me when sensitive, machine/location-identifiable information is sent to MS and the program is still being operated covertly.

The combination of ...

1. **Automatic activation** of telemetry with no notice to the dev/sysop, and
2. **No explicit notice** in the installer (or option to set the opt-out env var on install)

... makes the way that the program is being managed **dangerous** to Microsoft's objectives.

benaadams:
Just store an reuse a guid? So its specific but more anonymous

Rutix:
I have to agree with the ones before me. If the only purpose really is only to correlate data, storing a guid and using that would be way more privacy friendly than using the MAC address.

attilah:
It should be OPT_IN and not OPT_OUT. See the case of the Visual C++ team.

https://www.reddit.com/r/cpp/comments/4ibauu/visual_studio_adding_telemetry_function_calls_to/

Usually people takes telemetry as spying on them.

Yantrio:
Why does microsoft need to identify distinct machines ? Is there any documentation or discussion we can see to find out how and why this conclusion was drawn?

nil4:
Watch the last community standup, but turn down or mute the audio and focus on the body language. Does it look like the latest decisions were driven by consensus?

benaadams:
@Yantrio on would guess because ip address isn't a very good measure as you will get large activation clumping behind NATs/gateways/proxies. Much like you use a cookie in a browser rather than ip address to derive any sensible usage stats.

Yantrio:
My question isn't so much as to why it's required to use mac addresses vs IP, it's more as to why at all? is it important to know what a specific user is doing?

adamgorMSFT:
@Yantrio,  as already said in the other discussions linked below, "We're only interested in aggregate data that we can use to identify trends". Aggregated data can be quite useful for many different reasons. For example, it can help engineers prioritize features that will make the product better.

Also, there are several other discussions and documentation on this and other topics mentioned. 
https://github.com/dotnet/cli/pull/2145
https://github.com/dotnet/cli/issues/3093
https://github.com/dotnet/cli/issues/3404
https://blogs.msdn.microsoft.com/dotnet/2016/05/16/announcing-net-core-rc2/#telemetry

PinpointTownes:
> @Yantrio, as already said in the other discussions linked below, "We're only interested in aggregate data that we can use to identify trends". Aggregated data can be quite useful for many different reasons.

That doesn't explain why a random identifier wouldn't work in this case.

> For example, it can help engineers prioritize features that will make the product better.

You have GitHub, Uservoice and a bunch of other channels to hear what your community needs or wants :smile:

benaadams:
> You have GitHub, Uservoice and a bunch of other channels to hear what your community needs or wants

That doesn't capture what people use most, only what a louder subset moans about most:)

adamgorMSFT:
@PinpointTownes, Github and other community/user feedback sources are indeed leveraged quite a lot already.  Though they typically convey a different set of information.  It is more informative and accurate to have a complete picture from multiple sources, instead of just 1.    Having just 1 source can sometimes paint an inaccurate and biased scenario. 

Proof that the community feedback already influences decisions; you mentioned previously about "command line arguments". They were ultimately dropped from that earlier pull request. The decision to drop it was largely influenced because of community response.


Back on point as to why a random identifier won't work in this case at this time.  Though it would for the core usages, it won't for all.  But there are other scenarios it doesn't and that is a contributing factor why its done this specific way.  For 1 example, there is interest in VsCode/CLI correlation, as they are both popular to be used together.  VsCode collects a [hashed mac address](https://github.com/Microsoft/vscode/blob/master/src/vs/platform/telemetry/node/machineId.ts) as their MachineId, so if we want to correlate against that already collected data, that would be 1 reason to collect a comparable value. And as always, anyone can turn it off or compile a custom branch of the build.

richlander:
Thanks everyone for the feedback. I'm closing this PR now. I'll tell you why.

A little bit of context first. We added telemetry to the .NET Core Tools in RC2 I appreciated the feedback that folks made on that, which helped us to create a better product. #1 focus on the telemetry front is sharing usage data with you. We will not even consider making any more changes until that job is done. That's a promise.

Now, to this change. You can see that the change came in 20 days ago. This was during the height of shipping .NET Core 1.0. As a result, the PM team was super focussed on shipping 1.0 and didn't talk to the folks working on this PR. This is a poor excuse. I know you expect more from us.

With 1.0 out the door, we're going to take a closer look at our telemetry plan, both in terms of the actual telemetry and community engagement. We need telemetry for this product, however, "community engagement by PR" is not working. We need to adopt a different model for engagement. This is your .NET, and that needs to be more obvious from our engagement.

I ask for your continued engagement on this topic and to extend your patience a little bit longer.

Thanks everyone.

PR #3524, opened 2016-06-11
System.Private.CoreLib.dll IL binary is getting crossgen'd mistakenly.

### LINKED ISSUE #3522: Incorrect exclusion of System.Private.CoreLib for crossgen
See https://github.com/dotnet/core-setup/issues/150 for details.

System.Private.CoreLib.dll does not have “.dll” suffix in both core-setup (https://github.com/dotnet/core-setup/blob/6d22159ebd7508bf223d500593b2740d7d030d66/build_projects/shared-build-targets-utils/Utils/Crossgen.cs#L20) and CLI (https://github.com/dotnet/cli/blob/80d3476fd16609a41fde37e4cdaa78e15f659115/build_projects/shared-build-targets-utils/Utils/Crossgen.cs#L19).
 
As a result, https://github.com/dotnet/core-setup/blob/6d22159ebd7508bf223d500593b2740d7d030d66/build_projects/shared-build-targets-utils/Utils/Crossgen.cs#L186 check fails for System.Private.CoreLib.dll (since there is no such entry in the excluded file list) and we generate the NI and copy it as .dll

@brthor @piotrpMSFT PTAL.

CC @jkotas @eerhardt

eerhardt:
This causes both the IL ".dll" and the native image "ni.dll" to be the same size in the Shared Framework. This is incorrect, as the IL binary is much smaller.

The fix is to add ".dll" to the excludedLibraries list in the Crossgen util.

This ports the fix from core-setup:  https://github.com/dotnet/core-setup/pull/151

Fix #3522

@gkhanna79 @brthor

### DIFF
--- a/build_projects/shared-build-targets-utils/Utils/Crossgen.cs
+++ b/build_projects/shared-build-targets-utils/Utils/Crossgen.cs
@@ -17,7 +17,7 @@ namespace Microsoft.DotNet.Cli.Build
         {
             "mscorlib.dll",
             "mscorlib.ni.dll",
-            "System.Private.CoreLib",
+            "System.Private.CoreLib.dll",
             "System.Private.CoreLib.ni.dll"
         };
 


gkhanna79:
LGTM

eerhardt:

PR #3525, opened 2016-06-11
Add core dump generation

eerhardt:
We should be able to run the OSX leg of the PR over and over until we see the issue.

@gkhanna79 @jkotas @stephentoub @JohnChen0 @sergiy-k @janvorli

### DIFF
--- a/build_projects/dotnet-cli-build/build.sh
+++ b/build_projects/dotnet-cli-build/build.sh
@@ -102,13 +102,7 @@ then
 fi
 
 # Temporary logic to turn core dumps on so we can catch exit code 138 errors on OSX
-if [ "$(uname -s)" == "Darwin" ]; then
-  # If there are no core files already in /cores/, we'll enable core dump
-  # generation for this shell.
-  if [ "$(ls -A /cores)" ]; then 
-    ulimit -c unlimited
-  fi
-fi
+ulimit -c unlimited
 
 # Restore the build scripts
 echo "Restoring Build Script projects..."
--- a/build_projects/dotnet-cli-build/build.sh
+++ b/build_projects/dotnet-cli-build/build.sh
@@ -57,6 +57,28 @@ while [[ $# > 0 ]]; do
     shift
 done
 
+function print_info_from_core_file {
+  local core_file_name=$1
+  local executable_name=$2
+
+  if ! [ -e $executable_name ]; then
+    echo "Unable to find executable $executable_name"
+    return
+  elif ! [ -e $core_file_name ]; then
+    echo "Unable to find core file $core_file_name"
+    return
+  fi
+
+  # Check for the existence of GDB on the path
+  hash gdb 2>/dev/null || { echo >&2 "GDB was not found. Unable to print core file."; return; }
+
+  echo "Printing info from core file $1"
+
+  # Open the dump in GDB and print the stack from each thread. We can add more
+  # commands here if desired.
+  gdb --batch -ex "thread apply all bt full" -ex "quit" $executable_name $core_file_name
+}
+
 # Set nuget package cache under the repo
 export NUGET_PACKAGES="$REPOROOT/.nuget/packages"
 
@@ -101,7 +123,7 @@ then
     ulimit -n 1024
 fi
 
-# Temporary logic to turn core dumps on so we can catch exit code 138 errors on OSX
+# Temporary logic to turn core dumps on so we can catch segfaults on Unix
 ulimit -c unlimited
 
 # Restore the build scripts
@@ -121,4 +143,27 @@ echo "Invoking Build Scripts..."
 echo "Configuration: $CONFIGURATION"
 
 $DIR/bin/dotnet-cli-build ${targets[@]}
+
+# ======================= BEGIN Core File Inspection =========================
+if [ "$(uname -s)" == "Linux" ]; then
+  # Depending on distro/configuration, the core files may either be named "core"
+  # or "core.<PID>" by default. We read /proc/sys/kernel/core_uses_pid to 
+  # determine which it is.
+  core_name_uses_pid=0
+  if [ -e /proc/sys/kernel/core_uses_pid ] && [ "1" == $(cat /proc/sys/kernel/core_uses_pid) ]; then
+    core_name_uses_pid=1
+  fi
+
+  if [ $core_name_uses_pid == "1" ]; then
+    # We don't know what the PID of the process was, so let's look at all core
+    # files whose name matches core.NUMBER
+    for f in core.*; do
+      [[ $f =~ core.[0-9]+ ]] && print_info_from_core_file "$f" "dotnet" && rm "$f"
+    done
+  elif [ -f core ]; then
+    print_info_from_core_file "core" "dotnet"
+    rm "core"
+  fi
+fi
+
 exit $?
--- a/build_projects/dotnet-cli-build/build.sh
+++ b/build_projects/dotnet-cli-build/build.sh
@@ -101,6 +101,15 @@ then
     ulimit -n 1024
 fi
 
+# Temporary logic to turn core dumps on so we can catch exit code 138 errors on OSX
+if [ "$(uname -s)" == "Darwin" ]; then
+  # If there are no core files already in /cores/, we'll enable core dump
+  # generation for this shell.
+  if [ "$(ls -A /cores)" ]; then 
+    ulimit -c unlimited
+  fi
+fi
+
 # Restore the build scripts
 echo "Restoring Build Script projects..."
 (


build_projects/dotnet-cli-build/build.sh
```diff
@@ -101,6 +101,15 @@ then
     ulimit -n 1024
 fi
 
+# Temporary logic to turn core dumps on so we can catch exit code 138 errors on OSX
+if [ "$(uname -s)" == "Darwin" ]; then
+  # If there are no core files already in /cores/, we'll enable core dump
+  # generation for this shell.
+  if [ "$(ls -A /cores)" ]; then
```
gkhanna79: @sergiy-k Why would we need this check and not enable unconditionally?

gkhanna79: @eerhardt Please do the reverse (i.e. disable the generation) once the build script ends and do a "ls -l /cores" as well.

eerhardt: > Please do the reverse (i.e. disable the generation) once the build script ends

Is this necessary?  @adityamandaleeka didn't do the reverse in the buildtools change.

> and do a "ls -l /cores" as well.

Can you explain a little more what you want here?  What do you want to see happening?

gkhanna79: > Is this necessary?

We do not want crash dumps to be generated once the CLI build exits.

> What do you want to see happening?

Capture contents of cores folder (before and after) in the log to determine if we caught anything or not.

janvorli:
I was able to repro it locally on my OSX today. I have left it running in a loop for few hours and then there were two core files. Both had the same call stack, but since the release binaries have stripped symbols and I wasn't able to make lldb load the .dwarf symbol files, I was missing locals etc. So I have rebuilt the coreclr with the stripping and optimizations disabled (by patching the cmake / compiler options files) and copied that over the files in the dotnet tool subfolder. I am waiting for another repro with that.
This are top 10 frames from the stack:
```
(lldb) bt
* thread #1: tid = 0x0000, 0x000000010778a329 libcoreclr.dylib`AssemblyBindingHolder::CreateAssemblyBinding(LoaderHeap*) + 169, stop reason = signal SIGSTOP
  * frame #0: 0x000000010778a329 libcoreclr.dylib`AssemblyBindingHolder::CreateAssemblyBinding(LoaderHeap*) + 169
    frame #1: 0x00000001077891c3 libcoreclr.dylib`AssemblySpecBindingCache::StoreFile(AssemblySpec*, PEAssembly*) + 259
    frame #2: 0x00000001075d128e libcoreclr.dylib`AppDomain::AddFileToCache(AssemblySpec*, PEAssembly*, int) + 62
    frame #3: 0x00000001075d2721 libcoreclr.dylib`AppDomain::BindAssemblySpec(AssemblySpec*, int, int, StackCrawlMark*, AssemblyLoadSecurity*, int) + 1185
    frame #4: 0x0000000107713c20 libcoreclr.dylib`PEFile::LoadAssembly(unsigned int, IMDInternalImport*, char const*, char const*) + 352
    frame #5: 0x00000001075f50e8 libcoreclr.dylib`Module::LoadAssembly(AppDomain*, unsigned int, char const*, char const*) + 344
    frame #6: 0x00000001075e4add libcoreclr.dylib`Assembly::FindModuleByTypeRef(Module*, unsigned int, Loader::LoadFlag, int*) + 397
    frame #7: 0x000000010760902c libcoreclr.dylib`ClassLoader::LoadTypeDefOrRefThrowing(Module*, unsigned int, ClassLoader::NotFoundAction, ClassLoader::PermitUninstantiatedFlag, unsigned int, ClassLoadLevel) + 444
    frame #8: 0x000000010772b395 libcoreclr.dylib`SigPointer::GetTypeHandleThrowing(Module*, SigTypeContext const*, ClassLoader::LoadTypesFlag, ClassLoadLevel, int, Substitution const*, ZapSig::Context const*) const + 1797
    frame #9: 0x000000010777132f libcoreclr.dylib`ZapSig::DecodeType(Module*, Module*, unsigned char const*, ClassLoadLevel) + 111
    frame #10: 0x000000010771f38e libcoreclr.dylib`DynamicHelperFixup(TransitionBlock*, unsigned long*, unsigned int, Module*, CORCOMPILE_FIXUP_BLOB_KIND*, TypeHandle*, MethodDesc**, FieldDesc**) + 926
```
The failure point is here:
```asm
->  0x10778a329 <+169>: movq   $0x0, 0x70(%rcx)
```
The rcx + 0x70 seems to be a valid 16 byte aligned address in the dump, so the only thought that I have is that somehow the memory protection was set to read only (or maybe even PROT_NONE if the debugger would still show it from the dump in that case)

gkhanna79:
This is agreat @janvorli. Yesterday, the issue you ran into with checked CoreCLR was https://github.com/dotnet/coreclr/pull/5732, which is now fixed.

CC @JohnChen0 @fadimounir @jkotas since this is happening via DynamicHelperFixup.

eerhardt:
@janvorli - The CLI always uses the release build of CoreCLR. So if we get a core dump on a Jenkins machine, will it so us any good?  Will we be able to debug it?

janvorli:
@gkhanna79 great that it was fixed, I will then try to repro it with debug build too.

jkotas:
https://github.com/dotnet/coreclr/issues/5721 is around the same `AssemblySpecBindingCache` that where this AV is getting hit. I am wondering whether they have something in common.

eerhardt:
test OSX x64 Release Build

build_projects/dotnet-cli-build/build.sh
adityamandaleeka: @eerhardt @gkhanna79 

The reason why the BuildTools change doesn't reset ulimit is that each of those scripts is invoked in its own shell and the ulimit change is limited to the shell it's made in (and any children). I haven't looked into how this CLI script is invoked or by whom, but that may or may not apply here.

eerhardt:
test OSX x64 Release Build

This failed twice in a row on my other PR - and now it won't fail once...

eerhardt:
test OSX x64 Release Build

janvorli:
Strange - I've synced to the latest coreclr and I wasn't able to repro the issue anymore - running in a loop for at least two hours.

eerhardt:
test OSX x64 Release Build

eerhardt:
@janvorli - that's good, right?  ;)

I know we just took a fix for https://github.com/dotnet/coreclr/pull/5732 in release/1.0.0.  Which branch are you in?

eerhardt:
test OSX x64 Release Build

eerhardt:
We got what we needed out of this.

eerhardt:
@gkhanna79 @jkotas @stephentoub @JohnChen0 @sergiy-k @janvorli @adityamandaleeka  - I've re-opened this PR with the new logic for Linux.  I have made all of you collaborators on my fork, so you can push any change you want to this branch, if the logic isn't correct.  I may not be around very much today.

One place I didn't know exactly to put on this PR was passing in the executable name into `print_info_from_core_file`. The `dotnet` executable that is used at test time is contained under `$RepoRoot/artifacts/$RID/stage2`, I didn't know if I needed to pass the whole path to the executable or not.  There is a version of `dotnet` on the path which should match, but it isn't necessarily the path to the executable that failed.

janvorli:
@eerhardt - when I am trying to repro it locally, which cli branch should I use? I was using rel/1.0.0-preview2.

eerhardt:
Yes that is the correct branch where preview2 will be released from. It's the same branch that this PR is against.

eerhardt:
@dotnet-bot test this please

eerhardt:
test Ubuntu x64 Release Build

eerhardt:
Test Fedora23 x64 Debug Build

gkhanna79:
LGTM

eerhardt:
Looks like the RHEL leg failed with an exit code 139, but the core dump logging didn't log anything that I can see. Can someone take a look?

stephentoub:
I see:
```
Exit Code: 139
      StdOut:
      
      StdErr:
      
      
      Stack Trace:
           at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
           at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
        /mnt/resource/j/workspace/dotnet_cli/rel_1.0.0-preview2/release_rhel7.2_x64_prtest/test/Microsoft.DotNet.Tools.Tests.Utilities/Assertions/CommandResultAssertions.cs(32,0): at Microsoft.DotNet.Tools.Test.Utilities.CommandResultAssertions.Pass()
        /mnt/resource/j/workspace/dotnet_cli/rel_1.0.0-preview2/release_rhel7.2_x64_prtest/test/dotnet-publish.Tests/PublishStandaloneTests.cs(39,0): at Microsoft.DotNet.Tools.Publish.Tests.PublishStandaloneTests.StandaloneAppHasResourceDependency()
    Microsoft.DotNet.Tools.Publish.Tests.PublishStandaloneTests.StandaloneAppHasResourceDependency [FAIL]
```
Does that means it failed due to an unhandled exception?

eerhardt:
That's the random exit code we've been trying to capture. I had the core dump turned on, but it didn't look like the logging code logged anything.

stephentoub:
> I had the core dump turned on, but it didn't look like the logging code logged anything.

I'm wondering about the stack trace that was captured... where did that come from?

gkhanna79:
Yes, that usually comes when managed exception goes unhandled.

stephentoub:
> that usually comes when managed exception goes unhandled.

What exit code do we normally use when we crash due to an unhandled exception?  Is it that we normally use a different one, and something is causing us to seg fault while we're in the process of crashing due to the unhandled exception?

janvorli:
Unhandled exception aborts, so the exit code should be 134.

sergiy-k:
test OSX x64 Release Build
test OSX x64 Debug Build
test CentOS7.1 x64 Release Build
test RHEL7.2 x64 Debug Build

sergiy-k:

eerhardt:
We got what we needed out of this (for real this time).

PR #3526, opened 2016-06-11
Updating CoreFx, CoreSetup to rc3-24210-10, rc3-3004465-00 respectively

eerhardt:


eerhardt:
We will take another when the final Shared FX is available.

PR #3537, opened 2016-06-13
Updating CoreFx, CoreClr, CoreSetup to rc3-24212-01, 1.0.2, 5.0.1 respectively

eerhardt:


eerhardt:
We will take another when the Shared FX is available.

PR #3547, opened 2016-06-13
Updating NuGet to beta2-1484

eerhardt:
@zhili1208 @rrelyea @brthor @joshfree @piotrpMSFT


eerhardt:
test OSX x64 Release Build
test Ubuntu x64 Release Build
test  Windows_NT x64 Release Build

eerhardt:
Looks like we are going to have to wait for the RTM versions of CoreFX before this can get merged. We are getting errors that some packages are getting down/upgraded, and the result is that there is a mismatch of versions when crossgen'ing.

eerhardt:
This change was brought in by #3555.

PR #3551, opened 2016-06-14
[donotmerge] Enable ubuntu16 deb creation and publishing

brthor:
This cannot be fully tested, and will fail until a build is produced from https://github.com/dotnet/core-setup/pull/163, and we consume that in cli.

Putting this here for now though.

@piotrpMSFT


brthor:
Fixed via #3555

PR #3559, opened 2016-06-14
[DO NOT MERGE] Update known-issues.md

blackdwarf:
Add a better blurb about virtual networking. 

skip ci please

### DIFF
--- a/Documentation/known-issues.md
+++ b/Documentation/known-issues.md
@@ -27,6 +27,16 @@ sudo port -f uninstall openssl @0.9.8
 
 You can verify whether you have the right version using the  `openssl version` command from the Terminal. 
 
+## Running .NET Core CLI on Nano Server
+
+If you’re using Nano Server Technical Preview 5 with .NET Core CLI, you will need to copy all DLL files from c:\windows\system32\forwarders to c:\windows\system32, due to a bug that has since been fixed in later releases.
+ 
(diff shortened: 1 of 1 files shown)


Documentation/known-issues.md
```diff
@@ -12,6 +12,12 @@ brew install openssl
 brew link --force openssl
 ```
 
+Homebrew may also show the following warning:
+
+> Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries
+
+This warning is meant for the software that uses OpenSSL (in this case, .NET Core) and not for the end-user that is installing said software. Homebrew installation doesn't touch either the existing Apple crypto libraries or existing OpenSSL 0.9.8 version, so there is no impact on any software that uses either one of those crypto solutions and is already installed.
+
 MacPorts doesn't have the concept of linking, so it is reccomended that you uninstall 0.9.8 version of OpenSSL using the following command:
```
bartonjs: Existing typo in "recommended"

bartonjs:
The OpenSSL part LGTM.  Thanks, Z.  (https://github.com/dotnet/corefx/issues/9171#issuecomment-225690458 also has a bit about creating standalone OS X published applications, if you wanted to merge it in here.  Or any of my other words that you want to borrow)

blackdwarf:
Thanks @bartonjs. Merging this now.

PR #3563, opened 2016-06-14
update ubuntu 16 dockerfile from core-setup

brthor:
Update ubuntu 16 dockerfile.

I'm not sure how this was still working, llvm apt still added.

@livarcocc @ellismg @eerhardt @piotrpMSFT 

This is blocking our VSO builds


brthor:
didn't have latest locally, rebasing

Issue #3519, opened 2016-06-11
Assembly not found: Microsoft.CSharp


tmds:
After publishing my .net web app throws an exception when I try to start it:
```
Error: assembly specified in the dependencies manifest was not found -- package: 'Microsoft.CSharp', version: '4.0.1-rc3-24209-08', path: 'lib/netstandard1.3/Microsoft.CSharp.dll'
```

I use the 'latest' cli from github dotnet/cli on Debian 8.2: `1.0.0-preview2-003032`

This seems to be the same issue @GuardRex has: https://github.com/dotnet/cli/issues/3379

I am using the `https://www.myget.org/F/aspnetcidev/api/v3/index.json` feed. My dependencies in project.json are `-rc3-*`. I use the same dotnet version to publish and start the application.

In the dotnet folder, there is a Microsoft.CSharp.dll at `./shared/Microsoft.NETCore.App/1.0.0-rc3-004449-00/Microsoft.CSharp.dll`.

GuardRex:
My app worked for a few days and has regressed back to ...
```
Error: assembly specified in the dependencies manifest was not found -- 
package: 'Microsoft.CSharp', version: '4.0.1-rc3-24209-08', 
path: 'lib/netstandard1.3/Microsoft.CSharp.dll'
```
on the `aspnetvnext` feed. It's a vanilla portable app ...
```json
"dependencies": {
    "Microsoft.AspNetCore.Diagnostics": "1.0.0-*",
    "Microsoft.AspNetCore.Server.IISIntegration": "1.0.0-*",
    "Microsoft.Extensions.Logging.Console": "1.0.0-*",
    "Microsoft.AspNetCore.StaticFiles": "1.0.0-*",
    "Microsoft.AspNetCore.Mvc": "1.0.0-*",
    "Microsoft.AspNetCore.Server.Kestrel": "1.0.0-*",
    "Microsoft.NETCore.App": {
        "type": "platform",
        "version": "1.0.0-*"
    }
},
"frameworks": {
    "netcoreapp1.0": {}
},
```
Even if I convert it to an `rc2-final` app on a NuGet feed, I get the same issue, except it asks for a different version ...
```
Error: assembly specified in the dependencies manifest was not found -- 
package: 'Microsoft.CSharp', version: '4.0.1-rc2-24027', 
path: 'lib/netstandard1.3/Microsoft.CSharp.dll'
```
... and attempting to supply a dep for the package has no effect.

schellap:
[...]

PR #3574, opened 2016-06-14
Trim platfrom libraries from deps file

pakrym:


src/dotnet/commands/dotnet-publish/PublishCommand.cs
```diff
@@ -173,7 +173,10 @@ private bool PublishProjectContext(ProjectContext context, string buildBasePath,
             {
                 // Make executable in the new location
                 var executable = new Executable(context, buildOutputPaths, outputPath, buildOutputPaths.IntermediateOutputDirectoryPath, exporter, configuration);
-                executable.WriteConfigurationFiles(exports, FilterExports(exports, buildExclusionList), includeDevConfig: false);
+                var runtimeExports = exports.FilterExports(allExclusionList);
```
eerhardt: Can this result be cached above so we don't need to call this method twice?

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -159,7 +159,12 @@ private void WriteDepsFileAndCopyProjectDependencies(LibraryExporter exporter)
         {
             // When called this way we don't need to filter exports, so we pass the same list to both.
```
eerhardt: comment seems out of date now.

eerhardt:
looks good.  just a couple tiny comments

eerhardt:
:shipit: 

Fixes #3519

schellap:
@pakrym, confirming that build and publish use consistent code paths, correct? i.e., build=copy to deps file. publish=copy to file system.

Also wanted to confirm if native and resources were trimmed as well. Thanks!

LonghronShen:
Hi there, I'm using .NET Core 1.0 RTM, and the trimming problem is still there.
As far as I see, AutoMapper 5.1.1 and some portable class libraries won't load correctly. It says:

`Error: assembly specified in the dependencies manifest was not found -- package: 'AutoMapper', version: '5.1.1', path: 'lib/netstandard1.3/AutoMapper.dll'` 

or

'Error: assembly specified in the dependencies manifest was not found -- package: 'TestPortableClassLibrary', version: '1.0', path: 'lib/portable45-net45+win8+wpa81/CloOpenSDK.dll''

Any idea about this? Thank you!

piotrpMSFT:
@longhornshen can you open an issue and link this pr?

piotrpMSFT:
Also, a repro project would be wonderful!

LonghronShen:
Yeah, OK, I'll try to make a demo repo to help reproduce the problem and then create a new issue linking to this pr.

piotrpMSFT:
Thanks!

PR #3575, opened 2016-06-14
Fix eula.rtf formatting

### LINKED ISSUE #3573: eula.rtf is improperly formatted
The numbering in eula.rtf is incorrect. It goes from from 7->8->9->1->2->12. The 1->2 numbering is supposed to be 10->11. This renders incorrectly in our installers and if you just open the file in Word/Wordpad.

image

mellinoe:
The existing eula.rtf files that we are using in our installers are not formatted correctly. The numbering is wrong in a couple of places, and there a number in a numbered list which was not bold when it should have been. I've re-exported this file from our "master" docx file, and the formatting is now fixed.

Resolves #3573


mellinoe:

Issue #3621, opened 2016-06-16
Ubuntu 14.04 install directions are broken


BruceForstall:
## Steps to reproduce

Install by downloading 3 .deb packages as described here: https://github.com/dotnet/cli. Run

```
sudo dpkg -i <pkg.deb>
```

for each.

Error:
```
Unpacking dotnet-sharedframework-microsoft.netcore.app-1.0.0 (1.0.0-1) ...
dpkg: dependency problems prevent configuration of dotnet-sharedframework-microsoft.netcore.app-1.0.0:
 dotnet-sharedframework-microsoft.netcore.app-1.0.0 depends on dotnet-hostfxr-1.0.1; however:
  Package dotnet-hostfxr-1.0.1 is not installed.
```

So where are the instructions on how to download the hostfxr package?

## Environment data
`dotnet --info` output:

BruceForstall:
@schellap @gkhanna79

schellap:
Fixed by #3622

BruceForstall:
@schellap you only fixed the 1.0.0-preview2 branch. What about master? It's still broken, right?

schellap:
There is usually a mop-up RI between these branches... rel/1.0.0 has become the current home page branch, so fixed it there with https://github.com/dotnet/cli/pull/3654. @blackdwarf, please port to any other branches as necessary.

BruceForstall:
When you visit https://github.com/dotnet/cli, the fix isn't there.

PR #3622, opened 2016-06-16
Fix README to use hostfxr download links

schellap:
skip ci please

@brthor PTAL and confirm if this is the ordering you'd expect. Also, please let me know what the fourth column is for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,14 @@ You can download .NET Core as either an installer (MSI, PKG) or a zip (zip, gzip
 Ubuntu Installers
 ----------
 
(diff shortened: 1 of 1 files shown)


schellap:
Fixes #3621

README.md
```diff
@@ -58,6 +58,7 @@ Ubuntu Installers
 |         |Version |Installers|
```
brthor: right above this 

> we have three separate component packages  

Change to 
> we have four separate component packages

brthor:
Links look good, small comment on some of the wording, then :shipit:

README.md
schellap: I changed it to "we have separate component packages"

PR #3639, opened 2016-06-17
Introduce MSBuild to CLI

brthor:
This PR adds a command named `dotnet build3` which invokes msbuild to cli. 

As this PR sits, you can execute `/?` for help but not build anything useful. I've hacked things together locally to build a csproj, but work remains for that to be repeatable (hence WIP).

Work remaining here:
 - [x] Publish targets next to Microsoft.Build.dll so they're picked up correctly in `MSBuildToolsPath`
 - [x] Provide a property for the cli host path through `dotnet build3`
 - [x] Build & Consume a nuget package with NuGet.Build.Tasks converted to .Net Core. (I've done a conversion here: https://github.com/brthor/NuGet.BuildTasks )
 - [x] Add a test buiding a csproj and running the output

@eerhardt @livarcocc @piotrpMSFT @cdmihai


cdmihai:
@rainersigwald @AndyGerlicher

src/dotnet/commands/dotnet-build3/MSBuildForwardingApp.cs
```diff
@@ -0,0 +1,59 @@
...
+                depsFile: GetDepsFile(),
+                runtimeConfig: GetRuntimeConfig(),
+                environment: GetEnvironment()) { }
+
+        private static Dictionary<string, string> GetEnvironment()
+        {
+            return new Dictionary<string, string>
+            {
+                { "MSBuildExtensionsPath", GetDotnetMSBuildTargetsPath() },
+                { "DotnetHostPath", GetHostPath() }
```
brthor: @cdmihai @rainersigwald Calling this out because the cli needs to supply the path to the currently used host (which is not necessarily universal since multiple clis can be installed in different locations). 

Although this works, I'm not sure what the **right** thing to do here is. How will msbuild find this when it's not executed via the cli?

NuGet.Config
```diff
@@ -6,5 +6,7 @@
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="myget.org dotnet-buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
+    <add key="myget.org nugetbuildtasks netcore" value="https://myget.org/F/brthor-netcore-testpackages/api/v3/index.json" />
```
brthor: I hand rolled a .net core version of https://github.com/nuget/nuGet.BuildTasks and stuck it here. 
@jasonmalinowski Can we work together to come up with a long term plan for how these can be consumed in the cli, using .net core msbuild?

brthor: Post tests passing @eerhardt I will sync with you on moving msbuild packages and this build task package to cli-deps so we don't need to add these feeds here.

weshaggard: In general I thought we are trying to reduce the number of feeds as opposed to increasing them. Should these packages just be added to cli-deps?

brthor: Agreed @weshaggard my plan is not to check in with these feeds here.

src/dotnet/commands/dotnet-build3/Program.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.DotNet.Cli.Utils;
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+
+namespace Microsoft.DotNet.Cli
+{
+    public class Build3Command
```
weshaggard: Why build3? Why not just msbuild?

weshaggard:
cc @joperezr as he as experience with getting the .NET Core msbuild working on the shared framework for corefx builds.

src/dotnet/commands/dotnet-build3/Program.cs
brthor: Had msbuild at first, build3 via feedback from @Petermarcu and @piotrpMSFT

src/dotnet/commands/dotnet-build3/MSBuildForwardingApp.cs
brthor: The issue here is not only the host to execute msbuild, but the host msbuild should use to execute csc.exe. 

The first problem is like you described but we need additional data once we're inside msbuild to execute csc.exe successfully.

weshaggard: See https://github.com/Microsoft/msbuild/pull/681 where we essentially try using the host that msbuild is running under.

weshaggard: Also this issue https://github.com/dotnet/roslyn/issues/1159 might be more relevant to csc.

brthor:
@weshaggard @joperezr I based the msbuild dependency list off the list using for the tool-runtime in build tools minus the `NuProj` dependency. 

I didn't find it necessary to recase `Microsoft.CSharp.targets` however as you're doing here:
https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L96

And I'm not certain why we need the Portable Targets  (at least in the default installation)
`https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks/PackageFiles/init-tools.sh#L82`

The Csharp.Core targets with the NuGet.BuildTask were capable of building a .net core app runnable by `dotnet.exe` although there's still work to be done in producing deps/runtimeconfig files.

You have been dealing with this field much longer though so please let me know if I am missing something here.

joperezr:
> I didn't find it necessary to recase Microsoft.CSharp.targets however as you're doing here:

The reason why we had to do this is because we are using a private build of the portable targets, (which we do need in order to build corefx) and one of those targets has the wrong casing, which in windows is fine but x-plat it doesn't work.

> I used the tool-runtime project.json as a base for bringing in msbuild minus the NuProj dependency, 

right, actually that project.json contains not only the msbuild dependencies but also the buildtools dependencies, which is why there is a bunch of things there that you won't need.

build_projects/dotnet-cli-build/CompileTargets.cs
```diff
@@ -270,9 +274,30 @@ private static void RemovePdbsFromDir(string directory)
...
             }
 
+            CopyMSBuildTargetsToSDKRoot(sdkOutputDirectory);
+
             return c.Success();
         }
 
+        private static void CopyMSBuildTargetsToSDKRoot(string sdkOutputDirectory)
+        {
+            var msbuildTargetsDirectory = Path.Combine(sdkOutputDirectory, "runtimes", "any", "native");
```
eerhardt: "native"?  That seems odd since these aren't really native assets are they?

Also, why are they going into "runtimes\any", since these things aren't really "runtimes" assets.  It seems like they are more like normal content.

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
```diff
@@ -0,0 +1,106 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.Cli
+{
+    public class ForwardingApp
```
eerhardt: I think it would be good to have at least a high-level summary of what a "ForwardingApp" is/does.

eerhardt:
> using Microsoft.DotNet.Cli.Utils;

copyright on all new files.

---
Refers to: src/dotnet/commands/dotnet-build3/ForwardingApp.cs:1 in a73c35c. [](commit_id = a73c35cf70cb046d92ec5582a4268a8c5113c013, deletion_comment = False)

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            process.Start();
+            process.WaitForExit();
+
+            return process.ExitCode;
+        }
+
+        private string GetHostExeName()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
eerhardt: It would be nice to reuse some existing logic here.  How about `Microsoft.DotNet.ProjectModel.FileNameSuffixes.CurrentPlatform.Exe"?

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            if (_additionalProbingPath != null)
+            {
+                _allArgs = new string[] { "--additionalprobingpath", _additionalProbingPath }.Concat(_allArgs).ToArray();
+            }
+
+            _allArgs = new string[] { "exec" }.Concat(_allArgs).ToArray();
+
+            Console.WriteLine(_forwardApplicationPath);
```
eerhardt: Do we want to do this only for `--verbose`?

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
```diff
@@ -0,0 +1,106 @@
...
+            Dictionary<string, string> environment = null)
+        {
+            _forwardApplicationPath = forwardApplicationPath;
+            _argsToForward = argsToForward;
+            _depsFile = depsFile;
+            _runtimeConfig = runtimeConfig;
+            _additionalProbingPath = additionalProbingPath;
+            _environment = environment;
+
+            _allArgs = new string[] { _forwardApplicationPath }.Concat(_argsToForward).ToArray();
```
eerhardt: How about instead of new'ing up so many string[]s, we use a List<string> instead?  If you want something in the front, you can ".Insert(0)".

src/dotnet/commands/dotnet-build3/MSBuildForwardingApp.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        private static string GetHostPath()
+        {
+            return new Muxer().MuxerPath;
+        }
+
+        private static string GetDotnetMSBuildTargetsPath()
+        {
+            return Path.Combine(AppContext.BaseDirectory);
```
eerhardt: Any reason that you Path.Combine only 1 string?  Why not just return AppContext.BaseDirectory?

TestAssets/TestProjects/MSBuildTestApp/MSBuildTestApp.csproj
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
eerhardt: Can we slim this .csproj file down to the bare necessities?  That way can easily gauge what "hacks" are left in.

For example CodeAnalysisRuleSet, FileAlignment, Prefer32Bit seem like they can all be removed.

TestAssets/TestProjects/MSBuildTestApp/MSBuildTestApp.csproj
```diff
@@ -0,0 +1,51 @@
...
+    <DebugSymbols>true</DebugSymbols>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <NoStdLib>true</NoStdLib>
+    <DebugType>full</DebugType>
+    <PlatformTarget>x64</PlatformTarget>
+    <UseVSHostingProcess>false</UseVSHostingProcess>
+    <ErrorReport>prompt</ErrorReport>
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    <Prefer32Bit>true</Prefer32Bit>
+    <NoLogo>true</NoLogo>
```
eerhardt: At some point, it would be really nice to track each "hack" in this .csproj with an issue on GitHub.  Almost like:

```XML
<NoLogo>true</NoLogo>   <!--  URL to GitHub issue -->
```

test/dotnet-build3.Tests/GivenDotnetBuild3BuildsCsproj.cs
```diff
@@ -0,0 +1,41 @@
...
+            var testProjectDirectory = testInstance.TestRoot;
+
+            var build3Command = new TestCommand("dotnet");
+            build3Command.WorkingDirectory = testProjectDirectory;
+
+            build3Command.ExecuteWithCapturedOutput("build3")
+                .Should()
+                .Pass();
+
+            var outputDll = Path.Combine(testProjectDirectory, "bin", "Debug", "netcoreapp1.0", testAppName);
```
eerhardt: This variable doesn't appear to be used.

test/dotnet-build3.Tests/GivenDotnetBuild3BuildsCsproj.cs
```diff
@@ -0,0 +1,41 @@
...
+            var outputRunCommand = new TestCommand("dotnet");
+
+            outputRunCommand.ExecuteWithCapturedOutput("build3")
```
eerhardt: I'm not sure I follow this logic.  Shouldn't you be running the output assembly?

src/dotnet/commands/dotnet-pack/PackageGenerator.cs
```diff
@@ -58,7 +59,7 @@ public bool BuildPackage(IEnumerable<ProjectContext> contexts, List<DiagnosticMe
 
             var packageOutputPath = Path.Combine(
                 ArtifactPathsCalculator.PackageOutputPath,
-                GetPackageName() + NuGet.Constants.PackageExtension);
+                GetPackageName() + NuGetConstants.PackageExtension);
```
eerhardt: This change seems totally unrelated to the current work.  It really doesn't make sense to include it IMO.

brthor: It is necessary otherwise an error is thrown that Constants doesn't exist in the namespace Microsoft.NuGet which I believe is brought in by a build task

src/dotnet/commands/dotnet-build3/MSBuildForwardingApp.cs
brthor: Left over. Will take a cleanup pass here

test/dotnet-build3.Tests/GivenDotnetBuild3BuildsCsproj.cs
brthor: copypasta from above, fixing now

src/dotnet/commands/dotnet-pack/PackageGenerator.cs
eerhardt: Ah, I get it.  Thanks for clarifying.

build_projects/dotnet-cli-build/CompileTargets.cs
brthor: It was a convention that non-managed IL files could be placed in runtimes/any/native to be put in publish output. 

MSBuild has used this in their packages, however in a portable publish, at some point the subdirectory structure was created instead of files copied to root output.

eerhardt: Not sure I follow.  Is this a "point in time" thing?  Or do we expect this is how these packages will be for RTM?

brthor: I don't know, this is the way many packages are authored in the cli ecosystem currently. 

I think it's something we should fix though

rainersigwald: No reason to keep MSBuild's NuGet layout as it currently is if there's a better way. We'll just have to make sure changes get accounted for in dotnet/buildtools.

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
brthor: Perhaps a clearer name would be sufficient `NewProcessArgForwardingApp`

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
brthor: I think the perf impact here is minimal but will make the change.

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
brthor: this is here for debugging only

build_projects/dotnet-cli-build/CompileTargets.cs
brthor: I'm not certain what is the better way here. How do we publish non-IL files from packages with output? Do copyToOutput content files work? I had seem some issues suggesting they do not

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
eerhardt: Having one or two sentences in a comment that describe the high-level of what a class is/does isn't the end of the world.  Rarely does the original intention of a whole class change.

In general I agree with the approach of "strive to make your code as readable as possible without comments".  But I disagree with the approach of "never ever ever use comments".

`NewProcessArgForwardingApp` isn't very readable, IMO.

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
eerhardt: So will it be removed?  I think putting it under `--verbose` is a good idea and keeping it in, that way we can debug in the future.

piotrpMSFT:
please retarget to feature/msbuild branch.

build_projects/dotnet-cli-build/CompileTargets.cs
piotrpMSFT: @joelverhagen @emgarten what's the guidance from NuGet in this scenario?

src/dotnet/commands/dotnet-build3/Program.cs
piotrpMSFT: The eventual verb will continue to be `build`. We didn't want to introduce `msbuild` as a concept here since folks are likely to assume it will stick around. `dotnet build` is where we're headed.

build_projects/dotnet-cli-build/CompileTargets.cs
emgarten: ``contentFiles`` with ``copyToOuput=true`` sounds like what you want here. The file type can be anything. ``runtimes/{rid}/native`` on the other hand is only for native dlls.

src/dotnet/commands/dotnet-pack/PackageGenerator.cs
emgarten: There is a better constant for this in NuGet.Packaging.Core which you are already using. 

``PackagingCoreConstants.NupkgExtension``

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
```diff
@@ -0,0 +1,106 @@
...
+    public class ForwardingApp
+    {
+        private static readonly string s_hostExe = "dotnet";
+
+        private string _forwardApplicationPath;
+        private string[] _argsToForward;
+        private string _depsFile;
+        private string _runtimeConfig;
+        private string _additionalProbingPath;
+        private Dictionary<string, string> _environment;
```
emgarten: it looks like these can be readonly

build_projects/dotnet-cli-build/CompileTargets.cs
brthor: Per convo with @eerhardt contentFiles copy to output is working in publish today, I was mistaken. 

Is contentFiles the right approach for msbuild.exe though? which is a managed assembly

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
brthor: added

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
brthor: done.

src/dotnet/commands/dotnet-build3/ForwardingApp.cs
brthor: removed

TestAssets/TestProjects/MSBuildTestApp/MSBuildTestApp.csproj
brthor: cleaned this up quite a bit.

TestAssets/TestProjects/MSBuildTestApp/MSBuildTestApp.csproj
brthor: added comments where tracking issues exist.

test/dotnet-build3.Tests/GivenDotnetBuild3BuildsCsproj.cs
brthor: fixed

brthor:
Migrated to the feature branch
https://github.com/dotnet/cli/pull/3680

build_projects/dotnet-cli-build/CompileTargets.cs
eerhardt: Looking at the `Microsoft.Net.Compilers.netcore\1.3.0` package from Roslyn, the `csc.exe` and `vbc.exe` files come in `runtimes\any\native` as well.

Why wouldn't these binaries come via the "normal" way of `lib\<TFM>`?  That's how `NuGet.CommandLine.XPlat.dll` is packaged.

PR #3645, opened 2016-06-18
Publish sdk zips to Azure

schellap:
@piotrpMSFT this is to allow Antares to take new SDK zips without having the problem of overwriting other components in use. Since the M.N.A has stabilized to 1.0.0, the SDK+FX zips are attempting to overwrite the runtime libraries affecting their unzip deployment.

Cc @brthor 

Eventually, I assume we would have a file called "artifacts.txt" similar to the version badge, that is a summary of the build output and inputs in Azure that our consumers can use.


schellap:
Closing this PR, as this might cause SDK FX version mismatches...

eerhardt:
@schellap - this PR is for `preview2`, which is basically in lock down.  Do you intend for this change to go into `preview2`?  If so, we will need approval.  If not, can you move it to `rel/1.0.0`?

schellap:
I'll move it to `rel/1.0.0`. Will hand upload these for preview2. Thanks!

PR #3678, opened 2016-06-22
Fixing AzurePublisher to upload files to the correct URL.

eerhardt:
Previously it was duplicating the full URL inside of the blob container.

@brthor @livarcocc


brthor:
:shipit:

PR #3686, opened 2016-06-23
Fix the broken .NET Core Installer and Binaries links on README.

eerhardt:
My last checkin mistaken replaced the .NET Core links to point to the new SDK publish layout.  This changes it back.

I also removed the "phantom 4th column" in the Ubuntu installers table. This column was hidden because there is no header for it. The info it contained is in the full table.

skip ci please

@anurse @brthor


brthor:
clicked a bunch of the links, they seem to work :shipit:

anurse:
:shipit:

PR #3687, opened 2016-06-23
Don't use black color in the console

svick:
In the standard white text on black background console, text printed like this is invisible.

There don't seem to be any other uses of `.Black()` in the solution.


piotrpMSFT:
LGTM. Thanks @svick

PR #3696, opened 2016-06-24
msbuild package targets

brthor:
Creating this for early feedback, because Package has many dependencies on state generation that we need to iron out before we merge this. 

Archives are currently working here. Nuget package generation should work but is untested. Installers are not working yet.

I used conventions I thought made sense and made the msbuild files clean. Let's come to consensus on these.

@eerhardt @sokket @livarcocc @piotrpMSFT

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -5,6 +5,7 @@
     <clear />
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -5,7 +5,6 @@
     <clear />
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
-    <add key="buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>
--- a/build.proj
+++ b/build.proj
@@ -26,13 +26,21 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <DotnetCliBuildFrameworkInputs Include="build_projects/**/*.cs" />
-    <DotnetCliBuildFrameworkInputs Include="build_projects/**/*.json" />
+    <DotnetCliBuildFrameworkInputs Include="build_projects/**/*.cs" Exclude="build_projects/**/obj/**/*.cs" />
+    <DotnetCliBuildFrameworkInputs Include="build_projects/**/project.json;build_projects/**/project.lock.json" />
   </ItemGroup>
+  
+  <!-- Workaround to "Native image cannot be loaded multiple times" issue 
+       A target in the top level file needs to run and invoke a task 
+       https://github.com/Microsoft/msbuild/issues/750 -->
+  <Target Name="MSBuildWorkaroundTarget">
+    <Message Text="Dont remove this target" />
+  </Target>
 
   <Target Name="BuildDotnetCliBuildFramework" 
           Inputs="@(DotnetCliBuildFrameworkInputs)" 
-          Outputs="$(CLIBuildDll)">
+          Outputs="$(CLIBuildDll)"
+          DependsOnTargets="MSBuildWorkaroundTarget">
     <Exec Command="$(PlatformScriptHost) $(MSBuildThisFileDirectory)/build_projects/dotnet-cli-build/build$(PlatformScriptExtension) $(NoRunArg)" WorkingDirectory="$(MSBuildThisFileDirectory)"/>
   </Target>
 
@@ -40,7 +48,6 @@
 
   <UsingTask TaskName="CompileTargets" AssemblyFile="$(CLIBuildDll)" />
   <UsingTask TaskName="TestTargets" AssemblyFile="$(CLIBuildDll)" />
-  <UsingTask TaskName="PackageTargets" AssemblyFile="$(CLIBuildDll)" />
   <UsingTask TaskName="PublishTargets" AssemblyFile="$(CLIBuildDll)" />
 
   <Target DependsOnTargets="BuildDotnetCliBuildFramework" Name="Compile">
@@ -51,13 +58,10 @@
     <TestTargets />
   </Target>
 
-  <Target DependsOnTargets="BuildDotnetCliBuildFramework" Name="Package">
-    <PackageTargets />
-  </Target>
-
   <Target DependsOnTargets="BuildDotnetCliBuildFramework" Name="Publish">
     <PublishTargets />
   </Target>
 
   <Import Project="build/Microsoft.DotNet.Cli.Prepare.targets" />
+  <Import Project="build/Microsoft.DotNet.Cli.Package.targets" />
 </Project>
--- a/build/Microsoft.DotNet.Cli.Package.targets
+++ b/build/Microsoft.DotNet.Cli.Package.targets
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Layout" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildThisFileDirectory)/package/Microsoft.DotNet.Cli.Layout.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)/package/Microsoft.DotNet.Cli.Archive.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)/package/Microsoft.DotNet.Cli.Nupkg.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)/package/Microsoft.DotNet.Cli.Installer.DEB.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)/package/Microsoft.DotNet.Cli.Installer.MSI.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)/package/Microsoft.DotNet.Cli.Installer.PKG.targets" />
+
+  <Target Name="GenerateInstallers" 
+          DependsOnTargets="Init;Layout;GeneratePkgs;GenerateDebs;GenerateMsis" />
+
+  <Target Name="Package"
+          DependsOnTargets="BuildDotnetCliBuildFramework;
+                            Init;
+                            Layout;
+                            GenerateNugetPackages;
+                            GenerateArchives;
+                            GenerateInstallers"  />
+</Project>
--- a/build/Microsoft.DotNet.Cli.Prepare.targets
+++ b/build/Microsoft.DotNet.Cli.Prepare.targets
@@ -25,6 +25,7 @@
       <Stage2CompilationDirectory>$(BaseOutputDirectory)/stage2compilation</Stage2CompilationDirectory>
       <IntermediateDirectory>$(BaseOutputDirectory)/intermediate</IntermediateDirectory>
       <PackagesDirectory>$(BaseOutputDirectory)/packages</PackagesDirectory>
+      <SymbolsOutputDirectory>$(BaseOutputDirectory)/stage2symbols</SymbolsOutputDirectory>
     </PropertyGroup>
 
     <GenerateBuildVersionInfo RepoRoot="$(RepoRoot)">
@@ -46,6 +47,10 @@
     <PropertyGroup>
       <VersionBadge>$(BaseOutputDirectory)/$(VersionBadgeMoniker)_$(Configuration)_version_badge.svg</VersionBadge>
       <SdkVersion>$(NugetVersion)</SdkVersion>
+      
+      <SdkProjectJsonVersion>$(VersionMajor).$(VersionMinor).$(VersionPatch)-preview3-$(CommitCount)</SdkProjectJsonVersion>
+      <ProjectModelProjectJsonVersion>$(VersionMajor).$(VersionMinor).$(VersionPatch)-rc4-$(CommitCount)</ProjectModelProjectJsonVersion>
+      <DependencyModelAndInternalAbstractionsProjectJsonVersion>$(VersionMajor).$(VersionMinor).1-beta-$(CommitCount)</DependencyModelAndInternalAbstractionsProjectJsonVersion>
 
       <ArchiveExtension Condition=" '$(OSName)' == 'win' ">.zip</ArchiveExtension>
       <ArchiveExtension Condition=" '$(OSName)' != 'win' ">.tar.gz</ArchiveExtension>
@@ -61,18 +66,15 @@
       <ProductMonikerRid Condition=" '$(Rid)' == 'ubuntu.16.04-x64' Or '$(Rid)' == 'fedora.23-x64' Or '$(Rid)' == 'opensuse.13.2-x64' ">$(Rid)</ProductMonikerRid>
       <ProductMonikerRid Condition=" '$(ProductMonikerRid)' == '' ">$(OSName)-$(Architecture)</ProductMonikerRid>
 
-      <!-- Generated Installers + Archives -->
-      <SdkCompressedFile>$(PackagesDirectory)/dotnet-sdk-$(ProductMonikerRid).$(SdkVersion)$(ArchiveExtension)</SdkCompressedFile>
-      <SdkInstallerFile Condition=" '$(InstallerExtension)' != '' ">$(PackagesDirectory)/dotnet-sdk-$(ProductMonikerRid).$(SdkVersion)$(InstallerExtension)</SdkInstallerFile>
+      <ArtifactNameSdk>dotnet-sdk</ArtifactNameSdk>
+      <ArtifactNameSdkDebug>dotnet-sdk-debug</ArtifactNameSdkDebug>
+      <ArtifactNameCombinedHostHostFxrFrameworkSdk>dotnet-dev</ArtifactNameCombinedHostHostFxrFrameworkSdk>
+      <ArtifactNameCombinedFrameworkSdk>dotnet-sharedframework-sdk</ArtifactNameCombinedFrameworkSdk>
 
-      <CombinedFrameworkSDKHostCompressedFile>$(PackagesDirectory)/dotnet-dev-$(ProductMonikerRid).$(SdkVersion)$(ArchiveExtension)</CombinedFrameworkSDKHostCompressedFile>
-      <CombinedFrameworkSDKHostInstallerFile Condition=" '$(BundleExtension)' != '' ">$(PackagesDirectory)/dotnet-dev-$(ProductMonikerRid).$(SdkVersion)$(BundleExtension)</CombinedFrameworkSDKHostInstallerFile>
-
-      <CombinedFrameworkSDKCompressedFile>$(PackagesDirectory)/dotnet-sharedframework-sdk-$(ProductMonikerRid).$(SdkVersion)$(ArchiveExtension)</CombinedFrameworkSDKCompressedFile>
-      <CombinedFrameworkSDKInstallerFile Condition=" '$(BundleExtension)' != '' ">$(PackagesDirectory)/dotnet-sharedframework-sdk-$(ProductMonikerRid).$(SdkVersion)$(BundleExtension)</CombinedFrameworkSDKInstallerFile>
-
-      <SdkSymbolsCompressedFile>$(PackagesDirectory)/dotnet-sdk-debug-$(ProductMonikerRid).$(SdkVersion)$(ArchiveExtension)</SdkSymbolsCompressedFile>
-      <SdkSymbolsInstallerFile Condition=" '$(InstallerExtension)' != '' ">$(PackagesDirectory)/dotnet-sdk-debug-$(ProductMonikerRid).$(SdkVersion)$(InstallerExtension)</SdkSymbolsInstallerFile>
+      <ArtifactNameWithVersionSdk>$(ArtifactNameSdk)-$(ProductMonikerRid).$(SdkVersion)</ArtifactNameWithVersionSdk>
+      <ArtifactNameWithVersionSdkDebug>$(ArtifactNameSdkDebug)-$(ProductMonikerRid).$(SdkVersion)</ArtifactNameWithVersionSdkDebug>
+      <ArtifactNameWithVersionCombinedHostHostFxrFrameworkSdk>$(ArtifactNameCombinedHostHostFxrFrameworkSdk)-$(ProductMonikerRid).$(SdkVersion)</ArtifactNameWithVersionCombinedHostHostFxrFrameworkSdk>
+      <ArtifactNameWithVersionCombinedFrameworkSdk>$(ArtifactNameCombinedFrameworkSdk)-$(ProductMonikerRid).$(SdkVersion)</ArtifactNameWithVersionCombinedFrameworkSdk>
 
       <!-- Downloaded Installers + Archives -->
       <SharedHostCompressedFile>$(PackagesDirectory)/dotnet-host-$(ProductMonikerRid).$(SharedHostVersion)$(ArchiveExtension)</SharedHostCompressedFile>
--- a/build/package/Microsoft.DotNet.Cli.Installer.DEB.targets
+++ b/build/package/Microsoft.DotNet.Cli.Installer.DEB.targets
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    
+    <UsingTask TaskName="DebTargets" AssemblyFile="$(CLIBuildDll)" />
+
+    <Target Name="GenerateDebs" >
+        <DebTargets />
+    </Target>
+
+</Project>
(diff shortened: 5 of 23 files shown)


build/Microsoft.DotNet.Cli.Package.targets
```diff
@@ -0,0 +1,39 @@
...
+    <ProjectsToPack Include="Microsoft.DotNet.Files" />
+    <ProjectsToPack Include="Microsoft.DotNet.InternalAbstractions" />
+    <ProjectsToPack Include="Microsoft.DotNet.ProjectModel" />
+    <ProjectsToPack Include="Microsoft.DotNet.ProjectModel.Loader" />
+    <ProjectsToPack Include="Microsoft.DotNet.ProjectModel.Workspaces" />
+    <ProjectsToPack Include="Microsoft.Extensions.DependencyModel" />
+    <ProjectsToPack Include="Microsoft.Extensions.Testing.Abstractions" />
+  </ItemGroup>
+
+  <!-- these are all the things I need from Prepare -->
```
brthor: These need to come from prepare, which doesn't exist yet/

brthor:
> TestAssets/TestProjects/MSBuildTestApp.zip

Will remove this.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -1,60 +1,69 @@
...
         <CombinedFrameworkSdk>dotnet-sharedframework-sdk</CombinedFrameworkSdk>
       </ArtifactNames>
 
-      <ArtifactNamesWithVersions Include="None">
-        <Sdk>@(ArtifactNames -> '%(Sdk)-$(SdkVersion)')</Sdk>
-        <SdkDebug>@(ArtifactNames -> '%(SdkDebug)-$(SdkVersion)')</SdkDebug>
-        <CombinedHostHostFxrFrameworkSdk>@(ArtifactNames -> '%(CombinedHostHostFxrFrameworkSdk)-$(SdkVersion)')</CombinedHostHostFxrFrameworkSdk>
-        <CombinedFrameworkSdk>@(ArtifactNames -> '%(CombinedFrameworkSdk)-$(SdkVersion)')</CombinedFrameworkSdk>
+      <ArtifactNamesWithVersions Include="ArtifactNamesWithVersions">
+        <aSdk>dotnet-sdk-$(SdkVersion)</aSdk>
```
eerhardt: I'm wondering if these should just be properties, and not using an Item to group these properties. It feels weird to me because:
1. There aren't "many" of these items, just one.
2. The fact that we have "Include" with the same name as the XML element, kind of goes to show that there should only be a single one of these, and makes them feel more like properties to me.
(I'm no MSBuild expert, just wondering what the best practice is in situations like this.)

@weshaggard - my resident "MSBuild expert" to get his thoughts on this pattern.

weshaggard: Seeing how this is done I would just make them a set of properties. I would just call them ArtifactSDK and ArtifactWithVersionSDK. That will allow you to consume these more easily as well as consume them outside of targets if you need to (accessing metadata can only happen in targets in msbuild).

brthor: I was liking this pattern because it gave me a sort of namespacing of the properties, but ran into some trouble with bringing these properties into other itemgroup metadata and batching on that metadata. An empty metadata would somehow be included, so I will be converting these to properties

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -42,6 +43,17 @@
     <PropertyGroup>
       <SdkVersion>@(BuildVersion ->'%(NugetVersion)')</SdkVersion>
       <SdkBrandName>Microsoft .NET Core 1.0.0 - SDK Preview 2</SdkBrandName>
+
+      <ArtifactNameSdk>dotnet-sdk</ArtifactNameSdk>
```
brthor: @weshaggard @eerhardt 

I've moved these to properties but having properties that we can only differentiate in the global property bag with a prefix as part of the name feels more cluttered. 

Is this the recommended pattern? Are there cleaner ways to go about this?

eerhardt: The quick and dirty way I did this, I feel is an anti-pattern, but I don't know how to do it the "MSBuild way".  It was a lot of duplication.  I have a PR short-coming for "Init".

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.DotNet.InternalAbstractions;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Security.Cryptography;
+
+using static Microsoft.DotNet.Cli.Build.Framework.BuildHelpers;
+
+namespace Microsoft.DotNet.Cli.Build
+{
+    public class GenerateGuidFromName : Task
```
eerhardt: In general, for shared code like this, I'd say let's either:

1. Delete the existing code, and have it call into this Task class.
 - For example, anywhere that calls Utils.GenerateGuidFromName, could just call GenerateGuidFromName.GenerateGuid.
2. Have the Task class call into the existing shared util.

That way we don't have duplication as we go.

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: See here:  https://github.com/dotnet/cli/pull/3710/files#diff-d595ac18a42f088f16eefcd37b417b04R64

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
brthor: Existing code will be deleted.

brthor:
@livarcocc @eerhardt @sokket I'm going to disable these DTH tests..

build/package/Microsoft.DotNet.Cli.Nupkg.targets
```diff
@@ -0,0 +1,59 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Layout" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  
+  <Target Name="SetupGenerateNugetPackagesInputsOutputs"
+          DependsOnTargets="Init">
+    <PropertyGroup>
+      <NupkgVersionSuffix>@(BuildVersion -> '%(CommitCount)')</NupkgVersionSuffix>
+      <ProjectsSrcDirectory>$(RepoRoot)/src</ProjectsSrcDirectory>
+      <PackagingBuildBasePath>$(Stage2CompilationDirectory)/forPackaging</PackagingBuildBasePath>
+      <NupkgOutputDirectory>$(BaseOutputDirectory)/packages</NupkgOutputDirectory>
```
eerhardt: This can use the `<PackagesDirectory>$(BaseOutputDirectory)/packages</PackagesDirectory>` from the Init task.

build/package/Microsoft.DotNet.Cli.Installer.DEB.targets
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    
+    <UsingTask TaskName="DebTargets" AssemblyFile="$(CLIBuildDll)" />
+
+    <Target Name="GenerateDebs" >
+        <DebTargets />
+    </Target>
+
+</Project>
```
piotrpMSFT: trailing newlines throughout

brthor: They should be added?

build/package/Microsoft.DotNet.Cli.Nupkg.targets
brthor: done.

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
brthor: After thinking about this I cannot remove the existing code yet but I don't want this code to depend on it. When we make the build.cmd switch I will go through and clean up.

brthor:
Feeling good about this, please take a review pass.

build.proj
```diff
@@ -26,21 +26,27 @@
...
-    <DotnetCliBuildFrameworkInputs Include="build_projects/**/*.json" />
+    <DotnetCliBuildFrameworkInputs Include="build_projects/**/*.cs" Exclude="build_projects/**/obj/**/*.cs" />
+    <DotnetCliBuildFrameworkInputs Include="build_projects/**/project.json;build_projects/**/project.lock.json" />
   </ItemGroup>
+  
+  <!-- Workaround to "Native image cannot be loaded multiple times" issue 
+       A target in the top level file needs to run and invoke a task -->
+  <Target Name="MSBuildWorkaroundTarget">
+    <Message Text="Dont remove this target" />
+  </Target>
```
brthor: This works around that `Init` issue we ran into.

build/package/Microsoft.DotNet.Cli.Nupkg.targets
```diff
@@ -0,0 +1,69 @@
...
+        <ProjectName>Microsoft.DotNet.ProjectModel.Workspaces</ProjectName>
+        <Version>$(VersionMajor).$(VersionMinor).$(VersionPatch)-preview3-$(CommitCount)</Version>
+      </ProjectsToPack>
+      <ProjectsToPack Include="$(ProjectsSrcDirectory)/Microsoft.Extensions.DependencyModel">
+        <ProjectName>Microsoft.Extensions.DependencyModel</ProjectName>
+        <Version>1.0.1-beta-$(CommitCount)</Version>
+      </ProjectsToPack>
+      <ProjectsToPack Include="$(ProjectsSrcDirectory)/Microsoft.Extensions.Testing.Abstractions">
+        <ProjectName>Microsoft.Extensions.Testing.Abstractions</ProjectName>
+        <Version>$(VersionMajor).$(VersionMinor).$(VersionPatch)-preview3-$(CommitCount)</Version>
```
piotrpMSFT: We should create a BuildVersion property which defaults to CommitCount. This will give us flexibility to redefine versioning scheme if ever needed.

brthor: We do have one, the issue here is that we neglected to change the project.json versions from preview3 here. 

I can go ahead and fix that

build/package/Microsoft.DotNet.Cli.Nupkg.targets
```diff
@@ -0,0 +1,69 @@
...
+        <ProjectName>Microsoft.DotNet.Files</ProjectName>
+        <Version>$(VersionMajor).$(VersionMinor).$(VersionPatch)-preview3-$(CommitCount)</Version>
+      </ProjectsToPack>
+      <ProjectsToPack Include="$(ProjectsSrcDirectory)/Microsoft.DotNet.InternalAbstractions">
+        <ProjectName>Microsoft.DotNet.InternalAbstractions</ProjectName>
+        <Version>1.0.1-beta-$(CommitCount)</Version>
+      </ProjectsToPack>
+      <ProjectsToPack Include="$(ProjectsSrcDirectory)/Microsoft.DotNet.ProjectModel">
+        <ProjectName>Microsoft.DotNet.ProjectModel</ProjectName>
+        <Version>$(VersionMajor).$(VersionMinor).$(VersionPatch)-rc4-$(CommitCount)</Version>
```
piotrpMSFT: Can we provide logical names for these release variables? `preview3` seems to represent `ToolsVersionReleaseSuffix` while `rc4` represents `RuntimeVersionReleaseSuffix`? This would better capture the intent of the differences.

build_projects/dotnet-cli-build/TestTargets.cs
```diff
@@ -30,7 +30,7 @@ public class TestTargets : Task
             // "dotnet-compile-fsc.Tests",
             "dotnet-new.Tests",
             "dotnet-pack.Tests",
-            "dotnet-projectmodel-server.Tests",
+            //"dotnet-projectmodel-server.Tests",
```
piotrpMSFT: Whu?

eerhardt: I can't wait until we:

1. remove the projectmodel-server from the product, and that allows
2. we delete these tests for good.

test/Installer/testmsi.ps1
```diff
@@ -2,7 +2,8 @@
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 param(
-    [Parameter(Mandatory=$true)][string]$inputMsi
+    [string]$InputMsi,
```
eerhardt: Isn't this still mandatory?

eerhardt:
>  TestAssets/TestProjects/MSBuildTestApp.zip

I assume this really goes away at some point...

packaging/windows/clisdk/generatebundle.ps1
```diff
@@ -113,6 +113,4 @@ if(!(Test-Path $DotnetBundleOutput))
 
 Write-Host -ForegroundColor Green "Successfully created dotnet bundle - $DotnetBundleOutput"
 
-_ $RepoRoot\test\Installer\testmsi.ps1 @("$CLISDKMSIFile")
```
eerhardt: We still want this, right?

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
eerhardt: Why can't we delete Utils.GenerateGuidFromName, and move over the 2 callers of it to use GenerateGuidFromName.GenerateGuid()?

---
In reply to: [68872777](https://github.com/dotnet/cli/pull/3696#discussion_r68872777) [](ancestors = 68872777)

build_projects/dotnet-cli-build/DebTargets.cs
```diff
@@ -7,15 +7,25 @@
...
+            BuildTargetContext c = new BuildTargetContext(context, null, null);
+
+            return GenerateDebs(c).Success;
+        }
+
         public static BuildTargetResult GenerateDebs(BuildTargetContext c)
         {
             if (CurrentPlatform.IsPlatform(BuildPlatform.Ubuntu))
             {
+                PrepareTargets.Init(c);
```
eerhardt: What from Init is necessary?  Can we change it over to use the MSBuild Init properties instead, now that those are done?

eerhardt: i.e. We pass the input properties into this Task.

---
In reply to: [68874880](https://github.com/dotnet/cli/pull/3696#discussion_r68874880) [](ancestors = 68874880)

build_projects/dotnet-cli-build/PkgTargets.cs
```diff
@@ -45,10 +46,19 @@ public static BuildTargetResult InitPkg(BuildTargetContext c)
...
+            BuildTargetContext c = new BuildTargetContext(context, null, null);
+
+            return GeneratePkgs(c).Success;
+        }
+
         public static BuildTargetResult GeneratePkgs(BuildTargetContext c)
         {
             if (CurrentPlatform.IsPlatform(BuildPlatform.OSX))
             {
+                PrepareTargets.Init(c);
```
eerhardt: What from Init is necessary?  Can we change it over to use the MSBuild Init properties instead, now that those are done?

eerhardt:
> build_projects/DownloadFile.cs

This file is empty, should be reverted.

build.proj
eerhardt: Can we get a URL to a GH issue here?

---
In reply to: [68872852](https://github.com/dotnet/cli/pull/3696#discussion_r68872852) [](ancestors = 68872852)

build/package/Microsoft.DotNet.Cli.Installer.targets
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    
+    <Target Name="GenerateInstallers" 
+          DependsOnTargets="Init;Layout;GeneratePkgs;GenerateDebs;GenerateMsis" />
```
eerhardt: Does this really need its own file?

build/package/Microsoft.DotNet.Cli.Installer.PKG.targets
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    
+    <UsingTask TaskName="PkgTargets" AssemblyFile="$(CLIBuildDll)" />
+
+    <Target Name="GeneratePkgs" >
```
eerhardt: Does this need its own file?

build/package/Microsoft.DotNet.Cli.Layout.targets
```diff
@@ -0,0 +1,69 @@
...
+      <CombinedHostHostFxrFrameworkSdkOutputDirectory>$(LayoutDirectory)/$(ArtifactNameCombinedHostHostFxrFrameworkSdk)</CombinedHostHostFxrFrameworkSdkOutputDirectory>
+      <CombinedFrameworkSdkOutputDirectory>$(LayoutDirectory)/$(ArtifactNameCombinedFrameworkSdk)</CombinedFrameworkSdkOutputDirectory>
+    </PropertyGroup>
+
+    <!-- Set up Items Defining Layouts for easy change -->
+    <ItemGroup>
+      <LayoutDefinition Include="SdkName">
+        <InputFiles>@(SdkLayoutInput)</InputFiles>
+        <OutputFiles>@(SdkLayoutInput -> '$(SdkLayoutOutputDirectory)/$([MSBuild]::MakeRelative($(OutputDirectory), '%(Identity)'))')</OutputFiles>
+        <NameWithVersion>$(ArtifactNameWithVersionSdk)</NameWithVersion>
```
eerhardt: Who uses NameWithVersion?

build/package/Microsoft.DotNet.Cli.Installer.targets
brthor: As a unit of abstraction I would argue yes. 

What else did you have in mind

build/package/Microsoft.DotNet.Cli.Nupkg.targets
brthor: Notice we also have "beta" which represents neither. I'm not sure what that one falls into.

build_projects/dotnet-cli-build/DebTargets.cs
brthor: The point of leaving this mostly as is was to prevent the need to do that in this turn of the crank. In the next turn of the crank these won't be tasks at all. 

If you're saying that should happen before this goes in that is fine with me, just will take a little bit of time.

build_projects/dotnet-cli-build/PkgTargets.cs
brthor: Same reply as to deb targets.

packaging/windows/clisdk/generatebundle.ps1
brthor: It's still happening, check msi targeets.

I don't know why this call was here. It's a hack

test/Installer/testmsi.ps1
brthor: I switched it to follow the style in build.ps1 or dotnet-install.ps1. If you think it's important I'll add it back.

build_projects/dotnet-cli-build/TestTargets.cs
brthor: because these tests fail all the time for no reason and don't do anything but make our check in times longer. 

I can file an issue for re-enabling them if that sounds good

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
brthor: Yeah I can do that.

build/package/Microsoft.DotNet.Cli.Installer.PKG.targets
brthor: Yeah this is going to get way bigger. May as well put the patterns in place.

brthor:
> I assume this really goes away at some point... @eerhardt 

It's an easter egg :smile:

build/package/Microsoft.DotNet.Cli.Installer.PKG.targets
eerhardt: Ok, if this is going to have more stuff in it, then I'm fine with the separate file.

build_projects/dotnet-cli-build/DebTargets.cs
eerhardt: If the next step is to turn this whole custom task into an MSBuild target, then let's keep it as is.  I wasn't sure if this was the "final vision" of how this was going to look.

build/package/Microsoft.DotNet.Cli.Installer.targets
eerhardt: It just seems odd to have a file with a single method in it and no vision for adding any more methods to it.  I agree splitting things up into cohesive files makes sense, but there is a balance between everything being in its own file, and everything in one big file.

One big file == hard to read.  

But everything in its own file is also hard to read because you are constantly switching between many files trying to get the "big picture" in your head.

brthor: Alright I can see that, and yeah there is nothing else coming into this file, at least in the foreseaable future. I can place this in package

build.proj
brthor: Let me file a GH issue. Has previously been reported through contact with the team.

packaging/windows/clisdk/generatebundle.ps1
eerhardt: Ah yes, I missed it in the .targets file.  Thanks for the help pointing it out.

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
brthor: fixed.

build_projects/dotnet-cli-build/DebTargets.cs
brthor: :+1:

brthor:
> > build_projects/DownloadFile.cs
> This file is empty, should be reverted.

fixed.

NuGet.Config
```diff
@@ -5,6 +5,7 @@
     <clear />
     <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <add key="cli-deps" value="https://dotnet.myget.org/F/cli-deps/api/v3/index.json" />
+    <add key="buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
```
brthor: We need to take this dependency for now @eerhardt Pushing the new msbuild package to cli-deps every update is too tedious. The right fix here is to get the msbuild packages on their own feed.

brthor:
Want to get this checked in, so I can work on the next crank here. Take another pass please

build_projects/dotnet-cli-build/TestTargets.cs
eerhardt: Let's get an issue tracking this and put in a comment.  Let's make that the "default" behavior when we introduce a hack.

---
In reply to: [68990220](https://github.com/dotnet/cli/pull/3696#discussion_r68990220) [](ancestors = 68990220)

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        [Required]
+        public string Name { get; set; }
+
+        [Output]
+        public string OutputGuid { get; set; }
+
+        public override bool Execute()
+        {
+            OutputGuid = GenerateGuid(Name).ToString();
```
eerhardt: Our existing code does "ToUpper()" on this string, do we want it in the MSBuild world too?

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -46,6 +47,10 @@
     <PropertyGroup>
       <VersionBadge>$(BaseOutputDirectory)/$(VersionBadgeMoniker)_$(Configuration)_version_badge.svg</VersionBadge>
       <SdkVersion>$(NugetVersion)</SdkVersion>
+      
+      <SdkProjectJsonVersion>$(VersionMajor).$(VersionMinor).$(VersionPatch)-preview3-$(CommitCount)</SdkProjectJsonVersion>
```
eerhardt: I don't think "ProjectJsonVersion" is appropriate.  This should be "NuGetVersion" or something that is more long-term than "ProjectJson", which should be going away soon.

build/package/Microsoft.DotNet.Cli.Nupkg.targets
```diff
@@ -0,0 +1,69 @@
...
+      <ProjectsToPack Include="$(ProjectsSrcDirectory)/Microsoft.Extensions.DependencyModel">
+        <ProjectName>Microsoft.Extensions.DependencyModel</ProjectName>
+        <Version>$(DependencyModelAndInternalAbstractionsProjectJsonVersion)</Version>
+      </ProjectsToPack>
+      <ProjectsToPack Include="$(ProjectsSrcDirectory)/Microsoft.Extensions.Testing.Abstractions">
+        <ProjectName>Microsoft.Extensions.Testing.Abstractions</ProjectName>
+        <Version>$(SdkProjectJsonVersion)</Version>
+      </ProjectsToPack>
+
+      <ProjectPackTargetInputs Include="$(PackagingBuildBasePath)/src/%(ProjectsToPack.ProjectName)/**/netstandard*/%(ProjectName).dll" />
```
eerhardt: we also build for net451, not just netstandard.

eerhardt:
:shipit: once my latest comments are addressed.

build/package/Microsoft.DotNet.Cli.Nupkg.targets
brthor: This is true, but both dlls will be regenerated on a rebuild, and netstandard1.6 is the common denominator. Using this dll as an indicator of output for the whole package build is suitable for establishing 1:1 mapping.

build/Microsoft.DotNet.Cli.Prepare.targets
brthor: fair enough.

build.proj
brthor: done.

NuGet.Config
brthor: nevermind I'm downgrading this. It's breaking things.

build_projects/dotnet-cli-build/GenerateGuidFromName.cs
brthor: good catch I missed that.

PR #3706, opened 2016-06-27
Base Prepare Infrastructure

brthor:
Extracting this from #3696 so everyone can take advantage of the base work I needed to do for package targets.

Included is:
 - A task to get the current rid
 - A task to get the current arch
 - A task to generate an itemgroup with the build version info in metadata
 - base properties in prepare targets needed in other places

@eerhardt @sokket @livarcocc @piotrpMSFT

### DIFF
--- a/build/Microsoft.DotNet.Cli.Prepare.targets
+++ b/build/Microsoft.DotNet.Cli.Prepare.targets
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Layout" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  
+  <UsingTask TaskName="Rid" AssemblyFile="$(CLIBuildDll)" />
+  <UsingTask TaskName="Architecture" AssemblyFile="$(CLIBuildDll)" />
+  <UsingTask TaskName="GenerateBuildVersionInfo" AssemblyFile="$(CLIBuildDll)" />
+
+  <Target Name="Init" >
+    <!-- Current Runtime Information -->
+    <Rid>
+      <Output TaskParameter="OutputRid" PropertyName="Rid" />
+    </Rid>
+    <Architecture>
+      <Output TaskParameter="OutputArchitecture" PropertyName="Architecture" />
+    </Architecture>
+
+    <!-- Common Properties -->
+    <PropertyGroup>
+      <BaseOutputDirectory>$(RepoRoot)/artifacts/$(Rid)</BaseOutputDirectory>
+      <OutputDirectory>$(BaseOutputDirectory)/stage2</OutputDirectory>
+      <Stage2CompilationDirectory>$(BaseOutputDirectory)/stage2compilation</Stage2CompilationDirectory>
+      <IntermediateDirectory>$(BaseOutputDirectory)/intermediate</IntermediateDirectory>
+    </PropertyGroup>
+
+    <!-- Generates BuildVersion Item
+        BuildVersion Metadata:
+          - Major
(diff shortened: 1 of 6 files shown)


sokket:
LGTM! Thanks for putting this out there

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -0,0 +1,55 @@
...
+    <Rid>
+      <Output TaskParameter="OutputRid" PropertyName="Rid" />
+    </Rid>
+    <Architecture>
+      <Output TaskParameter="OutputArchitecture" PropertyName="Architecture" />
+    </Architecture>
+
+    <!-- Common Properties -->
+    <PropertyGroup>
+      <RepoRoot>$(MSBuildThisFileDirectory)/..</RepoRoot>
```
eerhardt: I think this property should go in build.proj, or even better a dirs.props file in the root.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -0,0 +1,55 @@
...
+      <BaseOutputDirectory>$(MSBuildThisFileDirectory)/../artifacts/$(Rid)</BaseOutputDirectory>
+      <OutputDirectory>$(BaseOutputDirectory)/stage2</OutputDirectory>
+      <Stage2CompilationDirectory>$(BaseOutputDirectory)/stage2compilation</Stage2CompilationDirectory>
+      <IntermediateDirectory>$(BaseOutputDirectory)/intermediate</IntermediateDirectory>
+      <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
```
eerhardt: Same here, this should go in build.proj or dirs.props file.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -0,0 +1,55 @@
...
+                ItemName="BuildVersion" />
+    </GenerateBuildVersionInfo>
+
+    <PropertyGroup>
+      <SdkVersion>@(BuildVersion ->'%(NugetVersion)')</SdkVersion>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <Monikers Include="Monikers">
+        <SdkBrandName>Microsoft .NET Core 1.0.0 - SDK Preview 2</SdkBrandName>
```
eerhardt: Similar to my comment here:  https://github.com/dotnet/cli/pull/3696/commits/5ce1fd6f4de0130f4ac62a3daa78838ffdc2aa7d#r68623008, it seems like this should be a property, and not inside an Item.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -0,0 +1,55 @@
...
+          - Patch
+          - ReleaseSuffix
+          - CommitHash
+          - CommitCount
+          - VersionSuffix
+          - SimpleVersion
+          - NugetVersion
+          - MsiVersion            -->
+    <GenerateBuildVersionInfo RepoRoot="$(RepoRoot)">
+        <Output TaskParameter="OutputBuildVersionInfo"
```
eerhardt: Same comment as https://github.com/dotnet/cli/pull/3696/commits/5ce1fd6f4de0130f4ac62a3daa78838ffdc2aa7d#r68623008.  It seems like we are using an Item here just for grouping multiple properties.  Would it be better if this Task generated a bunch of properties instead of an Item?

brthor: Talked offline, we'll leave buildversion as is for now

eerhardt:
:shipit:

PR #3025, opened 2016-05-13
Move all return statements to exit 0 in install.ps1

brthor:
fixes #3024 
cc @krwq @blackdwarf

PR #3708, opened 2016-06-27
Remove final 'Exit 0' 

Nick-Lucas:
skipciplease

Removed final 'Exit 0' as this is causing entire build processes to exit which reference the script as part of the overall build script. Exit 0 was previously added without any discussion on whether it was necessary, and it doesn't appear to be.

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -355,4 +355,3 @@ else {
 }
 
 Say "Installation finished"
-exit 0
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -329,7 +329,7 @@ $IsSdkInstalled = Is-Dotnet-Package-Installed -InstallRoot $InstallRoot -Relativ
 Say-Verbose ".NET SDK installed? $IsSdkInstalled"
 if ($IsSdkInstalled) {
     Say ".NET SDK version $SpecificVersion is already installed."
-    exit 0
+    return
 }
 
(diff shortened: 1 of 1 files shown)


brthor:
There was some internal discussion that led to this PR: https://github.com/dotnet/cli/pull/3025

I can't recall the exact issue now but I believe I found a set of conditions where "return" or the lack of "exit 0" did not return as a successful process despite the lack of an error. 

Why is exit 0 causing your build process to exit?

Nick-Lucas:
Hey thanks for the quick reply.

I may be using this script in a way not originally intended, however I've recently inherited it so am still learning this.

We execute through powershell on AppVeyor "iex" like
`iex ((New-Object Net.WebClient).DownloadString('https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1'))`

As soon as an 'Exit' returns, the entire build process exits as shown here:
https://ci.appveyor.com/project/Nick-Lucas/limebean/build/57
As you can see in this example I also tried wrapping in Try/Catch which didn't resolve the issue.

I am definitely not a powershell expert so it's possible there's a way around this, but right now I'm referencing my PR commit in the build script (where I've removed Exits from successful operations) which has resolve this particular issue.

Nick-Lucas:
We've managed to resolve this by using the following instead of `iex`
`  - appveyor DownloadFile https://raw.githubusercontent.com/dotnet/cli/v1.0.0-preview2/scripts/obtain/dotnet-install.ps1
  - powershell -File dotnet-install.ps1`

brthor:
I am also not a powershell or appveyor expert :smile: 

But I'm glad to hear you resolved the issue.

Issue #2960, opened 2016-05-09
Allow configuration of cli settings via a JSON file


ellismg:
Today, we have a collection of environment variables and XML files that provide configuration to the toolchain.  For example, we have an environment variable to to control if telemetry is reported.  We have another that you can use to change where the machine wide nuget configuration is read from (so folks can add additional sources).

It would be nice if this configuration could be centralized in some sort of config.json file SxS with the SDK files, to make it easier for folks to control instead of having to add tons of environment variables to their command lines or environment.

piotrpMSFT:
@ellismg do we have an explicit list of the configuration you care about? If this is going to make `preview2` it needs to come in fairly soon. You should also note that the env vars you mention, NuGet reads some and CLI reads others...

ellismg:
Can we start with the environment variables that CLI reads (I know the telemetry one, I was unaware there were others).

Ignoring the NuGet ones for preview fine is reasonable, we can figure out the best way to centralize them later.

blackdwarf:
Generally agree that this would be a good overall usability addition, so moving to RTM for further scoping and proper evaluation.

PR #3710, opened 2016-06-28
Convert 'Init' to MSBuild.

eerhardt:
Couple of notes:

1. I removed the "cmake" prereq since CLI doesn't require cmake anymore.
2. I left "CheckPrereq" as a custom task since there is so much C# code, and I don't see any benefit to moving that C# code into MSBuild.
3. I couldn't find a built-in "Set Environment Variable" task. I found one for VS 2015 C++ targets, but didn't want to depend on it.
4. I'm really looking for advice on how to do the `ExpectedBuildArtifacts` stuff in MSBuild.  It feels like a need a shared method, but I'm not sure how to do that. It is a lot of duplication.

@brthor @livarcocc @sokket @piotrpMSFT 

/cc @weshaggard in case you have some good ideas on how to generate those "CompressFile" and "InstallerFile" properties.


build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -1,47 +1,98 @@
...
+
+      <BundleExtension Condition=" '$(OSName)' == 'win' ">.exe</BundleExtension>
+      <BundleExtension Condition=" '$(OSName)' == 'osx' ">$(InstallerExtension)</BundleExtension>
+      <BundleExtension Condition=" '$(OSName)' == 'ubuntu' ">$(InstallerExtension)</BundleExtension>
+
+      <ProductMonikerRid Condition=" '$(Rid)' == 'ubuntu.16.04-x64' Or '$(Rid)' == 'fedora.23-x64' Or '$(Rid)' == 'opensuse.13.2-x64' ">$(Rid)</ProductMonikerRid>
+      <ProductMonikerRid Condition=" '$(ProductMonikerRid)' == '' ">$(OSName)-$(Architecture)</ProductMonikerRid>
+
+      <!-- Generated Installers + Archives -->
+      <SdkCompressedFile>$(PackagesDirectory)/dotnet-sdk-$(ProductMonikerRid).$(SdkVersion)$(ArchiveExtension)</SdkCompressedFile>
```
eerhardt: @weshaggard @brthor @sokket - looking for a better way to do this.  See https://github.com/dotnet/cli/blob/rel/1.0.0/build_projects/dotnet-cli-build/PrepareTargets.cs#L147 for the C# code I'm trying to port.

build_projects/dotnet-cli-build/SetEnvVar.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Microsoft.DotNet.Cli.Build
+{
+    public class SetEnvVar : Task
```
weshaggard: :(
You are correct that there isn't an MSBuild task that sets environment variables but I would argue that we shouldn't need one as depending on environment variables is always a bad idea. I think we can live with this temporarily but we should really try and avoid using environment variables.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -1,47 +1,98 @@
...
+      <Output TaskParameter="VersionPatch" PropertyName="VersionPatch" />
+      <Output TaskParameter="CommitHash" PropertyName="CommitHash" />
+      <Output TaskParameter="CommitCount" PropertyName="CommitCount" />
+      <Output TaskParameter="ReleaseSuffix" PropertyName="ReleaseSuffix" />
+      <Output TaskParameter="VersionSuffix" PropertyName="VersionSuffix" />
+      <Output TaskParameter="SimpleVersion" PropertyName="SimpleVersion" />
+      <Output TaskParameter="NugetVersion" PropertyName="NugetVersion" />
+      <Output TaskParameter="MsiVersion" PropertyName="MsiVersion" />
+      <Output TaskParameter="VersionBadgeMoniker" PropertyName="VersionBadgeMoniker" />
+      <Output TaskParameter="Channel" PropertyName="Channel" />
```
weshaggard: I would eventually move all these properties directly into a props file instead of needing to read them from the test file and pull them in this way. I understand it probably makes sense to keep them this way while the other build system is still in play and also reads the text file.

build/Microsoft.DotNet.Cli.Prepare.targets
weshaggard: While I didn't double check all the values here this doesn't look too unreasonable. Yes it might be a little more verbose but I think this is likely the best translation you could expect. It maybe possible to reduce a little more of the redundancy but I didn't dig too deeply.

brthor: @eerhardt Let's sync on this.

I made changes to include these elsewhere in the packaging targets and even so that the `CompressedFile` property should no longer be required. 

https://github.com/dotnet/cli/pull/3696/files#diff-3c15249555fc7f61346cba0dd00a99ccR14
https://github.com/dotnet/cli/pull/3696/files#diff-26682095fd81c9156a84091c95350ad9R34

There's no real reason we need these `CompressedFile` or `InstallerFile` properties present in init as they don't have purpose without having run `Package`.

build/Microsoft.DotNet.Cli.Prepare.targets
brthor: I agree with @weshaggard here for the long term.

For the short term why change all of this from itemgroup metadata to so many properties...

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -1,47 +1,98 @@
...
+      <Output TaskParameter="BranchName" PropertyName="BranchName" />
     </GenerateBuildVersionInfo>
 
     <PropertyGroup>
-      <SdkVersion>@(BuildVersion ->'%(NugetVersion)')</SdkVersion>
+      <VersionBadge>$(BaseOutputDirectory)/$(VersionBadgeMoniker)_$(Configuration)_version_badge.svg</VersionBadge>
+      <SdkVersion>$(NugetVersion)</SdkVersion>
+
+      <ArchiveExtension Condition=" '$(OSName)' == 'win' ">.zip</ArchiveExtension>
```
brthor: We don't need these properties

brthor:
:shipit: I will rebase on this

build_projects/dotnet-cli-build/SetEnvVar.cs
eerhardt: We need this because this is the only way to set "Telemetry" options (i.e. the telemetry profile).  We have an issue to be able to set these things using a config file:  https://github.com/dotnet/cli/issues/2960.

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: @brthor and I sync'd offline.  Some of these properties are used throughout multiple targets (Prepare, Package, Publish).  We decided that for now, straight conversion is acceptable, and we will refactor out the properties that aren't needed as we convert more of the build.

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: I also agree on the "move these properties out of the .txt file and into MSBuild", but we can't do that yet.

> For the short term why change all of this from itemgroup metadata to so many properties...

@brthor and I sync'd on this offline.  The main advantage is that the MSBuild code here is self-descriptive of which properties are available (you don't need a comment stating all the custom metadata available on the Item).  Also, since these are all single-valued values, it feels more correct to make these properties, than use an Item.

However, it would be nice if there was some sort of "grouping" mechanism in MSBuild properties (think namespaces), so you could say `BuildVersion_Major`, `BuildVersion_Minor`, etc.  The only grouping mechanism seems to be the underscore, but it doesn't appear to be a standard.  If we find a grouping mechanism in the future, we can change this.

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: I needed these properties for use in the properties below.

build/Microsoft.DotNet.Cli.Prepare.targets
weshaggard: I agree these shouldn't be an ItemGroup. As for grouping mechanism the best we have is naming convention as they are all global properties :(

PR #3721, opened 2016-06-28
Implement RestorePackages in MSBuild.

eerhardt:
This adds the DotNetTool custom task, with `dotnet restore` implemented.

@piotrpMSFT @brthor @sokket @livarcocc

### DIFF
--- a/build/Microsoft.DotNet.Cli.Prepare.targets
+++ b/build/Microsoft.DotNet.Cli.Prepare.targets
@@ -22,7 +22,8 @@
     <!-- Common Properties -->
     <PropertyGroup>
       <Stage0Path Condition=" '$(OSName)' == 'win' ">$(RepoRoot)/.dotnet_stage0/Windows/$(Architecture)</Stage0Path>
-      <Stage0Path Condition=" '$(OSName)' != 'win' ">$(RepoRoot)/.dotnet_stage0/$(OSName)</Stage0Path>
+      <Stage0Path Condition=" '$(OSName)' == 'osx' ">$(RepoRoot)/.dotnet_stage0/Darwin</Stage0Path>
+      <Stage0Path Condition=" '$(Stage0Path)' == '' ">$(RepoRoot)/.dotnet_stage0/Linux</Stage0Path>
       <DotNetPath>$(Stage0Path)</DotNetPath>
 
       <BaseOutputDirectory>$(RepoRoot)/artifacts/$(Rid)</BaseOutputDirectory>
--- a/build/Microsoft.DotNet.Cli.Prepare.targets
+++ b/build/Microsoft.DotNet.Cli.Prepare.targets
@@ -5,12 +5,13 @@
   <UsingTask TaskName="GenerateBuildVersionInfo" AssemblyFile="$(CLIBuildDll)" />
   <UsingTask TaskName="CheckPrereqs" AssemblyFile="$(CLIBuildDll)" />
   <UsingTask TaskName="SetEnvVar" AssemblyFile="$(CLIBuildDll)" />
+  <UsingTask TaskName="DotNetRestore" AssemblyFile="$(CLIBuildDll)" />
 
   <Target DependsOnTargets="BuildDotnetCliBuildFramework" Name="Prepare">
     <PrepareTargets />
   </Target>
 
-  <Target Name="Init" >
+  <Target Name="Init">
     <!-- Current Runtime Information -->
     <GetCurrentRuntimeInformation>
       <Output TaskParameter="Rid" PropertyName="Rid" />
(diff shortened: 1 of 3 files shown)


build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -95,4 +100,18 @@
     <!-- SetTelemetryProfile -->
     <SetEnvVar Name="DOTNET_CLI_TELEMETRY_PROFILE" Value="https://github.com/dotnet/cli;$(CommitHash)" />
   </Target>
+
+  <Target DependsOnTargets="Init" Name="RestorePackages">
+
+    <!-- Clean NuGet Temp Cache on Linux (seeing some issues on Linux) -->
+    <!-- TODO: do we still need this?  Seems unnecessary -->
+    <RemoveDir
```
eerhardt: I left this in here to get PR feedback.  I really think we should remove this.

piotrpMSFT: Let's kill it and bring back if we see these 'issues'. When it comes back, it does so with an issue URL.

build_projects/dotnet-cli-build/DotNetTool.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            base.LogToolCommand($"{GetWorkingDirectory()}> {message}");
+        }
+    }
+
+    public class DotNetRestore : DotNetTool
+    {
+        protected override string Command
+        {
+            get { return "restore"; }
```
piotrpMSFT: Does this disable the setter?

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -95,4 +100,18 @@
...
+      Condition=" '$(OSName)' != 'win' AND '$(OSName)' != 'osx' "
+      Directories="/tmp/NuGet" />
+
+    <DotNetRestore WorkingDirectory="$(RepoRoot)/src" ToolPath="$(DotNetPath)" />
+    <DotNetRestore WorkingDirectory="$(RepoRoot)/tools" ToolPath="$(DotNetPath)" />
+
```
piotrpMSFT: <Exec Command="$(DotNetPath) restore" WorkingDirectory="$(RepoRoot)/tools" />

Does this really gain us much? I think the alternative is strictly less characters... and it immediately shows us how we're invoking the CLI. One less place to look for potential bugs. What's the win?

build/Microsoft.DotNet.Cli.Prepare.targets
brthor: Man who knows when this hack went in, probably february, let's kill it

eerhardt: :+1:

build_projects/dotnet-cli-build/DotNetTool.cs
eerhardt: ?? there is no 'setter'.  Command is an abstract readonly property on the DotNetTool class.

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: The 'win' is when you start having args, like Verbosity.

eerhardt: Also, you can control your logging, like I did:

```C#
        protected override void LogToolCommand(string message)
        {
            base.LogToolCommand($"{GetWorkingDirectory()}> {message}");
        }
```

This allows us to see what directory we are restoring in the logs.

eerhardt:
I've responded to feedback, and fixed a simple bug @brthor pointed out.

build/Microsoft.DotNet.Cli.Prepare.targets
piotrpMSFT: Ok, lgtm. I want to be sure we don't start incorporating features just because they exist. Offline conversation showed some nice immediate logging wins. I do want to be sure that the Params we expose on these tasks match their cmd line equivalents so translation is easy.

piotrpMSFT:
🚢 🇮🇹

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: Agreed. There shouldn't be unnecessary translation between these params and the underlying dotnet params

eerhardt:

@troydai - DTH test failure:  http://dotnet-ci.cloudapp.net/job/dotnet_cli/job/feature_msbuild/job/release_windows_nt_x64_prtest/33/

```
MESSAGE:
System.TimeoutException : Response time out after 20 seconds.
+++++++++++++++++++
STACK TRACE:
at Microsoft.DotNet.ProjectModel.Server.Tests.DthTestClient.GetResponse(TimeSpan timeout) in D:\j\workspace\release_windo---5a97ca7c\test\dotnet-projectmodel-server.Tests\DthTestClient.cs:line 290 at Microsoft.DotNet.ProjectModel.Server.Tests.DthTests.DthNegative_BrokenProjectPathInLockFile() in D:\j\workspace\release_windo---5a97ca7c\test\dotnet-projectmodel-server.Tests\DthTests.cs:line 244
```

eerhardt:

Issue #3765, opened 2016-07-01
Add analyzer support for local project


nothrow:
https://github.com/dotnet/cli/issues/83 added support for referencing 3rd party code analyzers. 
I am developing solution, where I would like to add my own, solution-specific code analysis. It is not currently possible to reference `project.json`-project and use it as code analyzer.

piotrpMSFT:
In #3175 I showed how analyzers surface in the preview3 msbuild CLI. As you can see, you'll be able to add <Include>s in your project file and pull in other project outputs. Just mimic the pattern I've shown from the nuget.g.props file.

Issue #83, opened 2015-10-21
Add analyzer support


davidfowl:
We need to put some code in the ProjectModel that surfaces analyzers from the lock file. This also crosses over to development dependencies but that can be solved separately. We can likely port that code from the existing NuGet msbuild task

/cc @agocke

NikolayIT:
@davidfowl Any information when analyzers will be supported?

boxofnotgoodery:
Would really love to see this happen,
Code analysis and unit tests are both two tools I leverage to ensure my code is production ready, or at least qa ready :)

blackdwarf:
@jaredpar @agocke which milestone can we expect something like this to happen in?

giggio:
Yes, RC2! \o/

PR #3766, opened 2016-07-01
Add analyzer support for local project 

nothrow:
https://github.com/dotnet/cli/issues/3765

#83 added support for referencing 3rd party code analyzers. 
I am developing solution, where I would like to add my own, solution-specific code analysis. It is not currently possible to reference project.json-project and use it as code analyzer.

- Added 'type' field to project.json
- When 'type' is 'analyzer', referenced project is used for code analysis during build

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/Compilation/LibraryExporter.cs
+++ b/src/Microsoft.DotNet.ProjectModel/Compilation/LibraryExporter.cs
@@ -280,6 +280,7 @@ namespace Microsoft.DotNet.ProjectModel.Compilation
         {
             var builder = LibraryExportBuilder.Create(project);
             var compilerOptions = project.Project.GetCompilerOptions(project.TargetFrameworkInfo.FrameworkName, _configuration);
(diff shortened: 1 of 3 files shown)


eerhardt:
@blackdwarf @piotrpMSFT - what's the procedure for adding new schema elements to project.json?

piotrpMSFT:
Per [this](https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/) we are currently converging project.json with csproj/msbuild. We aren't able to modify the PJ schema until the conversion is complete as additions would create a moving target and slow down the long-term approach. 

@nothrow CLI deals with similar issues when building extensions. These must also be NuGet packages before they can be referenced. Can you take a similar approach for the time being? Specifically, restore & build the analyzers, then `restore --fallback` your primary solution?

I'm going to close this for the time being because we really cannot modify the PJ schema. Try out the suggestion above and if it's problematic let's chat on the associated issue you created!

PR #3812, opened 2016-07-08
Handle missing terminal capabilities.

josteink:
This patch tries to increase portability for the `dotnet-install.sh` script, especially with FreeBSD in mind.

When probing for terminal capabilities, `tput` will return exit-code `1` when the attribute is not valid for the current terminal. If we don't handle that exit-code, the shell-script too will exit with silent failure.

This patch simply ignores errors for attributes we cannot use, resulting in the same result as if we never probed for it in the first place.

### DIFF
--- a/scripts/obtain/dotnet-install.sh
+++ b/scripts/obtain/dotnet-install.sh
@@ -27,16 +27,16 @@ if [ -t 1 ]; then
     # see if it supports colors
     ncolors=$(tput colors)
     if [ -n "$ncolors" ] && [ $ncolors -ge 8 ]; then
-        bold="$(tput bold)"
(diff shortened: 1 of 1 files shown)


josteink:
@dotnet-bot please retest this

josteink:
@dotnet-bot test this please

josteink:
@dotnet-bot test this please

josteink:
Anyone feeling like taking a look at this one yet?

Cc: @eerhardt @livarcocc

eerhardt:
This looks good to me.

@krwq - any thoughts?

krwq:
lgtm :+1: 
@josteink - thanks for the fix!

PR #3813, opened 2016-07-08
 Add FreeBSD OS-detection for dotnet-installer.

josteink:
This patch tries to increase portability for the `dotnet-install.sh` script, especially with FreeBSD in mind.

While detecting FreeBSD for now will not result in any valid SDK or download, it is a baby-step in the general direction of getting FreeBSD build-suppot for the dotnet-cli.

### DIFF
--- a/scripts/obtain/dotnet-install.sh
+++ b/scripts/obtain/dotnet-install.sh
@@ -92,6 +92,9 @@ get_current_os_name() {
                 echo "opensuse.13.2"
                 return 0
             fi
+	elif [ "$(uname -s | grep -cim1 freebsd)" -eq 1 ]; then
+            echo "freebsd"
+            return 0
         fi
     fi
     
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -163,7 +163,6 @@ namespace Microsoft.DotNet.Cli
                 exitCode = result.ExitCode;
             }
 
-
             return exitCode;
 
         }


josteink:
Not sure why kloun's commit is showing up in this PR, but from what I can tell, it shouldn't really make any difference.

josteink:
Anyone feeling like taking a look at this one yet?

Cc: @eerhardt

eerhardt:
I'm not certain this is going to help at all, since we don't publish "freebsd" archives to our Azure blob.  The `get_current_os_name()` method will return "freebsd" now, but the script uses that result to build up a URL that is not going to resolve to an actual Azure blob.  Thus it will just fail later in the script with a "blob not found" error.

It seems better to keep the current behavior of failing with "OS name could not be detected" or even better "Current OS is not supported".

/cc @krwq

josteink:
I'm fully aware of that, but my intent was making FreeBSD a supported OS and I figured you had to start somewhere.

Since I've submitted this PR I've received some leads on `stage0` and how to build that manually.

Knowing what needs to be done there, it might make more sense starting in that order, before eventually updating `dotnet-install.sh`.

I'll see what i can get done with `stage0` and then get back to this PR later.

piotrpMSFT:
Sounds great! in the meantime, we'll think a bit about how to allow additional sources for the script so folks have a unified experience. Please reopen the PR when ready, @josteink

PR #3842, opened 2016-07-12
Piotrp msft/msbuild/incremental test

piotrpMSFT:
Adds incremental build of test asset packages and moves <Exec> to <DotNetTools>


build/Microsoft.DotNet.Cli.Test.targets
```diff
@@ -39,19 +39,21 @@
...
-      <TestArtifactsDir>$(TestOutputDir)/artifacts</TestArtifactsDir>
-      <TestResultXmlDir>$(TestOutputDir)/results</TestResultXmlDir>
-      <TestDirectory>$(RepoRoot)/test</TestDirectory>
-      <DotnetUnderTest>$(DotnetStage2)</DotnetUnderTest>
+      <TestOutputDir>$(BaseOutputDirectory)/tests/</TestOutputDir>
+      <TestPackagesBuildDir>$(TestOutputDir)/packagesBuild/</TestPackagesBuildDir>
+      <TestPackagesDir>$(TestOutputDir)/packages/</TestPackagesDir>
+      <TestArtifactsDir>$(TestOutputDir)/artifacts/</TestArtifactsDir>
+      <TestResultXmlDir>$(TestOutputDir)/results/</TestResultXmlDir>
+      <TestDirectory>$(RepoRoot)/test/</TestDirectory>
```
piotrpMSFT: Note, adding trailing slashes since that is the convention for directories in MSBuild.

build/Microsoft.DotNet.Cli.Test.targets
```diff
@@ -3,6 +3,9 @@
   <Import Project="test/ProjectsToTest.props" />
   <Import Project="test/TestPackageProjects.props" />
 
+  <UsingTask TaskName="DotNetBuild" AssemblyFile="$(CLIBuildDll)" />
```
eerhardt: Would it make sense to put all these UsingTask elements in build.proj?  That way they aren't duplicated in every .targets file that calls them.  (I see Prepare already has DotNetRestore.)

build/Microsoft.DotNet.Cli.Test.targets
```diff
@@ -60,15 +62,25 @@
...
   </Target>
  
   <Target Name="CleanTestProjectsBinObj" 
           DependsOnTargets="SetupRestoreTestsInputs;" 
           Inputs="@(RestoreTestsInputs)"
           Outputs="@(RestoreTestsInputs->'%(RelativeDir)project.lock.json')">
-    <Delete Files="$(TestDirectory)/**/bin/**/*.*')" />
-    <Delete Files="$(TestDirectory)/**/obj/**/*.*')" />
+    <ItemGroup>
+      <FilesToDelete Include="$(TestDirectory)**/bin/**/*.*" />
```
eerhardt: Why not just delete the folders?

build/test/TestPackageProjects.props
```diff
@@ -1,166 +1,268 @@
...
+      </TestPackageProject>
+
+      <TestPackageProject>
+        <PackInputs>%(BuildOutputs)</PackInputs>
+        <PackOutputs>$(TestPackagesDir)%(NuPkgName).%(VersionPrefix)%(VersionSuffix).nupkg</PackOutputs>
+      </TestPackageProject>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="AssociateOutputsWithProjects"
```
eerhardt: I don't think a ".props" file should contain Targets.

eerhardt:
:shipit:

build/Microsoft.DotNet.Cli.Test.targets
```diff
@@ -39,19 +39,21 @@
...
-      <TestResultXmlDir>$(TestOutputDir)/results</TestResultXmlDir>
-      <TestDirectory>$(RepoRoot)/test</TestDirectory>
-      <DotnetUnderTest>$(DotnetStage2)</DotnetUnderTest>
+      <TestOutputDir>$(BaseOutputDirectory)/tests/</TestOutputDir>
+      <TestPackagesBuildDir>$(TestOutputDir)/packagesBuild/</TestPackagesBuildDir>
+      <TestPackagesDir>$(TestOutputDir)/packages/</TestPackagesDir>
+      <TestArtifactsDir>$(TestOutputDir)/artifacts/</TestArtifactsDir>
+      <TestResultXmlDir>$(TestOutputDir)/results/</TestResultXmlDir>
+      <TestDirectory>$(RepoRoot)/test/</TestDirectory>
+      <DotnetUnderTest>$(Stage2Directory)</DotnetUnderTest>
```
brthor: What do you think about making this conditional, so it can be overridden?

For example `<DotnetUnderTest Condition=" '$(DotnetUnderTest)' == '' ">$(Stage2Directory)</DotnetUnderTest>`

Then we could easily reuse this for testing other dotnets, in perhaps the installers, or during dev time.

build/Microsoft.DotNet.Cli.Test.targets
brthor: will probably be faster to delete directories

https://msdn.microsoft.com/en-us/library/xyfz6ddb.aspx

brthor: If you keep this though, change your glob to include extension less files `$(TestDirectory)**/bin/**/*`

build/Microsoft.DotNet.Cli.Test.targets
```diff
@@ -60,15 +62,25 @@
...
-    <Delete Files="$(TestDirectory)/**/obj/**/*.*')" />
+    <ItemGroup>
+      <FilesToDelete Include="$(TestDirectory)**/bin/**/*.*" />
+      <FilesToDelete Include="$(TestDirectory)**/obj/**/*.*" />
+    </ItemGroup>  
+
+    <Delete Files="@(FilesToDelete)" />
+
+    <ItemGroup>
+      <FilesToDelete Remove="*" />
```
brthor: What does this do?

Is it a pattern for creating local items?

build/Microsoft.DotNet.Cli.Test.targets
```diff
@@ -142,7 +162,10 @@
                             SetupRestoreDesktopTestAssetProjectsInputs;"
           Inputs="@(RestoreDesktopTestAssetProjectsInputs)"
           Outputs="@(RestoreDesktopTestAssetProjectsInputs->'%(RelativeDir)/project.lock.json')">
-    <Exec Command="$(DotnetUnderTest) restore $(RepoRoot)TestAssets\DesktopTestProjects\ --fallbacksource $(TestPackagesDir)" />
+
+    <DotNetRestore  FallbackSource="$(TestPackagesDir)"
+                    ToolPath="$(DotnetUnderTest)" 
+                    WorkingDirectory="$(RepoRoot)TestAssets\DesktopTestProjects\" />
```
brthor: Forward slashes? We use the everywhere else

build/Microsoft.DotNet.Cli.Test.targets
piotrpMSFT: Finding folders requires finding files first, then deleting their parents, as far as I've seen. This is just less code and accomplishes nearly the same...

build/test/TestPackageProjects.props
piotrpMSFT: Will rename in next PR.

build/Microsoft.DotNet.Cli.Test.targets
piotrpMSFT: Will do in next PR.

build/Microsoft.DotNet.Cli.Test.targets
piotrpMSFT: Yes. FilesToDelete is a local that would ideally be discarded but <Delete> does not do wildcard expansion on its own.

build/Microsoft.DotNet.Cli.Test.targets
piotrpMSFT: These are `Desktop` projects so I chose to use Windows slashes.

build/test/TestPackageProjects.props
```diff
@@ -1,166 +1,268 @@
...
+      </BaseTestPackageProject>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="FindSourceFiles"
+          DependsOnTargets="SetupTestPackageProjectBaseData;"
+          Outputs="%(BaseTestPackageProject.Identity)">
+    <ItemGroup>
+      <__TestPackageProjectInputs Include="%(BaseTestPackageProject.BuildInputIncludeFilter)"
+                                  Exclude="%(BaseTestPackageProject.BuildInputExcludeFilter)">
```
brthor: I'm seeing `__` in some places and `_` in others, we should standardize here

brthor:
:shipit:

PR #3854, opened 2016-07-13
Incremental Test Part 2 - TestAssetProjects

piotrpMSFT:


piotrpMSFT:

piotrpMSFT:

piotrpMSFT:
removed the failing centos test machine.

piotrpMSFT:

build/test/TestAssetProjects.targets
```diff
@@ -0,0 +1,62 @@
...
+      <TestAssetProjects>
+        <BuildInputIncludeFilter>%(RelativeDir)**/*.*</BuildInputIncludeFilter>
+        <BuildInputExcludeFilter>%(RelativeDir)bin/**/*.*;%(RelativeDir)obj/**/*.*</BuildInputExcludeFilter>
+        <OutputDir>$([System.IO.Directory]::GetParent(%(Identity)))</OutputDir>
+        <ProjectPath>%(Identity)</ProjectPath>
+      </TestAssetProjects>
+
+      <TestAssetProjects>
+        <OutputName>$([System.IO.Path]::GetFileName(%(OutputDir)))</OutputName>
+        <OutputName Condition=" '%(RelativeDir)' == 'TestAssets/TestProjects/AppWithOutputAssemblyName/' ">MyApp</OutputName>
```
eerhardt: These seem weird - is it because the project.json specifies this name?

build/test/TestAssetProjects.targets
```diff
@@ -0,0 +1,62 @@
...
+        <OutputName Condition=" '%(RelativeDir)' == 'TestAssets/TestProjects/TestAppWithContentPackage/' ">AppWithContentPackage</OutputName>
+      </TestAssetProjects>
+
+      <TestAssetProjects>
+        <BuildOutput>%(TestAssetProjects.RelativeDir)bin/$(Configuration)/netcoreapp1.0/%(TestAssetProjects.OutputName).dll</BuildOutput>
```
eerhardt: Not all TestAssetProjects are 'netcoreapp1.0' are they?  Aren't some also desktop TFM?

build/test/TestAssetProjects.targets
piotrpMSFT: That's exactly right. This mechanism allows us to rename the expected output.

build/test/TestAssetProjects.targets
```diff
@@ -0,0 +1,62 @@
...
+    <ItemGroup>
+      <NoAutoBuildTestAssets Include="TestAssets$(PathSeparator)TestProjects$(PathSeparator)**$(PathSeparator).noautobuild" />
+      <NoAutoBuildTestAssetProjects Include="%(NoAutoBuildTestAssets.RelativeDir)project.json" />
+
+      <TestAssetProjects Include="TestAssets$(PathSeparator)TestProjects$(PathSeparator)**$(PathSeparator)project.json" 
+                         Exclude="@(NoAutoBuildTestAssetProjects)"/>
+      <TestAssetProjects>
+        <BuildInputIncludeFilter>%(RelativeDir)**/*.*</BuildInputIncludeFilter>
+        <BuildInputExcludeFilter>%(RelativeDir)bin/**/*.*;%(RelativeDir)obj/**/*.*</BuildInputExcludeFilter>
+        <OutputDir>$([System.IO.Directory]::GetParent(%(Identity)))</OutputDir>
```
eerhardt: I think "OutputDir" is not a great name for this variable, since it is capturing the project directory, right?

build/test/TestAssetProjects.targets
piotrpMSFT: https://github.com/dotnet/cli/blob/rel/1.0.0/build_projects/dotnet-cli-build/TestTargets.cs#L237

We moved the Desktop projects into DesktopTestProjects some time back.

build/test/TestAssetProjects.targets
piotrpMSFT: I could create a seperate itemgroup, OutputNameOverrides, and then merge them. I chose this approach because there are less MSBuild concepts in play. What are your thoughts on the alternative?

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -4,7 +4,10 @@
   <UsingTask TaskName="GenerateBuildVersionInfo" AssemblyFile="$(CLIBuildDll)" />
   <UsingTask TaskName="CheckPrereqs" AssemblyFile="$(CLIBuildDll)" />
   <UsingTask TaskName="SetEnvVar" AssemblyFile="$(CLIBuildDll)" />
+  <UsingTask TaskName="DotNetBuild" AssemblyFile="$(CLIBuildDll)" />
```
eerhardt: It feels like these belong in a more global place that Prepare.targets.  Prepare doesn't using any of these new tasks, so it seems weird for it to be declaring them.

Maybe adding a new "global.props", or similar name, would be a good place for these UsingTask elements that are used across all our .targets files.

build/test/TestAssetProjects.targets
eerhardt: IMO - this is the best approach until we actually have MSBuild working to build our projects.  I was just making sure I knew why this was doing what it was doing.  It is sad that we have to duplicate this in both the PJ and here, but I see nothing better in the short-term.

build/test/TestAssetProjects.targets
eerhardt: But some target both right?

Or are we saying it is "good enough" to just check 'netcoreapp1.0' in those situations?

eerhardt:
:shipit: after addressing the above comments.

build/test/TestAssetProjects.targets
piotrpMSFT: Fixed.

build/Microsoft.DotNet.Cli.Prepare.targets
piotrpMSFT: Let's do this in a separate PR. For now, I've put them in the same place as other tasks coming from CLIBuildDll. Let's rationalize our Init/build.proj/Prepare story in one fell swoop.

piotrpMSFT:
@dotnet-bot test Windows_NT x64 Release Build  please.

Issue #3784, opened 2016-07-05
dotnet publish - please show full path to publish folder


dazinator:
## Steps to reproduce

Run `dotnet publish -o some\relative\folder`
## Expected  behavior

The output in the console shows the absolute path to the publish folder: i.e 

```
publish: Published to C:\SomeDir\SomeFoler\some\relative\folder

```
## Actual behavior

The output in the console shows the relative path - which makes it harder to actually track down where the folder is, when the environment current path isn't set to what you might be expecting. For example on CI servers.

```
publish: Published to some\relative\folder

```
## Environment data

`dotnet --info` output:

![image](https://cloud.githubusercontent.com/assets/3176632/16591196/a6cdfe62-42d2-11e6-9463-a4ace8f1f898.png)

piotrpMSFT:
moved to the sdk repo

PR #3863, opened 2016-07-14
show absolute path to the publish folder

akoken:
The output in the console shows the absolute path to the publish folder.

Related issue:
 #3784


piotrpMSFT:
Can you please create an issue to discuss this first? I think that reflecting the user's input is the right behavior, but would like to get other's thoughts. 

This PR would also need to come with some tests. For example, how does it behave if the user provides relative vs. absolute path? What is the impact of the Working Directory?

Finally, the implementation of publish will change drastically once we move to MS Build, so unless this is blocking an important scenario we should avoid churning this code for the time being.

PR #3869, opened 2016-07-15
CLI Build Cleanup: Round 1

piotrpMSFT:
Deletes as much of the old build framework as was found to not break the build
Removes unnecessary usings to make future cleanup easier
Adds license headers

Once @livarcocc's Publish work comes in this effort will be finishable, with a huge cleanup coming in-hand.


eerhardt:
>         public static readonly string SharedFrameworkVersion = "1.0.0";

These are duped in the build.proj today.  It would be nice if we could delete the C# versions of these soon...

---
Refers to: build_projects/dotnet-cli-build/CliDependencyVersions.cs:8 in 7efaf21. [](commit_id = 7efaf21f2d31d61cf1fb948311f6bec716fc2607, deletion_comment = False)

eerhardt:
>             "coreSetupDownload", 

All the "Dirs" are duped in the MSBuild targets.  So at some point it would be great if we could remove them in the C# code.

---
Refers to: build_projects/dotnet-cli-build/CliDirs.cs:13 in 7efaf21. [](commit_id = 7efaf21f2d31d61cf1fb948311f6bec716fc2607, deletion_comment = False)

eerhardt:
:shipit:

PR #3871, opened 2016-07-15
Fix invalid json in runtimeconfig

devlead:
The json in  [appname].runtimeconfig.json example is invalid, this adds the missing `,`

skipciplease ( skipci )


piotrpMSFT:
Thanks @devlead. @blackdwarf FYI.

PR #3895, opened 2016-07-19
Move CheckPrereqs to MSBuild

piotrpMSFT:
/cc @brthor @eerhardt @livarcocc


build/prepare/CentosCoreclrAndCoreFxDependencies.props
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+	<CentosCoreclrAndCoreFxDependencies Include="unzip" />
```
livarcocc: I thought we were using 2 spaces for tabs in XML.

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	<Import Project="UbuntuCoreclrAndCoreFxDependencies.props" />
+
+	<Target Name="CheckPrereqs"
```
livarcocc: Your spacing is all messed up here.

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
...
+			DependsOnTargets="CheckUbuntuAptGetDependencies;
+												CheckCentOsYumDependencies;">
+
+		<Message Text="Using Stage0 version:" />
+		<Exec Command="$(DotnetStage0) --version" />
+	</Target>
+
+	<Target Name="CheckUbuntuAptGetDependencies"
+			Condition=" '$(Rid)' == 'ubuntu.14.04-x64' ">
```
livarcocc: Spacing.

build/prepare/CentosCoreclrAndCoreFxDependencies.props
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
```
livarcocc: Should you condition this item group?

livarcocc:
Overall looks good. Just fix spacing and add conditions if possible to the itemgroups.

build/prepare/CheckPrereqs.targets
eerhardt: Doesn't this run before $(Rid) is set?

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	<Import Project="UbuntuCoreclrAndCoreFxDependencies.props" />
```
eerhardt: Who is importing the other 2 .props files?

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
...
+
+		<Exec Command="dpkg -s %(UbuntuCoreclrAndCoreFxDependencies)" />
+
+		<Exec Command="dpkg -s %(DebianPackageBuildDependencies)" />
+
+		<OnError ExecuteTargets="ShowAptGetUsage" />
+	</Target>
+
+	<Target Name="ShowAptGetUsage">
+		<Message Text="Install the missing dependency with `sudo apt-get install {package}`." />
```
eerhardt: Does this actually tell the user which package is not installed?

eerhardt:
:clock1:

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+	<Import Project="UbuntuCoreclrAndCoreFxDependencies.props" />
+
+	<Target Name="CheckPrereqs"
+			DependsOnTargets="CheckUbuntuAptGetDependencies;
+												CheckCentOsYumDependencies;">
+
+		<Message Text="Using Stage0 version:" />
+		<Exec Command="$(DotnetStage0) --version" />
```
brthor: Nit: Seems like a strange place for this call

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
...
+	</Target>
+
+	<Target Name="CheckUbuntuAptGetDependencies"
+			Condition=" '$(Rid)' == 'ubuntu.14.04-x64' ">
+
+		<Exec Command="dpkg -s %(UbuntuCoreclrAndCoreFxDependencies)" />
+
+		<Exec Command="dpkg -s %(DebianPackageBuildDependencies)" />
+
+		<OnError ExecuteTargets="ShowAptGetUsage" />
```
brthor: What if we were to do target batching here instead of just task batching, and capture the package currently being checked into a property so that `ShowAptGetUsage` will know which package to print wasn't installed?

build/prepare/CheckPrereqs.targets
```diff
@@ -0,0 +1,37 @@
...
+
+	<Target Name="ShowAptGetUsage">
+		<Message Text="Install the missing dependency with `sudo apt-get install {package}`." />
+	</Target>
+
+	<Target Name="CheckCentOsYumDependencies"
+			Condition=" '$(OSName)' == 'centos' ">
+		<Exec Command="yum list installed %(CentosCoreclrAndCoreFxDependencies)" />
+
+		<OnError ExecuteTargets="ShowYumUsage" />
```
brthor: Same question here on using target batching.

build/prepare/CentosCoreclrAndCoreFxDependencies.props
piotrpMSFT: Why doesn't SublimeText magically keep its settings in sync???

eerhardt: Because it isn't VS. 

image

build/prepare/CheckPrereqs.targets
piotrpMSFT: great catch

build/prepare/CheckPrereqs.targets
piotrpMSFT: fixed

build/prepare/CheckPrereqs.targets
piotrpMSFT:
````
Task "Exec"
  dpkg -s libc6
````

It shows that during the exec statement. If I get time I'll create a tooltask later. This is not a high value scenario for us and we do currently get sufficient diagnostics.

build/prepare/CheckPrereqs.targets
piotrpMSFT: Just bringing along what we did before. This call will tell us that a Stage0 directory exists, that it has a dotnet, and that that dotnet has a version.

build/prepare/CheckPrereqs.targets
piotrpMSFT: This will make the code messy in that we'd need to combine the itemgroups or to do two targets. If we decide this is becoming a pain we'll add a ToolTask that will handle the scenario for us.

PR #3911, opened 2016-07-21
MSBuild end to end

eerhardt:
This change enables the following MSBuild end to end experience work from the command-line:

```
dotnet new -t msbuild
dotnet restore
dotnet build3
dotnet run3

"Hello, World!"
```

@piotrpMSFT @brthor @livarcocc


src/dotnet/Microsoft.DotNet.Core.Sdk.props
```diff
@@ -0,0 +1,67 @@
...
+    <!-- TODO: Need to figure out runtimes -->
+
+    <!-- TODO: Below are hacks eerhardt made on top of davkean's .props file -->
+    
+    <!-- Temp Hack: davkean had this, but that casues all references not to be copied to output dir: 
+    <CopyNuGetImplementations>false</CopyNuGetImplementations>
+         This will be set to false during "build", but set to true during "publish"
+    -->
+    <!--Setting PlatformTarget is needed when CopyNuGetImplementations is left on, or else you get 'Your project.json doesn't list 'win10' as a targeted runtime. You should add '"win10": { }' inside your "runtimes" section in your project.json, and then re-run NuGet restore' errors -->
+    <PlatformTarget>x64</PlatformTarget>
```
brthor: Shouldn't this have a condition so it can be overridden?

src/dotnet/Microsoft.DotNet.Core.Sdk.targets
```diff
@@ -0,0 +1,100 @@
...
+
+    <PublishDir Condition=" '$(PublishDir)' == '' ">$(OutDir)publish\</PublishDir>
+  </PropertyGroup>
+
+  <Target
+      Name="Publish"
+      DependsOnTargets="$(PublishDependsOn)" />
+
+  <Target Name="CopyPublishArtifacts">
+    <Message Text="TODO: Implement CopyPublishArtifacts" />
```
brthor: What will this target do when it's implemented?

src/dotnet/commands/dotnet-new/Program.cs
```diff
@@ -64,11 +65,37 @@ where resourceName.Contains(templateDir)
                             }
                         }
 
-                        archive.ExtractToDirectory(Directory.GetCurrentDirectory());
+                        string projectDirectory = Directory.GetCurrentDirectory();
+
+                        archive.ExtractToDirectory(projectDirectory);
+
+                        string projectJsonFile = Path.Combine(projectDirectory, "project.json");
```
brthor: Can you extract this new logic into some aptly named methods? Took me a minute to realize what was going on in this new logic.

Perhaps ReplaceProjectJsonTemplateValues and ReplaceFileTemplateNames ?

src/dotnet/Microsoft.DotNet.Core.Sdk.targets
eerhardt: It will do exactly what the name says:  it copies the necessary artifacts to the publish output folder.

---
In reply to: [71795755](https://github.com/dotnet/cli/pull/3911#discussion_r71795755) [](ancestors = 71795755)

src/dotnet/commands/dotnet-new/Program.cs
eerhardt: Good point.  I should have caught this myself.

---
In reply to: [71796369](https://github.com/dotnet/cli/pull/3911#discussion_r71796369) [](ancestors = 71796369)

src/Microsoft.DotNet.ProjectModel/Project.cs
```diff
@@ -84,6 +84,8 @@ public string ProjectDirectory
 
         public bool IsTestProject => !string.IsNullOrEmpty(TestRunner);
         
+        public bool OverrideIsRunnable { get; set; }
```
brthor: Why do we need this? Why not reuse the compilerOptions.EmitEntryPoint

eerhardt: Because that option doesn't go into project.json anymore.  It goes into .csproj <OutputType>Exe</>

---
In reply to: [71797075](https://github.com/dotnet/cli/pull/3911#discussion_r71797075) [](ancestors = 71797075)

src/Microsoft.DotNet.Compiler.Common/Executable.cs
```diff
@@ -97,13 +97,12 @@ private void MakeCompilationOutputRunnableForCoreCLR()
...
 
-            var emitEntryPoint = _compilerOptions.EmitEntryPoint ?? false;
+            var isRunnable = _compilerOptions.EmitEntryPoint ?? _context.ProjectFile.OverrideIsRunnable;
 
-            if (emitEntryPoint && !_context.IsPortable)
+            if (isRunnable && !_context.IsPortable)
             {
-                // TODO: Pick a host based on the RID
                 VerifyCoreClrPresenceInPackageGraph();
-                CoreHost.CopyTo(_runtimeOutputPath, _compilerOptions.OutputName + Constants.ExeSuffix);
```
brthor: Where does the host come from when running build/publish?

eerhardt: So publish it does the same as it did before.  I just moved that code into this Executable class.

Here, I changed "build" to do the same thing publish does - get it from the dependencies.

---
In reply to: [71797574](https://github.com/dotnet/cli/pull/3911#discussion_r71797574) [](ancestors = 71797574)

eerhardt: Note: `MakeCompilationOutputRunnableForCoreCLR` is only called from Build.  Publish doesn't call this method.

---
In reply to: [71797731](https://github.com/dotnet/cli/pull/3911#discussion_r71797731) [](ancestors = 71797731,71797574)

brthor: I guess I don't see where the host is coming from in build. What's getting it from the dependencies?

---
In reply to: [71797819](https://github.com/dotnet/cli/pull/3911#discussion_r71797819) [](ancestors = 71797819,71797731,71797574)

src/dotnet/Microsoft.DotNet.Core.Sdk.targets
brthor: I got that, What artifacts though? This ties in with the larger copy to output discussion. But if we use metadata to control what should be copied to output then Im not convinced we need a specific target to actually copy those things. 

It could just be Publish set's copy to output on the right things.

---
In reply to: [71796387](https://github.com/dotnet/cli/pull/3911#discussion_r71796387) [](ancestors = 71796387,71795755)

eerhardt: "Artifacts" are all the things that need to go in the publish output directory (.dlls, content files, etc.).  There has to be SOME target that copies all the files into the publish directory.  This is that target.  Who else is going to do that copying if this doesn't?

---
In reply to: [71798420](https://github.com/dotnet/cli/pull/3911#discussion_r71798420) [](ancestors = 71798420,71796387,71795755)

src/Microsoft.DotNet.ProjectModel/Project.cs
brthor: I get that, but I see this as a hack so we can force a non-synchronized project.json and csproj to behave in a certain way.
Why make a new property for this hack rather than reusing the property that is already there is what I'm getting at. 

If we see <OutputType>Exe</OutputType> why not set compilationOptions.EmitEntrypoint to true

---
In reply to: [71797149](https://github.com/dotnet/cli/pull/3911#discussion_r71797149) [](ancestors = 71797149,71797075)

eerhardt: Ah, I see what you mean now.  I thought you meant that the template (and users) need to set it in the project.json file.

fixed.

---
In reply to: [71798696](https://github.com/dotnet/cli/pull/3911#discussion_r71798696) [](ancestors = 71798696,71797149,71797075)

src/dotnet/Microsoft.DotNet.Core.Sdk.targets
brthor: Check out what is happening today:
[https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L3791](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L3791)
[https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L3985](https://github.com/Microsoft/msbuild/blob/master/src/XMakeTasks/Microsoft.Common.CurrentVersion.targets#L3985)

I'm proposing we can piggyback on the larger copy to output system.


---
In reply to: [71798669](https://github.com/dotnet/cli/pull/3911#discussion_r71798669) [](ancestors = 71798669,71798420,71796387,71795755)

src/Microsoft.DotNet.Compiler.Common/Executable.cs
brthor: For publish, it would auto output hte host when it was in the package graph, the same is not true for build (or at least so I thought). I had expected there would be changes to tell build, go copy this one thing from the package graph.

I might be incorrect here, just want to have the right understanding.

---
In reply to: [71798032](https://github.com/dotnet/cli/pull/3911#discussion_r71798032) [](ancestors = 71798032,71797819,71797731,71797574)

src/dotnet/Microsoft.DotNet.Core.Sdk.targets
eerhardt: Yes, this target is going to reuse as much of the existing msbuild logic as possible.  And change where it makes sense.

---
In reply to: [71799150](https://github.com/dotnet/cli/pull/3911#discussion_r71799150) [](ancestors = 71799150,71798669,71798420,71796387,71795755)

brthor: this will flesh out later on

---
In reply to: [71799484](https://github.com/dotnet/cli/pull/3911#discussion_r71799484) [](ancestors = 71799484,71799150,71798669,71798420,71796387,71795755)

src/Microsoft.DotNet.ProjectModel/Project.cs
eerhardt: Turns out this can't be fixed.  The "compilerOptions" object that is returned from GetCompilerOptions is a new object every time.  So if we set properties on it, they will be lost the next time GetCompilerOptions is invoked.

---
In reply to: [71799001](https://github.com/dotnet/cli/pull/3911#discussion_r71799001) [](ancestors = 71799001,71798696,71797149,71797075)

src/dotnet/Microsoft.DotNet.Core.Sdk.props
eerhardt: This is just a point-in-time hack.  And this props file gets included at the beginning of the project, so you can override it anytime afterwards.

But I've added the Condition just in case.

---
In reply to: [71795608](https://github.com/dotnet/cli/pull/3911#discussion_r71795608) [](ancestors = 71795608)

src/Microsoft.DotNet.Compiler.Common/Executable.cs
eerhardt: I think this was changed during an earlier prototype when all the dependencies were being copied to the build output folder.  I'm reverting this change.

---
In reply to: [71799445](https://github.com/dotnet/cli/pull/3911#discussion_r71799445) [](ancestors = 71799445,71798032,71797819,71797731,71797574)

src/Microsoft.DotNet.ProjectModel/Project.cs
brthor: Okay, makes sense for this property then

src/dotnet/Microsoft.DotNet.Core.Sdk.props
```diff
@@ -15,7 +15,8 @@
 
   <!-- User-facing configuration-specific defaults -->
   <PropertyGroup>
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <!-- TODO: This property can't be set in the CLI since ResolveNuGetPackageAssets needs an exact RID from the project.json, and the RID is made up of the PlatformTarget -->
+    <!--<PlatformTarget>AnyCPU</PlatformTarget>-->
```
brthor: Not sure what the right fix here is, perhaps something is smart enough to infer x86 and x64? Or we add support for non-concrete rids?

src/dotnet/Microsoft.DotNet.Core.Sdk.props
```diff
@@ -54,8 +55,6 @@
     <CopyNuGetImplementations>false</CopyNuGetImplementations>
          This will be set to false during "build", but set to true during "publish"
     -->
-    <!--Setting PlatformTarget is needed when CopyNuGetImplementations is left on, or else you get 'Your project.json doesn't list 'win10' as a targeted runtime. You should add '"win10": { }' inside your "runtimes" section in your project.json, and then re-run NuGet restore' errors -->
-    <PlatformTarget Condition=" '$(PlatformTarget)' == '' ">x64</PlatformTarget>
```
brthor: Wouldn't you still need x64?

src/dotnet/Microsoft.DotNet.Core.Sdk.props
eerhardt: For now, the CLI passes this in during "build3".  We need the new "ResolveNuGetPackageAssets" target that Nat is writing in order to make this correct.

---
In reply to: [71824700](https://github.com/dotnet/cli/pull/3911#discussion_r71824700) [](ancestors = 71824700)

src/dotnet/Microsoft.DotNet.Core.Sdk.props
eerhardt: See previous comment.

eerhardt:
test Ubuntu x64 Release Build

eerhardt:
OK, this is green now.  Please review.

eerhardt:
test Ubuntu x64 Release Build

eerhardt:

eerhardt:
test Fedora23 x64 Debug Build

eerhardt:
test Fedora23 x64 Debug Build

brthor:
Looks good, this is awesome progress, :shipit:

PR #3975, opened 2016-08-02
Remove temporary hacks from new NuGet

piotrpMSFT:
@eerhardt @brthor

### DIFF
--- a/build/Microsoft.DotNet.Cli.Compile.targets
+++ b/build/Microsoft.DotNet.Cli.Compile.targets
@@ -70,9 +70,6 @@
         <BinObj Include="$(SrcDirectory)/**/bin/**/*" />
         <BinObj Include="$(SrcDirectory)/**/obj/**/*" />
 
-        <FilesToRemoveFromStageDirectory Remove="*" />
-        <FilesToRemoveFromStageDirectory Include="$(StageDirectory)/**/*" />
-
         <SharedFramework Remove="*" />
         <SharedFramework Include="$(SharedFrameworkPublishDirectory)/**/*" />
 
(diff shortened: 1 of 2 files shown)


build/Microsoft.DotNet.Cli.Compile.targets
```diff
@@ -70,9 +70,6 @@
         <BinObj Include="$(SrcDirectory)/**/bin/**/*" />
         <BinObj Include="$(SrcDirectory)/**/obj/**/*" />
 
-        <FilesToRemoveFromStageDirectory Remove="*" />
-        <FilesToRemoveFromStageDirectory Include="$(StageDirectory)/**/*" />
```
eerhardt: I think this was in there before the new NuGet build came in.

build/compile/Microsoft.DotNet.Cli.LzmaArchive.targets
```diff
@@ -23,11 +23,6 @@
 
       <Message Text="Publishing Archiver" />
 
-      <!-- TEMP HACK: need to restore with the current stage, until we get a Stage0 with the new NuGet build -->
```
brthor: Do we need to replace this with restore from a stage0 dotnet?

build/Microsoft.DotNet.Cli.Compile.targets
piotrpMSFT: Correct. It's generating unnecessary spew in the msbuild logs.

build/compile/Microsoft.DotNet.Cli.LzmaArchive.targets
piotrpMSFT: We are not deleting the lock files anymore so a re-restore is not required.

eerhardt: Correct - we only restore once before compile.

build/Microsoft.DotNet.Cli.Compile.targets
eerhardt: But isn't it necessary to ensure the stage is clean before we start putting stuff in it?

Maybe calling "RemoveDir" instead will cut down on the spew.

brthor:
LGTM :shipit:

build/Microsoft.DotNet.Cli.Compile.targets
eerhardt: Ah, I see it now on line 83.

eerhardt:
:shipit:

PR #3984, opened 2016-08-03
fix the rest of featmsbuild versions

brthor:
@eerhardt @livarcocc @piotrpMSFT

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/project.json
+++ b/src/Microsoft.DotNet.ProjectModel/project.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-rc4-*",
+  "version": "1.0.0-featmsbuild-*",
   "buildOptions": {
     "keyFile": "../../tools/Key.snk"
   },
--- a/src/Microsoft.DotNet.Configurer/project.json
+++ b/src/Microsoft.DotNet.Configurer/project.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-rc3-*",
+  "version": "1.0.0-featmsbuild-*",
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",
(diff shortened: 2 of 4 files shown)


src/Microsoft.DotNet.ProjectModel/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-rc4-*",
+  "version": "1.0.0-featmsbuild-*",
```
piotrpMSFT: This doesn't seem right. ProjectModel RTM'd. Why do we have it at the same suffix as the preview components?

piotrpMSFT: it's strange that this nupkg was on an rc4 version. it seems delibirately different from others. Do you know why?

eerhardt: ProjectModel was scheduled to RTM at first.  We thought that ASP.NET runtime needed it.  However, ProjectModel references NuGet, and NuGet wasn't RTM'ing.  So we talked to the ASP.NET team, and they said their tooling references ProjectModel.  The runtime only needs DependencyModel.

Since we had already labeled ProjectModel as "rc2" as part of the whole "rc2" release, we bumped its version to "rc3" for Preview2.  After Preview2, we bumped it again to "rc4".

So that's the story of ProjectModel.  It has been in a weird state, but has never RTM'd.  Bringing it in line with the rest of our tooling makes sense to me.

eerhardt:
Note that InternalAbstractions and DependencyModel are being moved to core-setup by @livarcocc.  I'm not certain changing their prerelease suffixes from "beta" to "featmsbuild" right now is the best thing, since it will go back to "beta" when they are moved to core-setup.

brthor:
I'll remove the changes to dependency model and internal abstractions.

eerhardt:
Is this change still happening?  Any update?

brthor:
fell off my radar. We never landed on a version for ProjectModel. 

I think it should just be featmsbuild for the time being. This is going to be a migration only component in the near future and I imagine we'll stop shipping the nupkg at that time as well.

brthor:
Not important anymroe

PR #3989, opened 2016-08-04
Pass debugType thru to the command line compiler

tmat:
dotnet cli shouldn't automatically treat invalid value of "debugType" project setting as "full". Instead it should either pass the value as is to the compiler, which already reports a good error message (this PR), or report error itself.

This change indirectly enables "debugType" to be "embedded". See https://github.com/dotnet/roslyn/issues/12390 for background.

### DIFF
--- a/src/dotnet/commands/dotnet-compile-csc/Program.cs
+++ b/src/dotnet/commands/dotnet-compile-csc/Program.cs
@@ -198,19 +198,17 @@ namespace Microsoft.DotNet.Tools.Compiler.Csc
                 commonArgs.Add("-t:library");
             }
 
+            string debugType;
             if (string.IsNullOrEmpty(options.DebugType))
             {
-                commonArgs.Add(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
-                    ? "-debug:full"
-                    : "-debug:portable");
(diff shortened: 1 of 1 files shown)


tmat:
I haven't found any existing tests for "debugType". Please point me to a sample test that I can use and where to put it.

eerhardt:
> I haven't found any existing tests for "debugType". Please point me to a sample test that I can use and where to put it.

For some reason, the only ones I can find are for "dotnet test" when the test app is using full or portable pdbs. But this doesn't seem like a great place to put them.

I would suggest putting the test project.json files under `TestAssets\TestProjects` and either putting the test methods in `test\dotnet-compile.UnitTests` or `test\dotnet-compile.Tests`.

tmat:
@eerhardt I was thinking it'd be good to have more unit test approach - check that given a JSON file I expect certain command line to csc.

eerhardt:
Full disclosure: this code will be going away with the move to MSBuild.  So, I wouldn't suggest too much investment here.

tmat:
I see. Then I think I should not spend time writing new tests right now.

eerhardt:
> Then I think I should not spend time writing new tests right now.

That sounds fine to me.

tmat:
OK, I'm gonna merge then.

PR #4000, opened 2016-08-04
Modified dotnet-install.ps1 to support Nano Server

### LINKED ISSUE #3113: dotnet-install.ps1 should work on NanoServer
Download and run https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1 on a NanoServer.

**Actual Results:**
powershell : new-object : Cannot find type [net.webclient]: verify that the assembly 
    + CategoryInfo          : NotSpecified: (new-object : Ca...t the assembly :String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
containing this type is loaded.
At line:1 char:9
+ &{iex ((new-object net.webclient).DownloadString('https://raw.githubu ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidType: (:) [New-Object], PSArgumentExcepti 
   on
    + FullyQualifiedErrorId : TypeNotFound,Microsoft.PowerShell.Commands.NewOb 
   jectCommand

**Expected Results:**
Script should install the dotnet cli successfully.

The script is using Invoke-WebRequest which is based on net.webclient.  I believe you need to use HttpClient in order for it to work on Nano.

MichaelSimons:
Fixes #3113

cc @krwq @Priya91 @mellinoe

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -126,12 +126,23 @@ function Get-Latest-Version-Info([string]$AzureFeed, [string]$AzureChannel, [str
         $VersionFileUrl = "$AzureFeed/Sdk/$AzureChannel/latest.version"
     }
     
-    $Response = Invoke-WebRequest -UseBasicParsing $VersionFileUrl
+    try {
+        # HttpClient is used vs Invoke-WebRequest in order to support Nano Server which doesn't support the Invoke-WebRequest cmdlet.
+        Add-Type -Assembly System.Net.Http | Out-Null
+        $HttpClient = New-Object System.Net.Http.HttpClient
+        $Response = $HttpClient.GetAsync($VersionFileUrl).Result
+        $StringContent = $Response.Content.ReadAsStringAsync().Result
 
-    switch ($Response.Headers.'Content-Type'){
-        { ($_ -eq "application/octet-stream") } { $VersionText = [Text.Encoding]::UTF8.GetString($Response.Content) }
-        { ($_ -eq "text/plain") } { $VersionText = $Response.Content }
-        default { throw "``$Response.Headers.'Content-Type'`` is an unknown .version file content type." }
+        switch ($Response.Content.Headers.ContentType) {
+            { ($_ -eq "application/octet-stream") } { $VersionText = [Text.Encoding]::UTF8.GetString($StringContent) }
+            { ($_ -eq "text/plain") } { $VersionText = $StringContent }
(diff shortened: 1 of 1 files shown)


scripts/obtain/dotnet-install.ps1
```diff
@@ -281,7 +292,13 @@ function Get-List-Of-Directories-And-Versions-To-Unpack-From-Dotnet-Package([Sys
...
+    try {
+        Add-Type -Assembly System.IO.Compression.FileSystem | Out-Null
+    }
+    catch {
+        # Loading FileSystem is unnecessary and will fail on Nano Server
+    }
+
     Set-Variable -Name Zip
     try {
         $Zip = [System.IO.Compression.ZipFile]::OpenRead($ZipPath)
```
Priya91: how does powershell find this type, without loading ```System.Io.Compression.ZipFile``` assembly?

eerhardt:
:shipit:

scripts/obtain/dotnet-install.ps1
MichaelSimons: I am not an expert on this but my understanding is that Nano Server has a different flavor of Powershell - Core Edition built on .NET Core - vs the Desktop Edition build on .NET Framework.  Add-Type works on the Desktop Edition because it finds the assembly in the GAC.  On Nano, Add-Type is unable to resolve the assembly because it isn't GAC'd or does it reside in the active directory therefore an exception is thrown.  Core Edition is however able to resolve type references from the base libraries without having to explicitly load the assemblies.  They are automatically resolved from assemblies in C:\Windows\system32\DotNetCore\v1.0.

scripts/obtain/dotnet-install.ps1
```diff
@@ -281,7 +292,13 @@ function Get-List-Of-Directories-And-Versions-To-Unpack-From-Dotnet-Package([Sys
 function Extract-Dotnet-Package([string]$ZipPath, [string]$OutPath) {
     Say-Invocation $MyInvocation
 
-    Add-Type -Assembly System.IO.Compression.FileSystem | Out-Null
+    try {
```
MichaelSimons: This pattern needs to also be applies to the System.NET.Http logic that was added.  When I tested this, my PS environment already had the assembly loaded therefore it didn't fail on NanoServer.  If you run within a new PS session, it will fail.

MichaelSimons:
@dotnet-bot test Ubuntu16.04 x64 Debug Build please.

PR #4024, opened 2016-08-09
src/redist/project.json

piotrpMSFT:
Creates a redist/project.json project that pulls together all executables that make up the CLI
- To enable this, makes each tool [nuget, msbuild, csc, etc.] a standalone project
- Removes unnecessary dependencies from dotnet/project.json, which was turning into a dependency dumping ground
- Makes nuget run out-of-proc from dotnet.dll


build/Microsoft.DotNet.Cli.Compile.targets
```diff
@@ -208,12 +210,8 @@
       </ItemGroup>
 
       <Copy SourceFiles="@(MSBuildTargetsToCopy)"
-            DestinationFiles="@(MSBuildTargetsToCopy->'$(SdkOutputDirectory)\%(RecursiveDir)%(Filename)%(Extension)')" />
+            DestinationFiles="@(MSBuildTargetsToCopy->'$(SdkOutputDirectory)/%(RecursiveDir)%(Filename)%(Extension)')" />
 
-    <!-- Temporary workaround for MSBuild placing their .props files in the old 14.1 directory -->
-    <Move SourceFiles="$(SdkOutputDirectory)/14.1/Microsoft.Common.props"
```
eerhardt: I think you still need this workaround

src/dotnet/commands/dotnet-restore/NuGetForwardingApp.cs
```diff
@@ -0,0 +1,44 @@
...
+                GetNuGetExePath(),
+                argsToForward);
+        }
+
+        public int Execute()
+        {
+            return _forwardingApp.Execute();
+        }
+
+        private static string GetHostPath()
```
eerhardt: Is this method used?

src/nuget/project.json
```diff
@@ -0,0 +1,27 @@
...
+      "version": "3.5.0-rc1-1697",
+      "exclude": "compile"
+    },
+    "NuGet.CommandLine.XPlat": "3.5.0-rc1-1697"
+  },
+  "frameworks": {
+    "netcoreapp1.0": {
+      "imports": [
+        "dnxcore50",
+        "netstandardapp1.5",
```
eerhardt: Are all these imports still needed?  I thought NuGet fixed up all their packages to not require imports anymore.

src/nuget/project.json
```diff
@@ -0,0 +1,27 @@
...
+  "version": "1.0.0-featmsbuild-*",
+  "buildOptions": {
+    "emitEntryPoint": true
+  },
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0"
+    },
+    "NuGet.Commands": {
```
eerhardt: NuGet.CommandLine.XPlat references this.  Why do we need an explicit reference to it?

src/nuget/project.json
```diff
@@ -0,0 +1,27 @@
+{
```
eerhardt: Should this directory be "tool_nuget" to align with "tool_msbuild"

src/dotnet/commands/dotnet-restore/NuGetForwardingApp.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    public class NuGetForwardingApp
+    {
+        private const string s_nugetExeName = "NuGet.CommandLine.XPlat.dll";
+        private readonly ForwardingApp _forwardingApp;
+
+        public NuGetForwardingApp(string[] argsToForward)
+        {
+            _forwardingApp = new ForwardingApp(
+                GetNuGetExePath(),
```
eerhardt: What deps file and runtimeconfig are used here?

src/nuget/project.json
```diff
@@ -0,0 +1,27 @@
...
+  "version": "1.0.0-featmsbuild-*",
```
eerhardt: With this new project, are we getting a new .dll named "nuget.dll" in the output folder?  if yes, is that intentional?

build/Microsoft.DotNet.Cli.Compile.targets
```diff
@@ -121,36 +129,26 @@
...
 
-      <RemoveAssetFromDepsPackages DepsFile="$(SdkOutputDirectory)/csc.deps.json"
-                                   SectionName="runtimeTargets"
-                                   AssetPath="$(BinaryToCorehostifyRelDir)/%(BinariesToRemove.Identity).exe" />
-      <RemoveAssetFromDepsPackages DepsFile="$(SdkOutputDirectory)/dotnet.deps.json"
-                                   SectionName="runtimeTargets"
-                                   AssetPath="$(BinaryToCorehostifyRelDir)/%(BinariesToRemove.Identity).exe" />
-
-      <!-- cleanup compilers project output we don't need -->
-      <Delete Files="$(SdkOutputDirectory)/compilers%(PublishOutputExtensions.Identity)" />
```
eerhardt: Do we still need to delete the compilers output?  How about the src\nuget output?

src/dotnet/commands/dotnet-build3/MSBuildForwardingApp.cs
```diff
@@ -56,19 +54,18 @@ private static string GetHostPath()
 
         private static string GetRuntimeConfig()
```
eerhardt: This and GetDepsFile don't appear to be called anymore.

eerhardt:
>     public class Program

Is this file still needed?  We don't have it in the nuget, nor 'tool_msbuild'

---
Refers to: src/csc/Program.cs:5 in 3a55a41. [](commit_id = 3a55a4182f49c6c63072ea7b47f8efde2cdd7355, deletion_comment = False)

build/Microsoft.DotNet.Cli.Compile.targets
```diff
@@ -3,8 +3,16 @@
   <Import Project="compile/Microsoft.DotNet.Cli.LzmaArchive.targets" />
   <Import Project="crossgen/Microsoft.DotNet.Cli.Crossgen.targets" />
 
+  <PropertyGroup>
+    <BinaryToCorehostifyRelDir>runtimes/any/native</BinaryToCorehostifyRelDir>
+    <CoreSDKDir>$(RepoRoot)/resources/core-sdk</CoreSDKDir>
```
eerhardt: not sure I am crazy about 'core-sdk' as a name for this directory.

How about `$(RepoRoot)/resources/MSBuild`  or `$(RepoRoot)/resources/MSBuildImports`.

Note that our .props and .targets are getting deleted from our repo now that they live in the dotnet/sdk repo.

src/redist/project.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0"
+    },
+    "csc": "1.0.0-featmsbuild-*",
+    "dotnet": "1.0.0-featmsbuild-*",
+    "nuget": "1.0.0-featmsbuild-*",    
+    "tool_msbuild": "1.0.0-featmsbuild-*"
```
brthor: Should these have `"target": "project"`

build/Microsoft.DotNet.Cli.Compile.targets
```diff
@@ -76,6 +83,7 @@
         <BinariesToRemove Remove="*" />
         <BinariesToRemove Include="csc" />
         <BinariesToRemove Include="vbc" />
+        <BinariesToRemove Include="MSBuild" />
```
eerhardt: Is it weird that this Item is called "BinariesToRemove", but I don't see them being removed?

src/dotnet/commands/dotnet-restore/NuGetForwardingApp.cs
piotrpMSFT: build scripts duplicate dotnet.deps & dotnet.runtimeconfig for each tool we package with the sdk. This makes them independently runnable so e.g. msbuild doesn't need to understand the workarounds we have in place in CLI.

src/dotnet/commands/dotnet-restore/NuGetForwardingApp.cs
piotrpMSFT: fixed.

src/nuget/project.json
piotrpMSFT: fixed. tool_csc as well.

src/nuget/project.json
piotrpMSFT: removed.

src/nuget/project.json
piotrpMSFT: removed.

build/Microsoft.DotNet.Cli.Compile.targets
piotrpMSFT: fixed. Also now deleting the 14.1 directory.

src/redist/project.json
piotrpMSFT: that's implicit.

build/Microsoft.DotNet.Cli.Compile.targets
piotrpMSFT: No. we only publish redist, which transitively brings along the rest.

src/tool_* does not have a Program.cs so no binary gets generated, so nothing to clean up there either.

build/Microsoft.DotNet.Cli.Compile.targets
piotrpMSFT: sounds good. When this stuff lived under src/dotnet its intent was really unclear. I changed to MSBuildImports.

build/Microsoft.DotNet.Cli.Compile.targets
piotrpMSFT: They are removed from deps.json files.. I renamed to AssetsToRemoveFromDeps

src/dotnet/commands/dotnet-build3/MSBuildForwardingApp.cs
piotrpMSFT: I went through the file and removed all dead code.

src/nuget/project.json
piotrpMSFT: no, we're not. no .cs files.

src/redist/project.json
brthor: Is it guaranteed implicitly?

IIRC we've run into issues where packages were being pulled where we expected projects. Being explicit here is better IMO.

build/Microsoft.DotNet.Cli.Compile.targets
```diff
@@ -3,8 +3,16 @@
   <Import Project="compile/Microsoft.DotNet.Cli.LzmaArchive.targets" />
   <Import Project="crossgen/Microsoft.DotNet.Cli.Crossgen.targets" />
 
+  <PropertyGroup>
+    <BinaryToCorehostifyRelDir>runtimes/any/native</BinaryToCorehostifyRelDir>
+    <CoreSDKDir>$(RepoRoot)/resources/MSBuildImports</CoreSDKDir>
```
eerhardt: One last nit about this.  The $(CoreSDKDir) property is now out of date, and doesn't align with the folder it is referencing.

eerhardt: Same for $(CoreSDKContent) below.

---
In reply to: [74152883](https://github.com/dotnet/cli/pull/4024#discussion_r74152883) [](ancestors = 74152883)

src/redist/project.json
```diff
@@ -0,0 +1,33 @@
...
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0"
+    },
+    "tool_csc": "1.0.0-featmsbuild-*",
+    "dotnet": "1.0.0-featmsbuild-*",
+    "tool_nuget": "1.0.0-featmsbuild-*",
+    "tool_msbuild": "1.0.0-featmsbuild-*",
+
+    "MSBuild": "0.1.0-preview-00029-160805",
```
eerhardt: Shouldn't all these MSBuild dependencies come through the "tool_msbuild" dependency?  Why duplicate them here?

eerhardt:
:shipit:

Issue #2874, opened 2016-05-04
.NET CLI should look for restored assets using lowercase paths


joelverhagen:
I'm working on an effort to make the package ID and version components of file paths generated by NuGet case insensitive. The most prevalent example of this is packages within the global packages folder (e.g. `~/.nuget/packages/{id}/{version}`). This effort requires changes in NuGet , .NET CLI, and DotNetHost. The associated NuGet item is here:
https://github.com/NuGet/Home/issues/2522

Work in .NET CLI is:

1. When CLI is looking for restored assets, use lowercase paths (ideally using NuGet APIs or the lock file `path` property).
1. Include the `path` property in the `.deps.json` file.

piotrpMSFT:
@joelverhagen I just added this to the RTM milestone.

joelverhagen:
This is fixed is complete with @eerhardt's work in `feature/msbuild` and by https://github.com/dotnet/cli/pull/4053, which lays the groundwork for DotNetHost.

PR #4032, opened 2016-08-11
Add "path" property to .deps.json file libraries

joelverhagen:
This is step #3 in this NuGet issue: https://github.com/NuGet/Home/issues/2522. This work is also tracked by this CLI issue: https://github.com/dotnet/cli/issues/2874.

This PR adds a `"path"` property to the `"libraries"` entries in the `.deps.json` file. For packages, this is the relative path appended to the global packages folder (or fallback folder) to access the package's assets. For projects and other library types, this value is null.

After this work, the .NET Core host process need to be modified to prefer this property over the concatenation of the package ID and version.

/cc @eerhardt @emgarten @rrelyea @piotrpMSFT


eerhardt:
@joelverhagen, I hate to break this to you, but @livarcocc just removed the DependencyModel library from the dotnet/cli repo  in the feature/msbuild branch.  Instead, this library now lives in https://github.com/dotnet/core-setup/tree/master/src/Microsoft.Extensions.DependencyModel.

Is there any way you can port your changes to DependencyModel into that repo?

src/dotnet/commands/dotnet-publish/Program.cs
```diff
@@ -69,7 +69,7 @@ public static int Run(string[] args)
             }
             catch (Exception ex)
             {
-                Reporter.Error.WriteLine(ex.Message.Red());
+                Reporter.Error.WriteLine(ex.ToString().Red());
```
eerhardt: I'm not sure we want to always print the entire to string to the user.  We don't for other commands.

joelverhagen: Good catch, this was a debugging change that should not be merged.

joelverhagen:
> @joelverhagen, I hate to break this to you, but @livarcocc just removed the DependencyModel library from the dotnet/cli repo in the feature/msbuild branch.

No worries. I'll react to this change and close this PR if necessary.

eerhardt:
I looked through the changes, and they all look good to me.  Minus the part about this library being moved to another repo.

Maybe the best way forward is to make the change to DependencyModel and the DotNetHost in core-setup first.  And then make the CLI changes in the feature/msbuild branch.

joelverhagen:
Closing because this should be based off of the feature/msbuild branch (due to DependencyModel change).

Issue #2477, opened 2016-04-13
dotnet build fails - attempts to build *.aspx file in node_modules package


spboyer:
## Steps to reproduce
```
dotnet new
npm init 
npm install protractor --saveDev
dotnet build
```

## Expected  behavior
no errors

## Actual behavior
```
<projectname>/node_modules/protractor/node_modules/selenium-webdriver/lib/test/data/Page.aspx.cs(4,36): error CS0234: The type or namespace name 'Web' does not exist in the namespace 'System' (are you missing an assembly reference?)
```

## Environment data
```
dotnet --info
.NET Command Line Tools (1.0.0-beta-002202)

Product Information:
 Version:     1.0.0-beta-002202
 Commit Sha:  60b23d5115

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.11
 OS Platform: Darwin
 RID:         osx.10.11-x64
```

spboyer:
Is there an `ignore` setting in the **project.json** for **node_modules** to be set on *build, compile, etc*

spboyer:
```
"compileExclude": "node_modules"
```

geoemm:
compileExclude is deprecated now, use "compile": { "exclude": "node_modules/"  } in buildOptions

Dzivo:
This still throws the same error

piotrpMSFT:
I just tried this in preview3 with csproj and it worked quite nicely :) Here is the csproj I used, for reference. Notice the <Compile ... Exclude="..."/> tag.

```
<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
  
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp1.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="**\*.cs"
             Exclude="node_modules/selenium-webdriver/**/*.cs" />
    <EmbeddedResource Include="**\*.resx" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NETCore.App">
      <Version>1.0.1</Version>
    </PackageReference>
    <PackageReference Include="Microsoft.NET.Sdk">
      <Version>1.0.0-alpha-20161104-2</Version>
[...]

PR #4035, opened 2016-08-11
[WIP] Adding Restore3 command

emgarten:
This change adds the NuGet.Build.Tasks package containing NuGet's MSBuild task for restore. The Restore3 command will call the Restore target in NuGet.targets and convert the command line parameters into MSBuild properties.

### DIFF
--- a/build_projects/dotnet-cli-build/project.json
+++ b/build_projects/dotnet-cli-build/project.json
@@ -20,7 +20,8 @@
     "System.Runtime.Serialization.Primitives": "4.1.1",
     "System.Xml.XmlSerializer": "4.0.11",
     "WindowsAzure.Storage": "6.2.2-preview",
-    "NuGet.CommandLine.XPlat": "3.5.0-rc1-1697",
+    "NuGet.Build.Tasks": "3.6.0-beta.1.msbuild.1",
+    "NuGet.CommandLine.XPlat": "3.6.0-beta.1.msbuild.1",
     "Microsoft.Build.Framework": "0.1.0-preview-00029-160805",
     "Microsoft.Build.Utilities.Core": "0.1.0-preview-00029-160805",
     "Microsoft.DotNet.PlatformAbstractions": "1.0.1-beta-000914"
--- a/build_projects/dotnet-cli-build/project.json
+++ b/build_projects/dotnet-cli-build/project.json
@@ -20,7 +20,6 @@
     "System.Runtime.Serialization.Primitives": "4.1.1",
     "System.Xml.XmlSerializer": "4.0.11",
     "WindowsAzure.Storage": "6.2.2-preview",
-    "NuGet.Build.Tasks": "3.6.0-beta.1.msbuild.1",
     "NuGet.CommandLine.XPlat": "3.6.0-beta.1.msbuild.1",
     "Microsoft.Build.Framework": "0.1.0-preview-00029-160805",
     "Microsoft.Build.Utilities.Core": "0.1.0-preview-00029-160805",
--- a/src/dotnet/commands/dotnet-restore3/Program.cs
+++ b/src/dotnet/commands/dotnet-restore3/Program.cs
@@ -1,0 +1,111 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.CommandLine;
+using Microsoft.DotNet.Cli.Utils;
+
+namespace Microsoft.DotNet.Tools.Restore3
+{
+    public class Restore3Command
+    {
+        public static int Run(string[] args)
+        {
+            DebugHelper.HandleDebugSwitch(ref args);
+
+            CommandLineApplication cmd = new CommandLineApplication(throwOnUnexpectedArg: false)
+            {
+                Name = "restore3",
+                FullName = "restore3",
+                Description = "restore for msbuild"
+            };
+
+            cmd.HelpOption("-h|--help");
+
+            var argRoot = cmd.Argument(
+                    "[root]",
+                    "Optional path to a project file or MSBuild arguments.",
+                    multipleValues: true);
+
+            var sourceOption = cmd.Option(
+                    "-s|--source <source>",
+                    "Specifies a NuGet package source to use during the restore.",
+                    CommandOptionType.MultipleValue);
+
+            var packagesOption = cmd.Option(
+                    "--packages <packagesDirectory>",
+                    "Directory to install packages in.",
+                    CommandOptionType.SingleValue);
+
+            var disableParallelOption = cmd.Option(
+                    "--disable-parallel",
+                    "Disables restoring multiple projects in parallel.",
+                    CommandOptionType.NoValue);
+
+            var configFileOption = cmd.Option(
+                    "--configfile <file>",
+                    "The NuGet configuration file to use.",
+                    CommandOptionType.SingleValue);
+
+            var noCacheOption = cmd.Option(
+                    "--no-cache",
+                    "Do not cache packages and http requests.",
+                    CommandOptionType.NoValue);
+
+            var ignoreFailedSourcesOption = cmd.Option(
+                    "--ignore-failed-sources",
+                    "Treat package source failures as warnings.",
+                    CommandOptionType.NoValue);
+
+            cmd.OnExecute(() =>
+            {
+                var msbuildArgs = new List<string>()
+                {
+                     "/t:Restore"
+                };
+
+                if (sourceOption.HasValue())
+                {
+                    msbuildArgs.Add($"/p:RestoreSources={string.Join(";", sourceOption.Values)}");
+                }
+
+                if (packagesOption.HasValue())
+                {
+                    msbuildArgs.Add($"/p:RestorePackagesPath={packagesOption.Value()}");
+                }
+
+                if (disableParallelOption.HasValue())
+                {
+                    msbuildArgs.Add($"/p:RestoreDisableParallel=true");
+                }
+
+                if (configFileOption.HasValue())
+                {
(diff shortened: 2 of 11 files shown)


emgarten:
//cc @rrelyea

jasonwilliams200OK:
> This branch has conflicts that must be resolved

emgarten:
//cc @eerhardt @piotrpMSFT @rohit21agrawal

rohit21agrawal:
LGTM. :shipit:

src/dotnet/project.json
```diff
@@ -26,6 +26,8 @@
     }
   },
   "dependencies": {
+    "NuGet.CommandLine.XPlat": "3.6.0-beta.1.msbuild.1",
```
eerhardt: I don't think these references are needed anymore right?  We never actually load these assemblies into our process.  Instead we invoke MSBuild /t:Restore in a separate process.

src/dotnet/commands/dotnet-restore3/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.DotNet.Tools.Restore3
+{
+    public partial class Restore3Command
+    {
+        public static int Run(string[] args)
+        {
+            DebugHelper.HandleDebugSwitch(ref args);
+
+            return Restore3CommandApp.Execute(args);
```
eerhardt: Typically, the arg definitions happen in this Program.cs file, and the "logic" happens in the other file.  See dotnet-publish or dotnet-run for an example.

That being said, there really isn't "logic" in this command.  What do you think about collapsing these two files into a single file?  I think it would be easier to read.  We can spit them later, if more logic gets added.

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
```diff
@@ -0,0 +1,104 @@
...
+            CommandLineApplication cmd = new CommandLineApplication
+            {
+                Name = "restore3",
+                FullName = "restore3",
+                Description = "restore for msbuild"
+            };
+
+            cmd.HelpOption("-h|--help");
+
+            var argRoot = cmd.Argument(
```
eerhardt: Do we want this, or do we want to support 'pass through' args?  like "dotnet run" does.  Basically allow the user to pass any arg through to the MSBuild process.

To allow this, you need to set `throwOnUnexpectedArg` = false, and grab the RemainingArguments from the CommandLineApplication.

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
```diff
@@ -0,0 +1,104 @@
...
+                    "[root]",
+                    "Optional path to a project file or MSBuild arguments.",
+                    multipleValues: true);
+
+            var sourceOption = cmd.Option(
```
eerhardt: I don't see `-f|--fallbacksource <FEED>` as an option.  Is that still supported?

eerhardt: I also don't see the `-v|--verbosity` option.  Are we expecting to use msbuild's verbosity?

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                if (ignoreFailedSourcesOption.HasValue())
+                {
+                    msbuildArgs.Add($"/p:RestoreIgnoreFailedSources=true");
+                }
+
+                // Add any additional values passed
+                msbuildArgs.AddRange(argRoot.Values);
+
+                return new MSBuildForwardingApp(msbuildArgs.ToArray()).Execute();
```
eerhardt: There's no need that MSBuildForwardingApp strictly needs a `string[]`.  We should change that class to take an IEnumerable<string> instead, and then get rid of the ToArray() call here.

eerhardt:
:clock1:

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
emgarten: FallbackSource has been removed, I know it's used some places in the CLI build still. Is there a good reason to keep this instead of just having the user specify all sources with ``-s``?

``--verbosity`` ended up being a converter from the nuget level names to the msbuild level names, but going forward it seemed really unnecessary. Going forward I think this command should use whatever verbosity setting the rest of the commands use, and for now ``/v:`` lets you set it directly on msbuild, which controls the nuget restore output.

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
emgarten: That works today because the msbuild args start with ``/`` and this parser looks for ``-``. Unless you can also pass ``-`` to msbuild, is that possible?

I don't see a way to validate unexpected args in msbuild, so if we set this to false there wouldn't be any help.

src/dotnet/commands/dotnet-restore3/Program.cs
emgarten: I can do that, I was just trying to follow the pattern of the other commands.

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
eerhardt: > FallbackSource has been removed, I know it's used some places in the CLI build still. Is there a good reason to keep this instead of just having the user specify all sources with -s?

According to https://github.com/NuGet/Home/issues/2477, `-s` will completely override what is in NuGet.config.  This seems limiting since in a lot of cases you only want to add a source for this specific invocation.  Not completely overriding NuGet.config.  But there are other cases where you DO want to override.

Will both scenarios be supported by `-s`?

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
eerhardt: `dotnet build3 -v:diag` works today.  MSBuild knows how to look for `/` and for `-`.

I would assume MSBuild would validate the unexpected args.

src/dotnet/project.json
emgarten: The non-msbuild based restore command still exists to allow restoring xproj, this reference is needed for that.

eerhardt: @piotrpMSFT just removed this reference and changed the non-msbuild based restore command to call NuGet in a separate process.  See https://github.com/dotnet/cli/pull/4024

emgarten:
Updated

src/dotnet/project.json
emgarten: I'll remove them and see what is needed.

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
emgarten: ``-s`` just overrides, I think we need to look at fallback source, that scenario can be useful although it is rare, and at minimum it needs a new name to help with #2477. I'm not willing to add it back under the fallback name here since this is a new command and it should stay as clean as possible.

src/dotnet/project.json
eerhardt: Were you able to remove these?  I still can't see where they are necessary.

---
In reply to: [74501423](https://github.com/dotnet/cli/pull/4035#discussion_r74501423) [](ancestors = 74501423)

src/dotnet/commands/dotnet-restore3/Restore3CommandApp.cs
emgarten: Sounds good, I'll set ``throwOnUnexpectedArg`` to false

eerhardt:
:shipit:  once addressing the comment about needing the NuGet dependencies on src\dotnet\project.json

emgarten:
@eerhardt removing the dependencies from *src\dotnet\project.json* appears to cause several problems, the targets file doesn't get deployed and there is a compile error on a reference to a NuGet assembly. The dotnet project itself restores and builds fine, but there must be some *nearest package* changes that end up bringing in the wrong bits.

build_projects/dotnet-cli-build/project.json
```diff
@@ -20,7 +20,8 @@
     "System.Runtime.Serialization.Primitives": "4.1.1",
     "System.Xml.XmlSerializer": "4.0.11",
     "WindowsAzure.Storage": "6.2.2-preview",
-    "NuGet.CommandLine.XPlat": "3.5.0-rc1-1697",
+    "NuGet.Build.Tasks": "3.6.0-beta.1.msbuild.1",
```
eerhardt: Our private build tasks project should need to reference `NuGet.Build.Tasks`.  Why is this needed?

eerhardt:
> the targets file doesn't get deployed

Ah, right. I overlooked that part.  I think it would be better if the `NuGet.Build.Tasks` dependency was brought in by either https://github.com/dotnet/cli/blob/feature/msbuild/src/redist/project.json or https://github.com/dotnet/cli/blob/feature/msbuild/src/tool_nuget/project.json.  My thought is the 1st one.

/cc @piotrpMSFT, who added these new projects to get his opinion on which would be better.

eerhardt:
> there is a compile error on a reference to a NuGet assembly. The dotnet project itself restores and builds fine, but there must be some nearest package changes that end up bringing in the wrong bits.

I think that's because a new NuGet dependency was added in the latest bits that you haven't updated:

See https://github.com/dotnet/cli/blob/feature/msbuild/src/tool_nuget/project.json#L11

emgarten:
>  I think it would be better if the NuGet.Build.Tasks dependency was brought in by either https://github.com/dotnet/cli/blob/feature/msbuild/src/redist/project.json 

I'll try putting it there

> I think that's because a new NuGet dependency was added in the latest bits that you haven't updated:

That could be the problem, not sure how find and replace wasn't catching that one

emgarten:
Thanks for the pointers @eerhardt updating and moving the references solve the issue. I've checked this in, if the build task needs to go under another project I can move it.

PR #4038, opened 2016-08-11
Fix CI by not copying Microsoft.DotNet.PlatformAbstractions to stage0 since it is already loaded into the stage0 dotnet process.

eerhardt:
Eventually we should also remove the copying of WindowsAzure.Storage.dll since the MSBuild bug is fixed. But since Publish isn't tested in Jenkins, I didn't want to do that here.  I just want to get our builds off the floor.

@livarcocc @brthor @piotrpMSFT


eerhardt:
Merging this to unblock the CI.  Please take a look, and I'll fix any comments in another PR.

PR #4079, opened 2016-08-23
Update the Sdk version to the latest build.

eerhardt:
@brthor @livarcocc @Sridhar-MS @piotrpMSFT


brthor:
:shipit:

brthor:
Actually should we just wait on the rename to bump this version?

eerhardt:
> Actually should we just wait on the rename to bump this version?

Doesn't hurt to do it again... does it?

I mostly wanted to ensure nothing was broken with the latest.  So getting the latest in earlier seems better to me.

brthor:
sounds good

PR #4130, opened 2016-08-31
Update version of routing to 1.0.1

rynowak:
This is needed because MVC still references 1.0.0.


brthor:
feature/antares is an obsolete branch now. 

Can you make this PR directly to rel/1.0.0-preview2 or push to Sridhar's branch?

rynowak:
I pushed this instead of preview2-update

PR #4136, opened 2016-09-01
Set the "hashPath" when building the .deps.json file

joelverhagen:
1. Set the "hashPath" when building the .deps.json file
1. Use newer versions of `PlatformAbstractions`, `DependencyModel`, and `dotnet-test-xunit`

This is progress for this work item: https://github.com/NuGet/Home/issues/2522.

/cc @eerhardt @livarcocc @piotrpMSFT @gkhanna79 @emgarten

### DIFF
--- a/src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
+++ b/src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
@@ -147,7 +147,8 @@ namespace Microsoft.Extensions.DependencyModel
                     export.ResourceAssemblies.Select(CreateResourceAssembly),
                     libraryDependencies,
                     serviceable,
-                    GetLibraryPath(export.Library));
+                    GetLibraryPath(export.Library),
+                    GetLibraryHashPath(export.Library));
             }
             else
             {
@@ -169,7 +170,8 @@ namespace Microsoft.Extensions.DependencyModel
                     assemblies,
                     libraryDependencies,
                     serviceable,
-                    GetLibraryPath(export.Library));
+                    GetLibraryPath(export.Library),
+                    GetLibraryHashPath(export.Library));
             }
         }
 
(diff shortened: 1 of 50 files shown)


emgarten:
LGTM :shipit:

joelverhagen:

src/Microsoft.DotNet.ProjectModel/Resolution/PackageDependencyProvider.cs
```diff
@@ -17,13 +17,15 @@ namespace Microsoft.DotNet.ProjectModel.Resolution
...
         private readonly FallbackPackagePathResolver _packagePathResolver;
+        private readonly VersionFolderPathResolver _versionFolderPathResolver;
         private readonly FrameworkReferenceResolver _frameworkReferenceResolver;
 
         public PackageDependencyProvider(INuGetPathContext nugetPathContext, FrameworkReferenceResolver frameworkReferenceResolver)
         {
             if (nugetPathContext != null)
             {
                 _packagePathResolver = new FallbackPackagePathResolver(nugetPathContext);
+                _versionFolderPathResolver = new VersionFolderPathResolver(nugetPathContext.UserPackageFolder);
```
eerhardt: This looks odd to me.  Why does `VersionFolderPathResolver` need the user package folder?  If it cares about that, shouldn't it care about any other "fallback" folder?

src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
```diff
@@ -180,16 +182,27 @@ private string GetLibraryPath(LibraryDescription description)
...
             return null;
         }
 
+        private string GetLibraryHashPath(LibraryDescription description)
+        {
+            var packageDescription = description as PackageDescription;
+
+            if (packageDescription != null)
+            {
+                // This is hash path appended to the package path (much like package assets). This string should
```
eerhardt: (nit) small wording error:  "This string should mastered only by NuGet"

eerhardt:
:shipit:

src/Microsoft.DotNet.ProjectModel/Resolution/PackageDependencyProvider.cs
joelverhagen: Good point. I will pass a null path since this resolver is only used for building filenames.

src/Microsoft.DotNet.ProjectModel/DependencyContextBuilder.cs
joelverhagen: Fixed.

PR #4155, opened 2016-09-09
Ensure the project is built in `dotnet run3`.

eerhardt:
I also made a few minor clean ups along the way:
- Updating to a new SDK, so you can build twice in a row without deleting your obj folder.
- Fixing a new .xproj so you can build the CLI in VS.
- Fixing the error handling for run3.

This is part of https://github.com/dotnet/sdk/issues/118 "Implement dotnet run".

@brthor @Sridhar-MS @livarcocc


src/Microsoft.DotNet.Tools.Test/Microsoft.DotNet.Tools.Test.xproj
```diff
@@ -5,17 +5,16 @@
     <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
   </PropertyGroup>
 
-  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.Props" Condition="'$(VSToolsPath)' != ''"  />
+  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
```
nil4: Was the switch back to DNX here and below intended?

eerhardt: The intention is to match all the other xproj's in the repo.  I spot checked a few of them, and I only saw DNX being used, so that's what I used here.

I opened https://github.com/dotnet/cli/issues/4156 for fixing this everywhere.

eerhardt:

src/Microsoft.DotNet.Tools.Test/Microsoft.DotNet.Tools.Test.xproj
brthor: Had some issues using the dotnet targets as well.

src/dotnet/Program.cs
```diff
@@ -74,6 +74,22 @@ public static int Main(string[] args)
 
                 return 1;
             }
+            catch (Exception ex)
+            {
+#if DEBUG
```
brthor: 👍 A much needed behavior, didn't we have this before at some point?

eerhardt: We have it duplicated in all the individual commands' "Execute" methods, like it was for `dotnet run3`

src/dotnet/commands/dotnet-run3/Run3Command.cs
```diff
@@ -49,6 +52,27 @@ public int Start()
...
+
+            buildArgs.Add("/nologo");
+            buildArgs.Add("/verbosity:quiet");
+
+            if (!string.IsNullOrWhiteSpace(Configuration))
+            {
+                buildArgs.Add($"/p:Configuration={Configuration}");
+            }
+
+            var buildResult = new MSBuildForwardingApp(buildArgs).Execute();
```
brthor: I'm surprised to see this using MSBuild directly rather than through build3. 

I can't think of a downside at the moment, but what if we introduce some additional logic in build3?

brthor:
:shipit:

src/dotnet/commands/dotnet-run3/Run3Command.cs
eerhardt: We do the same thing for invoking restore3 and pack3:

https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-restore3/Program.cs#L105
https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-pack3/Pack3Command.cs#L81

PR #4162, opened 2016-09-12
Update MSBuild to 0.1.0-preview-00034-160909.

eerhardt:
I removed some redundant dependencies now that MSBuild's packages have their dependencies declared correctly.

@brthor @Sridhar-MS


eerhardt:
Merging to get the new MSBuild.

PR #4174, opened 2016-09-13
Update the CLI to the new Run static properties from the SDK.

eerhardt:
See https://github.com/dotnet/sdk/issues/98

### DIFF
--- a/src/dotnet/commands/dotnet-run3/Run3Command.cs
+++ b/src/dotnet/commands/dotnet-run3/Run3Command.cs
@@ -6,32 +6,25 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using Microsoft.Build.Execution;
-using Microsoft.Build.Framework;
 using Microsoft.DotNet.Cli;
 using Microsoft.DotNet.Cli.Utils;
-using NuGet.Frameworks;
 
 namespace Microsoft.DotNet.Tools.Run
 {
     public partial class Run3Command
     {
-        private const string GetRunInformationTaskName = "GetRunInformation";
-
+        private const string RunCommandPropName = "RunCommand";
+        private const string RunArgumentsPropName = "RunArguments";
+        private const string RunWorkingDirectoryPropName = "RunWorkingDirectory";
+        
         public string Configuration { get; set; }
         public string Project { get; set; }
         public IReadOnlyList<string> Args { get; set; }
 
-        private readonly ICommandFactory _commandFactory;
         private List<string> _args;
 
         public Run3Command()
-            : this(new RunCommandFactory())
         {
-        }
-
-        public Run3Command(ICommandFactory commandFactory)
-        {
-            _commandFactory = commandFactory;
         }
 
         public int Start()
@@ -40,14 +33,9 @@ namespace Microsoft.DotNet.Tools.Run
 
             EnsureProjectIsBuilt();
 
-            ITaskItem runInfoItem = GetRunInformation();
+            ICommand runCommand = GetRunCommand();
 
-            string commandName = runInfoItem.GetMetadata("CommandName");
-            string[] args = runInfoItem.GetMetadata("Args").Split(';');
-
-            ICommand command = _commandFactory.Create(commandName, Enumerable.Concat(args, _args));
-
-            return command
+            return runCommand
                 .Execute()
                 .ExitCode;
         }
@@ -73,7 +61,7 @@ namespace Microsoft.DotNet.Tools.Run
             }
         }
 
-        private ITaskItem GetRunInformation()
+        private ICommand GetRunCommand()
         {
             Dictionary<string, string> globalProperties = new Dictionary<string, string>()
             {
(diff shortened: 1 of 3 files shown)


eerhardt:
@brthor @Sridhar-MS

src/dotnet/commands/dotnet-run3/Run3Command.cs
```diff
@@ -87,30 +75,25 @@ private ITaskItem GetRunInformation()
...
-
-            BuildResult result = BuildManager.DefaultBuildManager.Build(buildParameters, buildRequestData);
-
-            TargetResult runInfoResult;
-            if (!result.ResultsByTarget.TryGetValue(GetRunInformationTaskName, out runInfoResult))
+            string runProgram = projectInstance.GetPropertyValue(RunCommandPropName);
+            if (string.IsNullOrEmpty(runProgram))
             {
-                throw new InvalidOperationException($"Could not find a target named '{GetRunInformationTaskName}' in your project. Please ensure 'dotnet run' supports this project.");
+                throw new InvalidOperationException($"The property named '{RunCommandPropName}' does not have a value in your project. Please ensure 'dotnet run' supports this project.");
```
eerhardt: Maybe I should check the `$(OutputType)` here and throw a decent error message saying that the $(OutputType) should be "Exe"?  Or not "Library"...  Or say "this is the output type: {OutputType}, ensure this is a runnable output type".

Thoughts?

src/dotnet/commands/dotnet-run3/Run3Command.cs
```diff
@@ -87,30 +75,25 @@ private ITaskItem GetRunInformation()
...
-            {
-                throw new InvalidOperationException($"Could not get the run information for project {Project}. An internal MSBuild error has occured" + Environment.NewLine +
-                    runInfoResult.Exception?.ToString());
-            }
+            string runArguments = projectInstance.GetPropertyValue(RunArgumentsPropName);
+            string runWorkingDirectory = projectInstance.GetPropertyValue(RunWorkingDirectoryPropName);
 
-            ITaskItem runInfoItem = runInfoResult.Items.FirstOrDefault(i => i.ItemSpec == projectInstance.FullPath);
-            if (runInfoItem == null)
+            string fullArguments = runArguments;
```
brthor: Are we expecting the user to escape these arguments themselves?

src/dotnet/commands/dotnet-run3/Run3Command.cs
brthor: Yeah I think that definitely makes sense
It'll prevent trivial github issues if nothing else.

src/dotnet/commands/dotnet-run3/Run3Command.cs
brthor: I think this is going to give us some trouble with escaping if it's widely used.

We'll be requiring users to know about how escaping works to do this right, things like you can't have a trailing backslash in your path if it's surrounded by quotes are non-obvious and cause user confusion frequently. 

What if the run arguments were a set of items that we could pass through the argument escaper?

src/dotnet/commands/dotnet-run3/Run3Command.cs
eerhardt: I've updated the error message to include the OutputType.

src/dotnet/commands/dotnet-run3/Run3Command.cs
eerhardt: We discussed this offline, and since the existing `$(StartArguments)` property doesn't have a separator between arguments other than a space.  Trying to parse a command-line string ourselves into separate args, and then put them back into a command-line string doesn't seem like a good approach.

So we will just take the string as it is given to us by MSBuild.  This behavior appears to be consistent with VS F5 - if you have a trailing backslash followed by a double quote, you get a double quote in your args.

src/dotnet/commands/dotnet-run3/Run3Command.cs
```diff
@@ -87,30 +71,30 @@ private ITaskItem GetRunInformation()
...
+                string outputType = projectInstance.GetPropertyValue("OutputType");
 
-            if (runInfoResult.ResultCode != TargetResultCode.Success)
-            {
-                throw new InvalidOperationException($"Could not get the run information for project {Project}. An internal MSBuild error has occured" + Environment.NewLine +
-                    runInfoResult.Exception?.ToString());
+                throw new GracefulException(string.Join(Environment.NewLine,
+                    "Unable to run your project.",
+                    "Please ensure you have a runnable project type and ensure 'dotnet run' supports this project.",
+                    $"The current OutputType is '{outputType}'."));
```
brthor: Like this syntax for join 👍

src/dotnet/commands/dotnet-run3/Run3Command.cs
```diff
@@ -87,30 +71,30 @@ private ITaskItem GetRunInformation()
...
+
+            string fullArguments = runArguments;
+            if (_args.Any())
             {
-                throw new InvalidOperationException($"'{GetRunInformationTaskName}' did not return an ITaskItem with the project's FullPath as the ItemSpec.");
+                fullArguments += " " + ArgumentEscaper.EscapeAndConcatenateArgArrayForProcessStart(_args);
             }
 
-            return runInfoItem;
+            CommandSpec commandSpec = new CommandSpec(runProgram, fullArguments, CommandResolutionStrategy.None);
```
brthor: We need an msbuild command resolution strategy, maybe not in this PR though

src/dotnet/commands/dotnet-run3/Run3Command.cs
```diff
@@ -87,30 +71,30 @@ private ITaskItem GetRunInformation()
...
+                throw new GracefulException(string.Join(Environment.NewLine,
+                    "Unable to run your project.",
+                    "Please ensure you have a runnable project type and ensure 'dotnet run' supports this project.",
+                    $"The current OutputType is '{outputType}'."));
             }
 
-            ITaskItem runInfoItem = runInfoResult.Items.FirstOrDefault(i => i.ItemSpec == projectInstance.FullPath);
-            if (runInfoItem == null)
+            string runArguments = projectInstance.GetPropertyValue("RunArguments");
+            string runWorkingDirectory = projectInstance.GetPropertyValue("RunWorkingDirectory");
```
brthor: What happens if these aren't set?

eerhardt: If these aren't set, it works as expected:

1. no args are passed into the command.
2. The "working directory" will be empty string - which means use the current working directory.  See https://msdn.microsoft.com/en-us/library/system.diagnostics.processstartinfo.workingdirectory(v=vs.110).aspx

src/dotnet/commands/dotnet-run3/Run3Command.cs
eerhardt: I don't think there is a "command resolution strategy" for this scenario at all.  The idea being that the MSBuild "Run" properties should tell the caller (dotnet run here) exactly what they should do to run the project.  There should be no need to resolve anything, that way the caller can be "dumb".  Also, that way no matter who the caller is (VS, dotnet run, VS Code, Project Rider, etc.) the project is run the exact same way.

PR #4184, opened 2016-09-15
Update msbuild

nguerrera:
Update msbuild to latest to make the common portion of cross-targeting targets available.

@brthor @livarcocc @eerhardt


brthor:
@333fred was going to submit a PR for moving over to the new msbuild feed, should we wait for that (not sure what the timeline is)

eerhardt:
> should we wait for that

No.  These are packages @nguerrera needs now.  @333fred can move to the new msbuild feed after this, whenever he likes.

333fred:
@brthor @eerhardt I'm still trying to resolve build issues with the new packages. I think I'll likely wait to upload changes until they're ready to open up the msbuild catalog.

PR #4192, opened 2016-09-16
CrossGen command, working code with some details pending

shhsu:
## Feature:

Given a published dotnet application, this tool can generate crossgen'd assemblies and arrange them in on of the two following folder structure

1. App - Make a copy of the application except all dotnet assemblies are crossgen'd
2. Optimization Cache - Populate into a structure that can be used as DOTNET_HOSTING_OPTIMIZATION_CACHE. (This feature is incomplete, right now it outputs the crossgen'd assemblies and sha file only. The rest of the structure still needs to be populated.)

Note that because this is a WIP feature, currently the user have to specify where crossgen.exe is

### How was runtime environments determined?
**Self contained:** deps.json provides RID
**Portable:** RID would be determined by detecting the RID of the current CLI. Framework would be determined by going into the shared framework directory and auto-roll-forward if necessary.

### Pending issues

1. This app does not currently generate complete DOTNET_HOSTING_OPTIMIZATION_CACHE hive. Only the crossgen'd assembly and sha file.

2. How important is the GeneratePDB option? Currently it only supports crossgen.exe version 1.1 and up. I can add logic to support GeneratePDB pre-1.1 if required but it would be a bit more messy.

3. We don't know how crossgen.exe would be shipped. Currently the command requires the location of crossgen.exe to be passed in.

### DIFF
--- a/src/dotnet/commands/dotnet-crossgen/CrossGenContext.cs
+++ b/src/dotnet/commands/dotnet-crossgen/CrossGenContext.cs
@@ -1,0 +1,244 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.PlatformAbstractions;
+using Microsoft.DotNet.ProjectModel;
+using Microsoft.DotNet.Tools.CrossGen.Outputs;
+using Microsoft.Extensions.DependencyModel;
+using NuGet.Frameworks;
+using NuGet.Versioning;
+
+using static Microsoft.DotNet.Tools.CrossGen.Operations.FileNameConstants;
+
+namespace Microsoft.DotNet.Tools.CrossGen
+{
+    public class CrossGenContext
+    {
+        private readonly string _appName;
+        private readonly string _appDir;
+        private readonly bool _generatePDB;
+        private CrossGenTarget _crossGenTarget;
+
+        /// <summary>
+        /// Context taken straight out of the app's deps file,
+        /// This context determine what assets to crossgen
+        /// </summary>
+        private DependencyContext _depsFileContext;
+
+        /// <summary>
+        /// Context that is merged with the shared framework (if applicable)
+        /// This context is the actual context used on run time and is
+        /// used to determine runtime fallbacks
+        /// </summary>
+        private DependencyContext _runtimeContext;
+        
+        public CrossGenContext(string appName, string appDir, bool generatePDB)
+        {
+            _appName = appName;
+            _appDir = appDir;
+            _generatePDB = generatePDB;
+        }
+
+        public void Initialize()
+        {
+            var depsFilePath = Path.Combine(_appDir, $"{_appName}.deps.json");
+            if (!File.Exists(depsFilePath))
+            {
+                throw new CrossGenException($"Deps {depsFilePath} file not found");
+            }
+
+            string sharedFrameworkDir;
+            NuGetFramework framework;
+            string rid;
+            using (var reader = new DependencyContextJsonReader())
+            {
+                using (var fstream = new FileStream(depsFilePath, FileMode.Open))
+                {
+                    _depsFileContext = reader.Read(fstream);
+                }
+
+                if (_depsFileContext == null)
+                {
+                    throw new CrossGenException($"Unexpected error while reading {depsFilePath}");
+                }
+
+                var runtimeConfigPath = Path.Combine(_appDir, $"{_appName}.runtimeconfig.json");
+                RuntimeConfig runtimeConfig = null;
+                if (File.Exists(runtimeConfigPath))
+                {
+                    runtimeConfig = new RuntimeConfig(runtimeConfigPath);
+                }
+
+                if (runtimeConfig != null && runtimeConfig.IsPortable)
+                {
+                    // This is portable app
+                    Reporter.Verbose.WriteLine($"This is a portable app, runtime config file: {runtimeConfigPath}");
+                    sharedFrameworkDir = LocateSharedFramework(runtimeConfig.Framework);
+                    var shreadFrameworkDepsFile = Path.Combine(sharedFrameworkDir, $"{runtimeConfig.Framework.Name}.deps.json");
+                    if (!File.Exists(shreadFrameworkDepsFile))
+                    {
+                        throw new CrossGenException($"Cannot locate share framework's deps file {shreadFrameworkDepsFile}");
+                    }
+
+                    using (var fstream = new FileStream(shreadFrameworkDepsFile, FileMode.Open))
+                    {
+                        _runtimeContext = reader.Read(fstream);
+                    }
+
+                    // After merging, framework and rid would be gone. So do this pre-merge.
+                    framework = NuGetFramework.Parse(_runtimeContext.Target.Framework);
+                    rid = _runtimeContext.Target.Runtime;
+
+                    if (_runtimeContext == null)
+                    {
+                        throw new CrossGenException($"Unable to load shared framework context from {shreadFrameworkDepsFile}");
+                    }
+
+                    _runtimeContext = _depsFileContext.Merge(_runtimeContext);
+                }
+                else
+                {
+                    Reporter.Verbose.WriteLine($"This is a standalone app, runtime config file: {(runtimeConfig == null ? runtimeConfigPath : "None")}");
+                    sharedFrameworkDir = null;
+                    _runtimeContext = _depsFileContext;
+                    framework = NuGetFramework.Parse(_runtimeContext.Target.Framework);
+                    rid = _runtimeContext.Target.Runtime;
+                }
+            }
+
+            if (framework.Framework != ".NETCoreApp")
+            {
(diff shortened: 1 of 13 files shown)


shhsu:
Including all stakeholders @piotrpMSFT @ericeil @brthor @livarcocc @JohnChen0 @gkhanna79 @glennc @sajaya @rynowak

ericeil:
@shhsu, did you mean to mention someone else?  I'm not familiar with this work at all....

src/dotnet/commands/dotnet-crossgen/Operations/CrossGenCmdUtil.cs
```diff
@@ -0,0 +1,180 @@
...
+            var outputFiles = new List<string>();
+
+            var fileName = Path.GetFileName(assemblyLocation);
+            var module = fileName.Substring(0, fileName.Length - 4);
+            var crossgenArgs = string.Join(" ", GetArgs(appPath, assemblyLocation, module, platformAssembliesPaths, outputDirectory, false));
+            var cmd = Command.Create(new CommandSpec(_crossgenPath, crossgenArgs, CommandResolutionStrategy.None))
+                .WorkingDirectory(appPath)
+                // disable partial ngen
+                .EnvironmentVariable("COMPlus_PartialNGen", "0")
+                .CaptureStdOut()
```
eerhardt: Why are we capturing the std out here and below?  Why not just let the output flow up?

src/dotnet/commands/dotnet-crossgen/Exceptions/CrossGenException.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Microsoft.DotNet.Tools.CrossGen.Exceptions
+{
+    public class CrossGenException : Exception
+    {
+        public CrossGenException(string msg, Exception cause = null)
```
eerhardt: usually "cause" is called "innerException".

shhsu: fixed

src/dotnet/commands/dotnet-crossgen/Operations/CrossGenCmdUtil.cs
shhsu: It might be a bit verbose. I choose to only write it out if verbose is on if that's ok.

src/dotnet/commands/dotnet-crossgen/CrossGenContext.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            var targetVersion = NuGetVersion.Parse(version);
+            var candidateNames = Directory.GetDirectories(root).Select(d => Path.GetFileName(d));
+
+            string bestMatch = null;
+            NuGetVersion bestMatchVersion = null;
+            foreach (var candidateName in candidateNames)
+            {
+                var currentVersion = NuGetVersion.Parse(candidateName);
+                if (currentVersion.Major == targetVersion.Major && currentVersion.CompareTo(targetVersion) > 0)
```
eerhardt: The dotnet host only rolls forward on patch versions (Major.minor.patch).  Do we want to keep the same behavior in crossgen?  Or are we saying that any major versions of the shared framework will be compatible with crossgen?

/cc @gkhanna79

src/dotnet/commands/dotnet-crossgen/CrossGenTarget.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using NuGet.Frameworks;
+
+namespace Microsoft.DotNet.Tools.CrossGen
+{
+    public class CrossGenTarget
+    {
+        public NuGetFramework Framework { get; private set; }
+        public string RID { get; private set; }
```
eerhardt: (minor) Should be cased as "Rid".

eerhardt: Or "RuntimeIdentifier"

---
In reply to: [80278936](https://github.com/dotnet/cli/pull/4192#discussion_r80278936) [](ancestors = 80278936)

src/dotnet/commands/dotnet-crossgen/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            var outputDirParam = app.Option("--outputDir <DIR>", "(Optional) Directory to write the output cache to, default is DOTNET_HOSTING_OPTIMIZATION_CACHE", CommandOptionType.SingleValue);
+            var outputStructureParam = app.Option("--output-structure", "(Optional) Structure of the output. \"APP\" - Default option, obtain app's directory structure; \"CACHE\" - Arrange the crossgen'd assemblies in the structure of the optimization cache.", CommandOptionType.SingleValue);
+            var crossGenExeLocationParam = app.Option("--crossgenExe <FILE>", "(Optional) Location of crossgen executable.", CommandOptionType.SingleValue);
+            var generatePDBParam = app.Option("--generatePDB", "(Optional) Option to generate PDB", CommandOptionType.NoValue);
+            var diaSymReaderLocationParam = app.Option("--diasymreader <FILE>", "(Optional) Location of diasymreader", CommandOptionType.SingleValue);
+            var overwriteHashParam = app.Option("--overwrite-hash", "(Optional) If package hash value conflicts with existing cache, the program will exit unless this option is given.", CommandOptionType.NoValue);
```
eerhardt: I'm not sure I understand this option completely.  What is the scenario for it?

src/dotnet/commands/dotnet-crossgen/Operations/CrossGenCmdUtil.cs
```diff
@@ -0,0 +1,180 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.CrossGen.Exceptions;
+
+/// <summary>
+/// This is a wrapper of crossgen.exe native image generation utiltites
```
eerhardt: This summary comment should go on the class, not the namespace.

src/dotnet/commands/dotnet-crossgen/Exceptions/CrossGenException.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Microsoft.DotNet.Tools.CrossGen.Exceptions
```
eerhardt: It seems odd to me to put this single class in its own namespace.  Can it just go in `Microsoft.DotNet.Tools.CrossGen`?

src/dotnet/commands/dotnet-crossgen/Operations/NativeImageType.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DotNet.Tools.CrossGen.Operations
+{
+    public enum NativeImageType
+    {
+        Ready2Run, FragileNonVersionable
```
eerhardt: (minor) normally these values are on their own lines, for readability

src/dotnet/commands/dotnet-crossgen/CrossGenContext.cs
shhsu: Let's be safe and keep minor version as well

src/dotnet/commands/dotnet-crossgen/CrossGenTarget.cs
shhsu: Renamed the property

src/dotnet/commands/dotnet-crossgen/Exceptions/CrossGenException.cs
shhsu: Sure

src/dotnet/commands/dotnet-crossgen/Operations/CrossGenCmdUtil.cs
shhsu: Moved

src/dotnet/commands/dotnet-crossgen/Operations/NativeImageType.cs
shhsu: Done

src/dotnet/commands/dotnet-crossgen/Program.cs
shhsu: This is for the cache priming scenario. Say that if the hash value for a package on the deps file differs from the current hash value on the current optimization cache, the desired behavior should be to bail out of the crossgen process because there's something wrong with the assembly. If we overwrite this package cache, all other apps wouldn't be able to use this package's cache. (Come to think of it, maybe I should have left the hive entirely untouched. Currently the behavior is that every packages visited before the failed hash check would still be updated.)

What I had in mind is that if the customer want to choose to overwrite the entire package if there's a hash conflict. They can do so.

shhsu:
As discussed with @piotrpMSFT, closing PR for now until we figure out what to do

PR #4193, opened 2016-09-16
add includesymbols and includesource to pack3 command as symbols package is not created by default

rohit21agrawal:
This PR adds --include-symbols and --include-source command line option to  the pack3 command. pack3 does not build symbols package by default and these commands and provide a way to control what goes into the packages.

CC: @rrelyea @joelverhagen @eerhardt @livarcocc

### DIFF
--- a/src/dotnet/commands/dotnet-pack3/Pack3Command.cs
+++ b/src/dotnet/commands/dotnet-pack3/Pack3Command.cs
@@ -30,6 +30,12 @@ namespace Microsoft.DotNet.Tools.Pack3
             var noBuild = cmd.Option("--no-build",
                 "Do not build project before packing", 
                 CommandOptionType.NoValue);
+            var includeSymbols = cmd.Option("--include-symbols",
+                "Include PDBs along with the DLLs in the output folder",
+                CommandOptionType.NoValue);
+            var includeSource = cmd.Option("--include-source",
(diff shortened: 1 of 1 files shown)


rohit21agrawal:
@eerhardt how do i get dotnet bot to start the one failing test again?

rohit21agrawal:

src/dotnet/commands/dotnet-pack3/Pack3Command.cs
```diff
@@ -30,6 +30,12 @@ public static int Run(string[] args)
             var noBuild = cmd.Option("--no-build",
                 "Do not build project before packing", 
                 CommandOptionType.NoValue);
+            var includeSymbols = cmd.Option("--include-symbols",
+                "Include PDBs along with the DLLs in the output folder",
+                CommandOptionType.NoValue);
+            var includeSource = cmd.Option("--include-source",
+                "Include PDBs and source files. Source files go into the src folder in the resulting nuget package",
```
eerhardt: Is this description correct?  Why would `--include-source` mean that I should include PDBs?  Why not have the user specify both `--include-symbols` and `--include-source` if they want both?

rohit21agrawal: @eerhardt does it make sense to have source files without PDBs ?
@yishaigalatzer thoughts?

PR #4200, opened 2016-09-19
Fix the VC Redist links in the installer.

Sridhar-MS:
The windows installer downloads the VC redist bundles as a remote payload from a URL. This URL owned by the VC++ team recently changed to point to updated CRTs. This broke our windows installers. So we decided to upload the CRTs to a URL owned by the .Net Core team and use that for chaining the VC Redist bundle.

cc @ellismg @leecow @livarcocc


ellismg:
LGTM.

Sridhar-MS:

PR #4234, opened 2016-09-23
Support building with custom container name

naamunds:
Currently, using a custom container name is supported for only some
publish steps.  This commit makes it supported for all publish steps.

This commit also ensures that the Overwrite property is specified on all
calls to the UploadToAzure task.

cc: @livarcocc @brthor

### DIFF
--- a/build/Microsoft.DotNet.Cli.Publish.targets
+++ b/build/Microsoft.DotNet.Cli.Publish.targets
@@ -97,11 +97,13 @@
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
       ContainerName="$(ContainerName)"
-      Items="@(CliVersionBadgeToUpload)" />
+      Items="@(CliVersionBadgeToUpload)"
+      Overwrite="$(OverwriteOnPublish)" />
 
     <SetBlobPropertiesBasedOnFileType
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
+      ContainerName="$(ContainerName)"
       Items="@(CliVersionBadgeToUpload)" />
   </Target>
 </Project>
--- a/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
+++ b/build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
@@ -22,42 +22,44 @@ namespace Microsoft.DotNet.Cli.Build
             Sdk,
         }
 
-        private const string s_dotnetBlobRootUrl = "https://dotnetcli.blob.core.windows.net/" + s_dotnetBlobContainerName;
         private const string s_dotnetBlobContainerName = "dotnet";
 
         private string _connectionString { get; set; }
+        private string _containerName { get; set; }
         private CloudBlobContainer _blobContainer { get; set; }
 
-        public AzurePublisher()
+        public AzurePublisher(string containerName = s_dotnetBlobContainerName)
         {
             _connectionString = EnvVars.EnsureVariable("CONNECTION_STRING").Trim('"');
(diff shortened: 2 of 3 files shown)


build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
```diff
@@ -219,35 +221,9 @@ private void DeleteBlob(string path)
...
-        {
-            return $"{s_dotnetBlobRootUrl}/{CalculateRelativePathForFile(file, product, version)}";
-        }
-
         private static string CalculateRelativePathForFile(string file, Product product, string version)
         {
             return $"{product}/{version}/{Path.GetFileName(file)}";
         }
-
-        public static async Task DownloadFile(string blobFilePath, string localDownloadPath)
```
brthor: dead code?

build/Microsoft.DotNet.Cli.Publish.targets
```diff
@@ -97,11 +97,13 @@
...
       AccountName="$(CloudDropAccountName)"
       ContainerName="$(ContainerName)"
-      Items="@(CliVersionBadgeToUpload)" />
+      Items="@(CliVersionBadgeToUpload)"
+      Overwrite="$(OverwriteOnPublish)" />
 
     <SetBlobPropertiesBasedOnFileType
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
+      ContainerName="$(ContainerName)"
```
brthor: Where does this come from? The environment?

build/Microsoft.DotNet.Cli.Publish.targets
```diff
@@ -97,11 +97,13 @@
       AccountKey="$(CloudDropAccessToken)"
       AccountName="$(CloudDropAccountName)"
       ContainerName="$(ContainerName)"
-      Items="@(CliVersionBadgeToUpload)" />
+      Items="@(CliVersionBadgeToUpload)"
+      Overwrite="$(OverwriteOnPublish)" />
```
brthor: Same for overwrite on publish. Where does this come from? Is this something we need here?

build_projects/shared-build-targets-utils/Publishing/AzurePublisher.cs
naamunds: Yeah.  The methods `CalculateFullUrlForFile` and `DownloadFile` do not appear to be used anywhere.  It made sense to remove them in this PR because they are using a hard-coded "dotnet" container and would have to be refactored if we were to keep them.

brthor: sounds good

build/Microsoft.DotNet.Cli.Publish.targets
naamunds: Yes. It's already used elsewhere in this file and in PublishContent.targets.

naamunds: Because SetBlobPropertiesBasedOnFileType is hard-coded to "dotnet" without the changes in this PR, this task fails if ContainerName is set to something other than "dotnet", because the "dotnet" container won't exist or won't have the expected CLI build artifacts.

brthor: I see, just wanted to check. that makes sense

build/Microsoft.DotNet.Cli.Publish.targets
naamunds: It's also an environment variable, and it's currently used in the other call to the `UploadToAzure` task, which is in PublishContent.targets.  It didn't make sense to me to use that variable that call to UploadToAzure and not for this one.

brthor:
:shipit:

PR #4242, opened 2016-09-23
Add support for Migration defaults

### LINKED ISSUE #4100: Migration: Redesign TransformApplicator
Based on feedback from @livarcocc

### LINKED ISSUE #4120: Migration: Add capability for consuming a Defaults Snapshot
Currently, the Migration tool will output properties even if their value is the same as the default in the sdk.

Add capability for creating a snapshot of these defaults and consuming it in the migration tool so these properties won't be output when the value is equal to the default.

brthor:
Still working on the tests, but functionality is there.

Fixes #4120 
Fixes #4100 

@eerhardt @Sridhar-MS @livarcocc

### DIFF
--- a/src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
+++ b/src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
@@ -1,0 +1,128 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+using Microsoft.DotNet.ProjectJsonMigration.Transforms;
+using Microsoft.Build.Construction;
+using System.Collections.Generic;
+using Microsoft.DotNet.ProjectJsonMigration.Models;
+using System;
+using System.IO;
+using System.Reflection;
+using Newtonsoft.Json;
+
+namespace Microsoft.DotNet.ProjectJsonMigration.Rules
+{
+    public class AddDefaultsToProjectRule : IMigrationRule
+    {
+        internal const string c_DefaultsProjectElementContainerLabel = "MigrationDefaultsTempContainer";
+        internal const string c_SdkDefaultsJsonFileName = "sdkdefaults.json";
+
+        private readonly ITransformApplicator _transformApplicator;
+
+        public AddDefaultsToProjectRule(ITransformApplicator transformApplicator = null)
+        {
+            _transformApplicator = transformApplicator ?? new TransformApplicator();
+        }
+
+        public void Apply(MigrationSettings migrationSettings, MigrationRuleInputs migrationRuleInputs)
+        {
+            SerializableMigrationDefaultsInfo defaults = ResolveDefaults(migrationSettings);
+
+            var project = migrationRuleInputs.OutputMSBuildProject;
+            var defaultsPropertyGroups = new Dictionary<string, ProjectPropertyGroupElement>();
+            var defaultsItemGroups = new Dictionary<string, ProjectItemGroupElement>();
+
+            AddDefaultPropertiesToProject(defaults, project, defaultsPropertyGroups);
+            AddDefaultItemsToProject(defaults, project, defaultsItemGroups);
+        }
+
+        private void AddDefaultItemsToProject(
+            SerializableMigrationDefaultsInfo defaults, 
+            ProjectRootElement project, 
+            Dictionary<string, ProjectItemGroupElement> defaultsItemGroups)
+        {
+            foreach (var itemInfo in defaults.Items)
+            {
+                ProjectItemGroupElement itemGroup;
+                var parentCondition = itemInfo.ParentCondition ?? string.Empty;
+
+                if (!defaultsItemGroups.TryGetValue(parentCondition, out itemGroup))
+                {
+                    itemGroup = project.AddItemGroup();
+                    itemGroup.Label = c_DefaultsProjectElementContainerLabel;
+                    itemGroup.Condition = parentCondition;
+
+                    defaultsItemGroups[parentCondition] = itemGroup;
+                }
+
+                var item = itemGroup.AddItem(itemInfo.ItemType, itemInfo.Include);
+                item.Exclude = itemInfo.Exclude;
+                item.Remove = itemInfo.Remove;
+                item.Condition = itemInfo.Condition;
+            }
+        }
+
+        private static void AddDefaultPropertiesToProject(
+            SerializableMigrationDefaultsInfo defaults, 
+            ProjectRootElement project, 
+            Dictionary<string, ProjectPropertyGroupElement> defaultsPropertyGroups)
+        {
+            foreach (var propertyInfo in defaults.Properties)
+            {
+                ProjectPropertyGroupElement propertyGroup;
+                var parentCondition = propertyInfo.ParentCondition ?? string.Empty;
+
+                if (!defaultsPropertyGroups.TryGetValue(parentCondition, out propertyGroup))
+                {
+                    propertyGroup = project.AddPropertyGroup();
+                    propertyGroup.Label = c_DefaultsProjectElementContainerLabel;
+                    propertyGroup.Condition = parentCondition;
+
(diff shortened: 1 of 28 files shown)


brthor:

tools/MigrationDefaultsConstructor/Program.cs
```diff
@@ -0,0 +1,196 @@
...
+
+namespace MigrationDefaultsConstructor
+{
+    public class Program
+    {
+        private const string c_temporaryDotnetNewMSBuildProjectName = "p";
+
+        static void Main(string[] args)
+        {
+            var dotnetPath = Command.Create("which", new [] { "dotnet" }).CaptureStdOut().Execute().StdOut.Trim();
```
eerhardt: Is this used?

tools/MigrationDefaultsConstructor/Program.cs
```diff
@@ -0,0 +1,196 @@
...
+                    Condition = property.Condition,
+                    ParentCondition = property.Parent.Condition
+                };
+
+                allProperties.Add(propertyInfo);
+            }
+            
+        }
+
+        private static string CreateDotnetNewMSBuild(string projectName, string dotnet)
```
eerhardt: Is this used?

tools/MigrationDefaultsConstructor/Program.cs
```diff
@@ -0,0 +1,196 @@
...
+{
+    public class Program
+    {
+        private const string c_temporaryDotnetNewMSBuildProjectName = "p";
+
+        static void Main(string[] args)
+        {
+            var dotnetPath = Command.Create("which", new [] { "dotnet" }).CaptureStdOut().Execute().StdOut.Trim();
+            
+            var targetsFile = ProjectRootElement.Open(Path.Combine(Directory.GetCurrentDirectory(), "Microsoft.NETCore.Sdk.targets"));
```
eerhardt: I don't see where the SDK files are coming from.

src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
```diff
@@ -0,0 +1,123 @@
...
+            Dictionary<string, ProjectItemGroupElement> defaultsItemGroups)
+        {
+            foreach (var itemInfo in defaults.Items)
+            {
+                ProjectItemGroupElement itemGroup;
+                var parentCondition = itemInfo.ParentCondition ?? string.Empty;
+
+                if (!defaultsItemGroups.TryGetValue(parentCondition, out itemGroup))
+                {
+                    itemGroup = project.AddItemGroup();
```
eerhardt: We don't have any rules (or plan on having any rules in the future) that will add things to these groups, right?

Because if we add these groups here, and then some other rule adds to this group, and then we delete the 'defaults' group at the end of migration, we will lose that other rules' changes.

src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
```diff
@@ -0,0 +1,123 @@
...
+                using (StreamReader reader = new StreamReader(resource))
+                {
+                    return DeserializeDefaults(reader.ReadToEnd());
+                }
+            }
+        }
+
+        private string TryResolveSdkDefaultsFileFromAppBase()
+        {
+            var candidate = Path.Combine(AppContext.BaseDirectory, c_SdkDefaultsJsonFileName);
```
eerhardt: Is it weird to put "configuration" type things next to the CLI?  The CLI could be installed in a location that isn't writeable by the user, right?  What if we added this as an option to the `dotnet migrate` command?

eerhardt:
:shipit:

src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
brthor: I think this is a pretty advanced scenario, probably used for debugging or hot fixes between releases. I'm not sure a cmd line option would be used often enough not just be confusing.

src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
brthor: Yes this is a good realization. Nothing adds to these groups.

tools/MigrationDefaultsConstructor/Program.cs
brthor: Magic, I'll stick a readme with instructions, you just need to put them there yourself. 
This tool should be run infrequently.

brthor:
This is the latest project output here:

```
  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
  <PropertyGroup>
    <OutputPath>bin\$(Configuration)</OutputPath>
  </PropertyGroup>
  <PropertyGroup>
    <TargetFrameworks>netstandard1.5</TargetFrameworks>
    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="NETStandard.Library">
      <Version>1.6.0</Version>
    </PackageReference>
    <PackageReference Include="System.Runtime.Analyzers">
      <Version>1.1.0</Version>
      <PrivateAssets>All</PrivateAssets>
    </PackageReference>
  </ItemGroup>
  <ItemGroup>
    <Compile Include="**\*.cs" Exclude="bin\**;obj\**;**\*.xproj;packages\**" />
    <EmbeddedResource Include="**\*.resx" Exclude="bin\**;obj\**;**\*.xproj;packages\**" />
    <EmbeddedResource Include="compiler\resources\**\*" Exclude="bin\**;obj\**;**\*.xproj;packages\**" />
    <PackageReference Include="Microsoft.NET.Sdk">
      <Version>1.0.0-alpha-20161010-1</Version>
      <PrivateAssets>All</PrivateAssets>
    </PackageReference>
  </ItemGroup>
  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.5' ">
    <DefineConstants>$(DefineConstants);NETSTANDARD1_5</DefineConstants>
  </PropertyGroup>
    <DefineConstants>$(DefineConstants);RELEASE</DefineConstants>
  </PropertyGroup>
  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
</Project>
```

Lot's of good stuff getting defaulted including the define constants for debug. 

If the defaults change, the sdk revision commit can be adjusted to target the defaults at. For example, if there was a desire to include the project.json `Compile` excludes as a default in the sdk

brthor:

src/Microsoft.DotNet.ProjectJsonMigration/Rules/AddDefaultsToProjectRule.cs
brthor: I double checked on this to be sure, things might get deleted, but nothing will get added.

eerhardt:
```
  <PropertyGroup>
    <OutputPath>bin\$(Configuration)</OutputPath>
  </PropertyGroup>
```

That can be removed, can't it?

eerhardt:
test Ubuntu x64 Release Build

tools/MigrationDefaultsConstructor/project.json
```diff
@@ -0,0 +1,33 @@
...
+    "netcoreapp1.0": {
+      "dependencies": {
+        "Microsoft.NETCore.App": {
+          "type": "platform",
+          "version": "1.0.1"
+        },
+        "dotnet": {
+          "target": "project"
+        },
+        "MSBuild": "0.1.0-preview-00038-160914",
```
eerhardt: These should all be "15" versions now.  We've been on that version for a while.

brthor:
test Ubuntu x64 Release Build

brthor:
`    <OutputPath>bin\$(Configuration)</OutputPath>` 

Is coming in because it's in the new template. I think it's a separate fix.

brthor:
test  CentOS7.1 x64 Debug Build

brthor:

PR #4261, opened 2016-09-27
PackageReference Migration

brthor:
WIP PackageReference, won't merge until we figure out what the plan is for imports.

We can use this branches bits to test PackageReference in the meantime though.

@livarcocc @Sridhar-MS @eerhardt @srivatsn


src/Microsoft.DotNet.ProjectJsonMigration/MigrationNuGetFrameworkExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using NuGet.Frameworks;
+
+namespace Microsoft.DotNet.ProjectJsonMigration
+{
+    public static class MigrationNuGetFrameworkExtensions
+    {
+        public static string GetMSBuildCondition(this NuGetFramework framework)
+        {
+            return $" '$(TargetFrameworkIdentifier),Version=$(TargetFrameworkVersion)' == '{framework.DotNetFrameworkName}' ";
```
eerhardt: I think you can just say `'$(TargetFramework)' == '{framework.ShortName}'`

/cc @nguerrera - to confirm if this is true.

src/Microsoft.DotNet.ProjectJsonMigration/Rules/TemporaryMutateProjectJsonRule.cs
```diff
@@ -39,8 +40,8 @@ public void Apply(MigrationSettings migrationSettings, MigrationRuleInputs migra
             }
 
             var json = CreateDestinationProjectFile(sourceProjectFile, destinationProjectFile);
-            InjectSdkReference(json, ConstantPackageNames.CSdkPackageName, migrationSettings.SdkPackageVersion);
             RemoveRuntimesNode(json);
+            RemoveAllDependencies(json);
```
eerhardt: Why?  I would think now we wouldn't need to mutate the project.json at all anymore.

brthor: That's a good point, I was thinking that we still need it for tools, but that is not going to work

src/Microsoft.DotNet.ProjectJsonMigration/MigrationNuGetFrameworkExtensions.cs
brthor: Is this true @nguerrera? Will make the change

nguerrera: As long as you put the same spelling in TargetFrameworks, then yes, and that is the idiomatic way within a project.

brthor: 👍

brthor: changed.

brthor:
test Windows_NT x64 Release Build please

Issue #4165, opened 2016-09-12
Migration: Create Migration Report


brthor:
- [ ] ProjectJsonMigration ProjectMigrator returns a Migration Report Object
- [ ] dotnet-migrate serializes the migration report to disk

Issue #4164, opened 2016-09-12
Migration: Clean up Text Output


brthor:
- [x] Clean up error output to be more clear on success or failure
- [x] Output Errors to StdErr
- [x] Create an output on Success with a summary of project migrated
- [x] Add Line + File Information to deprecated errors

PR #4310, opened 2016-10-04
Migration Report

brthor:
@livarcocc @srivatsn @333fred 

/cc @eerhardt @Sridhar-MS @blackdwarf 

The migration report including options for outputting to a file or stdout in a user friendly or json format.

Fixes: #4165 
Fixes:  #4164

Successful Example:
 
screen shot 2016-10-04 at 3 09 17 pm

Failure Example:

screen shot 2016-10-04 at 3 30 01 pm


src/Microsoft.DotNet.ProjectJsonMigration/Report/ProjectMigrationReport.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+namespace Microsoft.DotNet.ProjectJsonMigration
+{
+    public class ProjectMigrationReport
```
333fred: Can we put a comment here that we need to update the xproj migration tools in the roslyn-project-system if this is ever updated?

brthor: Can you share a link to what file or files should be updated?

333fred: It'll probably be https://github.com/dotnet/roslyn-project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/Xproj/MigrateXprojProjectFactory.cs. At the very least, looking at that will provide the context for what needs to be updated, I may factor reading the log out.

src/Microsoft.DotNet.ProjectJsonMigration/Report/MigrationReport.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public int MigratedProjectsCount => ProjectMigrationReports.Count;
+
+        public int SucceededProjectsCount => ProjectMigrationReports.Count(p => !p.Errors.Any());
+
+        public int FailedProjectsCount => ProjectMigrationReports.Count(p => p.Errors.Any());
+
+        public bool AllSucceeded => ! ProjectMigrationReports.Any(p => p.Errors.Any());
+
+        public MigrationReport MergeReport(MigrationReport otherReport)
```
eerhardt: (nit) Could just be named "Merge".  "Report" is a little redundant.

src/dotnet/commands/dotnet-migrate/MigrateCommand.cs
```diff
@@ -41,18 +54,81 @@ public int Execute()
...
+                }
+                else
+                {
+                    migrationReport = migrationReport.MergeReport(projectMigrationReport);
+                }
             }
 
+            WriteReport(migrationReport);
+
             return 0;
```
eerhardt: Do we want to return non-zero when the migration fails?

src/dotnet/commands/dotnet-migrate/MigrateCommand.cs
```diff
@@ -41,18 +54,81 @@ public int Execute()
...
+            }
+
+            StringBuilder sb = new StringBuilder();
+
+            foreach (var projectMigrationReport in migrationReport.ProjectMigrationReports)
+            {
+                if (projectMigrationReport.Errors.Any())
+                {
+                    sb.AppendLine($"Project {projectMigrationReport.ProjectName} migration failed ({projectMigrationReport.ProjectDirectory})");
+                    sb.AppendLine(string.Join("\n", projectMigrationReport.Errors.Select(e => e.GetFormattedErrorMessage())));
```
eerhardt: Actually, why not just `foreach (line in Errors) { sb.AppendLine(line); }`?

---
In reply to: [81990635](https://github.com/dotnet/cli/pull/4310#discussion_r81990635) [](ancestors = 81990635)

src/Microsoft.DotNet.ProjectJsonMigration/ProjectContextExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.DotNet.ProjectModel;
+using System.IO;
+
+namespace Microsoft.DotNet.ProjectJsonMigration
+{
+    public static class ProjectContextExtensions
+    {
+        public static string GetProjectName(this ProjectContext projectContext)
+        {
+            return Path.GetFileNameWithoutExtension(projectContext.GetOutputPaths("_").CompilationFiles.Assembly);
```
eerhardt: SaveOutputProjectRule has this same logic duplicated.  Can it call this new method instead?

[https://github.com/dotnet/cli/blob/3a567e595773b61650f2dcd1c7ec2b53b91cd7a8/src/Microsoft.DotNet.ProjectJsonMigration/Rules/SaveOutputProjectRule.cs#L13-L14](https://github.com/dotnet/cli/blob/3a567e595773b61650f2dcd1c7ec2b53b91cd7a8/src/Microsoft.DotNet.ProjectJsonMigration/Rules/SaveOutputProjectRule.cs#L13-L14)

src/Microsoft.DotNet.ProjectJsonMigration/Report/MigrationReport.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class MigrationReport
+    {
+        public List<ProjectMigrationReport> ProjectMigrationReports { get; }
+
+        public int MigratedProjectsCount => ProjectMigrationReports.Count;
+
+        public int SucceededProjectsCount => ProjectMigrationReports.Count(p => !p.Errors.Any());
+
+        public int FailedProjectsCount => ProjectMigrationReports.Count(p => p.Errors.Any());
```
eerhardt: We could make "Passed" and "Failed" properties on `ProjectMigrationReport`, which would make this a little more readable.

`public int FailedProjectsCount => ProjectMigrationReports.Count(p => p.Failed);`

src/Microsoft.DotNet.ProjectJsonMigration/Report/ProjectMigrationReport.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public ProjectMigrationReport(string projectDirectory, string projectName, string outputMSBuildProject, List<string> warnings)
+            : this(projectDirectory, projectName, outputMSBuildProject, null, warnings) { }
+
+        private ProjectMigrationReport(string projectDirectory, string projectName, string outputMSBuildProject, List<MigrationError> errors, List<string> warnings, bool skipped=false)
+        {
+            ProjectDirectory = projectDirectory;
+            ProjectName = projectName;
+            OutputMSBuildProject = outputMSBuildProject;
+            Errors = errors ?? Enumerable.Empty<MigrationError>().ToList();
```
eerhardt: no need to `Enumerable.Empty<MigrationError>().ToList()` - just call `new List<X>()`, less code, and less perf overhead.

src/Microsoft.DotNet.ProjectJsonMigration/Report/ProjectMigrationReport.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        public string ProjectDirectory { get; }
+
+        public string ProjectName { get; }
+
+        public string OutputMSBuildProject { get; }
+
+        public List<MigrationError> Errors { get; }
+
+        public List<string> Warnings { get; }
```
eerhardt: Do we ever actually use Warnings?  I didn't see them used.

eerhardt:
:shipit:

src/Microsoft.DotNet.ProjectJsonMigration/Report/MigrationReport.cs
brthor: will fix.

src/dotnet/commands/dotnet-migrate/MigrateCommand.cs
brthor: I was thinking about this, I think so as well.

src/dotnet/commands/dotnet-migrate/MigrateCommand.cs
brthor: I'll do the foreach loop, a little more readable.

src/dotnet/commands/dotnet-migrate/MigrateCommand.cs
```diff
@@ -41,18 +54,81 @@ public int Execute()
...
+                    sb.AppendLine();
+                }
+                else
+                {
+                    sb.AppendLine($"Project {projectMigrationReport.ProjectName} migration succeeded ({projectMigrationReport.ProjectDirectory})");
```
brthor: Yeah, I'm going to make a change to always bring the report even if you specify a file and then that should be doable. The issue I ran into was with using the TextWriter rather than Console to write.

src/Microsoft.DotNet.ProjectJsonMigration/ProjectContextExtensions.cs
brthor: `_` is a common pattern in functional languages for something that you don't care about.

I can put a comment in there though.

src/Microsoft.DotNet.ProjectJsonMigration/Report/MigrationReport.cs
brthor: Yeah I like that.

src/Microsoft.DotNet.ProjectJsonMigration/Report/ProjectMigrationReport.cs
brthor: Good catch

src/dotnet/commands/dotnet-migrate/MigrateCommand.cs
brthor: fixed.

Now this pr fixes this issue too: https://github.com/dotnet/cli/issues/4164

brthor:
Added changes to also tackle https://github.com/dotnet/cli/issues/4164
Addressed all feedback
Please take another pass on Code review.

brthor:
New Samples: 
screen shot 2016-10-05 at 5 40 53 pm
screen shot 2016-10-05 at 5 41 13 pm

brthor:

brthor:

PR #4323, opened 2016-10-05
Migrate Args update comment

333fred:
The new project system calls dotnet migrate on the command line. When we update the cli, we also need to update the VS side. I'm skipping CI since this is purely a comment change. Tagging @brthor for review.

skip ci please


brthor:
thanks for adding this 👍

PR #4361, opened 2016-10-10
Remove dotnet-compile-fsc.

eerhardt:
It's new home is https://github.com/dotnet/netcorecli-fsc/

@piotrpMSFT @livarcocc @KevinRansom @enricosada


piotrpMSFT:
@KevinRansom FYI. We're removing fsc from the MSBuild port of the CLI since we don't yet have a MSBuild-driven alternative. Once that comes online we'll look at adding back fsc templates.

eerhardt:
@piotrpMSFT - I didn't remove the templates in this PR.

Issue #4273, opened 2016-09-29
dotnet-install.ps1 fails behind a proxy


mendhak:
## Steps to reproduce

In an environment where Internet access requires an outbound proxy, run:

    dotnet-install.ps1  --version 1.0.0-preview2-003133

Internet Explorer proxy has been set
netsh/winhttp proxy has been set

Is there some other environment variable I should be setting to make this work?

## Expected  behavior

The script should download the binaries for running dotnet.  

## Actual behavior

Download fails:

>D:\BuildAgent\work\11440641e6edc7c0\BuildTools>powershell -executionpolicy bypass -file dotnet-install.ps1  --version 1.0.0-preview2-003133
dotnet-install: Downloading https://dotnetcli.blob.core.windows.net/dotnet/preview/Binaries/1.0.0-preview2-003133/dotnet-dev-win-x64.1.0.0-preview2-003133.zip
Invoke-WebRequest : Unable to connect to the remote server
At D:\BuildAgent\work\11440641e6edc7c0\BuildTools\dotnet-install.ps1:345 char:13
     $resp = Invoke-WebRequest -UseBasicParsing $DownloadLink -OutFile $ZipPath
`~~~~~~~~~~~~~~
CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebException
FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand

## Environment data
`dotnet --info` output:  n/a
Internet Explorer proxy has been set
netsh/winhttp proxy has been set

piotrpMSFT:
@mendhak thanks for reporting this. You're right that install.ps1 has not been tested in proxy scenarios. It looks like we should be passing proxy info when calling Invoke-WebRequest:

https://social.technet.microsoft.com/Forums/office/en-US/1cd75340-c22c-4c37-a2b0-8cd6051473d8/invokewebrequest-and-proxy?forum=winserverpowershell

We won't have cycles to update the script in the near future, but perhaps you would like to experiment with this and create a PR? The change should be fairly straightforward.

PR #4364, opened 2016-10-10
Allow dotnet-install.ps1 to work behind an outbound proxy

mendhak:

The `dotnet-install.ps1` script for obtaining dotnet currently does not work behind a proxy (example, corporates with outbound proxies).  This PR adds an optional parameter, `-ProxyAddress`, which takes the address of a proxy server, such as `http://servername:3128/` and uses it when initializing the `System.Net.Http.HttpClient`.   

Issue #4273

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -47,7 +47,7 @@
 .PARAMETER AzureFeed
     Default: https://dotnetcli.azureedge.net/dotnet
     This parameter should not be usually changed by user. It allows to change URL for the Azure feed used by this installer.
-.PARAMETER Proxy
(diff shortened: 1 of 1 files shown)


scripts/obtain/dotnet-install.ps1
```diff
@@ -47,6 +47,8 @@
 .PARAMETER AzureFeed
     Default: https://dotnetcli.azureedge.net/dotnet
     This parameter should not be usually changed by user. It allows to change URL for the Azure feed used by this installer.
+.PARAMETER ProxyAddress
```
piotrpMSFT: https://github.com/dotnet/cli/issues/4368

PR #4377, opened 2016-10-11
bump nuget version

brthor:
@eerhardt @livarcocc @jgoshi @piotrpMSFT

### DIFF
--- a/build_projects/dotnet-cli-build/project.json
+++ b/build_projects/dotnet-cli-build/project.json
@@ -20,7 +20,7 @@
     "System.Runtime.Serialization.Primitives": "4.1.1",
     "System.Xml.XmlSerializer": "4.0.11",
     "WindowsAzure.Storage": "6.2.2-preview",
-    "NuGet.CommandLine.XPlat": "3.6.0-rc-1954",
+    "NuGet.CommandLine.XPlat": "3.6.0-rc-1976",
     "Microsoft.Build.Utilities.Core": "0.1.0-preview-00043-160929",
     "Microsoft.DotNet.PlatformAbstractions": "1.0.1-beta-000933"
   },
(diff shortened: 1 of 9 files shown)


brthor:
strange error, debugging this

brthor:
I fixed the null ref here. 

What it looked like was that the CLI.Utils -> ProjectModel P2P export expressed a library range with a null min and max version. 

@emgarten have there been any changes in nuget that sound like they may affect this? I'm not certain what changed since the last version of nuget.

emgarten:
@brthor do you have an example of what NuGet is writing out that is wrong? There have been some changes to ``projectFileDependencyGroups`` for csproj + package references. Xproj should still have the same result.

emgarten:
The latest nuget build is now ``3.6.0-rc-1978``

emgarten:
I've tracked the problem down to LibraryRange.VersionRange now defaulting to VersionRange.All instead of null. It looks like there is still a lot of code in this repo expecting null. I'm going to revert this back in NuGet, it was just done as part of other clean up. Using null to mean the All range in the past has been dangerous since a range that fails TryParse can easily turn into All instead of whatever it was intended to be.

The VersionUtility.Render method here is missing a lot of scenarios. Not having a min version is perfectly valid, and so is having a non-min inclusive version which is also not covered here.

Consider copying or calling the NuGet equivalent here: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.LibraryModel/LibraryRange.cs#L71-L111

emgarten:
I've created a new PR here with the updated NuGet packages: https://github.com/dotnet/cli/pull/4385

eerhardt:
Closing in favor of #4385.

brthor:
Awesome thanks @emgarten

PR #4379, opened 2016-10-11
Updating NuGet to 3.6.0-rc-1978

emgarten:
Adding support for non-conditioned targets/props with $(TargetFramework)


eerhardt:
@emgarten - I believe the error you are hitting is the same one @brthor is fighting with in #4377

PR #4387, opened 2016-10-12
remove autounify, fix langversion

### LINKED ISSUE #4386: dotnet migrate: buildOptions/languageVersion is not migrated properly
## Steps to reproduce

1. `dotnet new`
1. Add `"languageVersion": "csharp5"` to the `"buildOptions"` node.
1. `dotnet migrate`
1. `dotnet restore3`
1. `dotnet build3`

## Expected  behavior

The build should succeed. The .csproj should contain `<LangVersion>5</LangVersion>`.

## Actual behavior

The build fails:

```
CSC : error CS1617: Invalid option 'csharp5' for /langversion; must be ISO-1, ISO-2, Default or an integer in range 1 to 6. [C:\Users\joelv\Desktop\migrate\langver\langver.csproj]
```

## Workaround

Edit the .csproj `<LangVersion>csharp5</LangVersion>` to `<LangVersion>5</LangVersion>`.

## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (1.0.0-preview3-003830)

Product Information:
 Version:            1.0.0-preview3-003830
 Commit SHA-1 hash:  90a51bdd00

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
```

brthor:
Fixes #4386 (@joelverhagen)
Fixes auto unify being set in migration (@nguerrera)

/cc @livarcocc @eerhardt

### DIFF
--- a/src/Microsoft.DotNet.ProjectJsonMigration/Rules/MigrateBuildOptionsRule.cs
+++ b/src/Microsoft.DotNet.ProjectJsonMigration/Rules/MigrateBuildOptionsRule.cs
@@ -76,7 +76,7 @@ namespace Microsoft.DotNet.ProjectJsonMigration.Rules
 
         private AddPropertyTransform<CommonCompilerOptions> LanguageVersionTransform =>
             new AddPropertyTransform<CommonCompilerOptions>("LangVersion",
-                compilerOptions => compilerOptions.LanguageVersion,
+                compilerOptions => FormatLanguageVersion(compilerOptions.LanguageVersion),
                 compilerOptions => !string.IsNullOrEmpty(compilerOptions.LanguageVersion));
 
         private AddPropertyTransform<CommonCompilerOptions> DelaySignTransform =>
@@ -354,6 +354,16 @@ namespace Microsoft.DotNet.ProjectJsonMigration.Rules
                     new JObject(),
                     null,
                     ProjectFilesCollection.DefaultPublishExcludePatterns);
+        }
+
+        private string FormatLanguageVersion(string langVersion)
+        {
+            if (langVersion.StartsWith("csharp"))
+            {
+                return langVersion.Replace("csharp", "");
+            }
+
+            return langVersion;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.ProjectJsonMigration/Rules/MigrateBuildOptionsRule.cs
```diff
@@ -355,5 +355,15 @@ private IncludeContext GetCopyToOutputIncludeContext(CommonCompilerOptions compi
                     null,
                     ProjectFilesCollection.DefaultPublishExcludePatterns);
         }
+
+        private string FormatLanguageVersion(string langVersion)
+        {
+            if (langVersion.StartsWith("csharp"))
```
joelverhagen: Are there other languages to consider here?

brthor: Migration only migrates to csharp, there shouldn't be any cases where we're migrating a different value unless you can think of something.

joelverhagen: What about case sensitivity? If `CSHARP5` is in the project.json, does `dotnet build` still do the right thing?

src/Microsoft.DotNet.ProjectJsonMigration/Rules/MigrateBuildOptionsRule.cs
```diff
@@ -355,5 +355,15 @@ private IncludeContext GetCopyToOutputIncludeContext(CommonCompilerOptions compi
...
+            {
+                return langVersion.Replace("csharp", "");
```
joelverhagen: and if you want replace (not sure why) nit: string.Empty

brthor: fixed

src/Microsoft.DotNet.ProjectJsonMigration/Rules/MigrateBuildOptionsRule.cs
brthor: good catch here

brthor:
Also added a fix for the target framework property on project refs.

brthor:

PR #4391, opened 2016-10-12
Updating CoreFx, CoreSetup to servicing-24611-07, 1.0.2 respectively

eerhardt:


eerhardt:
@livarcocc - these are the corefx changes I would expect #4396 to have.

eerhardt:
@livarcocc - maybe not.  I'm still confused at our servicing model....  Did any new CoreFX packages get produced for Microsoft.NETCore.App 1.0.2?

/cc @ericstj @gkhanna79 @weshaggard ?  How is the CLI supposed to know which CoreFX/CoreCLR packages have been updated between `1.0.0` and `1.0.2`?  And what their new versions are?

TestAssets/DesktopTestProjects/DesktopAppWithNativeDep/project.json
```diff
@@ -2,7 +2,7 @@
   "version": "1.0.0-*",
   "dependencies": {
     "PackageWithFakeNativeDep": "1.0.0-*",
-    "Microsoft.NETCore.Platforms": "1.0.1"
+    "Microsoft.NETCore.Platforms": "1.0.2-servicing-24611-07"
```
gkhanna79: This update should not be there.

gkhanna79: And likewise for other instances of the same.

gkhanna79:
The only change in 1.0.2 build of Core-Setup was to produce Microsoft.NETCore.App 1.0.2 package, which has the same dependencies of CoreCLR and CoreFX as 1.0.1.

eerhardt:
>The only change in 1.0.2 build of Core-Setup was to produce Microsoft.NETCore.App 1.0.2 package, which has the same dependencies of CoreCLR and CoreFX as 1.0.1.

Now I'm super confused.  Why are we making a new version?  Just to update the version number?  What code actually changed between `1.0.1` and `1.0.2`?

TestAssets/DesktopTestProjects/DesktopAppWithNativeDep/project.json
gkhanna79: The file is not correct. Until yesterday, CoreSetup 1.0.2 was supposed to contain the contents of corresponding 1.0.0 branches of CoreFX and CoreCLR.

However, the POR was updated to ensure that .NET Core 1.0.2 (from Core-Setup 1.0.0 branch) should only contain the fix for OSX PATH issue for Sierra and nothing else. Everything else will come in a subsequent LTS release (say 1.0.3).

Thus, Core-Setup 1.0.0 had bunch of commits reverted to no longer pickup the updated CoreCLR and CoreFX 1.0.0 packages, which is what the list above points to.

For CLI Preview 2, the change should simply be to refer to 1.0.2 M.N.App and that is it.

eerhardt: Ah, OK.  That explains it.  We had an installer fix and that's why we are shipping a new version.

Thanks.  I'll close this PR.

gkhanna79:
> Just to update the version number? 

Yes.

PR #4405, opened 2016-10-14
Change cli build to csproj files (First step on one project)

jgoshi:
First doing this just for dotnet-cli-build


build_projects/dotnet-cli-build/dotnet-cli-build.csproj
```diff
@@ -0,0 +1,72 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
brthor: There is an issue for migration defaults, create comments there for cleaned up properties.

brthor: https://github.com/dotnet/cli/issues/4120

jgoshi: @brthor Can you review the file and let me know what are the properties I can remove (those that are not needed)? From me looking at the file, I see two things (I didn't add them to the issue yet until you let me know if they are valid):
1. Remove <?xml version="1.0" encoding="utf-8"?>
2. Make one property group: <OutputPath>bin\$(Configuration)</OutputPath> gets put into its own PropertyGroup

jgoshi: Should I write any of this as comments in #4120

PR #4421, opened 2016-10-14
Removing the old Microsoft.NuGet.targets in ImportsBefore/After.

eerhardt:
These files are no longer part of the CLI, so these imports are no longer needed.


eerhardt:
test OSX x64 Release Build

Issue #4433, opened 2016-10-17
dotnet restore3 should not display "Build started" and "Build succeeded" in the restore log


balachir:
INSTALL STEPS
WPT.25810.00 (includes CLI version 1.0.0-preview3-003748)
Add private MyGet feeds to Package Manager Settings

REPRO STEPS
1. New folder, type "dotnet new -t msbuild"
2. Type "dotnet restore3"

ACTUAL
Restore shows "Build started" and "Build succeeded" in the log. This might confuse end-users regarding why calling a 'restore' seems to invoke a 'build'. I initially thought it was the project getting built, then I realized that it's probably the standard output from MSBuild.

[From Rob - Nuget team] They should consider changing the normal output of msbuild … perhaps they could use the target name that is executed, instead of always saying "Build". Pack and Publish and Test will have the same problems.

![image](https://cloud.githubusercontent.com/assets/8246794/19457053/62aa4b98-9479-11e6-99d1-ebabe268c490.png)

PR #4438, opened 2016-10-18
Reduce MSBuild verbosity for restore

jeffkl:
Wanted to run this by everyone.  This is the alternative to a custom logger but might address the concern in #4433.

* Remove the MSBuild logo
* Set verbosity to minimal

Sample output now with only high importance messages:

image

The user could still see the normal output with the file logger:

`dotnet restore3 /fl /flp:Verbosity:Normal`

### DIFF
--- a/src/dotnet/commands/dotnet-restore3/Program.cs
+++ b/src/dotnet/commands/dotnet-restore3/Program.cs
@@ -69,7 +69,9 @@ namespace Microsoft.DotNet.Tools.Restore3
             {
                 var msbuildArgs = new List<string>()
                 {
-                     "/t:Restore"
+                     "/NoLogo",
+                     "/t:Restore",
+                     "/ConsoleLoggerParameters:Verbosity=Minimal"
                 };
 
                 if (sourceOption.HasValue())


davidfowl:
Can you show a screenshot with errors as well.

jeffkl:
Good idea!

image

jeffkl:
Here's the error if the package is incompatible:

image

jeffkl:
FYI @andygerlicher, @rainersigwald

livarcocc:
LGTM. But we talked about doing minimal verbose for other commands yesterday. Is this something that we still want to do?

jeffkl:
> But we talked about doing minimal verbose for other commands yesterday. Is this something that we still want to do?

I mostly wanted to get this out there as an example.  In my opinion, some commands should have lower verbosity and others shouldn't.  Restore should really just show what's being restored but build should show me everything.  I'll leave that up to your team.

It might make sense to add a flag to the `MSBuildForwardingApp` class like `Verbosity` rather than add this stuff to the list of arguments for each command.

davidfowl:
Do the package sources get printed out? That seems like a critical piece of information missing...

jeffkl:
They don't get printed out with the console logger verbosity at minimal because the NuGet targets log that information as "normal" verbosity.  These are the options:

1. Users can specify `/flp:Verbosity=Normal` which will emit an `MSBuild.log` file of the normal output.
2. Modify the NuGet targets to log the package sources as `High` verbosity instead of `Normal`.
3. Implement a custom console logger in dotnet CLI which does not emit the BuildStarted and BuildEnded events, as well as any other events that are not useful.

livarcocc:

livarcocc:

livarcocc:
Unless someone objects, I am going to merge this PR.

I have opened this issue on the NuGet repo for them to raise the verbosity on the package sources: https://github.com/NuGet/Home/issues/3719.

dasMulli:
The "original" project.json based restore command had a verbosity setting. It would be nice if restore3 had the same option for compatibility, defaulting to minimal, instead of hardcoding the msbuild argument.

Currently the restore command exposes this:
```
$ dotnet restore --help
...
-v|--verbosity <verbosity>      The verbosity of logging to use. Allowed values: Debug, Verbose, Information, Minimal, Warning, Error.
```

rainersigwald:
@dasMulli As with all MSBuild-pass-through commands, you can add additional MSBuild parameters to be appended to the `MSBuild.exe` command line. Since console verbosity is last-argument-wins, you can use that to override this default.

(It might be nice to make this clear in the help or munge a `dotnet` verbosity argument into MSBuilds, but at least there's a way.)

:shipit:

jeffkl:
@livarcocc do you want me to have an explicit `--verbosity` argument which defaults to minimal?  Or should we keep my current implementation?

livarcocc:
Don't add it here. I think if we add it, we should make it a general argument and each command should be able to set its own default.

PR #4460, opened 2016-10-20
turn on signing test

brthor:


brthor:
have this passing locally on windows, I threw up two commits in a sequence here to see if the tests pass xplat with KeyOriginator or if they need AssemblyKeyOriginator

PR #4516, opened 2016-10-27
Use --disable-parallel during xplat restores

dagood:
Works around NuGet parallel restore timeouts becoming build failures when the machine has a slow connection, for example in our Docker Fedora containers.

See https://github.com/dotnet/core-setup/issues/327

Upgrades buildtools to take https://github.com/dotnet/buildtools/pull/1114

Restore takes a very long time (> 7 minutes) in my Fedora 23 Docker container, but I haven't compared before/after yet. Having 4 NuGet sources seems like it would hurt even more with `--disable-parallel`. Some followup work to this would be to add a switch to let devs enable parallel restores if their connection can handle it.

I noticed a manually-queued (I think) Fedora build succeeded recently: https://mseng.visualstudio.com/dotnetcore/_build/index?buildId=3380864&_a=summary. Getting a lucky connection speed is a factor in this bug, so it's possible that instead of taking this PR, CLI could start spinning builds repeatedly to get successful ones.

@livarcocc @piotrpMSFT 
/cc @dleeapho @mellinoe

### DIFF
--- a/build_projects/dotnet-cli-build/DotNetRestore.cs
+++ b/build_projects/dotnet-cli-build/DotNetRestore.cs
@@ -12,7 +12,7 @@ namespace Microsoft.DotNet.Cli.Build
 
         protected override string Args
         {
-            get { return $"{GetSource()} {GetPackages()} {GetSkipInvalidConfigurations()}"; }
+            get { return $"{GetSource()} {GetPackages()} {GetSkipInvalidConfigurations()} {GetDisableParallel()}"; }
         }
 
         public string Source { get; set; }
@@ -49,6 +49,11 @@ namespace Microsoft.DotNet.Cli.Build
             }
 
             return null;
+        }
+
+        private string GetDisableParallel()
+        {
+            return "--disable-parallel";
         }
     }
 }
--- a/.toolversions
+++ b/.toolversions
@@ -1,2 +1,2 @@
-microsoft.dotnet.buildtools=1.0.26-prerelease-00802-01
+microsoft.dotnet.buildtools=1.0.26-prerelease-00926-02
 microsoft.dotnet.buildtools.run=1.0.0-prerelease-00807-04
--- a/build.proj
+++ b/build.proj
@@ -68,7 +68,11 @@
           Inputs="$(DotnetCliBuildDirectory)/dotnet-cli-build.csproj"
           Outputs="@(RestoreDotnetCliBuildFrameworkOutputs)">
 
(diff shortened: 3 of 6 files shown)


build_projects/dotnet-cli-build/DotNetRestore.cs
```diff
@@ -50,5 +50,10 @@ private string GetVerbosity()
 
             return null;
         }
+
+        private string GetDisableParallel()
+        {
+            return "--disable-parallel";
+        }
```
dagood: Looking for input: what's the best way to add this in a way that it doesn't happen on Windows, where it isn't needed? Adding an extra property would work, but it's tedious in msbuild--don't know if there's another convention in CLI.

dagood:
Merged, resolved conflicts, and added `--disable-parallel` to the tests, as far as I can tell. I got a successful build on my Fedora 23 container. I'll do my before/after measurements on OSX (since a "before" build on Fedora would fail).

dagood:
My OSX msbuild timings are:
 * `00:19:01.12` (with parallel)
 * `01:01:12.25` (--disable-parallel)

I believe buildtools setup isn't included in that, which takes up a decent chunk of time in both cases.

Any ideas what the windows errors are? https://ci.dot.net/job/dotnet_cli/job/rel_1.0.0/job/release_windows_nt_x64_prtest/458/

piotrpMSFT:

livarcocc:

dagood:
@livarcocc Did you see a build where https://github.com/dotnet/cli/pull/4527 didn't make restores work? (@mellinoe suggested merging only the NSS upgrade until it was proven to be not enough.)

saineshwar:
Still facing same issue after run command :- dotnet restore --disable-parallel

piotrpMSFT:
Looks like you're having issues reaching NuGet servers. Can you open this issue in the nuget/home repo, @saineshwar ?

dagood:
Maybe there's some context I'm missing, but it's possible @saineshwar is using a version of `dotnet` that doesn't have the parallel HTTP workaround implemented--and unfortunately there's no way to check without manually comparing the versions or peeking at the IL. (In any case, it's unrelated to this PR, because this PR was only meant to start using `--disable-parallel` in CLI.)

@saineshwar what version of `dotnet` are you using? (`dotnet --version`)

Here's the NuGet issue about parallel restore timeout failures that `--disable-parallel` is the workaround for: https://github.com/NuGet/Home/issues/2945

PR #4543, opened 2016-10-29
[WiP] enable /m by default

piotrpMSFT:
Enable /m by default


piotrpMSFT:
@cdmihai @rainer @jeffkl @AndyGerlicher @Sarabeth-Jaffe-Microsoft

This change makes `/m` the default for invoking MSBuild via CLI. Our implementation pre-pends the `/m` to the user's argument list. I've found through experimentation that msbuild honors only the `last` `/m` passed to it, enabling users to simply pass in `/m:1` when they want to disable parallel execution. So long as this behavior is consistent, we have the ability to control multiproc execution without introducing any new command-line concepts.

/cc @blackdwarf

rainersigwald:
Belated 👏 👍 :shipit: 

And last-one-wins is by design, so user-override-by-explicit-`/m:1` sounds like a good plan to me.

PR #4551, opened 2016-10-31
Custom logger for MSBuild to receive telemetry events

jeffkl:
1. Upgraded MSBuild version to latest
2. Had to make the telemetry session ID a static variable so that the forwarding app could use it.
3. Simple Logger implementation that forwards telemetry events to the `Telemetry` pipeline

Best to look at 27458107daec6c1158fe5481b7927528f8395db5 since it doesn't have the MSBuild version upgrade.  I've tested this all manually and will be writing tests after everyone signs off on the implementation here.  How can I find the events in the back-end?  Any place where an exception could occur, I'm just ignoring it so that telemetry related stuff never fails an operation.  Should I trace something instead?

@piotrpMSFT @livarcocc @rainersigwald @cdmihai

### DIFF
--- a/src/dotnet/Telemetry.cs
+++ b/src/dotnet/Telemetry.cs
@@ -15,6 +15,7 @@ namespace Microsoft.DotNet.Cli
 {
     public class Telemetry : ITelemetry
     {
+        internal static string CurrentSessionId = null;
         private TelemetryClient _client = null;
 
         private Dictionary<string, string> _commonProperties = null;
@@ -34,7 +35,9 @@ namespace Microsoft.DotNet.Cli
 
         public Telemetry () : this(null) { }
 
-        public Telemetry(INuGetCacheSentinel sentinel)
+        public Telemetry(INuGetCacheSentinel sentinel) : this(sentinel, null) { }
+
+        public Telemetry(INuGetCacheSentinel sentinel, string sessionId)
         {
             Enabled = !Env.GetEnvironmentVariableAsBool(TelemetryOptout) && PermissionExists(sentinel);
 
@@ -42,6 +45,9 @@ namespace Microsoft.DotNet.Cli
             {
                 return;
             }
+
+            // Store the session ID in a static field so that it can be reused
+            CurrentSessionId = sessionId ?? Guid.NewGuid().ToString();
 
             //initialize in task to offload to parallel thread
             _trackEventTask = Task.Factory.StartNew(() => InitializeTelemetry());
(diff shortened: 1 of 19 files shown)


src/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs
```diff
@@ -36,7 +46,16 @@ public MSBuildForwardingApp(IEnumerable<string> argsToForward)
 
         public int Execute()
         {
-            return _forwardingApp.Execute();
+            try
+            {
+                Environment.SetEnvironmentVariable(TelemetrySessionIdEnvironmentVariableName, Telemetry.CurrentSessionId);
```
cdmihai: Who are the intended consumers of this env var? If it is just the MSBuildLogger constructor, would it be cleaner to set it as a static on that type instead? Remember the MSBuild tests and environment variables :)

jeffkl: It's really just there in case someone executes another instance of `dotnet` under this instance.

piotrpMSFT:

piotrpMSFT:
@dotnet-bot test  RHEL7.2 x64 Release Build please.

piotrpMSFT:
@dotnet-bot test RHEL7.2 x64 Release Build please.

src/dotnet/Telemetry.cs
```diff
@@ -34,7 +35,9 @@ public class Telemetry : ITelemetry
 
         public Telemetry () : this(null) { }
 
-        public Telemetry(INuGetCacheSentinel sentinel)
+        public Telemetry(INuGetCacheSentinel sentinel) : this(sentinel, null) { }
```
piotrpMSFT: why not `string sessionId=null`?

src/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs
piotrpMSFT: As an example of this, CLI build invokes <Exec> of CLI all over the place. This may be a common pattern, and the env var automatically gets plumbed through to child processes. Are the issues you're referring to largely in unit tests clashing with one another?

cdmihai: Yup, we've had some bad experiences with shared state in MSBuild's unit tests, because a couple of core msbuild types get configured via env variables or read the current working directory :)

jeffkl:
I'll work on some tests today.  I think I just need to verify that the command-line argument is added and that the Logger is attached to the event

PR #4575, opened 2016-11-02
Fix dotnet test help typo

jgoshi:
@piotrpMSFT @livarcocc @krwq


jgoshi:
Something went wrong. Was supposed to just have one change

PR #4584, opened 2016-11-02
Rename test templates for casing consistency

jgoshi:
@livarcocc @piotrpMSFT @krwq


jgoshi:

jgoshi:
Doing this change in a new branch so I don't pollute our history

PR #4600, opened 2016-11-03
Fix for issue  https://github.com/Microsoft/vstest/issues/175

Faizan2304:


Faizan2304:
@livarcocc: Can you please take a look.

PR #4617, opened 2016-11-04
Removed Project_Readme from CSharp_Web

seancpeters:
Added the package fallback target to CSharp_Web


seancpeters:
Good catch, that’s a mistake from my copy-paste ☹… fixing

seancpeters:
The $if$ is in an odd spot to keep spacing consistent with the VS version of the template.

src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
```diff
@@ -5,7 +5,11 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp1.0</TargetFramework>
     <PreserveCompilationContext>true</PreserveCompilationContext>
-  </PropertyGroup>
+  </PropertyGroup>$if$ ($context$ == WebCore)
```
livarcocc: We do not handle these in the CLI today. I don't understand what this is doing in the CLI template.

seancpeters:
Sorry about that. I misunderstood. I've removed the if-else entirely.

livarcocc:
@seancpeters All right. We will have those capabilities soon I hope, but not yet.

livarcocc:
Though, why are we doing this now? Do we need approval for this?

piotrpMSFT:
This is a pre-req for the web sdk update we're doing later tonight/tomorrow morning.

PR #4729, opened 2016-11-15
Extract localizable strings from dotnet-new

piotrpMSFT:
@krwq @livarcocc @jgoshi @nguerrera 

This is a prototype for how we will extract localizable strings from existing product code. I plan to use this as a template for folks who will help us do the full extraction. Focus on design, naming, usability... this pattern will be all over the place.

### DIFF
--- a/src/dotnet/commands/dotnet-new/LocalizableStrings.cs
+++ b/src/dotnet/commands/dotnet-new/LocalizableStrings.cs
@@ -1,0 +1,35 @@
+namespace Microsoft.DotNet.Tools.New
+{
+    internal static class LocalizableStrings
+    {
+        public static const string AppName = "dotnet new";
+
+        public static const string AppFullName = ".NET Initializer";
+
+        public static const string AppDescription = "Initializes empty project for .NET Platform"; 
+
+        public static const string CreateFailedBecauseProjectExists = "Creating new {0} project failed, project already exists.";
+
+        public static const string CreateFailedBecauseContentExists = "Creating new {0} project failed, directory already contains {1}";
+
+        public static const string CreateFailed = "Creating new {0} project failed.";
+
+        public static const string CreateSucceeded = "Created new {0} project in {1}.";
+
+        public static const string LanguageOptionValueName = "LANGUAGE";
+
+        public static const string LanguageOptionDescription = "Language of project    Valid values: {0}.";
+
+        public static const string LanguageOptionUnrecognizedValue = "Unrecognized language: {languageValue}";
+
+        public static const string TypeOptionValueName = "TYPE";
+
+        public static const string TypeOptionDescription = "Type of project        {0}";
+
+        public static const string TypeOptionUnrecognizedValue = "Unrecognized type: {typeValue}";
+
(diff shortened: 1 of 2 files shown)


src/dotnet/commands/dotnet-new/LocalizableStrings.cs
```diff
@@ -0,0 +1,35 @@
+namespace Microsoft.DotNet.Tools.New
+{
+    internal static class LocalizableStrings
+    {
+        public static const string AppName = "dotnet new";
+
```
krwq: do we need the extra spaces? I don't think they make it easier to read

src/dotnet/commands/dotnet-new/LocalizableStrings.cs
```diff
@@ -0,0 +1,35 @@
...
+        public static const string AppFullName = ".NET Initializer";
+
+        public static const string AppDescription = "Initializes empty project for .NET Platform"; 
+
+        public static const string CreateFailedBecauseProjectExists = "Creating new {0} project failed, project already exists.";
+
+        public static const string CreateFailedBecauseContentExists = "Creating new {0} project failed, directory already contains {1}";
+
+        public static const string CreateFailed = "Creating new {0} project failed.";
```
krwq: When I look only on this string I believe it is hard to read the context how to translate this sentence (I wouldn't know how to translate this to Polish without more context or some example word I can put there) - I think it should be something like {ProgrammingLanguageName} or at least a comment explaining: {0} -> i.e. "C#"

piotrpMSFT: We'll get into the nitt-gritty of translation and adjust. For now we are focusing on the calling convention from CLI into the localization objects. I agree we'll likely need to give help, but it's unclear what that format will be.

src/dotnet/commands/dotnet-new/LocalizableStrings.cs
piotrpMSFT: I don't want to change style subjectively. Distinct statements get a line in between to show they're not meaningfully related. Consistency avoids subjective fights over "I don't think these are related!"

src/dotnet/commands/dotnet-new/LocalizableStrings.cs
nguerrera: The loc team will file bugs if they need clarification.

src/dotnet/commands/dotnet-new/LocalizableStrings.cs
nguerrera: In what context is "dotnet new" localizable. Isn't it what I always have to type?

nguerrera: Also, I'm sure you've seen the build errors, but it should be just `const` not `static const` ;)

nguerrera:
:+1:

piotrpMSFT:
I'll close this PR at the end of the day, once everyone has had a chance to comment. Will then come back with the `Real` PR.

Issue #4739, opened 2016-11-16
Migration: Delete .xproj.user file


dasMulli:
## Steps to reproduce
* Have a pj/xproj solution you have worked on in VS
   * Alt: Forget a proper gitignore and end up with *.user files in your source control.
* Run `dotnet migrate`

## Expected  behavior
Only a csproj file in project folder.

## Actual behavior
There is a leftover `.xproj.user` file in the directory

## Environment data
`dotnet --info` output:
```
.NET Command Line Tools (1.0.0-preview4-004079)

Product Information:
 Version:            1.0.0-preview4-004079
 Commit SHA-1 hash:  43dfa6b8ba

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.11-x64
```

piotrpMSFT:
Sure, we can move this one to backup as well.

PR #4741, opened 2016-11-16
add *.xproj.user to backup file types

piotrpMSFT:
Fixes https://github.com/dotnet/cli/issues/4739

/cc @jgoshi @livarcocc @krwq

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -296,3 +296,8 @@ test/PackagedCommands/Consumers/*/project.json
 
 # VS generated files
 launchSettings.json
+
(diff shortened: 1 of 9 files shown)


dasMulli:
@piotrpMSFT  On a side-note: I also ended up with `project.lock.json` still being there.. If you're at it, consider it a weakly-filed issue.

piotrpMSFT:
user file was an easier sell since just deleting it is a little destructive IIRC. Feel free to PR the lock file... previous PR sets the example.

PR #4750, opened 2016-11-16
Extract Localizable strings from dotnet-build

mattscheffer:
This commit extracts localizable strings from dotnet-build and leaves them in a separate file. The majority of the strings from the file are extracted except for tags or MSbuild arguments. If any of those need to be extracted, please let me know.

### DIFF
--- a/src/dotnet/commands/dotnet-build/LocalizableStrings.cs
+++ b/src/dotnet/commands/dotnet-build/LocalizableStrings.cs
@@ -22,9 +22,7 @@
 
         public const string OutputOptionName = "OUTPUT_DIR";
 
-        public const string ProjectArgumentDescription1 = "The MSBuild project file to build. If a project file is not specified,";
-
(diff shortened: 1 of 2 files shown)


piotrpMSFT:

piotrpMSFT:

piotrpMSFT:

mattscheffer:
Thanks for throwing re-tests at those @piotrpMSFT . Are they generally flaky? Can you and/or @nguerrera give approval?

nguerrera:
Just so all reviewers have context, we're starting by extracting the strings to constants because we need some work to get resx -> c# working outside VS and x plat. Once we have that, we'll convert to resx (can be done with a one off script) and there will be no changes needed to the string consuming code.

src/dotnet/commands/dotnet-build/LocalizableStrings.cs
```diff
@@ -0,0 +1,38 @@
+﻿namespace Microsoft.DotNet.Tools.Build
+{
+    internal class LocalizableStrings
```
nguerrera: Nit: should be internal static class, but it doesn't matter since these files won't live long.

PR #4780, opened 2016-11-18
Web project template now includes content in "views" and "wwwroot" fo…

robertsundstrom:
skipciplease

In the generated "Web" project file, the directories "Views" and "wwwroot" are not shown when edited in an IDE (VS for Mac etc.), so I added them as Content.

Probably a project system issue. Is there a planned approach? Or is this there another preferred way to add them?

### DIFF
--- a/src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
+++ b/src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
@@ -13,6 +13,8 @@
 
   <ItemGroup>
     <Compile Include="**\*.cs" Exclude="$(GlobalExclude)" />
+    <Content Include="Views\**\*" Exclude="$(GlobalExclude)" />
+    <Content Include="wwwroot\**\*" Exclude="$(GlobalExclude)" />
     <EmbeddedResource Include="**\*.resx" Exclude="$(GlobalExclude)" />
   </ItemGroup>
 


piotrpMSFT:
@mlorbetske can you take a look?

mlorbetske:
The patterns for picking up those files when in an IDE are carried by the Microsoft.Net.SDK.Web package (more accurately, the project system package it pulls in). When you saw the files missing in the IDE, had NuGet restore completed?

piotrpMSFT:
It sounds like this change may not be needed. @mlorbetske can you confirm?

mlorbetske:
That's correct, these changes should not be required if restore completed.

piotrpMSFT:
@robertsundstrom I'll close the PR based on @mlorbetske's feedback. If you find this is still failing after a restore please do open an issue and tag us both, we'll look into it asap.

robertsundstrom:
Sorry for not being active in the discussion in my own pull request. 

Yes, I thought that the tooling would handle that. I loaded the project in VS for Mac Preview, but the directories did not appear when restoring its dependencies. But it is just a preview, after all. Maybe it would have worked if I created the project in VS for Mac. But that would seem strange.

piotrpMSFT:
@robertsundstrom can you open an issue specific to the VS4Mac issue? This is actually our first bug specific to that IDE and I want to use your issue to figure out how we get the right engagement on those.

mrward:
I have opened a bug against VS for Mac on [bugzilla](https://bugzilla.xamarin.com/show_bug.cgi?id=47825). The ASP.NET Core support in VS for Mac is missing a lot of features in the first preview.

PR #4858, opened 2016-11-28
Adding a library for working with sln files

jgoshi:


src/Microsoft.DotNet.Cli.Sln/FileManpulation/ByteOrderMark.cs
```diff
@@ -0,0 +1,121 @@
+//
```
krwq: This file seems to be copied from somewhere. Also I believe you should `git commit --author "Zlatko ...."`

krwq:
1. Please put all the files from the other repo to a single commit
2. Zlatko's changes should be a diff of what he actually changed and commited as him (git commit --author ...)
3. We need to figure out if the license is all right

PR #4911, opened 2016-12-04
Fix output race

piotrpMSFT:
TestCommand starts the test process before wiring up stderr & stdout. This change delays process start until after the wireup is finished so that the test process cannot shut down before we have wired up output redirection.

Note, we haven't seen this failure on OS's I use regularly, so pushing to CI without local run.

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/CommandResolution/ProjectToolsCommandResolver.cs
@@ -248,14 +248,12 @@ namespace Microsoft.DotNet.Cli.Utils
 
             try
             {
-                File.Copy(tempDepsFile, depsPath);
+                File.Move(tempDepsFile, depsPath);
             }
             catch (Exception e)
             {
                 Reporter.Verbose.WriteLine($"unable to generate deps.json, it may have been already generated: {e.Message}");
-            }
-            finally
-            {
+                
                 try
                 {
                     File.Delete(tempDepsFile);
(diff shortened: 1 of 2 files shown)


piotrpMSFT:
@MattGertz @livarcocc @jgoshi @krwq I'd love to get approvals for this ASAP :) 

The PR addresses two issues that have been plaguing build stability since RC1:
- Tests failing due to StreamForwarder "Collection alerady disposed"
- Tests failing due to malformed deps.json

Both of these issues were caused by fairly subtle races across processes. 

The fix for the first is a patch-up, with a real fix requiring a fairly big design change to our Stream Forwarding capability. This is tracked by #4913 

The second fix is much more complete, though not bulletproof. File.Move is typically atomic, unlike File.Copy which is always non-atomic. We can still see issues when folks' tmp directory sits on a different physical disk than their nuget package cache AND they try to run a particular extension simultaneously in multiple processes AND the extension in question is being run for the first time. A proper fix for this will require something akin to NuGet's filesystem-based semaphore system [likely exactly that system] but the change would take on a new conceptual dependency and be much more complicated to validate, so leaving it off until we get real reports of folks being troubled by this.

MattGertz:
Doing string comparisons in the exception handlers seems gross.  I take it that these exceptions don't populate the HResult member?

piotrpMSFT:
@MattGertz good point. Unfortunately I only have CI repro at the moment so working with what we know...

piotrpMSFT:
I'm also concerned about stability of hresults across platforms. Since this is test code and we dont have a local repro id really like to get this approved and merged @MattGertz.

MattGertz:
OK, but I want to revisit this later.

PR #4912, opened 2016-12-04
Add missing SDKs

mlorbetske:
@rohit21agrawal @piotrpMSFT @vijayrkn @srivatsn 

NuGet.BuildTasks.Pack 4.0.0-rc2
Microsoft.NET.Sdk.Publish (parameterized version)
Microsoft.NET.Sdk.Web.ProjectSystem (parameterized version)

PR #4915, opened 2016-12-04
Adjust the C# web template for the SDK attribute

mlorbetske:
Use the attribute
Remove the initial and final imports
Extend GlobalExclude
Update the include patterns
Remove the PackageReference for the SDK

@piotrpMSFT @phenning @vijayrkn @srivatsn @MattGertz

### DIFF
--- a/src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
+++ b/src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
@@ -1,6 +1,4 @@
-<Project ToolsVersion="15.0">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
-
+<Project ToolsVersion="15.0" Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp1.0</TargetFramework>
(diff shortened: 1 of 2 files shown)


src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
```diff
@@ -11,16 +9,22 @@
     <PackageTargetFallback>$(PackageTargetFallback);portable-net45+win8+wp8+wpa81;</PackageTargetFallback>
```
davidfowl: Why do we still need these imports?

mlorbetske: Scaffolding I think, @prafullbhosale is that right?

mlorbetske:
This change depends on https://github.com/dotnet/cli/pull/4912

src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
phenning: Correct, we add it to the base StarterWeb.csproj file in the vs templates since we don't have a way to conditionally inject this via a template engine rule and because this import should go away after we eventually move to the later runtime packages according to @prafullbhosale

phenning:
I think we will want to update https://github.com/dotnet/cli/blob/d2d0353e60919faffecc4052040a0817c18286cd/build/Microsoft.DotNet.Cli.BundledSdks.props
for build 1203-1, as that is what the corresponding VS build will have,

phenning:
Other then potentially updating the SDK version, looks good!

MattGertz:
Approved pending tests passing.

mlorbetske:
@phenning the SDK version is updated in the PR this depends on

phenning:
Looks good then!

piotrpMSFT:
doesn't look like CI is coming back tonight. Try again in the morning...

src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
prafullbhosale: Yes, Scaffolding needs it because of Microsoft.Composition.

piotrpMSFT:
@dotnet-bot test this please

piotrpMSFT:

PR #4917, opened 2016-12-04
Rename centralized LocalizableStrings file

piotrpMSFT:
Rename LocalizableStrings in CLI root directory to CommonLocalizableStrings to avoid conflict with command-specific loc files.


piotrpMSFT:
@MattGertz

Issue #4513, opened 2016-10-27
Add dotnet add|list|remove project commands for solution files


blackdwarf:
The CLI needs to provide a way for users to manipulate the solution file (SLN file). 

Below is a mini-spec of the command. 

### Usage
`dotnet add <project_or_solution> project <project>

### Top level Verbs

* `add`
* `list`
* `remove`

### Arguments

- &lt;project_or_solution&gt;
  - A project or a solution path 
- &lt;project&gt;
  - Path to the project file to include in the SLN file. Has to be a relative path. 

### Defaults
-   If the &lt;project&gt; is not present, we will try to see if there is a project in the current directory; if there is, it will be added, if not, an error is displayed
- If the &lt;project_or_solution&gt; is not present, we will check the current directory for either a project file or a sln file. 
  - If neither are found, an error is displayed
  - If more than one is found or if there is both sln and a project file present, an error is displayed prompting the user to specify which file the operation should work on. 

### Notes
* This command adds a project to a solution if it works on an SLN file. In case a project file, it will add a `<ProjectReference>` to the project in question. 
* The command will make absolute paths of the project being added relative to the solution file/project being added. This is done to avoid messy breaks when moving machines. 

### Examples
- `dotnet add project "../path/to/project.csproj"`
  - Add a project found at the specified path to the solution file
- `dotnet remove project "../path/to/project.csproj"`
  - Remove a project found at the specified path to the solution file
- `dotnet list path/to/solution.sln project`
  - List out all of the projects and configurations from the solution;

alexkreskiyan:
@blackdwarf If the following makes sense and is not too late yet =)

Would be nice to operate not only sln file and not only VS-specific stuff.
At least, would be cool to have one commang group, providing through specific commands possibility to generate minimalistic environment for solution.

We have `global.json` and several `project.json` files. So, it's possible to locate where to place `.sln` and `.xproj` files.

### Examples
- `dotnet editor vs` - in folder with `global.json`. Creates `.sln`/`.xproj`/ or refreshes them carefully (for case of manually changed `.sln`/`.xproj`
- `dotnet editor vs -f "path/to/global.json"` - in other folder
- `dotnet editor vs add <project_name>` - as far as it's possible to evaluate available names through `global.json` and `project.json`
- `dotnet editor vs remove <project_name>` - same

[...]

PR #4978, opened 2016-12-09
Implement dotnet add proj

jgoshi:

### DIFF
--- a/TestAssets/NonRestoredTestProjects/DotnetAddProjProjects/Invalid.sln
+++ b/TestAssets/NonRestoredTestProjects/DotnetAddProjProjects/Invalid.sln
@@ -1,0 +1,1 @@
+Testing an invalid solution file
--- a/src/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs
+++ b/src/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs
@@ -40,6 +40,8 @@ namespace Microsoft.DotNet.Cli.Sln.Internal
 {
     public class SlnFile
     {
+        public const string CSharpProjectTypeGuid = "{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}";
+
         SlnProjectCollection projects = new SlnProjectCollection();
         SlnSectionCollection sections = new SlnSectionCollection();
         SlnPropertySet metadata = new SlnPropertySet(true);
@@ -99,12 +101,22 @@ namespace Microsoft.DotNet.Cli.Sln.Internal
 
         static Regex slnVersionRegex = new Regex(@"Microsoft Visual Studio Solution File, Format Version (\d?\d.\d\d)");
 
+        public string FileName
+        {
+            get { return FilePath.FileName; }
+        }
+
+        public string FullPath
+        {
+            get { return FilePath.FullPath; }
+        }
+
         /// <summary>
         /// The directory to be used as base for converting absolute paths to relative
         /// </summary>
-        public FilePath BaseDirectory
+        public string BaseDirectory
         {
-            get { return FileName.ParentDirectory; }
+            get { return FilePath.ParentDirectory.FullPath; }
         }
 
         /// <summary>
@@ -135,11 +147,11 @@ namespace Microsoft.DotNet.Cli.Sln.Internal
             get { return projects; }
         }
 
-        public FilePath FileName { get; set; }
+        private FilePath FilePath { get; set; }
 
         public void Read(string file)
         {
-            FileName = file;
+            FilePath = file;
             format = FileUtil.GetTextFormatInfo(file);
 
             using (var sr = new StreamReader(new FileStream(file, FileMode.Open)))
@@ -213,12 +225,15 @@ namespace Microsoft.DotNet.Cli.Sln.Internal
                 throw new InvalidSolutionFormatException(curLineNum, "File header is missing");
         }
 
-        public void Write(string file)
+        public void Write(string file = null)
         {
-            FileName = file;
+            if (!string.IsNullOrEmpty(file))
+            {
+                FilePath = file;
+            }
             var sw = new StringWriter();
             Write(sw);
-            File.WriteAllText(file, sw.ToString());
+            File.WriteAllText(FullPath, sw.ToString());
         }
 
         public void Write(TextWriter writer)
--- a/src/dotnet/MsbuildSolution.cs
+++ b/src/dotnet/MsbuildSolution.cs
@@ -102,10 +102,11 @@ namespace Microsoft.DotNet.Tools
         {
             int numberOfAddedProjects = 0;
 
-            foreach (var projectPath in projects.Select((p) => MsbuildUtilities.NormalizeSlashes(p)))
+            foreach (var projectPath in projects)
             {
+                var projectPathNormalized = MsbuildUtilities.NormalizeSlashes(projectPath);
                 if (_slnFile.Projects.Any((p) =>
-                     string.Equals(p.FilePath, projectPath, StringComparison.OrdinalIgnoreCase)))
+                     string.Equals(p.FilePath, projectPathNormalized, StringComparison.OrdinalIgnoreCase)))
                 {
                     Reporter.Output.WriteLine(string.Format(
                         CommonLocalizableStrings.SolutionAlreadyHasAProject,
@@ -140,7 +141,7 @@ namespace Microsoft.DotNet.Tools
                         Id = projectGuid.ToString("B").ToUpper(),
                         TypeGuid = SlnFile.CSharpProjectTypeGuid,
                         Name = Path.GetFileNameWithoutExtension(projectPath),
-                        FilePath = projectPath
+                        FilePath = projectPathNormalized
                     };
 
                     _slnFile.Projects.Add(slnProject);
--- a/src/dotnet/MsbuildSolution.cs
+++ b/src/dotnet/MsbuildSolution.cs
@@ -1,0 +1,162 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
+using Microsoft.DotNet.Cli.Sln.Internal;
+using Microsoft.DotNet.Cli.Utils;
+
+namespace Microsoft.DotNet.Tools
+{
+    internal class MsbuildSolution
+    {
+        private SlnFile _slnFile;
+
+        public string SolutionFullPath
+        {
+            get
+            {
+                return _slnFile.FullPath;
+            }
+        }
+
+        private MsbuildSolution(string solutionPath)
+        {
+            _slnFile = new SlnFile();
+            try
+            {
+                _slnFile.Read(solutionPath);
+            }
+            catch
+            {
+                throw new GracefulException(CommonLocalizableStrings.InvalidSolution, solutionPath);
+            }
+        }
+
+        public static MsbuildSolution FromFileOrDirectory(string fileOrDirectory)
+        {
+            if (File.Exists(fileOrDirectory))
+            {
+                return FromFile(fileOrDirectory);
+            }
+            else
+            {
+                return FromDirectory(fileOrDirectory);
+            }
+        }
+
+        private static MsbuildSolution FromFile(string solutionPath)
+        {
+            if (!File.Exists(solutionPath))
+            {
+                throw new GracefulException(CommonLocalizableStrings.SolutionDoesNotExist, solutionPath);
+            }
+
+            return new MsbuildSolution(solutionPath);
+        }
+
+        private static MsbuildSolution FromDirectory(string solutionDirectory)
+        {
+            DirectoryInfo dir;
+            try
+            {
+                dir = new DirectoryInfo(solutionDirectory);
+            }
+            catch (ArgumentException)
+            {
+                throw new GracefulException(CommonLocalizableStrings.CouldNotFindSolutionOrDirectory, solutionDirectory);
+            }
+
+            if (!dir.Exists)
+            {
+                throw new GracefulException(CommonLocalizableStrings.CouldNotFindSolutionOrDirectory, solutionDirectory);
+            }
+
+            FileInfo[] files = dir.GetFiles("*.sln");
+            if (files.Length == 0)
+            {
+                throw new GracefulException(CommonLocalizableStrings.CouldNotFindSolutionIn, solutionDirectory);
+            }
+
+            if (files.Length > 1)
+            {
+                throw new GracefulException(CommonLocalizableStrings.MoreThanOneSolutionInDirectory, solutionDirectory);
+            }
+
+            FileInfo solutionFile = files.First();
+
+            if (!solutionFile.Exists)
+            {
+                throw new GracefulException(CommonLocalizableStrings.CouldNotFindSolutionIn, solutionDirectory);
+            }
+
+            return new MsbuildSolution(solutionFile.FullName);
+        }
+
+        public int AddProjectToSolution(IEnumerable<string> projects)
+        {
+            int numberOfAddedProjects = 0;
+
+            foreach (var projectPath in projects.Select((p) => MsbuildUtilities.NormalizeSlashes(p)))
+            {
+                if (_slnFile.Projects.Any((p) =>
+                     string.Equals(p.FilePath, projectPath, StringComparison.OrdinalIgnoreCase)))
+                {
+                    Reporter.Output.WriteLine(string.Format(
+                        CommonLocalizableStrings.SolutionAlreadyHasAProject,
+                        SolutionFullPath,
+                        projectPath));
+                }
+                else
+                {
+                    string projectGuidString = null;
+                    if (File.Exists(projectPath))
+                    {
+                        var projectElement = ProjectRootElement.Open(
+                            projectPath,
+                            new ProjectCollection(),
+                            preserveFormatting: true);
+
+                        var projectGuidProperty = projectElement.Properties.Where((p) =>
+                            string.Equals(p.Name, "ProjectGuid", StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
+
+                        if (projectGuidProperty != null)
+                        {
+                            projectGuidString = projectGuidProperty.Value;
+                        }
+                    }
+
+                    var projectGuid = (projectGuidString == null)
+                        ? Guid.NewGuid()
+                        : new Guid(projectGuidString);
+
+                    var slnProject = new SlnProject
+                    {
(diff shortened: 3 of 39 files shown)


jgoshi:
Fixes #4513 (partial fix to implement dotnet add proj)

jgoshi:
test Ubuntu x64 Release Build

jgoshi:

test/dotnet-add-p2p.Tests/GivenDotnetAddP2P.cs
```diff
@@ -13,45 +12,13 @@ namespace Microsoft.DotNet.Cli.Add.P2P.Tests
 {
     public class GivenDotnetAddP2P : TestBase
     {
+        const string TestGroup = "NonRestoredTestProjects";
+        const string ProjectName = "DotnetAddP2PProjects";
```
piotrpMSFT: I really don't like this growing pattern of "throw all my pet projects in a single test asset". It makes it very difficult to understand the intent of a test asset, to reuse them across scenarios. It also makes for messier unit tests since we specify the actual asset under test in two places.

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+        const string ProjectName = "DotnetAddProjProjects";
+
+        [Theory]
+        [InlineData("--help")]
+        [InlineData("-h")]
+        public void WhenHelpOptionIsPassedItPrintsUsage(string helpArg)
+        {
+            var cmd = new AddProjCommand().Execute(helpArg);
+            cmd.Should().Pass();
+            cmd.StdOut.Should().Contain("Usage");
```
piotrpMSFT: Not to solve here, but I'm really worried about this when we get localized text...

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+        [Theory]
+        [InlineData("Invalid.sln")]
+        [InlineData("")]
+        public void WhenInvalidSolutionIsPassedItPrintsErrorAndUsage(string slnName)
+        {
+            var setup = CreateTestSetup(TestGroup, ProjectName);
+
+            var cmd = new AddProjCommand()
+                .WithWorkingDirectory(setup.TestRoot)
+                .WithSolution(slnName)
```
piotrpMSFT: I'm nervous this will lead to too much magic over time. The helper methods are nice, but they make it really tough to figure out "What was actually executed on the command line that hit the bug I'm seeing" or to point folks at tests and say "Here's how the scenario works".

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            cmd.StdOut.Should().Contain("Usage");
+        }
+
+        [Fact]
+        public void WhenNoProjectIsPassedItPrintsErrorAndUsage()
+        {
+            var setup = CreateTestSetup(TestGroup, ProjectName);
+
+            var cmd = new AddProjCommand()
+                .WithWorkingDirectory(Path.Combine(setup.TestRoot, "MoreThanOne"))
```
piotrpMSFT: This is a good example. The Unit Test now has two specifiers for the actual project under test. What's more, the second specifier requires a Path.Combine to get it right. The concept count in this test is strictly increased because of the TestSetup.

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            cmd.StdOut.Should().Contain("Usage");
+        }
+
+        [Fact]
+        public void WhenMoreThanOneSolutionExistsInTheDirectoryItPrintsErrorAndUsage()
+        {
+            var setup = CreateTestSetup(TestGroup, ProjectName);
+
+            var cmd = new AddProjCommand()
+                .WithWorkingDirectory(Path.Combine(setup.TestRoot, "MoreThanOne"))
```
piotrpMSFT: Here too

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+                .Execute($"\"{setup.GetProjectFullPath("Lib")}\"");
```
piotrpMSFT: And here

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+        [InlineData("Lib", "", true)]
+        [InlineData("LibWithProjectGuid", "{84A45D44-B677-492D-A6DA-B3A71135AB8E}", true)]
+        public void ItAddsProjectAndPrintsStatus(string projectName, string projectGuid, bool makeRelative)
+        {
+            var setup = CreateTestSetup(TestGroup, ProjectName);
+            string slnDir = Path.Combine(setup.TestRoot, "MoreThanOne");
+            string slnName = "App.sln";
+
+            var projectToAdd = makeRelative
+                ? $"\"..{Path.DirectorySeparatorChar}{projectName}{Path.DirectorySeparatorChar}{projectName}.csproj\""
```
piotrpMSFT: Why not PathUtilities.MakeRelative?

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+                .WithWorkingDirectory(slnDir)
+                .WithSolution(slnName)
+                .Execute(projectToAdd);
+            cmd.Should().Pass();
+            cmd.StdOut.Should().Contain("added to the solution");
+            cmd.StdErr.Should().BeEmpty();
+
+            var slnFile = new SlnFile();
+            slnFile.Read(Path.Combine(slnDir, slnName));
+            VerifyProjectInSolution(slnFile, projectName, projectGuid);
```
piotrpMSFT: slnFile.Projects.Should().Contain(projectName, projectGuid)

These kinds of Verify functions hide what the test is actually validating, making it difficult to review and verify.

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            Restore(Path.Combine(setup.TestRoot, "App"), "App.csproj");
```
piotrpMSFT: If this were using TestAssets, instead of TestAssetsManger, and didn't 'tostring' the Root, you could:

```
   Root.GetDirectory("App")
           .GetFile("App.csproj);
```

Quite a bit more clear of what is going on...

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+        {
+            var setup = CreateTestSetup(TestGroup, ProjectName);
+            string slnDir = Path.Combine(setup.TestRoot, "MoreThanOne");
+            string slnName = "AppAndLib.sln";
+            string slnFullPath = Path.Combine(slnDir, slnName);
+
+            var contentBefore = File.ReadAllText(slnFullPath);
+            var cmd = new AddProjCommand()
+                .WithWorkingDirectory(slnDir)
+                .WithSolution(slnName)
```
piotrpMSFT: We're using an absolute path in one place and a relative path in another. Is that meaningful to the test? Otherwise, it's extra cognitive load that we could eliminate...

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+                .Execute($"\"{setup.GetProjectFullPath("Lib")}\"");
+            cmd.Should().Pass();
+            cmd.StdOut.Should().Contain("already contains project");
```
piotrpMSFT: @blackdwarf I think returning '0' in this scenario makes sense but want to get your sense on the broader UX. If we're told to add something, and its already added, is that an error or is that a success?

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            cmd.StdErr.Should().BeEmpty();
+
+            var contentAfter = File.ReadAllText(slnFullPath);
+            contentAfter.Should().BeEquivalentTo(contentBefore);
```
piotrpMSFT: I'd reduce this to:

```
File.ReadAllText(...)
  .Should().BeEquivalentTo(contentBefore);
```

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            var setup = CreateTestSetup(TestGroup, ProjectName);
+            string slnDir = Path.Combine(setup.TestRoot, "MoreThanOne");
+            string slnName = "App.sln";
+            string slnFullPath = Path.Combine(slnDir, slnName);
+
+            var contentBefore = File.ReadAllText(slnFullPath);
+            var cmd = new AddProjCommand()
+                .WithWorkingDirectory(slnDir)
+                .WithSolution(slnName)
+                .Execute($"\"{setup.GetProjectFullPath("Lib")}\" \"IDoNotExist.csproj\"");
```
piotrpMSFT: @blackdwarf thoughts here? Should we fail the full operation, or add the ones that exist and report on the ones we failed to add? While I like the determinism here, I can imagine being frustrated if I have to navigate through an already long string to fix something broken in the middle.

I also think we should talk about the "add many" UX pattern. It creates a bunch of ambiguous UX decisions like this one, and I cannot imagine myself using it except in scripts... which may as well invoke the command multiple times...

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            string slnFullPath = Path.Combine(slnDir, slnName);
+
+            var contentBefore = File.ReadAllText(slnFullPath);
+            var cmd = new AddProjCommand()
+                .WithWorkingDirectory(slnDir)
+                .WithSolution(slnName)
+                .Execute("--force \"IDoNotExist.csproj\"");
+            cmd.Should().Pass();
+            cmd.StdOut.Should().Contain("added to the solution");
+            cmd.StdErr.Should().BeEmpty();
```
piotrpMSFT: Help text said something about relative paths, but I don't see that reflected in the test...

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            var projects = slnFile.Projects
+                .Where((p) => p.Name == projectName)
+                .ToList();
+
+            projects.Count.Should().Be(1);
+            if (!string.IsNullOrEmpty(projectGuid))
+            {
+                projects[0].Id.Should().Be(projectGuid);
+            }
+            projects[0].TypeGuid.Should().Be(SlnFile.CSharpProjectTypeGuid);
```
piotrpMSFT: Actually, we need to chat about the project guid. I'm sure @enricosada will have some thoughts about using a csproj GUID for fsproj...

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            projects[0].FilePath.Should().Be($"..\\{projectName}\\{projectName}.csproj");
+        }
+
+        private void Restore(string projectDirectory, string projectName)
+        {
+            var command = new RestoreCommand()
```
piotrpMSFT: Again, we should be using TestAssets. It already handles restore for us, and preserves restore artifacts to speed up test execution.

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+                .WithWorkingDirectory(projectDirectory);
+
+            if (!Path.HasExtension(projectName))
+            {
+                projectName += ".csproj";
```
piotrpMSFT: why?

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,236 @@
...
+            }
+
+            command.Execute($"{projectName}")
+                .Should().Pass();
+        }
+
+        private void Build(string solutionDirectory, string solutionName, string configuration = "Debug")
+        {
+            if (!Path.HasExtension(solutionName))
```
piotrpMSFT: TestAssets.

test/dotnet-add-proj.Tests/MSBuild.exe
```diff
@@ -0,0 +1 @@
+https://github.com/Microsoft/msbuild/issues/927
```
piotrpMSFT: newline

test/dotnet-add-proj.Tests/MSBuild.exe.config
```diff
@@ -0,0 +1 @@
+https://github.com/Microsoft/msbuild/issues/927
```
piotrpMSFT: and here

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
enricosada: @piotrpMSFT is right, as italian i always get issues with test suites about that.
Problem is, it's difficult to change default language unless it's done INSIDE the called program, usually with a flag (for exaple csc/fsc use `--preferred-lang`).
If someone have a better solution, i'd like to hear it, becaue i have the same issue in other suites

TestAssets/NonRestoredTestProjects/DotnetAddProjProjects/Invalid.sln
```diff
@@ -0,0 +1 @@
+Testing an invalid solution file
```
jgoshi: You want a newline? Is there a particular reason? The file is supposed to represent an invalid sln file...

src/dotnet/MsbuildSolution.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Linq;
+
+using Microsoft.Build.Construction;
+using Microsoft.Build.Evaluation;
+using Microsoft.DotNet.Cli.Sln.Internal;
+using Microsoft.DotNet.Cli.Utils;
+
+namespace Microsoft.DotNet.Tools
+{
+    internal class MsbuildSolution
```
jgoshi: So there is another class (that already existed before this PR) called MsbuildProject. I called this one MsbuildSolution to be consistent.

src/dotnet/MsbuildUtilities.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools.Common;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.DotNet.Tools
+{
+    internal class MsbuildUtilities
```
jgoshi: I can move them to PathUtility

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
enricosada: @jgoshi atm you can use current f# guid `F2A71F9B-5D33-465A-A702-920D77279786` at least is not the c# one 😄

cartermp: I think the current F# project type guid is fine; we're in a bit of a funky situation for a while since we're still on our own project system, but will be moving to the Roslyn project system in the next year (at least that's the plan...).

jgoshi:

jgoshi:
Because this is a UX blocker and I have others (like completing the CRUD operations and changing help text) I want to merge this and clean up later. Right after the UX blockers are in. Here is the tracking issue:
https://github.com/dotnet/cli/issues/4989

src/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs
```diff
@@ -40,6 +40,8 @@ namespace Microsoft.DotNet.Cli.Sln.Internal
 {
     public class SlnFile
     {
+        public const string CSharpProjectTypeGuid = "{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}";
```
krwq: @piotrpMSFT const property? do you mean static readonly?

test/Microsoft.DotNet.Tools.Tests.Utilities/TestBase.cs
```diff
@@ -68,6 +69,20 @@ protected static TestAssetsManager GetTestGroupTestAssetsManager(string testGrou
             return testAssetsMgr;
         }
 
+        protected static TestSetup CreateTestSetup(
+            string testGroup,
+            string projectName,
+            [CallerMemberName] string callingMethod = nameof(CreateTestSetup),
```
krwq: [CallerMemberName] will have precedence here - default value is required when you use CallerMemberName

test/Microsoft.DotNet.Tools.Tests.Utilities/TestBase.cs
```diff
@@ -178,5 +193,32 @@ private bool IsPortable(string executablePath)
...
+            {
+                new NewCommand()
+                    .WithWorkingDirectory(dir.Path)
+                    .ExecuteWithCapturedOutput("-t Lib")
+                .Should().Pass();
+            }
+            catch (System.ComponentModel.Win32Exception e)
+            {
+                throw new Exception(
+                    $"Intermittent error in `dotnet new` occurred when running it in dir `{dir.Path}`\nException:\n{e}");
```
krwq: that is some combination of dotnet new and test assets manager doing some stuff at the same time

test/dotnet-add-p2p.Tests/GivenDotnetAddP2P.cs
krwq: well, this is testing adding references - having them in separate folders is also not good

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
krwq: I've done similar comment on my PR when adding it and was reluctant to test any exception messages. I believe we can enforce locale in our tests though or do something similar to what corefx does - when building Debug we always get English

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
krwq: the first two were initially used for most common projects used by almost every other test in dotnet add p2p

PR #5022, opened 2016-12-14
Implement dotnet add project

jgoshi:


test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,200 @@
...
+                .ExecuteWithCapturedOutput($"add InvalidSolution.sln project {projectToAdd}");
+            cmd.Should().Fail();
+            cmd.StdErr.Should().Contain("Invalid solution ");
+            cmd.StdOut.Should().Contain("Usage:");
+        }
+
+        [Fact]
+        public void WhenInvalidSolutionIsFoundItPrintsErrorAndUsage()
+        {
+            var projectDirectory = TestAssetsManager.CreateTestInstance("InvalidSolution")
```
piotrpMSFT: Consider TestAssets.cs instead of TestAssetsManager.cs. I'm putting together a PR that moves us over in bulk, but it may be helpful to you here to use the new impl.

jgoshi:
@jonsequitur

src/dotnet/SlnFileFactory.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private static SlnFile FromFile(string solutionPath)
+        {
+            SlnFile slnFile = null;
+            try
+            {
+                slnFile = SlnFile.Read(solutionPath);
+            }
+            catch
+            {
```
jonsequitur: What kinds of errors are we expecting here? It looks like `SlnFile.Read` could potentially throw a few different types of exceptions, not all of which would be accurately described by `CommonLocalizableStrings.InvalidSolution`.

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
jgoshi: I'll wait to see your PR, then fix these while I'm implementing more tests (like the dotnet list or remove project).

src/dotnet/SlnFileFactory.cs
jgoshi: I wanted to provide a general error in case of any failure. We also have unit tests for the SlnFile reader (and I have an issue open to add more comprehensive tests). So basically whenever we cannot open the file for now we just say invalid solution until we have the ability to detect specific cases.

src/dotnet/SlnFileFactory.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            if (files.Length > 1)
+            {
+                throw new GracefulException(
+                    CommonLocalizableStrings.MoreThanOneSolutionInDirectory,
+                    solutionDirectory);
+            }
+
+            FileInfo solutionFile = files.First();
```
jonsequitur: `files.Single()` would be a bit more correct and robust in the event of upstream code changes, e.g. in the absence of line 68 - 73, this line would do the wrong thing instead of throwing.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
```diff
@@ -95,7 +95,8 @@ internal static CommandLineApplication CreateApplication(CommandLineApplication
                         }
                     }
 
-                    msbuildProj.ConvertPathsToRelative(ref references);
+                    references = references.Select((r) =>
+                        PathUtility.GetRelativePath(msbuildProj.ProjectDirectory, Path.GetFullPath(r))).ToList();
```
jonsequitur: This is effectively changing the meaning of this variable from (implictly) something like `absolutePathReferences` to `relativePathReferences`. Two variables with explicit names would make the code clearer.

src/dotnet/commands/dotnet-add/dotnet-add-proj/LocalizableStrings.cs
```diff
@@ -0,0 +1,11 @@
+namespace Microsoft.DotNet.Tools.Add.ProjectToSolution
```
jonsequitur: Copyright header?

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            var cmd = new DotnetCommand()
+                .ExecuteWithCapturedOutput($"add project {helpArg}");
+            cmd.Should().Pass();
+            cmd.StdOut.Should().Contain("Usage");
+        }
+
+        [Theory]
+        [InlineData("idontexist.sln")]
+        [InlineData("ihave?inv@lid/char\\acters")]
```
jonsequitur: If one of these characters were accidentally treated as valid, then this test will hide the defect. Perhaps a separate `InlineData` per invalid character?

src/dotnet/SlnFileFactory.cs
jgoshi: Ok.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jgoshi: Ok

test/dotnet-add-proj.Tests/GivenDotnetAddProj.cs
jgoshi: Ok

jgoshi:

jgoshi:
@MattGertz I got signoff from @piotrpMSFT and addressed the comments from @jonsequitur

PR #5056, opened 2016-12-16
Fix no build option in dotnet test task

codito:


codito:
@livarcocc @piotrpMSFT

PR #5057, opened 2016-12-16
Implement dotnet remove project

jgoshi:
@piotrpMSFT @livarcocc @krwq @jonsequitur @blackdwarf


src/dotnet/CommandLine/CommandLineApplication.cs
```diff
@@ -64,7 +64,21 @@ public CommandLineApplication Command(string name, bool throwOnUnexpectedArg = t
...
+
+        public CommandLineApplication Command(CommandLineApplication command, bool throwOnUnexpectedArg = true)
+        {
+            return Command(command, _ => { }, throwOnUnexpectedArg);
+        }
+
+        public CommandLineApplication Command(
+            CommandLineApplication command,
+            Action<CommandLineApplication> configuration,
+            bool throwOnUnexpectedArg = true)
```
jonsequitur: It's good policy to have null argument checks on these so that if/when a null is passed the exception is thrown at the point where the coding error was made rather than on some later invocation.

src/dotnet/DotNetTopLevelCommandBase.cs
```diff
@@ -3,57 +3,80 @@
...
-            app.Argument(
+            command.Argument(
                 Constants.ProjectOrSolutionArgumentName,
                 CommonLocalizableStrings.ArgumentsProjectOrSolutionDescription);
 
             foreach (var subCommandCreator in SubCommands)
             {
-                subCommandCreator(app);
+                var subCommand = subCommandCreator();
+                command.Command(subCommand);
```
jonsequitur: The purpose of the `Command` method is a bit unclear. This line of code shows the problem by using the same word for subject, object, and verb. Can we come up with some names that clarify the intent?

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
```diff
@@ -13,114 +14,101 @@
...
-                        projectOrDirectory = PathUtility.EnsureTrailingSlash(Directory.GetCurrentDirectory());
-                    }
+            return command;
+        }
 
-                    var projects = new ProjectCollection();
-                    var msbuildProj = MsbuildProject.FromFileOrDirectory(projects, projectOrDirectory);
+        public override int Run(string fileOrDirectory)
+        {
+            ProjectCollection projects = new ProjectCollection();
```
jonsequitur: `var`?

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
```diff
@@ -13,114 +14,101 @@
...
+                throw new GracefulException(CommonLocalizableStrings.SpecifyAtLeastOneReferenceToAdd);
+            }
 
-                    string frameworkString = frameworkOption.Value();
-                    List<string> references = app.RemainingArguments;
-                    PathUtility.EnsureAllPathsExist(references, CommonLocalizableStrings.ReferenceDoesNotExist);
-                    IEnumerable<MsbuildProject> refs = references.Select((r) => MsbuildProject.FromFile(projects, r));
+            string frameworkString = _frameworkOption.Value();
+            PathUtility.EnsureAllPathsExist(RemainingArguments, CommonLocalizableStrings.ReferenceDoesNotExist);
+            IEnumerable<MsbuildProject> refs = RemainingArguments.Select((r) => MsbuildProject.FromFile(projects, r));
```
jonsequitur: Since you're doing multiple enumerations of this in the nested loop below and because `MsBuildProject.FromFile` is presumably doing I/O, there's probably a perf benefit to resolving this `ToArray` or `ToList` here.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
```diff
@@ -13,114 +14,101 @@
...
-                            if (!@ref.CanWorkOnFramework(framework))
-                            {
-                                Reporter.Error.Write(GetProjectNotCompatibleWithFrameworksDisplayString(
+                            Reporter.Error.Write(GetProjectNotCompatibleWithFrameworksDisplayString(
                                     @ref,
-                                    new string[] { frameworkString }));
-                                return 1;
-                            }
+                                    msbuildProj.GetTargetFrameworks().Select((fx) => fx.GetShortFolderName())));
+                            return 1;
```
jonsequitur: Is the intention here to show all of the target frameworks in the error?

src/dotnet/DotNetTopLevelCommandBase.cs
jgoshi: Do you have any suggestions?

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jgoshi: @krwq can you answer?

src/dotnet/CommandLine/CommandLineApplication.cs
jgoshi: I agree with you, but don't think it's necessary to add it here. It should be clear enough to the caller they shouldn't be passing in a null here.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jgoshi: Ok

src/dotnet/CommandLine/CommandLineApplication.cs
jgoshi: If you really feel strongly about this I can make the change. But what I'm saying is there has to be some discretion on when to add the null check vs when not too. We don't have all our methods doing null checks. And I felt this particular one doesn't need it.

jonsequitur: It's more of a code hygiene habit, not critical. Code doesn't need it until there's a bug upstream, at which time it becomes a pain to track down the issue.

jgoshi:

src/dotnet/DotNetTopLevelCommandBase.cs
jonsequitur: Per the behavior of the method, `CreateCommandLineApplication` would be more explicit. It doesn't express the real intention a lot better, though. 

Sometimes a connecting word can help, like:

```
command.Create<what?>For(subCommand);
```

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
krwq: Yes, this should list all of the frameworks in the error

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jonsequitur: I tend to prefer `var` because it makes refactoring easier and I get the sense that the coding convention is shifting in that direction but I'm new here. If you're of the `var`-sometimes opinion, this is still a pretty clear place where you're not losing information by using var, since the type being instantiated is specified right there. The class name is redundant in this example.

src/dotnet/DotNetTopLevelCommandBase.cs
jgoshi: Ok, so you mean change the name of the Command method? To be command.AddSubCommand or something like that?

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jgoshi: Ok, will change that. Thank you both for the insight.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
```diff
@@ -13,114 +14,101 @@
...
+            if (numberOfAddedReferences != 0)
+            {
+                msbuildProj.ProjectRootElement.Save();
+            }
+
+            return 0;
         }
 
-        private static string GetProjectNotCompatibleWithFrameworksDisplayString(MsbuildProject project, IEnumerable<string> frameworksDisplayStrings)
+        private string GetProjectNotCompatibleWithFrameworksDisplayString(MsbuildProject project, IEnumerable<string> frameworksDisplayStrings)
```
krwq: any particular reason for this change?

jonsequitur:
Yup. Method names tend to be easier to understand if they have verbs in
them.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
```diff
@@ -13,114 +14,101 @@
...
-                catch (GracefulException e)
+            }
+            else
+            {
+                var framework = NuGetFramework.Parse(frameworkString);
+                if (!msbuildProj.IsTargettingFramework(framework))
                 {
-                    Reporter.Error.WriteLine(e.Message.Red());
-                    app.ShowHelp();
+                    Reporter.Error.WriteLine(string.Format(
```
krwq: just wanna confirm - are we still gonna get help for `dotnet add p2p` here? I'm assuming this is handled by command line parser now

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jgoshi: No, I guess I can change it back.

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
jgoshi: I believe so. What is the command line I can try to verify this?

src/dotnet/DotNetTopLevelCommandBase.cs
jgoshi: Ok will do

src/dotnet/CommandLine/CommandLineApplication.cs
jgoshi: Will add the null check

livarcocc:
@dotnet-bot Test this

livarcocc:
@dotnet-bot Test this please

jgoshi:

src/dotnet/commands/dotnet-add/dotnet-add-p2p/Program.cs
piotrpMSFT: `dotnet add p2p --help`

jgoshi: Oh, then yes that works. We have a test for it too.

src/dotnet/CommandLine/CommandLineApplication.cs
```diff
@@ -56,15 +56,34 @@ public CommandLineApplication(bool throwOnUnexpectedArg = true)
...
 
-        public CommandLineApplication Command(string name, bool throwOnUnexpectedArg = true)
+        public CommandLineApplication AddCommand(string name, bool throwOnUnexpectedArg = true)
         {
-            return Command(name, _ => { }, throwOnUnexpectedArg);
+            return AddCommand(name, _ => { }, throwOnUnexpectedArg);
         }
 
-        public CommandLineApplication Command(string name, Action<CommandLineApplication> configuration,
+        public CommandLineApplication AddCommand(string name, Action<CommandLineApplication> configuration,
```
livarcocc: NIT: When we put one parameter in a new line, we put each in a new line. Like this

```
AddCommand(
    string name,
    Action<CommandLineApplication> configuration,
    bool throwOnUnexpectedArg = true)
```

test/dotnet-remove-proj.Tests/GivenDotnetRemoveProj.cs
```diff
@@ -0,0 +1,280 @@
...
+using FluentAssertions;
+using Microsoft.DotNet.Cli.Sln.Internal;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using System;
+using System.IO;
+using Xunit;
+
+namespace Microsoft.DotNet.Cli.Remove.P2P.Tests
+{
+    public class GivenDotnetRemoveProj : TestBase
```
livarcocc: This class and the one above are pretty similar in name in my opinion. I think we should call this one GivenDotnetRemoveProjFromSolution, or better yet, GivenThatWeWantToRemoveProjectsFromASolution.

PR #5112, opened 2016-12-21
Allow selecting with either MIT or MSFT License

ellismg:
This change introduces a new configuration knob to the build
`InstallerLicenseType` which controls if the installers that we build are
licenses using the MIT or .NET Library license. The default is to continue
to use the .NET Library license.

For our zips/tar.gz's we will be moving to using the MIT license instead.
However, due to how these zips are built, that change will actually happen
in dotnet/core-setup (since the zips/tarballs from the host/host + shared
framework combo carry the license with them and that's just laid down on
disk when we pull artifacts from Azure).


ellismg:
/cc @piotrpMSFT @livarcocc 

The ability to choose what license was used at build time is something that that @richlander wants.  The major issue is getting the zip's/tar.gz's to be MIT licensed, but due to how we build our zips, we need changes in core-setup to support that (since cli just gets whatever license core-setup lays down in the zips).

I will need to make similar changes (in spirit but not implementation, due to the different build systems) in the servicing branches for LTS and FTS.  Can you tell me what branches we are using for this?  preview-2.1 and preview-2?

krwq:
@ellismg what's going on with the file renames in here? is all of that expected?

ellismg:
> what's going on with the file renames in here? is all of that expected? The diff looks like some localization files are being moved to bad folders.

Git makes this a little confusing.  Because we have to pass a resources folder to the mac packaging stuff, I decided have MIT and MSFT versions of this folder.  Some assets (like images) are the same, others (like the EULA) differ between the two.

You can see in [the build scripts](https://github.com/dotnet/cli/pull/5112/files#diff-39c8c000e4762b830505bae74daca979R22) how we now use `$(InstallerLicenseType)` to select the correct resources folder.

ellismg:
The other option would be construct the resouces directory at build time, selecting different versions of some assets (e.g. the rtf's) while retaining others (e.g. the html files and images).  Since internally git will de-duplicate these files (at least in the object database, but not the work tree) I didn't try to do that.  I can image that you may want different versions of the installers to look different long term (e.g. the MIT licensed installer has a different logo or color) and having them split out explicitly makes that flexibility easier.  I'm not wed to the implementation, if there's something you all would prefer I'm happy to make the changes.

tmds:
@ellismg just saw that the tar.gz files on dotnet/cli still contain the .NET Library license. Looking forward to see this PR merged and get the MIT license instead :+1:

ellismg:
Fixed by #6780

PR #5195, opened 2017-01-03
change default verbosity to normal (to enable build status)

krwq:
https://github.com/dotnet/cli/issues/5110

Making sure that build status is visible. The additional messages are not desirable but the only other solution for this is to implement our own logger or fix that in msbuild logger as currently printing only build status + warnings  + errors is not supported ( Microsoft/msbuild#1513 )

### DIFF
--- a/src/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs
+++ b/src/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs
@@ -31,7 +31,7 @@ namespace Microsoft.DotNet.Tools.MSBuild
             };
         
         private readonly IEnumerable<string> _msbuildRequiredParameters = 
-            new List<string> { "/m", "/v:m" };
+            new List<string> { "/m" };
 
         public MSBuildForwardingApp(IEnumerable<string> argsToForward)
         {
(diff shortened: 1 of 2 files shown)


src/dotnet/commands/dotnet-msbuild/MSBuildForwardingApp.cs
```diff
@@ -31,7 +31,7 @@ public class MSBuildForwardingApp
             };
         
         private readonly IEnumerable<string> _msbuildRequiredParameters = 
-            new List<string> { "/m", "/v:m" };
+            new List<string> { "/m" };
```
piotrpMSFT: Issue #5110 only talks about `dotnet build`. I do not think we should change the verbosity for `run`, `restore`, etc. @blackdwarf ?

krwq: IMO it makes sense for build and run. I have mixed feelings for restore and others. @blackdwarf - please comment here.

piotrpMSFT: `run` is explicitly why we switched to `/m` in the first place.

blackdwarf: Indeed. Just `build` please, not `run`. We've gotten focused feedback since Preview 1 days that `run` was too chatty and **that** chattiness level was way below what a normal MSBuild output looks like.

blackdwarf:
@krwq please do not merge this yet. I need to follow up with some people on some comments that were made that this is not an adequate solution to the problem. Stay tuned...

piotrpMSFT:
Closing for the moment. Let's re-open once the plan firms up.

PR #5223, opened 2017-01-05
Added links to project system uses of serialized objects.

333fred:
This adds links to the parts of the project system where we're deserializing JSON objects defined by CLI. We need to make sure that these are kept up date with each other. Tagging @livarcocc @piotrpMSFT for review.

skip ci please

### DIFF
--- a/src/Microsoft.DotNet.ProjectJsonMigration/MigrationError.cs
+++ b/src/Microsoft.DotNet.ProjectJsonMigration/MigrationError.cs
@@ -3,6 +3,8 @@
 
 namespace Microsoft.DotNet.ProjectJsonMigration
 {
+    /// Any changes to this need to be reflected in roslyn-project-system
+    // https://github.com/dotnet/roslyn-project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/Xproj/MigrationError.cs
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.ProjectJsonMigration/MigrationError.cs
```diff
@@ -3,6 +3,8 @@
 
 namespace Microsoft.DotNet.ProjectJsonMigration
 {
+    /// Any changes to this need to be reflected in roslyn-project-system
+    // https://github.com/dotnet/roslyn-project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/ProjectSystem/VS/Xproj/MigrationError.cs
```
jonsequitur: It's all too easy and common for people not to read the comments before making a code change. It's pretty easy to enforce this sort of thing with a simple unit test. Serialize an instance of the class and compare it to a hard-coded JSON string. This will prevent accidental breaking changes to the serialization format.

piotrpMSFT:
yea, agreed. If we want this to be a contract, let's make it a test contract. We don't expect folks to scan for random comments before making code changes.

piotrpMSFT:
Let's reopen this once the comments are replaced with tests.

PR #5246, opened 2017-01-09
Update xlf files

nguerrera:
@livarcocc @piotrpMSFT @srivatsn This needs to get in ASAP for handoff.


nguerrera:
@MattGertz for approval. Note that changes to these files do not affect the product yet (change to consume translations still pending for cli), but need to be in the repo ASAP so that the loc team can pick up these late changes today.

PR #5290, opened 2017-01-12
Support cross build

jyoungyun:
Add '--env-var' option to build script.
With this option, the RID value can be changed using 'DOTNET_RUNTIME_ID' environment.

TestCommand:
```
./build.sh --env-vars DOTNET_RUNTIME_ID=ubuntu.14.04-arm -c Release
```

Related issue: #5289


run-build.sh
```diff
@@ -93,11 +93,18 @@ while [[ $# > 0 ]]; do
             args=( "${args[@]/$2}" )
             shift
             ;;
+        --env-vars)
```
eerhardt: Can we support this option on `run-build.ps1` as well?  That way our Windows and Unix builds stay in lock step.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -29,6 +29,8 @@
     <PropertyGroup>
       <IsDesktopAvailable>False</IsDesktopAvailable>
       <IsDesktopAvailable Condition=" '$(OSName)' == 'win' ">True</IsDesktopAvailable>
+      <IsCross>False</IsCross>
+      <IsCross Condition=" !$(Rid.Contains('$(OSName)')) Or !$(Rid.Contains('$(Architecture)')) ">True</IsCross>
```
eerhardt: What does this property do?  It doesn't appear to be read by any other part of the build.

jyoungyun: I will use this feature in the next patch not to download debian packages and not to delete redist.dll and tool_msbuild.dll (https://github.com/dotnet/cli/blob/rel/1.0.0/build/Microsoft.DotNet.Cli.Compile.targets#L162#L163). Without these two dlls, our sdk didn't work normally.

jyoungyun:
@eerhardt I understood what you were concerned about. `IsCross` related code was modified to be included in the next patch. Thank you :)

run-build.sh
jonsequitur: @jyoungyun Could you describe the use case for this? My first impression is that this could be solved using an  external script that sets the needed variables rather than having the CLI's scripts take on this responsibility, but I'm probably missing some of the context.

Thanks.

am11: Conventionally, the env-vars vs. argument style is either:

```sh
./build.sh --runtime-id ubuntu.14.04-arm -c Release

# or
DOTNET_RUNTIME_ID=ubuntu.14.04-arm  ./build.sh -c Release
# windows powershell equivalent of:
$env:DOTNET_RUNTIME_ID=ubuntu.14.04-arm; ./build.ps1 -c Release
```

In case both are supported and provided, command-line argument will take the precedence.

jyoungyun: @am11 Yes, I know. But just wanted to modify it to the same structure as core-setup. The core-setup build process has recently changed, so this patch should wait a bit longer.

qmfrederik:
@jyoungyun I think this needs to go to master instead of 1.0.0.

You'll hit some merge conflicts when going to master. I've rebased your branch on master - see https://github.com/qmfrederik/cli/commit/14d8f808fc74181a8ec38c4173c0ccf74f2f6bc9. Hope to be able to cross-compile from master soon :)

jyoungyun:
@qmfrederik Thank you for resolving the merge conflict issue :) I really hope to be able to cross-compile from master soon!!

jyoungyun:
Sorry to interrupt, I will update pr for cross build.

PR #5293, opened 2017-01-12
Migration: do not inject built in compile includes that the SDK does

### LINKED ISSUE #5253: Migrate needs to turn off default globs if adding globs
Migrate https://github.com/robotdotnet/NetworkTables, NetworkTables.xproj.sln. The resulting csproj's have include globs with `**\*.cs`, but do not have `<EnableDefaultItems>false</EnableDefaultItems>`. This causes https://github.com/dotnet/sdk/issues/604 to show up in all projects. There are a few issues with the resulting csproj:

* Multiple compile lines - one for the default globs and one for the migrated includes. Not sure how easy it will be to condense them with how we do migration.
* Default items are not turned off, resulting in broken projects.

I'm using 1.0.0-preview5-004468. Tagging @livarcocc. /cc @thadhouse.

jgoshi:
@piotrpMSFT @livarcocc @krwq @jonsequitur @eerhardt @piotroko 

Fixes #5253

### DIFF
--- a/src/Microsoft.DotNet.ProjectJsonMigration/Microsoft.DotNet.Internal.ProjectModel/Files/ProjectFilesCollection.cs
+++ b/src/Microsoft.DotNet.ProjectJsonMigration/Microsoft.DotNet.Internal.ProjectModel/Files/ProjectFilesCollection.cs
@@ -9,14 +9,9 @@ namespace Microsoft.DotNet.Internal.ProjectModel.Files
 {
     internal class ProjectFilesCollection
     {
-        public static readonly string[] DefaultCompileBuiltInPatterns = new[] { @"**/*.cs" };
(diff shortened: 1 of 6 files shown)


piotrpMSFT:
@MattGertz tagging for shiproom review

MattGertz:
I agree.

jgoshi:
test Ubuntu x64 Release Build

piotrpMSFT:
@MattGertz `proactive`?

jgoshi:
@livarcocc This case is already tested for in GivenThatIWantToMigrateBuildOptions.cs. There are two tests: MigratingGroupIncludeExcludePopulatesAppropriateProjectItemElement, MigratingGroupIncludeOnlyPopulatesAppropriateProjectItemElement which I didn't need to change (because it was testing this case). But what I did change was the GetDefaultIncludePatterns (which is not expected to return nothing). The test passing with that change is sufficient right?

Issue #5297, opened 2017-01-13
dotnet-migrate shows wrong stack trace when showing error for unhandled exceptions


krwq:
the problem is with Program.cs for dotnet-migrate command. Instead of
```
throw e;
```

we should just
```
throw;
```

currently the stack is being shown in a context of Main instead of where the exception occured.

PR #5298, opened 2017-01-13
fix stack trace for unhandled exceptions in dotnet-migrate

krwq:
Fixes https://github.com/dotnet/cli/issues/5297

This will improve debuggability of dotnet-migrate.

@jonsequitur @livarcocc @piotrpMSFT


krwq:
@MattGertz - can we merge this? This will make debugging problems with dotnet-migrate much easier as we will get the right stack track. I haven't catched this problem in my previous PR and this is non reliably testable scenario considering that we should not have unhandled exceptions in our code base

MattGertz:
Proactively approved for RTM.

Issue #5345, opened 2017-01-16
Can't import xUnit.net xproj => csproj, fails to update the solution


bradwilson:
## Steps to reproduce

* Clone https://github.com/xunit/xunit (this is on SHA `f400140333e6b15065c16ed9a1b5d0b8c13da69a` which is the current master)
* Update `global.json` SDK version to `1.0.0-rc3-004504`
* Run `dotnet migrate xunit.vs2015.sln`

## Expected  behavior

Migration succeeds.

## Actual behavior

```
  $ dotnet migrate xunit.vs2015.sln
Project xunit.core migration succeeded (C:\Dev\xunit\xunit\src\xunit.core)
Project xunit.assert migration succeeded (C:\Dev\xunit\xunit\src\xunit.assert)
Project xunit.execution.desktop migration succeeded (C:\Dev\xunit\xunit\src\xunit.execution)
Project xunit.console migration succeeded (C:\Dev\xunit\xunit\src\xunit.console)
Project xunit.runner.utility.desktop migration succeeded (C:\Dev\xunit\xunit\src\xunit.runner.utility)
Project xunit.runner.reporters.desktop migration succeeded (C:\Dev\xunit\xunit\src\xunit.runner.reporters)
Project xunit.runner.tdnet migration succeeded (C:\Dev\xunit\xunit\src\xunit.runner.tdnet)
Project xunit.runner.msbuild migration succeeded (C:\Dev\xunit\xunit\src\xunit.runner.msbuild)
Project test.xunit.assert migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.assert)
Project test.utility migration succeeded (C:\Dev\xunit\xunit\test\test.utility)
Project test.xunit.execution migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.execution)
Project test.xunit.console migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.console)
Project test.xunit.runner.msbuild migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.runner.msbuild)
Project test.xunit.runner.reporters migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.runner.reporters)
Project test.xunit.runner.tdnet migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.runner.tdnet)
Project test.xunit.runner.utility migration succeeded (C:\Dev\xunit\xunit\test\test.xunit.runner.utility)
Summary
Total Projects: 16
[...]

PR #5357, opened 2017-01-17
Piotrp msft/issues/5343

piotrpMSFT:
@livarcocc


piotrpMSFT:
@MattGertz

piotrpMSFT:
This fix is incomplete due to backup directory location logic. The real fix is quite a bit more involved and requires changes slated for RTM. I'm closing the PR for the time being.

I also chatted with @onovotny who called out that the fix is not immediately necessary for his scenario, given we fixed #5345

PR #5424, opened 2017-01-22
Change new3 to new

mlorbetske:
Checking CI status with the change, please do not merge yet

@seancpeters @piotrpMSFT @livarcocc


mlorbetske:
@dontet-bot test this please

build/compile/Microsoft.DotNet.Cli.LzmaArchive.targets
```diff
@@ -72,7 +72,12 @@
       <MakeDir Directories="@(NuGetPackagesArchiveDirectory)"/>
 
       <DotNetNew ToolPath="$(Stage2Directory)"
-                 TemplateType="Web"
+                 TemplateType="--debug:reinit"
```
piotrpMSFT: that seems like a funny TemplateType. Please do change the implementation of DotNetNew if the current parameters no longer make sense...

build/compile/Microsoft.DotNet.Cli.LzmaArchive.targets
```diff
@@ -72,7 +72,12 @@
...
+                 WorkingDirectory="$(NuGetPackagesArchiveProject)" />
+
+      <DotNetNew ToolPath="$(Stage2Directory)"
+                 TemplateType="Mvc"
+                 TemplateArgs="-au Individual"
```
piotrpMSFT: Here as well. Is `-au` unique to the Mvc template? or it's applicable to all. I'd prefer to avoid `insert arbitrary string here` params if we can!

build_projects/dotnet-cli-build/DotNetNew.cs
```diff
@@ -12,16 +12,18 @@ protected override string Command
...
         protected override string Args
         {
-            get { return $"{GetTemplateType()}"; }
+            get { return $"{GetTemplateType()} {TemplateArgs}"; }
         }
 
         public string TemplateType { get; set; }
 
+        public string TemplateArgs { get; set; }
+
```
piotrpMSFT: Ok cool. Perhaps --debug above should be passed as a TemplateArgs?

src/dotnet/commands/dotnet-migrate/TemporaryDotnetNewTemplateProject.cs
```diff
@@ -65,8 +65,15 @@ private ProjectRootElement GetMSBuildProject()
 
         private void RunCommand(string commandToExecute, IEnumerable<string> args, string workingDirectory)
         {
+            List<string> argsWithDirectory = new List<string>(args);
```
piotrpMSFT: Can we do this back up on line 53? It seems we have the necessary context there, and I'd like to avoid generating args in two places if we can.

src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
```diff
@@ -1,14 +1,21 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
```
piotrpMSFT: Does this template need to exist at all anymore? shouldn't it be coming out of new3?

src/dotnet/commands/dotnet-new/CSharp_Web1.1/$projectName$.csproj
```diff
@@ -1,14 +1,21 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
```
piotrpMSFT: Same here...

src/dotnet/commands/dotnet-new/CSharp_Xunittest/$projectName$.csproj
```diff
@@ -7,8 +7,8 @@
 
   <ItemGroup>
```
piotrpMSFT: and here. I expected all these template directories to go away.

build/compile/Microsoft.DotNet.Cli.LzmaArchive.targets
mlorbetske: We added a second parameter to make the additional arguments more clear - this item in particular though has been removed in the commit I'm testing locally now.

build/compile/Microsoft.DotNet.Cli.LzmaArchive.targets
mlorbetske: It's unique to MVC, it's the auth type

build_projects/dotnet-cli-build/DotNetNew.cs
mlorbetske: Yep, it's that way in the next commit that's coming

src/dotnet/commands/dotnet-migrate/TemporaryDotnetNewTemplateProject.cs
mlorbetske: Done

src/dotnet/commands/dotnet-new/CSharp_Web/$projectName$.csproj
mlorbetske: Let's talk about this one in person tomorrow if possible. There seem to be 3 phases in LZMA archive creation/extraction at first run.

1. The package references from a template (this one) are used to compute a hash that's used as a blob name to fetch, allowing the creation of the archive to be skipped if it already exists
2. The template (via an invocation of new) is used to catalog the actual references to use, these are restored into a directory, packed up & uploaded to a blob (iff phase 1 couldn't find a matching archive)
3. At first run, a set of templates from invocations of new (I think) is restored to the user cache of packages

I'm not sure why phase 1 didn't use an invocation of new in the first place, but this now poses a synchronization issue if it doesn't.

mlorbetske:
Replaced by #5430

PR #5465, opened 2017-01-26
Update CLI to use shared framework 2.0.0-beta-001386-00

eerhardt:
@jonsequitur @piotrpMSFT @livarcocc @jgoshi


eerhardt:
merging now that this is green.

PR #5489, opened 2017-01-27
Add checked-in build definitions

naamunds:
These build definitions will be used by pipebuild to build the product.

The signing now uses MicroBuild because the builds are being moved to
the DevDiv VSTS instance, and the Microsoft Code Signing system which
was used in MSEng is not supported there.

MicroBuild does not support the .NET Core MSBuild, so the full-framework
MSBuild is used to build the signing targets.  However, the new build
machines do not have MSBuild 15, so they use 14. This means that
projects which use the new MSBuild 15 syntax fail to be imported.
They're not needed for the signing, so their importing is skipped by
using a new SkipMsBuild15Projects property.

Thanks to @crummel for doing most of the work on the signing logic.

/cc @piotrpMSFT @livarcocc @chcosta

### DIFF
--- a/build/Microsoft.DotNet.Cli.Prepare.targets
+++ b/build/Microsoft.DotNet.Cli.Prepare.targets
@@ -62,7 +62,7 @@
       <Output TaskParameter="MsiVersion" PropertyName="MsiVersion" />
       <Output TaskParameter="VersionBadgeMoniker" PropertyName="VersionBadgeMoniker" />
       <Output TaskParameter="Channel" PropertyName="Channel" />
-      <Output TaskParameter="BranchName" PropertyName="BranchName" />
+      <Output TaskParameter="BranchName" PropertyName="BranchName" Condition=" '$(BranchName)' == '' " />
     </GenerateBuildVersionInfo>
 
     <PropertyGroup>
--- a/build/Microsoft.DotNet.Cli.Signing.proj
+++ b/build/Microsoft.DotNet.Cli.Signing.proj
@@ -1,0 +1,114 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project InitialTargets="SetSigningProperties" DefaultTargets="SignFiles" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))/dir.props" />
+  <Import Project="$(RepoRoot)/build_tools/MicroBuild.Core.props" />
+
+  <!-- This will be overridden if we're building with MicroBuild. -->
+  <Target Name="SignFiles">
+    <Message Text="Fake sign target.  Would sign: @(FilesToSign)" />
+  </Target>
+
+  <Target Name="SetSigningProperties">
+    <Error Condition="'$(Rid)' == ''" Text="Missing required property 'Rid'." />
+    <PropertyGroup>
+      <BaseOutputDirectory Condition="'$(BaseOutputDirectory)' == ''">$(RepoRoot)/artifacts/$(Rid)</BaseOutputDirectory>
+      <Stage2Directory Condition="'$(Stage2Directory)' == ''">$(BaseOutputDirectory)/stage2</Stage2Directory>
+      <Stage2CompilationDirectory Condition="'$(Stage2CompilationDirectory)' == ''">$(BaseOutputDirectory)/stage2compilation</Stage2CompilationDirectory>
+      <PackagesDirectory Condition="'$(PackagesDirectory)' == ''">$(BaseOutputDirectory)/packages</PackagesDirectory>
+
+      <!-- The OutDir and IntermediateOutputPath properties are required by MicroBuild. MicroBuild only
+           signs files that are under these paths. -->
+      <OutDir Condition="'$(OutDir)' == ''">$(BaseOutputDirectory)</OutDir>
+      <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == ''">$(BaseOutputDirectory)/intermediate</IntermediateOutputPath>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="PostCompileSign" DependsOnTargets="GetPostCompileSignFiles;SignFiles" />
+
+  <Target Name="GetPostCompileSignFiles">
+    <ItemGroup>
+      <!-- External files -->
+      <FilesToSign Include="$(Stage2Directory)/sdk/**/Newtonsoft.Json.dll;
+                            $(Stage2Directory)/shared/**/libuv.dll">
+        <Authenticode>$(ExternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <!-- Built binaries -->
+      <FilesToSign Include="$(Stage2Directory)/sdk/**/csc.exe;
+                            $(Stage2Directory)/sdk/**/csc.dll;
+                            $(Stage2Directory)/sdk/**/dotnet.dll;
+                            $(Stage2Directory)/sdk/**/System.*.dll;
+                            $(Stage2Directory)/sdk/**/Microsoft.*.dll;
+                            $(Stage2Directory)/sdk/**/NuGet*.dll;
+                            $(Stage2Directory)/sdk/**/datacollector.dll;
+                            $(Stage2Directory)/sdk/**/MSBuild.dll;
+                            $(Stage2Directory)/sdk/**/testhost.dll;
+                            $(Stage2Directory)/sdk/**/vstest.console.dll">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <!-- Built files for the packages -->
+      <FilesToSign Include="$(Stage2CompilationDirectory)/forPackaging/**/*dotnet*.dll;
+                            $(Stage2CompilationDirectory)/forPackaging/**/Microsoft.Extensions.DependencyModel.dll;
+                            $(Stage2CompilationDirectory)/forPackaging/**/Microsoft.Extensions.Testing.Abstractions.dll">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignNuPkgContents" DependsOnTargets="GetSignNuPkgContentsFiles;SignFiles" />
+
+  <Target Name="GetSignNuPkgContentsFiles">
+    <PropertyGroup>
+      <OutDir>$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <!-- NuPkg contents -->
+      <FilesToSign Include="$(OutDir)/bin/**/Microsoft.DotNet.Cli.Utils.dll">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignMsiAndCab" DependsOnTargets="GetSignMsiAndCabFiles;SignFiles" />
+
+  <Target Name="GetSignMsiAndCabFiles">
+    <PropertyGroup>
+      <OutDir>$(PackagesDirectory)</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToSign Include="$(PackagesDirectory)/**/*.msi">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(PackagesDirectory)/**/*.cab">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignEngine" DependsOnTargets="GetSignEngineFiles;SignFiles" />
+
+  <Target Name="GetSignEngineFiles">
+    <PropertyGroup>
+      <OutDir>$(PackagesDirectory)</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToSign Include="$(PackagesDirectory)/**/*engine.exe">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignCliBundle" DependsOnTargets="GetSignCliBundleFiles;SignFiles" />
+
+  <Target Name="GetSignCliBundleFiles">
+    <PropertyGroup>
+      <OutDir>$(PackagesDirectory)</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToSign Include="$(PackagesDirectory)/**/*.exe">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Import Project="$(RepoRoot)/build_tools/MicroBuild.Core.targets" />
+</Project>
--- a/buildpipeline/DotNet-CLI-CentOS-x64.json
+++ b/buildpipeline/DotNet-CLI-CentOS-x64.json
@@ -1,0 +1,172 @@
+{
+  "build": [
+    {
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Shell Script build.sh",
+      "timeoutInMinutes": 0,
+      "task": {
+        "id": "6c731c3c-3c68-459a-a5c9-bde6e6595b5b",
+        "versionSpec": "2.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptPath": "build.sh",
+        "args": "$(BuildArguments)",
+        "disableAutoCwd": "false",
+        "cwd": "",
+        "failOnStandardError": "false"
+      }
+    }
+  ],
+  "options": [
+    {
+      "enabled": false,
+      "definition": {
+        "id": "5bc3cfb7-6b54-4a4b-b5d2-a3905949f8a6"
+      },
+      "inputs": {
+        "additionalFields": "{}"
+      }
+    },
+    {
+      "enabled": false,
+      "definition": {
+        "id": "7c555368-ca64-4199-add6-9ebaf0b0137d"
+      },
+      "inputs": {
+        "multipliers": "[]",
+        "parallel": "false",
+        "continueOnError": "true",
+        "additionalFields": "{}"
+      }
+    },
+    {
+      "enabled": false,
+      "definition": {
+        "id": "a9db38f9-9fdc-478c-b0f9-464221e58316"
+      },
+      "inputs": {
+        "workItemType": "4777",
+        "assignToRequestor": "true",
+        "additionalFields": "{}"
+      }
+    },
+    {
+      "enabled": false,
+      "definition": {
+        "id": "57578776-4c22-4526-aeb0-86b6da17ee9c"
+      },
+      "inputs": {
+        "additionalFields": "{}"
+      }
+    }
+  ],
+  "variables": {
+    "BuildConfiguration": {
+      "value": "Release",
+      "allowOverride": true
+    },
+    "BuildArguments": {
+      "value": "--skip-prereqs --configuration $(BuildConfiguration)  --docker centos --targets Default",
+      "allowOverride": true
+    },
+    "CONNECTION_STRING": {
+      "value": "PassedViaPipeBuild"
+    },
+    "PUBLISH_TO_AZURE_BLOB": {
+      "value": "true",
+      "allowOverride": true
+    },
+    "GITHUB_PASSWORD": {
+      "value": "PassedViaPipeBuild"
+    },
+    "ARTIFACT_STORAGE_KEY": {
+      "value": "PassedViaPipeBuild"
+    },
+    "ARTIFACT_STORAGE_ACCOUNT": {
+      "value": "dotnetcli"
+    },
+    "ARTIFACT_STORAGE_CONTAINER": {
+      "value": "dotnet"
+    },
+    "CHECKSUM_STORAGE_KEY": {
+      "value": "PassedViaPipeBuild"
+    },
+    "CHECKSUM_STORAGE_ACCOUNT": {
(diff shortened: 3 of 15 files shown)


piotrpMSFT:
🥇 😍

build.proj
```diff
@@ -80,8 +80,9 @@
 
   <Import Project="build/Microsoft.DotNet.Cli.Prepare.targets" />
   <Import Project="build/Microsoft.DotNet.Cli.Compile.targets" />
+  <Import Project="build/Microsoft.DotNet.Cli.Signing.targets" />
   <Import Project="build/Microsoft.DotNet.Cli.Package.targets" />
-  <Import Project="build/Microsoft.DotNet.Cli.Test.targets" />
+  <Import Project="build/Microsoft.DotNet.Cli.Test.targets" Condition="'$(SkipMsBuild15Projects)' != 'True'" />
```
piotrpMSFT: Can you share the purpose of this condition, @naamunds ? The property name is not immediately intuitive to me.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -2,8 +2,8 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="Microsoft.DotNet.Cli.tasks" />
   <Import Project="prepare/CheckPrereqs.targets" />
-  <Import Project="Microsoft.DotNet.Cli.BundledSdks.props" />
-  <Import Project="Microsoft.DotNet.Cli.BundledTemplates.props" />
+  <Import Project="Microsoft.DotNet.Cli.BundledSdks.props" Condition="'$(SkipMsBuild15Projects)' != 'True'" />
+  <Import Project="Microsoft.DotNet.Cli.BundledTemplates.props" Condition="'$(SkipMsBuild15Projects)' != 'True'" />
```
piotrpMSFT: Same here.

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -68,7 +68,7 @@
       <Output TaskParameter="BranchName" PropertyName="BranchName" />
     </GenerateBuildVersionInfo>
 
-    <GenerateNuGetPackagesArchiveVersion RepoRoot="$(RepoRoot)">
+    <GenerateNuGetPackagesArchiveVersion RepoRoot="$(RepoRoot)" Condition="'$(SkipMsBuild15Projects)' != 'True'">
```
piotrpMSFT: and here

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
piotrpMSFT: Why ToolsVersion14? is this not running on VS 2017?

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+
+  <Target Name="SetSignFilesOutDir" BeforeTargets="SignFiles">
+    <PropertyGroup>
+      <OutDir Condition="'$(OutDir)' == ''">$(OutputDirectory)</OutDir>
+      <IntermediateOutputPath Condition="'$(IntermediateOutputPath)' == ''">$(IntermediateDirectory)</IntermediateOutputPath>
+    </PropertyGroup>
+  </Target>
+
+  <Import Project="$(RepoRoot)\build_tools\MicroBuild.Core.props" />
+  <Import Project="$(RepoRoot)\build_tools\MicroBuild.Core.targets" />
```
piotrpMSFT: Wohoo! a use case of build_tools :D

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+
+  <Target Name="PostCompileSign" DependsOnTargets="GetPostCompileSignFiles">
+    <CallTarget Targets="SignFiles" />
```
piotrpMSFT: Why do we need to use CallTarget? It's a pattern we've been trying to avoid in CLI due to the nuances [properties defined in calling targets not being available, etc]. If necessary is fine, but why not have PostCompileSign depend on SignFiles depend on GetPostCompileSignFiles?

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+  </Target>
+
+  <Target Name="GetPostCompileSignFiles" DependsOnTargets="Init">
+    <ItemGroup>
+      <!-- External files -->
+      <FilesToSign Include="$(OutputDirectory)sdk/**/Newtonsoft.Json.dll">
+        <Authenticode>$(ExternalCertificateId)</Authenticode>
```
piotrpMSFT: Are we able to use "Metadata as attributes" here? Or is an older MSBuild expected to run these targets?

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+      </FilesToSign>
+      <FilesToSign Include="$(OutputDirectory)shared/**/libuv.dll">
```
piotrpMSFT: This is pretty verbose. Can we have one "FilesToSign" per CertificateId and `;` separate all of these paths in a single `Include`?

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(Stage2CompilationDirectory)/forPackaging/**/Microsoft.Extensions.Testing.Abstractions.dll">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignNuPkgContents" DependsOnTargets="GetSignNuPkgContentsFiles">
+    <CallTarget Targets="SignFiles" />
```
piotrpMSFT: Same CallTarget comment as above.

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+  </Target>
+
+  <Target Name="GetSignNuPkgContentsFiles" DependsOnTargets="Init">
+    <PropertyGroup>
+      <OutDir>$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <!-- NuPkg contents -->
+      <FilesToSign Include="$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils/bin/**/*.dll">
```
piotrpMSFT: Same `;` delimiter comment as above.

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils/obj/**/*.exe">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignSharedFrameworkCrossgenedAssemblies" DependsOnTargets="GetSignSharedFrameworkCrossgenedAssembliesFiles">
+    <CallTarget Targets="SignFiles" />
```
piotrpMSFT: same

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+  </Target>
+
+  <Target Name="GetSignSharedFrameworkCrossgenedAssembliesFiles" DependsOnTargets="Init">
+    <PropertyGroup>
+      <OutDir>$(OutputDirectory)shared</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToSign Include="$(OutputDirectory)shared/**/Microsoft.*.dll">
```
piotrpMSFT: Same

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignSharedFrameworkCrossgenedAssemblies" DependsOnTargets="GetSignSharedFrameworkCrossgenedAssembliesFiles">
+    <CallTarget Targets="SignFiles" />
+  </Target>
+
+  <Target Name="GetSignSharedFrameworkCrossgenedAssembliesFiles" DependsOnTargets="Init">
```
piotrpMSFT: This is a good time to think through this. Should CLI still be signing SharedFramework assemblies? Aren't those signed & crossgen'd in the Shared Framework packages already?

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(OutputDirectory)shared/**/System.*.dll">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignMsiAndCab" DependsOnTargets="GetSignMsiAndCabFiles">
+    <CallTarget Targets="SignFiles" />
```
piotrpMSFT: Same

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+  </Target>
+
+  <Target Name="GetSignMsiAndCabFiles" DependsOnTargets="Init">
+    <PropertyGroup>
+      <OutDir>$(PackagesDirectory)</OutDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToSign Include="$(PackagesDirectory)/**/*.msi">
```
piotrpMSFT: Same

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(PackagesDirectory)/**/*.cab">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignEngine" DependsOnTargets="GetSignEngineFiles">
+    <CallTarget Targets="SignFiles" />
```
piotrpMSFT: Same

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToSign Include="$(PackagesDirectory)/**/*engine.exe">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignCliBundle" DependsOnTargets="GetSignCliBundleFiles">
+    <CallTarget Targets="SignFiles" />
```
piotrpMSFT: Same

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+  </Target>
+
+  <Target Name="BeforeBuild" BeforeTargets="InitializeCommonProps" Condition="'$(SkipMsBuild15Projects)' == 'True'">
+    <HandleAssemblyResolve SearchPath="$(RepoRoot)\build_tools" />
+  </Target>
+  <UsingTask TaskName="HandleAssemblyResolve" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll" Condition="'$(SkipMsBuild15Projects)' == 'True'" >
+    <ParameterGroup>
+      <SearchPath ParameterType="System.String" Required="true" />
+    </ParameterGroup>
+    <Task>
```
piotrpMSFT: This is the first inline task in CLI repo. Can we instead put this in dotnet-cli-build.csproj?

buildpipeline/Dotnet-CLI-Ubuntu16.04-x64.json
```diff
@@ -0,0 +1,180 @@
...
+  "name": "Dotnet-CLI-Ubuntu16.04-x64",
+  "project": {
+    "id": "0bdbc590-a062-4c3f-b0f6-9383f67865ee",
+    "name": "DevDiv",
+    "description": "Visual Studio and DevDiv team project for git source code repositories.  Work items will be added for Adams, Dev14 work items are tracked in vstfdevdiv.  ",
+    "url": "https://devdiv.visualstudio.com/DefaultCollection/_apis/projects/0bdbc590-a062-4c3f-b0f6-9383f67865ee",
+    "state": "wellFormed",
+    "revision": 418097459
+  }
+}
```
piotrpMSFT: Please make sure we have all of these trailing newline issues resolved.

buildpipeline/DotNet-CLI-CentOS-x64.json
```diff
@@ -0,0 +1,172 @@
+{
```
piotrpMSFT: This is hopeful: Can we collapse these down to a single file for common file? having these files be disjoint will make it difficult to understand where our builds differ...

build/Microsoft.DotNet.Cli.Signing.targets
eerhardt: This pattern is a bit odd to me.  Usually `.props` files are imported at the top of a project file, and `.targets` files are imported at the bottom.

This gives you the advantage of being able to override any property set in a `.props` file, and for the `.targets` file to be able to read properties set in your project.

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+    </ItemGroup>
+  </Target>
+
+  <Target Name="SignNuPkgContents" DependsOnTargets="GetSignNuPkgContentsFiles">
+    <CallTarget Targets="SignFiles" />
+  </Target>
+
+  <Target Name="GetSignNuPkgContentsFiles" DependsOnTargets="Init">
+    <PropertyGroup>
+      <OutDir>$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils</OutDir>
```
eerhardt: seems like an odd "outdir".  Is this used and correct?

build.proj
naamunds: MicroBuild does not support the .NET Core MSBuild, so the full-framework MSBuild is used to build the signing targets. However, the new build machines do not have MSBuild 15, so they use 14. This means that projects which use the new MSBuild 15 syntax fail to be imported.
They're not needed for the signing, so their importing is skipped by using a new `SkipMsBuild15Projects` property which is set to true in the signing build steps.

naamunds: I can look into the possibility of getting MSBuild 15 on the build agents and then remove this code at that point, but I wanted to get this working now.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: MSBuild 15 is not installed on the build agents; see the above comment thread.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: It looks like the "metadata as attributes" feature was supported starting with MSBuild 15, and currently the build agents don't have MSBuild 15.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: Yes; this property refers to the directory in which the files to sign are located. MicroBuild only signs files that are in a subdirectory of either the `OutDir` or `IntermediateOutputPath`. For this task, the files to sign are in the `$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils` directory.

build/Microsoft.DotNet.Cli.Signing.targets
```diff
@@ -0,0 +1,199 @@
...
+    </PropertyGroup>
+    <ItemGroup>
+      <!-- NuPkg contents -->
+      <FilesToSign Include="$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils/bin/**/*.dll">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils/bin/**/*.exe">
+        <Authenticode>$(InternalCertificateId)</Authenticode>
+      </FilesToSign>
+      <FilesToSign Include="$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils/obj/**/*.dll">
```
eerhardt: Why are we signing files in the `obj` folder?  Seems like only files in the `bin` folder are necessary.

eerhardt: Also, won't this sign a bunch of referenced assemblies as well?  I don't think we need them signed - only what gets put in the nupkg.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: I've confirmed that these assemblies are already signed, so I'll remove that step from the build.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: I'm fine with removing the obj files and only signing `$(RepoRoot)/src/Microsoft.DotNet.Cli.Utils/bin/**/Microsoft.DotNet.Cli.Utils.dll` here.  I tested out that change, and it results in every file in the nupkg being signed. Do we know that Microsoft.DotNet.Cli.Utils.dll will always be the only file in the nupkg, though?

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: I just tried that, but `AppDomain` isn't going to be added to .NET Core until 2.0 (see https://apisof.net/catalog/System.AppDomain), so the build of dotnet-cli-build.csproj fails. I think for this reason we'll have to leave it as an inline task, where it can use the full-framework .NET.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: Changed as suggested.

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: I've changed it so that `SignFiles` is also in the `DependsOnTargets`, e.g. `DependsOnTargets="GetPostCompileSignFiles;SignFiles"`

build/Microsoft.DotNet.Cli.Signing.targets
naamunds: Changed as suggested.

buildpipeline/Dotnet-CLI-Ubuntu16.04-x64.json
naamunds: Fixed.

buildpipeline/DotNet-CLI-CentOS-x64.json
naamunds: Yes, that's the long-term plan. I wanted to get this in quickly by doing the work iteratively; I'll log an issue for consolidating the build definitions.

naamunds:
In addition to addressing the comments here, I've updated the Windows x64 build definition to use a PowerShell script for publishing the NuGet packages to the dotnet-cli MyGet feed.  This is because the NuGet Server Endpoint used by the NuGet Publisher build task in MSEng does not exist in DevDiv, and because the PS script as easier to maintain.

naamunds:
@piotrpMSFT I've moved the signing logic to a standalone project due to it needing to be built with the .NET desktop MSBuild, as we discussed offline.  This decoupling gives a couple simplifying benefits:

- We no longer need to use `Condition="'$(SkipMsBuild15Projects)' != 'True'"`
- The HandleAssemblyResolve inline task is no longer necessary; it was needed for calculating the RID as part build.proj.

When we have a common props file with the properties at https://github.com/naamunds/cli/blob/94648355485c08c2f3776eb128ec3b28c8be2554/build/Microsoft.DotNet.Cli.Signing.proj#L17-L20, the code can be refactored to use them.

build/Microsoft.DotNet.Cli.Signing.proj
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project InitialTargets="SetSigningProperties" DefaultTargets="SignFiles" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <RepoRoot>$(MSBuildThisFileDirectory)/..</RepoRoot>
```
piotrpMSFT: Consider using CLI's `dir.props`. It's great for properties that we already define statically like `<RepoRoot>`

naamunds: Good idea.

naamunds:
@piotrpMSFT @livarcocc I've added a commit porting over https://github.com/dotnet/core-setup/pull/1406.  This issue would impact CLI also if not fixed; I saw it in some of my test RHEL builds.  As I mentioned there, I see this as a short-term fix.  Let me know if you have any questions or concerns.

naamunds:
I've added https://github.com/dotnet/cli/pull/5489/commits/4322dd5601e4b7ddc32398502b3e7886524106a2 for temporarily using sandboxes for publishing. @piotrpMSFT @livarcocc PTAL.

/cc @dagood

piotrpMSFT:

piotrpMSFT:

piotrpMSFT:
@MattGertz merging the infra PR we approved last night.

Issue #5470, opened 2017-01-26
Exception thrown while parsing command line arguments with dotnet new


MichaelSimons:
## Steps to reproduce
`dotnet new Web1.1`
e.g. forgot the `-t`

## Expected  behavior
I expect some output that would indicate the specified arguments were invalid possibly even listing what part was invalid.  Possibly even display the --help content.  I do not expect to see an unhandled exception

## Actual behavior

```
Specify --help for a list of available options and commands.

Unhandled Exception: Microsoft.DotNet.Cli.CommandLine.CommandParsingException: Unrecognized command or argument 'Web1.1'
   at Microsoft.DotNet.Cli.CommandLine.CommandLineApplication.HandleUnexpectedArg(CommandLineApplication command, String[] args, Int32 index, String argTypeName)
   at Microsoft.DotNet.Cli.CommandLine.CommandLineApplication.Execute(String[] args)
   at Microsoft.DotNet.Tools.New.NewCommand.Run(String[] args)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, ITelemetry telemetryClient)
   at Microsoft.DotNet.Cli.Program.Main(String[] args)
```

## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (1.0.0-rc4-004616)

Product Information:
 Version:            1.0.0-rc4-004616
 Commit SHA-1 hash:  56d8071361

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\1.0.0-rc4-004616
```

dasMulli:
Looks like `new3` handles it well and new will be replaced by it eventually. But unhandled exceptions are very bad on macOS!:
<img width="830" alt="screen shot 2017-01-26 at 20 38 17" src="https://cloud.githubusercontent.com/assets/6129250/22347236/86b365f0-e407-11e6-8f17-1028355bc5d2.png">

piotrpMSFT:
Fixed by #5493

PR #5493, opened 2017-01-27
display CommandParsingException gracefully

jonsequitur:
This set of changes handles `CommandParsingException` gracefully (so as not to show the user a stack trace) and generalizes graceful exception display somewhat away from being type-specific.

It specifically addresses #5470.

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/ExceptionExtensions.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/ExceptionExtensions.cs
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.DotNet.Cli.Utils.ExceptionExtensions
+namespace Microsoft.DotNet.Cli.Utils
 {
     internal static class ExceptionExtensions
     {
@@ -11,5 +11,10 @@ namespace Microsoft.DotNet.Cli.Utils.ExceptionExtensions
         {
             Reporter.Verbose.WriteLine($"Warning: Ignoring exception: {e.ToString().Yellow()}");
         }
+
+        public static bool ShouldBeDisplayedAsError(this Exception e) =>
+            e.Data.Contains(CLI_User_Displayed_Exception);
+
+        internal const string CLI_User_Displayed_Exception = "CLI_User_Displayed_Exception";
     }
 }
--- a/test/dotnet-new.Tests/NewCommandTests.cs
+++ b/test/dotnet-new.Tests/NewCommandTests.cs
@@ -1,0 +1,33 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using FluentAssertions;
+using Microsoft.DotNet.Tools.Test.Utilities;
+using System;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.New.Tests
+{
+    public class NewCommandTests
+    {
+		private readonly ITestOutputHelper output;
+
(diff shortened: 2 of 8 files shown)


jonsequitur:
@livarcocc @piotrpMSFT @eerhardt @MichaelSimons

test/dotnet-new.Tests/NewCommandTests.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.DotNet.New.Tests
+{
+    public class NewCommandTests
+    {
+		private readonly ITestOutputHelper output;
```
eerhardt: (double nit) spacing and private fields should start with `_`.  Then you can get rid of `this.` in the constructor too.  :)

test/dotnet-new.Tests/NewCommandTests.cs
```diff
@@ -0,0 +1,33 @@
...
+
+		public NewCommandTests(ITestOutputHelper output)
+		{
+			this.output = output;
+		}
+
+        [Fact]
+        public void WhenSwitchIsSkippedThenItPrintsError()
+        {
+            var cmd = new DotnetCommand().Execute("new Web1.1");
```
eerhardt: Since we know that "dotnet new3" is coming in and replacing "dotnet new", it feels like we should use a different command and switch here.

test/dotnet-new.Tests/NewCommandTests.cs
jonsequitur: Old team's convention. Working on changing that habit. 👍

test/dotnet-new.Tests/NewCommandTests.cs
jonsequitur: The bug being fixed is specific to the "new" command regardless of the underlying implementation, which is why I wrote this test. Are you concerned we'll lose coverage when new3 is swapped in? Some other concern?

We've been discussing a more consolidated approach to command line parsing and validation. That is out of scope for this change, though.

eerhardt: My concern is that it will just add more work for #5430 to get merged in.  Where if we picked a different command and switch, we wouldn't have to rewrite this test when #5430 is merged.

jonsequitur:

jonsequitur:

test/dotnet-new.Tests/NewCommandTests.cs
jonsequitur: This test is not specific to New or New3. Is it an invalid test of New3?

eerhardt: Yes, New3 doesn't take the '-t'. So 'dotnet new3 Web1.1' is actually a valid command line.

jonsequitur:

src/Microsoft.DotNet.Cli.Utils/ExceptionExtensions.cs
```diff
@@ -3,13 +3,18 @@
...
     {
         public static void ReportAsWarning(this Exception e)
         {
             Reporter.Verbose.WriteLine($"Warning: Ignoring exception: {e.ToString().Yellow()}");
         }
+
+        public static bool ShouldBeDisplayedAsError(this Exception e) =>
+            e.Data.Contains(CLI_User_Displayed_Exception);
+
+        internal const string CLI_User_Displayed_Exception = "CLI_User_Displayed_Exception";
```
krwq: this feels like it should be an enum value

jonsequitur:

PR #5509, opened 2017-01-29
remove unused and incorrect localizations

jonsequitur:
This addresses #4977. Along the way, I removed a number of localization strings that are not in use.


jonsequitur:
@nguerrera @piotrpMSFT @livarcocc @eerhardt @jgoshi

src/dotnet/commands/dotnet-run/RunCommand.cs
```diff
@@ -119,14 +121,15 @@ private void Initialize()
 
                 if (projectFiles.Length == 0)
                 {
+                    var project = "--project";
+
                     throw new InvalidOperationException(
-                        $"{LocalizableStrings.RunCommandInvalidOperationException1} {directory}." + Environment.NewLine +
-                        LocalizableStrings.RunCommandInvalidOperationException2);
+                        $"Couldn\'t find a project to run. Ensure a project exists in  {directory}, or pass the path to the project using {project}");
```
eerhardt: This and the message below are messages to the user - so they should be localized.

Given our work with exceptions, the exception type should probably change here too.

(nit) Apostrophes don't need to be escaped.

jonsequitur: I was hoping to get some opinions on introducing a loc const for this. In a couple of instances like this one, multi-line strings had been broken into several consts. Loc strings are not supposed to be programmatically concatenated. It makes translation difficult by taking pieces of text out of context and it increases the likelihood that the concatenated string will be wrong in some languages. 

So with that in mind, I inlined and combined the strings but was unsure of the process for introducing new loc consts at this stage, since they won't have matching values in the xlf / resx files.

@nguerrera @piotrpMSFT

src/dotnet/CommonLocalizableStrings.cs
```diff
@@ -5,183 +5,65 @@ namespace Microsoft.DotNet.Tools
 {
     internal class CommonLocalizableStrings
     {
-        public const string UnsupportedProjectType = "Unsupported project type. Please check with your sdk provider.";
         public const string ProjectAlreadyHasAreference = "Project already has a reference to `{0}`.";
         public const string ProjectReferenceCouldNotBeFound = "Project reference `{0}` could not be found.";
         public const string ProjectReferenceRemoved = "Project reference `{0}` removed.";
 
         // Project related
-        public const string Project = "Project";
```
krwq: cc: @blackdwarf - do we still need the backup strings or can we remove them?

jonsequitur:

jonsequitur:

src/dotnet/CommandLine/CommandLineApplication.cs
```diff
@@ -372,11 +372,11 @@ public void ShowHelp(string commandName = null)
                 if (cmd != this && cmd.Arguments.Any())
                 {
                     var args = string.Join(" ", cmd.Arguments.Select(arg => arg.Name));
-                    headerBuilder.Insert(usagePrefixLength, string.Format(LocalizableStrings.UsageItemWithArgs, cmd.Name, args));
+                    headerBuilder.Insert(usagePrefixLength, string.Format(" {0} {1}", cmd.Name, args));
```
piotrpMSFT: These templates were originally in LocalizableStrings to allow for string ordering reversals [RTL languages, etc]. Is that no longer needed?

src/dotnet/CommonLocalizableStrings.cs
piotrpMSFT: UnsupportedProjectType will be needed once we have the ProjectTypeGuid lookup in place. @jonsequitur we added many of these strings proactively because the turnaround time for Loc is non-0. Can we move them to a separate section of the file instead of deleting?

src/dotnet/CommandLine/CommandLineApplication.cs
jonsequitur: I haven't seen this approach used before. Translators typically do the ordering reversals by swapping the token positions within the larger, translated string. A loc string that's nothing but tokens is very hard to understand and the extra indirection will lead to bugs in specific languages that the dev team is unlikely to catch.

src/dotnet/CommonLocalizableStrings.cs
jonsequitur: Which ones specifically? Should we move them into the `LocalizableStrings` section for a specific command?

jonsequitur:
This PR has been superseded by #5618.

Issue #5448, opened 2017-01-25
dotnet add <project> package no longer works


danroth27:
I used to be able to specify the project path when running 'dotnet add package', but it's no longer working:

```
C:\>dotnet add test\test.csproj package Microsoft.Extensions.Configuration
Please specify one package reference to add.

C:\>dotnet --info
.NET Command Line Tools (1.0.0-rc4-004597)

Product Information:
 Version:            1.0.0-rc4-004597
 Commit SHA-1 hash:  0a62481cc0

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\1.0.0-rc4-004597
```

It used to work:

```
C:\>dotnet add test\test.csproj package Microsoft.Extensions.Configuration
Microsoft (R) Build Engine version 15.1.523.56541
Copyright (C) Microsoft Corporation. All rights reserved.

  Writing C:\Users\daroth\AppData\Local\Temp\tmp8822.tmp
info : Adding PackageReference for package 'Microsoft.Extensions.Configuration' into project 'test\test.csproj'.
log  : Restoring packages for C:\test\test.csproj...
info :   GET https://api.nuget.org/v3-flatcontainer/microsoft.extensions.configuration/index.json
info :   OK https://api.nuget.org/v3-flatcontainer/microsoft.extensions.configuration/index.json 236ms
info :   GET https://api.nuget.org/v3-flatcontainer/microsoft.extensions.configuration/1.1.0/microsoft.extensions.configuration.1.1.0.nupkg
info :   OK https://api.nuget.org/v3-flatcontainer/microsoft.extensions.configuration/1.1.0/microsoft.extensions.configuration.1.1.0.nupkg 368ms
info :   GET https://www.myget.org/F/aspnetcoremodules/api/v3/flatcontainer/microsoft.extensions.configuration/index.json
info :   NotFound https://www.myget.org/F/aspnetcoremodules/api/v3/flatcontainer/microsoft.extensions.configuration/index.json 181ms
info :   GET https://www.myget.org/F/danroth27/api/v3/flatcontainer/microsoft.extensions.configuration/index.json
info :   NotFound https://www.myget.org/F/danroth27/api/v3/flatcontainer/microsoft.extensions.configuration/index.json 198ms
[...]

PR #5533, opened 2017-01-31
refresh enumerator when moving to next subcommand

jonsequitur:
Subcommand argument parsing has a bug that causes some arguments to be ignored due to the same command enumerator instance was being used for different subcommands. 

This addresses #5448 specifically so the test I added uses the command line pattern from the repro, but the problem is general to all subcommands.


jonsequitur:
@piotrpMSFT @livarcocc @eerhardt @jgoshi

jonsequitur:

piotrpMSFT:

PR #5581, opened 2017-02-04
Update MSBuild

piotrpMSFT:
@livarcocc


piotrpMSFT:
@MattGertz updating MSBuild in CLI per 4:30 call.

PR #5600, opened 2017-02-06
Merge rel/1.0.0 to Master

piotrpMSFT:
@eerhardt @livarcocc @jonsequitur @krwq @jgoshi


eerhardt:
@piotrpMSFT - did you resolve that "dotnet new" is now "dotnet new3"?  Thus we need to update the templates to lay down "2.0" projects.  And also fixing up the nuget package archiver.

piotrpMSFT:
@eerhardt for the moment I resolved any merge conflicts and want to see what CI shows. Ack on `new` templates needing to lay down 2.0 references. @mlorbetske what's the fastest way for us to make that happen? @eerhardt what about the lzma archiver would not have been addressed by a merge?

piotrpMSFT:
and clearly there is work to be done here :)

eerhardt:
I've been working with @seancpeters on his ideas for getting the templates updated for 2.0.  Specifically we need to:

1. Change the TFM to `netcoreapp2.0`
1. Insert a new property `<RuntimeFrameworkVersion>$CurrentSharedFrameworkVersion$</RuntimeFrameworkVersion>` where `$CurrentSharedFrameworkVersion$` is changed to the current shared framework's version.
1. Lay down a NuGet.config file to point to the `dotnet-core` feed because the 2.0 packages aren't officially shipped.

As for the archiver, today it is using the checked-in "2.0 console" template since there are no "2.0 web" templates/packages. Plus it uses the current SharedFX version to generate the archive hash.  In "dotnet new3" world, it is using the ephemeral-hive and the Web templates (which aren't 2.0).  We need to resolve this so the 2.0 templates are used.

piotrpMSFT:
ok got it. should I consider this merge PR to be something your'e driving and close ?

eerhardt:
> ok got it. should I consider this merge PR to be something your'e driving and close ?

I really haven't been spending much time on it outside of exchanging a few conversations with @seancpeters.  I haven't tried any of his suggestions, yet.  So if someone else wants to take this, feel free.

seancpeters:
+ @mlorbetske 
Are any of you in the office today? If so, we should try getting together to work out what needs to go into the new templates for netcoreapp2.0. Mike & I can wander over to 25 if anyone is there.

piotrpMSFT:
@seancpeters most of us are remote today due to snow. how about a call @4:00?

seancpeters:
@piotrpMSFT @mlorbetske 
Heh, yeah, my wishful thinking people would be in. 4 works for me.

eerhardt:
@piotrpMSFT @livarcocc @jonsequitur - this is looking green.  Can you guys review the 2 commits I made?  Then I can get this in to 'master' and we should be golden.

PR #5613, opened 2017-02-07
Removing extra PackageReference workaround for tests

emgarten:
Removing PackageReferences which were added to work around a test issue. I don't see any reason why these would be needed. When trying these locally restore works fine without the extra references.

Fixes https://github.com/NuGet/Home/issues/4416

//cc @rohit21agrawal


emgarten:
@livarcocc @piotrpMSFT it looks like this extra workaround in the test code is no longer needed. I'm not seeing any issues around this on my side, and the CI looks happy with removing it.

piotrpMSFT:
@MattGertz, this is a test cleanup PR that doesn't affect product code. Are you comfortable with us merging?

MattGertz:
Yup, fine.

PR #5618, opened 2017-02-08
Remove incorrect localizations

jonsequitur:
I've rebased these localization fixes that I made against rel/1.0.0 onto master. 

This replaces #5509.


src/dotnet/commands/dotnet-run/LocalizableStrings.cs
```diff
@@ -16,26 +16,14 @@ internal class LocalizableStrings
...
-        public const string RunCommandMSBuildExtensionsPath = "MSBuildExtensionsPath";
-
-        public const string RunCommandConfiguration = "Configuration";
-
-        public const string RunCommandTargetFramework = "TargetFramework";
-
-        public const string RunCommandProjectInstance = "RunCommand";
-
-        public const string RunCommandOutputType = "OutputType";
+        public const string RunCommandExceptionUnableToRun = "Unable to run your project\nPlease ensure you have a runnable project type and ensure '{0}' supports this project.\nThe current {1} is '{2}'";
```
mairaw: Missing a period after project. According to Microsoft's style guide, you should avoid using please unless you're asking something to the user that is inconvenient or the app is to blame for this. 

Suggested change:
"Unable to run your project.\n Ensure you have a runnable project type and that '{0}' supports this project.\nThe current {1} is '{2}'."

src/dotnet/commands/dotnet-sln/LocalizableStrings.cs
```diff
@@ -12,7 +12,7 @@ internal class LocalizableStrings
         public const string AppHelpText = "Projects to add or to remove from the solution.";
 
         public const string AddAppFullName = ".NET Add project(s) to a solution file Command";
-        public const string AddSubcommandHelpText = "Add a specified project(s) to the solution.";
+        public const string AddSubcommandHelpText = "Add one or more specified projects to the solution.";
```
mairaw: Add one or more specified projects is a bit awkward. It should be "Add the specified projects to the solution" or "Add one or more projects to the solution". From what I can I see in the SDK today, the latter would be fine.

src/dotnet/commands/dotnet-test/LocalizableStrings.cs
```diff
@@ -36,7 +36,7 @@ internal class LocalizableStrings
 
         public const string CmdConfiguration = "CONFIGURATION";
 
-        public const string CmdConfigDescription = "Configuration to use for building the project.  Default for most projects is  \"Debug\".";
+        public const string CmdConfigDescription = "Configuration to use for building the project.  Default for most projects is \"Debug\".";
```
mairaw: I find this string "Default for most projects" which is used in a lot of places  really confusing for customers. It doesn't tell me when I would not get the default.

test/dotnet-sln-add.Tests/GivenDotnetSlnAdd.cs
```diff
@@ -25,8 +26,14 @@ public class GivenDotnetSlnAdd : TestBase
   -h|--help  Show help information
 
 Additional Arguments:
- Add a specified project(s) to the solution.
+ Add one or more specified projects to the solution.
```
mairaw: same as the comment before about a similar string.

piotrpMSFT:
@blackdwarf can you help out with the string comments? I think they probably should be addressed in a separate PR but want to be sure we track anything we want to change.

blackdwarf:
@mairaw let's open a separate PR against master and go through and fix all of the strings in one fell swoop, so to speak. Sounds ok?

src/dotnet/commands/dotnet-run/LocalizableStrings.cs
akoeplinger: The original code used Environment.NewLine, `\n` is probably not correct on Windows in the new string.

piotrpMSFT:
@jonsequitur are we going to merge this?

mairaw:
Yes, sounds good @blackdwarf

src/dotnet/commands/dotnet-run/LocalizableStrings.cs
jonsequitur: @akoeplinger Concatenation of multiple strings is something to be avoided in localization. I'm unaware of `\n` causing issues in this kind of usage.

jonsequitur:
@mairaw Was an issue opened to track this? I'm planning to do some refactoring of the localization code and would like to keep those two changes coordinated.

mairaw:
@jonsequitur I missed your message earlier. I haven't opened an issue for this specific change. But I was talking to @blackdwarf about this and I'd like to verify the strings a little bit more extensively since there are a lot of inconsistencies, missing punctuation, etc. I'll open an issue.

PR #5623, opened 2017-02-08
Remove project.json-based omnisharp.json

piotrpMSFT:
/cc @jonsequitur @krwq @livarcocc @eerhardt @jgoshi @DustinCampbell

### DIFF
--- a/omnisharp.json
+++ b/omnisharp.json
@@ -1,6 +1,0 @@
-{
-    "dotnet": {
-        "projects": "src/*/project.json;test/*/project.json;scripts/*/project.json",
-        "enablePackageRestore": false
-    }
-}


omnisharp.json
```diff
@@ -1,6 +0,0 @@
-{
-    "dotnet": {
-        "projects": "src/*/project.json;test/*/project.json;scripts/*/project.json",
```
DustinCampbell: To be honest, I think "projects" was a hold-out from the days of DNX-support in OmniSharp. When that was dropped, this no longer worked.

omnisharp.json
```diff
@@ -1,6 +0,0 @@
...
-        "enablePackageRestore": false
```
DustinCampbell: The default was false anyways.

PR #5627, opened 2017-02-08
Fix the CLI .sln file to load and build in VS 2017.

eerhardt:
@jonsequitur @livarcocc @piotrpMSFT


eerhardt:

PR #5661, opened 2017-02-10
Refactor tasks used in generate static props files

jgoshi:
@piotrpMSFT @livarcocc @eerhardt @krwq @jonsequitur


build_projects/dotnet-cli-build/GetVersionBadgeMoniker.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.DotNet.Cli.Build
+{
+    public class GetVersionBadgeMoniker : Task
+    {
+        [Output]
+        public string VersionBadgeMoniker { get; set; }
+
+        public override bool Execute()
+        {
+            VersionBadgeMoniker = Monikers.GetBadgeMoniker();
```
eerhardt: I don't think we need a task for this.  I think the logic in GetBadgeMoniker can be moved to MSBuild.

run-build.ps1
```diff
@@ -110,11 +110,11 @@ $env:PATH = "$env:DOTNET_INSTALL_DIR;$env:PATH"
...
 {
     Write-Host "Not building due to --nobuild"
-    Write-Host "Command that would be run: 'dotnet msbuild build.proj /m /p:Architecture=$Architecture $ExtraParameters'"
+    Write-Host "Command that would be run: 'dotnet msbuild build.proj /m /p:OverrideArchitecture=$Architecture $ExtraParameters'"
 }
 else
 {
-    dotnet msbuild build.proj /p:Architecture=$Architecture /p:GeneratingPropsFile=true /t:WriteDynamicPropsToStaticPropsFiles
-    dotnet msbuild build.proj /m /v:diag /p:Architecture=$Architecture $ExtraParameters
+    dotnet msbuild build.proj /p:OverrideArchitecture=$Architecture /p:GeneratingPropsFile=true /t:WriteDynamicPropsToStaticPropsFiles
```
eerhardt: Why are we introducing this new property?

jgoshi: I wanted to rename it. So that passing something in from the command-line can override that value stored in the props file. To allow x86 build on x64 machine for example.

eerhardt: Usually passing something in from the command line overrides what is stored in the props file by default.  Doesn't it?

build_projects/dotnet-cli-build/GetVersionBadgeMoniker.cs
jgoshi: Where can I find the implementation of GetBadgeMoniker?

run-build.ps1
jgoshi: When we import the props file wouldn't that overwrite the property value we passed in?

eerhardt: @rainersigwald @jeffkl @cdmihai - is there a ruling on this?  Doesn't the commandline win?

build_projects/dotnet-cli-build/GetVersionBadgeMoniker.cs
eerhardt: https://github.com/dotnet/cli/search?utf8=%E2%9C%93&q=GetBadgeMoniker

run-build.ps1
rainersigwald: Properties passed in via `/p:PropName=value` on the command line are immutable within the build, so @eerhardt's right that they always* win.

* There is an exception: if the project specifies that a property is one of its `TreatAsLocalProperties`, then it can be changed within the build. That's rare and not the default.

jgoshi: Ok cool. I'll remove this part then.

PR #5679, opened 2017-02-12
Read default language for `dotnet new` from env var

enricosada:
Read `DOTNET_NEW_PREFERRED_LANG` as default language for `dotnet new`.
If not set, the default continue to be `C#`

Currently to create F# projects, we need to specify `-lang f#` foreach `dotnet new` invocation.
It's really annoying, and atm `dotnet new` doesnt yet support set the preference property.

This option can be removed later, when `dotnet new` support set the default language (as normal config)

With this pr

```
set DOTNET_NEW_PREFERRED_LANG=F#
dotnet new console
dotnet new lib
```

/cc @cartermp @dsyme

### DIFF
--- a/src/dotnet/commands/dotnet-new/NewCommandShim.cs
+++ b/src/dotnet/commands/dotnet-new/NewCommandShim.cs
@@ -49,9 +49,11 @@ namespace Microsoft.DotNet.Tools.New
                 { new Guid("10919118-4E13-4FA9-825C-3B4DA855578E"), () => typeof(CaseChangeMacro) }
             }.ToList();
 
+            string preferredLang = Environment.GetEnvironmentVariable("DOTNET_NEW_PREFERRED_LANG") ?? "C#";
+
             var preferences = new Dictionary<string, string>
             {
-                { "prefs:language", "C#" },
+                { "prefs:language", preferredLang },
                 { "dotnet-cli-version", Product.Version }
             };
 


piotrpMSFT:
@enricosada this should go to the dotnet/templating repo. That's where the templating engine is developed. The configuration of default language should be consistent across usages of the teplating engine.

/cc @mlorbetske FYI

mlorbetske:
@piortpMSFT the preference is currently set in the CLI, if it moved in to the engine the CLI would be overriding the behavior as it stands. If that's the behavior we want then we'll need to coordinate the change to remove the preference from CLI & add it to the engine - any thoughts on timing for that?

piotrpMSFT:
@mlorbetske let's discuss tomorrow.

enricosada:
@piotrpMSFT If i need to implement anything, just tell me. I tried that to have a small diff. 
I'd like something to set language preference for 1.0 Rtm because is really awesome to have it

piotrpMSFT:
@mlorbetske and I didn't run across eachother today. The 1.0 bar is quite high at this point, @enricosada. I don't think we'll be able to merge it for 1.0 in CLI.

enricosada:
@piotrpMSFT i understand, i was late (finalizing fsproj is priority for me atm, working on that). 
That's why i tried with a simple env var (low risk) instead of implement `--set`, but is ok i think. 
Next patch version maybe :)

piotrpMSFT:
btw @enricosada you should chat with @livarcocc tomorrow. There is a change we need to make to the dotnet-compile-fsc tool...

enricosada:
@piotrpMSFT for sure, when you want @livarcocc .

Btw i'll send an update template really soon (in test now https://github.com/dotnet/netcorecli-fsc/pull/61, so max 1-2 days), ref [F# sdk roadmap](https://github.com/dotnet/netcorecli-fsc/wiki/Roadmap)

That will:

- completly remove `dotnet-compile-fsc` tool requirement
- use `fsc` from `FSharp.Compiler.Tools` packaged (fsc it's published as normal framework dependent app, and invoked like `dotnet packagepath/fsc.dll`)
- use msbuild task `FscTask` (from same package), that's the same task used by net40 code previously.

But let's chat if you can @livarcocc so i'll check if i need to change new code too.

PR #5695, opened 2017-02-14
Get update-dependenices working again

ellismg:
- Fix build issues due to .props file not existing on a clean
  enlistment
- Fix path separator issues on *NIX
- Add update-dependencies.sh
- Support -Update in update-dependencies.ps1

### DIFF
--- a/build_projects/update-dependencies/update-dependencies.ps1
+++ b/build_projects/update-dependencies/update-dependencies.ps1
@@ -4,14 +4,16 @@
 #
 
 param(
-    [switch]$Help)
+    [switch]$Help,
+    [switch]$Update)
 
 if($Help)
 {
     Write-Host "Usage: .\update-dependencies.ps1"
     Write-Host ""
     Write-Host "Options:"
-    Write-Host "  -Help                              Display this help message"
+    Write-Host "  -Help                 Display this help message"
+    Write-Host "  -Update               Update dependencies (but don't open a PR)"
     exit 0
 }
 
@@ -19,6 +21,12 @@ $Architecture='x64'
 
 $RepoRoot = "$PSScriptRoot\..\.."
 $ProjectPath = "$PSScriptRoot\update-dependencies.csproj"
+$ProjectArgs = ""
+
+if ($Update)
+{
+    $ProjectArgs = "--Update"
+}
 
 # Use a repo-local install directory (but not the artifacts directory because that gets cleaned a lot
 if (!$env:DOTNET_INSTALL_DIR)
@@ -34,6 +42,11 @@ if($LASTEXITCODE -ne 0) { throw "Failed to install stage0" }
 # Put the stage0 on the path
 $env:PATH = "$env:DOTNET_INSTALL_DIR;$env:PATH"
 
+# Generate some props files that are imported by update-dependencies
+Write-Host "Generating property files..."
+dotnet msbuild $RepoRoot\build.proj /p:Architecture=$Architecture /p:GeneratingPropsFile=true /t:WriteDynamicPropsToStaticPropsFiles
+if($LASTEXITCODE -ne 0) { throw "Failed to generate intermidates" }
+
 # Restore the app
 Write-Host "Restoring $ProjectPath..."
 dotnet restore "$ProjectPath"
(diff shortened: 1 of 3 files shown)


ellismg:
@eerhardt, can you please take a look?

FYI, I noticed that the pended changes from this script have a trailing carriage return at the end of each line, but it wasn't obvious to me why, I assume it has something to do with how VersionTools is updating this stuff.  @dagood might know more.  I was seeing this on both Ubuntu 14.04 and Windows.  It didn't seem to cause any issues, however.

build_projects/update-dependencies/update-dependencies.sh
```diff
@@ -0,0 +1,63 @@
...
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ "$SOURCE" != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+REPO_ROOT="$DIR/../.."
+PROJECT_PATH="$DIR/update-dependencies.csproj"
+
+# Some things depend on HOME and it may not be set. We should fix those things, but until then, we just patch a value in
```
eerhardt: Is this section necessary?  What needs this?

build_projects/update-dependencies/update-dependencies.sh
```diff
@@ -0,0 +1,63 @@
...
+if [ $? -ne 0 ]; then
+    echo "Failed to install stage 0"
+    exit 1
+fi
+
+# Generate some props files that are imported by update-dependencies
+echo "Generating property files..."
+dotnet msbuild "$REPO_ROOT/build.proj" /p:Architecture=x64 /p:GeneratingPropsFile=true /t:WriteDynamicPropsToStaticPropsFiles
+
+# Put the stage 0 on the path
```
eerhardt: Do we need to do this before we call `dotnet msbuild` above?

build_projects/update-dependencies/update-dependencies.sh
ellismg: Are you talking about setting `$HOME`?  I was seeing cases where $HOME is unset and that causing issues for `dotnet restore` (since it can't find the home directory to put `.nuget/packages` in during restore).

I'm happy to remove this as all the scinerios I care about have $HOME set, but this could be useful when running the script in Jenkins or maybe Docker.

build_projects/update-dependencies/update-dependencies.sh
ellismg: Yeah, I will swap them.

build_projects/update-dependencies/update-dependencies.ps1
```diff
@@ -34,12 +42,17 @@ if($LASTEXITCODE -ne 0) { throw "Failed to install stage0" }
 # Put the stage0 on the path
 $env:PATH = "$env:DOTNET_INSTALL_DIR;$env:PATH"
 
+# Generate some props files that are imported by update-dependencies
```
eerhardt: I don't think update-dependencies actually needs to import the dirs.props in the root.  Can we just remove the Import in its .csproj instead?  That way this call isn't duplicated.

build_projects/update-dependencies/update-dependencies.sh
eerhardt: ok, I'm fine with keeping this if it is needed.  I just didn't know why it was necessary.

---
In reply to: [100941962](https://github.com/dotnet/cli/pull/5695#discussion_r100941962) [](ancestors = 100941962)

build_projects/update-dependencies/update-dependencies.ps1
ellismg: I'll give it a shot.

ellismg: For the rest of the internet, it continues to be needed because update-dependencies.csproj has a P2P reference to Microsoft.DotNet.Cli.Utils/Microsoft.DotNet.Cli.Utils.csproj, so I left this in.

ellismg:
@eerhardt Do you know if the OpenSUSE 13.2 leg is supposed to pass?  I know for other repos for 2.0 we've been pulling that out because it has been end of life'd and Azure didn't have VMs we could build on.  Are we doing a similar thing for CLI 2.0?

eerhardt:

it should be passing.  We are a little behind responding to distro changes in the CLI, But CI should always be passing.

dagood:
> trailing carriage return at the end of each line

I tried running your branch on Windows (after downgrading stuff in DependencyVersions.props) and this didn't seem to repro. Could you maybe push the commit somewhere so I can see what you're seeing?

eerhardt:
@ellismg - feel free to merge at your will.

PR #5706, opened 2017-02-15
Remove import of dir.props from Signing.proj

naamunds:
The build agents in our VSTS pool don't yet have MSBuild 15, so the
signing project can't import any MSBuild 15 projects. (The signing
project is built with the desktop MSBuild because MicroBuild depends on
it.)  Once the agents have MSBuild 15, this import can be added back.


naamunds:
I'll log an issue for removing this workaround when the build agents have MSBuild 15, which should happen shortly after VS 2017 RTMs.

piotrpMSFT:
@MattGertz this is an infra-only change that will enable us to move forward on CI infra replacement. We'll merge once you've had a chance to approve.

MattGertz:
Thanks, agreed.

PR #5705, opened 2017-02-15
Rename props files

jgoshi:
@piotrpMSFT @livarcocc @eerhardt @krwq @jonsequitur

PR #5693, opened 2017-02-13
Remove more properties from build.proj

jgoshi:
@piotrpMSFT @krwq @livarcocc @jonsequitur @eerhardt

PR #5713, opened 2017-02-15
Misc build changes

ellismg:
For injestion into a composed build, I'm adding some minor tweaks to
our build system:

 - Teach DownloadFile task how to consume file:// URLs so we can pass
   around artifacts using local sources.
 - Add `SkipBuildingInstallers`, an MSBuild property that can be set
   to true when installers (pkgs, msis, debs, rpms and the like) do
   not need to be built.
 - Add `IncludeAdditionalSharedFrameworks`, an MSBuild property that
   can be set to prevent additional shared frameworks (i.e. shared
   frameworks that the CLI does not use at runtime) from being
   downloaded and included in our payload.
 - Add `IncludeNuGetPackageArchive` an MSBuild property that can be
   set to prevent the lzma archive containing all the nupkgs we
   restore on first run from being included in the final output.
 - Provide a way to change the Uri used to pull down the shared
   framework archive (so that during a composed buid we can point it
   at a local source).

### DIFF
--- a/build/Microsoft.DotNet.Cli.Prepare.targets
+++ b/build/Microsoft.DotNet.Cli.Prepare.targets
@@ -35,7 +35,7 @@
     </PropertyGroup>
 
     <!-- Additional Shared Framework to be installed -->
-    <PropertyGroup>
+    <PropertyGroup Condition=" '$(IncludeAdditionalSharedFrameworks)' == 'true' ">
       <AdditionalCoreSetupChannel>preview</AdditionalCoreSetupChannel>
       <AdditionalSharedFrameworkVersion>1.0.3</AdditionalSharedFrameworkVersion>
       <AdditionalSharedHostVersion>1.0.1</AdditionalSharedHostVersion>
@@ -61,7 +61,7 @@
 
   <Target Name="SetupDownloadHostAndSharedFxInputsOutputs" DependsOnTargets="Init">
     <PropertyGroup>
-      <CoreSetupBlobRootUrl>https://dotnetcli.azureedge.net/dotnet/</CoreSetupBlobRootUrl>
+      <CoreSetupBlobRootUrl Condition="'$(CoreSetupBlobRootUrl)' == ''">https://dotnetcli.azureedge.net/dotnet/</CoreSetupBlobRootUrl>
       <CoreSetupBlobRootUrlWithChannel>$(CoreSetupBlobRootUrl)$(CoreSetupChannel)</CoreSetupBlobRootUrlWithChannel>
       <SharedFrameworkArchiveBlobRootUrl>$(CoreSetupBlobRootUrlWithChannel)/Binaries/$(SharedFrameworkVersion)</SharedFrameworkArchiveBlobRootUrl>
       <CoreSetupInstallerBlobRootUrl>$(CoreSetupBlobRootUrlWithChannel)/Installers</CoreSetupInstallerBlobRootUrl>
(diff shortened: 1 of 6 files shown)


build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -145,37 +145,37 @@
       </_DownloadAndExtractItem>
 
       <_DownloadAndExtractItem Include="DownloadedSharedFrameworkInstallerFile"
-                               Condition="!Exists('$(DownloadedSharedFrameworkInstallerFile)') And '$(InstallerExtension)' != ''">
+                               Condition="'$(SkipBuildingInstallers)' != 'true' And !Exists('$(DownloadedSharedFrameworkInstallerFile)') And '$(InstallerExtension)' != ''">
```
ellismg: I noticed these conditions did not embrace the spaces between the opening and closing quotes of the Condition attribute.  I'm happy to add them if that's the style here.

piotrpMSFT:
@livarcocc FYI

eerhardt:
> Add IncludeAdditionalSharedFrameworks, an MSBuild property that
can be set to prevent additional shared frameworks (i.e. shared
frameworks that the CLI does not use at runtime) from being
downloaded and included in our payload.

FYI - today `dotnet test` requires having the `1.0.x` runtime in order to work.  This is probably a point-in-time issue, but not having the `1.0.x` runtime installed will mean `dotnet test` won't work at all.  I've logged https://github.com/Microsoft/vstest/issues/494 for this issue in `dotnet test`.

build/Microsoft.DotNet.Cli.Prepare.targets
eerhardt: I will defer to @piotrpMSFT and @livarcocc, but IMO I've seen a mix of both styles all over that I don't think one way or the other is "the preferred style".

build/Microsoft.DotNet.Cli.Prepare.targets
```diff
@@ -145,37 +145,37 @@
...
+                               Condition="'$(SkipBuildingInstallers)' != 'true' And !Exists('$(DownloadedHostFxrInstallerFile)') And '$(InstallerExtension)' != ''">
         <Url>$(CoreSetupInstallerBlobRootUrl)/$(HostFxrVersion)/$(DownloadedHostFxrInstallerFileName)</Url>
         <DownloadFileName>$(DownloadedHostFxrInstallerFile)</DownloadFileName>
         <ExtractDestintation></ExtractDestintation>
       </_DownloadAndExtractItem>
     </ItemGroup>
 
     <!-- Additional Shared Framework to be installed -->
-    <PropertyGroup>
+    <PropertyGroup Condition=" '$(IncludeAdditionalSharedFrameworks)' != 'false' ">
```
eerhardt: The is probably a personal preference thing, but to me a blank value and "false" go together.  There's no surprises that if a property isn't set, then its logic doesn't get run.

So if we want the default value of a boolean property to be "true", it is better to to have a setter that defaults the value to "true", and then keep the Condition logic the normal `$(Property) == 'true'`.

ellismg: This makes sense.  I'll add this (and the other configuration properties (SkipBuildingInstallers and IncludeNuGetPackageArchive)) to `build.proj` and give them correct defaults.  Is build.proj the right place to set up this general config info?

eerhardt: No, we are trying to refactor our properties into logical groupings and not have `dirs.props` and `build.proj` be dumping grounds for random things.  See #5693 for an example.

I'm not sure where a good place would be right now. @piotrpMSFT @jgoshi - any ideas?

ellismg: It looks like `build/Microsoft.DotNet.Cli.BuildDefaults.props` may be a reasonable place?

eerhardt: That sounds good to me.

eerhardt: FYI - @jgoshi is working on renaming all those .props files to drop "Microsoft.DotNet.Cli" from the names.  See #5705

ellismg:
@eerhardt I rebased on top of master to pick up the build.proj refactoring and added my defaults to `build/Microsoft.DotNet.Cli.BuildDefaults.props`

eerhardt:
test Ubuntu x64 Release Build 

flaky test: WhenToolAssetsFileIsInUseThenCLIRetriesLaunchingTheCommandForAtLeastOneSecond

PR #5722, opened 2017-02-15
don't try to make crossgen executable if crossgen is disabled (it might not exist!)

stevedesmond-ca:
During an attempted "cross-compile" build (using #5290) for ubuntu.16.04-arm on ubuntu.16.04-x64, the build fails when trying to set crossgen as executable, because the crossgen file doesn't exist.

With this PR, the build script will ignore that step (thus completing the build) if /p:DISABLE_CROSSGEN=1 is set. 

Perhaps of note: the step immediately after this one uses the same condition.


eerhardt:
test OSX x64 Release Build
test Ubuntu x64 Release Build

flaky tests

PR #5728, opened 2017-02-16
Attempt to fix intermittent failures with ...CLIRetriesLaunchingTheCommandForAtLeastOneSecond 

krwq:
We are seeing some intermittent failures with WhenToolAssetsFileIsInUseThenCLIRetriesLaunchingTheCommandForAtLeastOneSecond 

I'm checking if this simple fix will help - I will retry this on CI several times so please do not merge until we can prove this is fixing the problem

My suspicion is that because we do not lock when checking if the file exists it may be failing before it tries to retry or while the file is being written


krwq:
already seeing one more place to which needs lock when checking if file exists (ProjectToolsCommandResolver.TryGetToolLockFile)

PR #5731, opened 2017-02-16
remove dotnet restore-projectjson

piotrpMSFT:


piotrpMSFT:
@livarcocc @jonsequitur @jgoshi @krwq

PR #5738, opened 2017-02-16
Remove BundledSdks.props

jgoshi:

I also wanted to confirm something. I was going to remove the properties from BundledSdks.proj, but think they both need to remain.

This one is a workaround for issue https://github.com/Microsoft/msbuild/issues/885
&lt;CLIBuildDllPath&gt;$([MSBuild]::Unescape($(CLIBuildDll)))&lt;/CLIBuildDllPath&gt;

This one is computed as part of the build based on properties set by the caller.
&lt;SdkNuPkgPath&gt;$(NuGetPackagesDir)/$(SdkPackageName.ToLower())/$(SdkPackageVersion.ToLower())&lt;/SdkNuPkgPath&gt;

So there are no properties to separate out of BundledSdks.proj


jgoshi:
@dotnet-bot test this please

piotrpMSFT:
Why do we want to make this change?

jgoshi:
I can cancel this PR if you want. I was looking at these props files and they didn't have any properties, just ItemGroup. And we were moving away from creating props files with the same name as a targets files (creating props files based on type of property rather than a props file per target file). So I thought it was best to remove these props files. If you think this change is unnecessary let me know and I'll cancel it.

jgoshi:
Closing this as it is unnecessary

PR #5742, opened 2017-02-16
Convert redist.csproj outputdir to static property

piotrpMSFT:
Also some small proj file cleanups

/cc @livarcocc @krwq @jonsequitur @eerhardt @jgoshi

### DIFF
--- a/build/OutputDirectories.props
+++ b/build/OutputDirectories.props
@@ -7,7 +7,7 @@
       <DotnetStage1>$(Stage1Directory)/dotnet$(ExeExtension)</DotnetStage1>
       <Stage2Directory>$(BaseOutputDirectory)/stage2</Stage2Directory>
       <Stage2SymbolsDirectory>$(BaseOutputDirectory)/stage2symbols</Stage2SymbolsDirectory>
-      <Stage2SdkDirectory>$(Stage2Directory)/sdk/$(CommitCount)</Stage2SdkDirectory>
+      <Stage2SdkDirectory>$(Stage2Directory)/sdk/$(SdkVersion)</Stage2SdkDirectory>
       <DotnetStage2>$(Stage2Directory)/dotnet$(ExeExtension)</DotnetStage2>
       <OutputDirectory>$(Stage2Directory)/</OutputDirectory>
       <Stage2CompilationDirectory>$(BaseOutputDirectory)/stage2compilation</Stage2CompilationDirectory>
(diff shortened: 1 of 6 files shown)


src/redist/redist.csproj
```diff
@@ -7,23 +7,16 @@
...
-      <Version>$(CLI_NuGet_Version)</Version>
-    </PackageReference>
-    <PackageReference Include="Microsoft.TestPlatform.CLI">
-      <Version>$(CLI_TestPlatform_Version)</Version>
-    </PackageReference>
-    <PackageReference Include="Microsoft.TestPlatform.Build">
-      <Version>$(CLI_TestPlatform_Version)</Version>
-    </PackageReference>
+    <PackageReference Include="Microsoft.NETCore.App" Version="$(CLI_SharedFrameworkVersion)" />
+    <PackageReference Include="Microsoft.Build.Runtime" Version="$(CLI_MSBuild_Version)" />
```
krwq: this is probably not something we want in this PR but we could update the versions in a target for all projects - that would be much less error prone

build/OutputDirectories.props
```diff
@@ -7,6 +7,7 @@
       <DotnetStage1>$(Stage1Directory)/dotnet$(ExeExtension)</DotnetStage1>
       <Stage2Directory>$(BaseOutputDirectory)/stage2</Stage2Directory>
       <Stage2SymbolsDirectory>$(BaseOutputDirectory)/stage2symbols</Stage2SymbolsDirectory>
+      <Stage2SdkDirectory>$(Stage2Directory)/sdk/$(CommitCount)</Stage2SdkDirectory>
```
eerhardt: Should be able to remove this duplication in the Compile.targets file by changing this line:

'<SdkOutputDirectory>$(StageDirectory)/sdk/$(SdkVersion)</SdkOutputDirectory>'

to

'<SdkOutputDirectory>$(Stage2SdkDirectory)</SdkOutputDirectory>'

eerhardt: That way these properties can't get out of sync.

src/redist/redist.csproj
piotrpMSFT: how would that differ from what what we do today? the versions come from a unified place...

krwq: i mean that the target could add version metadata and you would just say the reference name

piotrpMSFT: I see. I like the idea! The problem is that we need PackageReferences to be defined during the static property evaluation phase. by the time targets run it's too late because the itemgroup may have already been examined.

Perhaps we can do something with ItemGroupDefinitions?

piotrpMSFT:

PR #5759, opened 2017-02-17
Refactor properties out of test.targets

jgoshi:


build/Test.targets
```diff
@@ -16,12 +11,11 @@
       <PathListSeparator Condition=" '$(OSName)' == 'win' ">%3b</PathListSeparator>
       <ExecPath>$(Stage2Directory)$(PathListSeparator)$(PATH)</ExecPath>
       <ExecPath Condition=" '$(OSName)' == 'win' ">$(Stage2Directory)</ExecPath>
-      <TestArtifactsEnv>$(TestArtifactsDir)</TestArtifactsEnv>
     </PropertyGroup>
```
jgoshi: I felt that these properties could remain here since they depend on environment variables. Or should they also be separated out into a static props file?

build/Test.targets
```diff
@@ -16,12 +11,11 @@
...
 
     <ItemGroup>
       <TestTaskEnvironmentVariables Include="PATH=$(ExecPath)" />
-      <TestTaskEnvironmentVariables Include="TEST_ARTIFACTS=$(TestArtifactsEnv)" />
+      <TestTaskEnvironmentVariables Include="TEST_ARTIFACTS=$(TestArtifactsDir)" />
     </ItemGroup>
 
     <PropertyGroup>
```
jgoshi: Same comment as above.

jgoshi:

build/Test.targets
piotrpMSFT: At a minimum, move the property group to be a direct descendant of <Project>. No need for them to be evaluated in a target, as far as I can see.

build/Test.targets
piotrpMSFT: Same comment as above.

build/test/TestProjects.targets
```diff
@@ -10,20 +10,20 @@
...
-                                      test$(PathSeparator)Msbuild.Tests.Utilities$(PathSeparator)Msbuild.Tests.Utilities.csproj;
-                                      test$(PathSeparator)Performance$(PathSeparator)Performance.csproj" />
+                             Include="test/ArgumentsReflector/ArgumentsReflector.csproj;
+                                      test/Microsoft.DotNet.Tools.Tests.Utilities/Microsoft.DotNet.Tools.Tests.Utilities.csproj;
+                                      test/Msbuild.Tests.Utilities/Msbuild.Tests.Utilities.csproj;
+                                      test/Performance/Performance.csproj" />
 
       <PreTestProjectsToExclude Condition=" 'Executed after primary test phase.' != ' Consider moving.' "
-                             Include="test$(PathSeparator)Installer$(PathSeparator)Microsoft.DotNet.Cli.Msi.Tests$(PathSeparator)Microsoft.DotNet.Cli.Msi.Tests.csproj;" />
+                             Include="test/Installer/Microsoft.DotNet.Cli.Msi.Tests/Microsoft.DotNet.Cli.Msi.Tests.csproj;" />
```
piotrpMSFT: This is not needed anymore?? GREAT!

build/Test.targets
jgoshi: Ok will do.

build/Test.targets
jgoshi: Ok

PR #5777, opened 2017-02-20
Consume local info support from VersionTools

ellismg:
Davis added support for VersionTools to be able to pull from a local
filesystem layout that looks like what we have in GitHub on
dotnet/versions.

Start to consume this so a composed build can use this logic to update
dependencies using local version information produced during the build.

### DIFF
--- a/build_projects/update-dependencies/Config.cs
+++ b/build_projects/update-dependencies/Config.cs
@@ -10,19 +10,20 @@ namespace Microsoft.DotNet.Scripts
     /// </summary>
     /// <remarks>
     /// The following Environment Variables are required by this script:
-    /// 
+    ///
     /// GITHUB_USER - The user to commit the changes as.
     /// GITHUB_EMAIL - The user's email to commit the changes as.
     /// GITHUB_PASSWORD - The password/personal access token of the GitHub user.
     ///
     /// The following Environment Variables can optionally be specified:
-    /// 
-    /// ROSLYN_VERSION_URL - The Url to get the current Roslyn version. (ex. "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/roslyn/netcore1.0")
-    /// CORESETUP_VERSION_URL - The Url to get the current dotnet/core-setup package versions. (ex. "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/core-setup/master")
+    ///
+    /// DOTNET_VERSION_URL - The Url to the root of the version information (this is combined with the fragments bellow) (ex. "https://raw.githubusercontent.com/dotnet/versions/master/build-info")
+    /// ROSLYN_VERSION_FRAGMENT - The fragment to combine with DOTNET_VERSION_URL to get the current dotnet/roslyn package versions. (ex. "dotnet/roslyn/netcore1.0")
+    /// CORESETUP_VERSION_FRAGMENT - The fragment to combine with DOTNET_VERSION_URL to get the current dotnet/core-setup package versions. (ex. "dotnet/core-setup/master")
     /// GITHUB_ORIGIN_OWNER - The owner of the GitHub fork to push the commit and create the PR from. (ex. "dotnet-bot")
     /// GITHUB_UPSTREAM_OWNER - The owner of the GitHub base repo to create the PR to. (ex. "dotnet")
     /// GITHUB_PROJECT - The repo name under the ORIGIN and UPSTREAM owners. (ex. "cli")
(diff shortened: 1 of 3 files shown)


build_projects/update-dependencies/Program.cs
```diff
@@ -60,6 +60,20 @@ public static void Main(string[] args)
             }
         }
 
+        private static BuildInfo GetBuildInfo(string name, string buildInfoFragment, bool fetchLatestReleaseFile = true)
+        {
+            const string FileUrlProtocol = "file://";
+
+            if (s_config.DotNetVersionUrl.StartsWith(FileUrlProtocol, StringComparison.Ordinal))
+            {
+                return BuildInfo.LocalFileGetAsync(name, s_config.DotNetVersionUrl.Substring(FileUrlProtocol.Length), buildInfoFragment.Replace('/', Path.DirectorySeparatorChar), fetchLatestReleaseFile).Result;
```
dagood: [nit] Very long line.

build_projects/update-dependencies/Program.cs
```diff
@@ -60,6 +60,20 @@ public static void Main(string[] args)
             }
         }
 
+        private static BuildInfo GetBuildInfo(string name, string buildInfoFragment, bool fetchLatestReleaseFile = true)
+        {
+            const string FileUrlProtocol = "file://";
```
dagood: Huh, if you're able to get a valid `file://` URI, doesn't that work with `BuildInfo.Get`?

build_projects/update-dependencies/Config.cs
```diff
@@ -32,25 +32,26 @@ public class Config
         private Lazy<string> _userName = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_USER"));
         private Lazy<string> _email = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_EMAIL"));
         private Lazy<string> _password = new Lazy<string>(() => GetEnvironmentVariable("GITHUB_PASSWORD"));
-        
-        private Lazy<string> _roslynVersionUrl = new Lazy<string>(() => GetEnvironmentVariable("ROSLYN_VERSION_URL", "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/roslyn/netcore1.0"));
-        private Lazy<string> _coreSetupVersionUrl = new Lazy<string>(() => GetEnvironmentVariable("CORESETUP_VERSION_URL", "https://raw.githubusercontent.com/dotnet/versions/master/build-info/dotnet/core-setup/master"));
+
+        private Lazy<string> _dotNetVersionUrl = new Lazy<string>(() => GetEnvironmentVariable("DOTNET_VERSION_URL", "https://raw.githubusercontent.com/dotnet/versions/master/build-info"));
```
eerhardt: Can you update the class comments for these changes?

build_projects/update-dependencies/Program.cs
```diff
@@ -60,6 +60,20 @@ public static void Main(string[] args)
             }
         }
 
+        private static BuildInfo GetBuildInfo(string name, string buildInfoFragment, bool fetchLatestReleaseFile = true)
```
eerhardt: It seems odd that `fetchLatestReleaseFile` is defaulted to 'true', but it is only called in 2 places, both passing in 'false'.

eerhardt:
:shipit:

build_projects/update-dependencies/Program.cs
ellismg: `BuildInfo.Get` calls into HttpClient, which (at least on *nix) does not accept the `file://` protocol (just `http://` and `https://`) so we do this here.

build_projects/update-dependencies/Program.cs
ellismg: I wanted to retain the signature of Version Tools's BuildInfo.Get.  Logically, this behaves like that method except that it just also handles the case where `DOTNET_VERSION_URL` has a `file://` protocol.  That's why I picked the default that I did.

eerhardt: Dumb question:  Why aren't we updating Version Tool's `BuildInfo.Get` to handle `file://`?

dagood: Moving this logic there makes sense to me. I didn't know we were adopting this convention at the time, and I didn't want to get in the business of determining whether a string is a URL or a path--but the `file://` scheme makes it easy.

ellismg: I can make that change, Davis.

PR #5781, opened 2017-02-21
Refactor properties from LzmaCompile.targets

jgoshi:


build/compile/LzmaArchive.targets
```diff
@@ -115,11 +115,7 @@
       </GenerateNuGetPackagesArchiveVersion>
       <PropertyGroup>
         <NugetPackagesArchiveName>nuGetPackagesArchive.$(NuGetPackagesArchiveVersion).lzma</NugetPackagesArchiveName>
```
jgoshi: I couldn't remove the rest of these because they depend on the GenerateNuGetPackagesArchiveVersion task. The task looks complicated and I wasn't sure if I could compute that value without a task.

piotrpMSFT: Yea, looks like the math should stay in C#. Can we move that task to our pre-build, and generate the props there?

build/BuildOutputs.props
```diff
@@ -3,5 +3,7 @@
     <DotnetHostBaseName>dotnet$(ExeExtension)</DotnetHostBaseName>
     <DotnetHostFxrBaseName>$(DynamicLibPrefix)hostfxr$(DynamicLibExtension)</DotnetHostFxrBaseName>
     <HostPolicyBaseName>$(DynamicLibPrefix)hostpolicy$(DynamicLibExtension)</HostPolicyBaseName>
+    <ArchiverDll>$(ToolsOutputDirectory)/Archiver.dll</ArchiverDll>
```
piotrpMSFT: This is not really a Build Output. It's an intermediate file that we use to produce the real build output. I'd keep this in the LZMA csproj as a static [outside of target] property.

build/compile/LzmaArchive.targets
jgoshi: Unfortunately, the GenerateNuGetPackagesArchiveVersion task takes Stage2Directory as input. And Stage2Directory isn't computed in the pre-build. Unless I move that and what it depends on as well? It depends on $(RepoRoot)

jgoshi: Actually I'm unable to move this to the pre-build since it depends on using the dotnet.exe that exists in the stage2 directory. This means we have to build first.

eerhardt: Yes, just leave these alone.  There is no value in trying to refactor this.  The `$(NuGetPackagesArchiveVersion)` isn't a static property - it changes depending on which `<PackageReference>` items are in the `dotnet new` templates.  So yes, you'd have to build the product first, before you could generate this `$(NuGetPackagesArchiveVersion)`.

PR #5791, opened 2017-02-22
RootProject property

piotrpMSFT:
Defines a property for the root project of the CLI. This is the project that should be published when CLI is being published. @jonsequitur thoughts on naming?

/cc @livarcocc @jgoshi @krwq @eerhardt


piotrpMSFT:

PR #5797, opened 2017-02-22
Add unit tests around command line parsing

krwq:
dotnet run and dotnet test will be tested separately since they will likely require more work

cc: @jonsequitur @livarcocc @piotroko @piotrpMSFT @jgoshi 

also removed few unused files i noticed


src/dotnet/commands/dotnet-pack/PackCommand.cs
```diff
@@ -109,10 +119,43 @@ public static int Run(string[] args)
...
+            return cmd.Execute();
+        }
+
+        public ProcessStartInfo GetProcessStartInfo()
+        {
+            return _forwardingApp.GetProcessStartInfo();
+        }
+
+        public int Execute()
+        {
```
jonsequitur: This method is repeated in a bunch of places and doesn't rely on internals, so it can be deduplicated by making it an extension method.

src/dotnet/commands/dotnet-cache/Program.cs
```diff
@@ -71,10 +74,43 @@ public static int Run(string[] args)
                 cache.ExtraMSBuildArguments = app.RemainingArguments;
                 cache.ProjectArgument = projectArgument.Value();
                
-                return cache.Execute();
+                return 0;
             });
 
-            return app.Execute(args);
+            int exitCode = app.Execute(args);
```
krwq: I think we would need some kind of static virtual methods to put FromArgs in a base class - I think this would be to hacky to do. @jonsequitur @piotrpMSFT - any ideas how to implement that without adding much complexity? T4 template which would generate this for each instance perhaps?

krwq: I think it would be better to solve this on different occasion - this PR already has grew enough and it is not clear to me yet how to do it without adding more code than it would reduce

PR #5801, opened 2017-02-22
Refactor properties from files in the package subdir

jgoshi:


build/OutputDirectories.props
```diff
@@ -25,5 +25,15 @@
       <TestPackagesDir>$(TestOutputDir)/packages/</TestPackagesDir>
       <TestArtifactsDir>$(TestOutputDir)/artifacts/</TestArtifactsDir>
       <TestResultXmlDir>$(TestOutputDir)/results/</TestResultXmlDir>
+      <ArchiveOutputDirectory>$(PackagesDirectory)</ArchiveOutputDirectory>
```
eerhardt: I'm starting to see this work as being degrading the cohesiveness of our build code.  If I have a property that is only used in one .targets file, pulling that property into a global .props file is spreading out the code and making it harder to reason about our build logic.

build/Version.props
```diff
@@ -17,6 +17,9 @@
     <NugetVersion Condition=" '$(NugetVersion)' == '' ">$(VersionMajor).$(VersionMinor).$(VersionPatch)</NugetVersion> 
 
     <SdkVersion>$(NugetVersion)</SdkVersion>
-    <SdkNugetVersion>$(NugetVersion)</SdkNugetVersion> 
+    <SdkDebianPackageVersion>$(SdkVersion)</SdkDebianPackageVersion>
+    <SdkNugetVersion>$(NugetVersion)</SdkNugetVersion>
+
+    <DotnetDebToolVersion Condition=" '$(DotnetDebToolVersion)' == '' ">1.0.0-*</DotnetDebToolVersion>
```
eerhardt: This property doesn't belong here.  IMO it really belongs in the DEB specific file.  But if we feel all properties need to be removed from .targets files, then this better belongs in `DependencyVersions.props`

build/InputDirectories.props
```diff
@@ -4,6 +4,7 @@
     <DotnetCliBuildDirectory>$(RepoRoot)/build_projects/dotnet-cli-build</DotnetCliBuildDirectory>
     <MSBuildImportsDir>$(RepoRoot)/resources/MSBuildImports</MSBuildImportsDir>
     <SrcDirectory>$(RepoRoot)/src</SrcDirectory>
+    <ProjectsSrcDirectory>$(RepoRoot)/src</ProjectsSrcDirectory>
```
eerhardt: This is a duplicate property of the one directly above it.  Why do we need 2 duplicate global properties?  It makes it hard to figure out which one I should use.

build/OutputDirectories.props
jgoshi: That's a good point.

jgoshi: Are you ok with letting it go for this PR (since it's a pretty big one). Once it goes through I'll take a pass at cleaning up our global props files. Basically if we have duplicate properties I'll remove them, and if a property is used in only one file I'll move it back into the file.

jgoshi: Nevermind, I'm getting a test failure. I think I'll do the clean up now (and close this PR).

jgoshi: I checked with @piotrpMSFT about this. First we'll go ahead and finish the work to get properties out of targets. Then take a second pass to clean up stuff.

eerhardt: > First we'll go ahead and finish the work to get properties out of targets

When you say "targets", do you mean .targets files or <Target> nodes?  I'm fine with moving them out of <Target> nodes as a rule/guideline.  I'm not fine with moving them out of .targets files as a rule/guidline.  They should only be moved into .props files if they are shared between multiple files.

build/InputDirectories.props
jgoshi: Removing it

build/Version.props
jgoshi: Moving it back to deb.targets file

build/OutputDirectories.props
jgoshi: I mean moving them out of &lt;Target&gt; (not necessarily the .targets file). Moving forward I want to do what you suggested, only move it to a global file if it's used in multiple places. But I'm sure I missed some of those (with previous PRs that have already merged). So I'll also take a separate pass to clean that up.

jgoshi:
Going to redo this PR

PR #5804, opened 2017-02-23
Refactor publish.target

jgoshi:


jgoshi:

build/InstallerInfo.props
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <PropertyGroup>
+    <SdkDebianPackageVersion>$(SdkVersion)</SdkDebianPackageVersion>
```
piotrpMSFT: It seems like the Debian properties should probably go together, perhaps in a PG in the debian installer targets file?

build/Publish.targets
```diff
@@ -5,6 +5,25 @@
   <Import Project="$(MSBuildThisFileDirectory)/publish/Badge.targets" />
   <Import Project="$(MSBuildThisFileDirectory)/publish/Checksum.targets" />
 
+  <ItemGroup>
+    <SdkInstallerFileItemGroup Include="$(SdkInstallerFile)" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <SdkDebianUploadUrl>$(DotnetBlobRootUrl)/$(Product)/$(FullNugetVersion)/%(SdkInstallerFileItemGroup.Filename)%(SdkInstallerFileItemGroup.Extension)</SdkDebianUploadUrl>
```
piotrpMSFT: Yea, these seem related to the ones above. I wonder if publishing to the debian repo isn't a separate concern. Perhaps this target should move to its own targets file with the associated properties?

PR #5814, opened 2017-02-23
Remove corehost from CLI layout

piotrpMSFT:
@davidfowl @anurse @pakrym this PR removes `corehost` from the 2.0 CLI layout. There was some intent early in CLI 1.0 to use this file for back-compat scenarios, but to the best of my knowledge no one takes advantage of it and it only creates bloat in the redist. Any concerns with removing in 2.0?

/cc @livarcocc @jonsequitur @jgoshi @krwq @blackdwarf


piotrpMSFT:
Thanks for taking a look, @anurse!

PR #5822, opened 2017-02-23
Don't look for Debian package dependencies when skipping installers

ellismg:
Today, we always check to see if DebianPackageBuildDependencies are
present on the machine doing a build, because we will need them to
build the .deb files later on.

Since we now allow skipping of building packages, don't check for
these dependencies if we weren't going to use them.

### DIFF
--- a/build/prepare/CheckPrereqs.targets
+++ b/build/prepare/CheckPrereqs.targets
@@ -17,7 +17,7 @@
 
     <Exec Command="dpkg -s %(UbuntuCoreclrAndCoreFxDependencies.Identity)" />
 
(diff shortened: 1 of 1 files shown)


ellismg:
/cc @eerhardt @livarcocc 

I ran into this when building the CLI as part of a composed source build on a Ubuntu 14.04 machine that didn't have devscripts installed on it.  Since I already told the build to skip building the packages, we should also skip the prereq checks.

eerhardt:
Another option would have been to pass `--skip-prereqs` into build.sh.

PR #5828, opened 2017-02-24
First pass at cleaning up the global properties files

jgoshi:
If the global properties file had a clear purpose (like FileExtensions.props) I left it. Otherwise if a global property is only used in one file I moved it back to that file. This moves in the direction of a global props file only containing properties that are shared.


piotrpMSFT:
I'll leave this open so others can take a look. My only concern was that `<channel>` is now pretty burried. We may want to extract out a `<branchinfo>` concept to ease the creation of new branches. However, this will impact branding and versioning as well so we can leave it to another change.

eerhardt:
> We may want to extract out a <branchinfo> concept to ease the creation of new branches.

Agreed. We had BranchInfo.props now. Should we just leave channel and branch name in that file?

jgoshi:
I'll put BranchInfo.props back

jgoshi:

PR #5832, opened 2017-02-24
Remove workaround for multi-proc builds

eerhardt:
The SDK bug with building a .sln file is now fixed.  We should remove our workaround so we test that building a .sln file works as expected, with no multi-proc issues.

This is a test infrastructure only change.  It may require the latest SDK change to be merged first, which is #5813.

@livarcocc @piotrpMSFT


eerhardt:
@MattGertz for approval to merge.  This is a test infrastructure only change that is necessary to test our end product works correctly with building .sln files.

PR #5833, opened 2017-02-24
Fix debian_config.json on ubuntu16.10

eerhardt:
When we only have a single shared framework, the python json parser throws an error since there is a hanging comma on the last property.

Fixing this by removing the comma when there is only a single Shared Framework dependency.

@piotrpMSFT @livarcocc @jgoshi


eerhardt:
Merging to unblock the official 1.0.1 build.

PR #5840, opened 2017-02-24
Bump Roslyn to 2.0.0-rc5-61424-02

piotrpMSFT:
/cc @livarcocc @jaredpar


piotrpMSFT:
FYI, @jasonmalinowski will be confirming this version number at the end of the day. We will be making it slightly higher, but do not yet know which # we want to use.

jasonmalinowski:
Unless anything unexpected happens, the magic number for Roslyn is going to be 2.0.0-rc5-61427-04. The packages for that build are not on MyGet but should be in about two hours.

piotrpMSFT:
Updated to 2.0.0-rc5-61427-04 per @jasonmalinowski

jasonmalinowski:
The package should now exist on MyGet.

piotrpMSFT:
@jasonmalinowski yep. Some legs are already green :)

PR #5860, opened 2017-02-27
Remove duplicate installer suffix

piotrpMSFT:
/cc @livarcocc @blackdwarf


blackdwarf:
Is there a way we can get a PKG with this change to review the text?

blackdwarf:
Looks good on the PKG as well. :shipit:

PR #5872, opened 2017-02-28
Add CI support for the Linux Portable build

jgoshi:


jgoshi:
test ci please

netci.groovy
```diff
@@ -34,6 +34,10 @@ platformList.each { platform ->
     else if (os == 'Ubuntu') {
         buildCommand = "./build.sh --skip-prereqs --configuration ${configuration} --docker ubuntu.14.04 --targets Default"
     }
+	else if (os == 'Linux') {
+		os = 'Ubuntu16.04';
```
eerhardt: I don't think this is correct.  Core-setup uses RHEL7.2 according to this comment:

https://github.com/dotnet/core-setup/blob/master/netci.groovy#L66-L69

jgoshi:
test Linux x64 Release Build

netci.groovy
jgoshi: But I thought the whole point of this work is that it should run on any distro. So as a first test we were going to try it on Ubuntu 16.04

ellismg: The reason that core-setup uses RHEL 7.2 is that it needs to build native code and we want to do that on a machine that has the oldest version of GLIBC we support (and in our cases that's on RHEL 7).  In CLI's case, you aren't building any native code, so you should be able to do your managed linux build anywhere.

eerhardt: I think for our official build and for Jenkins we should match core-setup unless we have a valid reason that requires us to use a different distro.

piotrpMSFT: @Petermarcu and I agreed to use Ubuntu 16.04 because it is a popular distro for CLI and because it now doubles our distro coverage across corefx and cli.

PR #5877, opened 2017-03-01
Clean up chmod usage

piotrpMSFT:
This PR:
- Moves CLI chmod'ing to be a post-publish step of redist.csproj
- Removes complexity from CLI chmod'ing. We at this point have no native components in the CLI, and we no longer have any remaining executables in the `Sdk` directory. The command, therefore, puts all folders at 755 [rwxrw-rw-] for folders, 644 [rw-r--r--] for files.

I've also mostly eliminated the usage of the chmod task as it was quite slow at dealing with large numbers of files [23s last I looked].

/cc @livarcocc @jgoshi @jonsequitur @eerhardt

### DIFF
--- a/build_projects/dotnet-cli-build/Chmod.cs
+++ b/build_projects/dotnet-cli-build/Chmod.cs
@@ -9,26 +9,12 @@ namespace Microsoft.DotNet.Cli.Build
     public class Chmod : ToolTask
     {
         [Required]
-        public string File { get; set; }
+        public string Glob { get; set; }
 
         [Required]
         public string Mode { get; set; }
 
         public bool Recursive { get; set; }
-
-        protected override bool ValidateParameters()
-        {
-            base.ValidateParameters();
-
-            if (!System.IO.File.Exists(File))
-            {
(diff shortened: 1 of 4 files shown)


piotrpMSFT:
Added back +x for RunCsc.sh. There is still a little complexity after all :)

build_projects/dotnet-cli-build/Chmod.cs
```diff
@@ -47,12 +33,12 @@ protected override string GenerateFullPathToTool()
 
         protected override string GenerateCommandLineCommands()
         {
-            return $"{GetRecursive()} {GetMode()} {GetFilePath()}";
+            return $"{GetRecursive()} {GetMode()} {GetGlob()}";
         }
 
-        private string GetFilePath()
+        private string GetGlob()
```
jonsequitur: Nitpick: Why the extra indirection, as opposed to calling `Glob` in `GenerateCommandLineCommands`?

piotrpMSFT:
@jonsequitur we follow this pattern across all tasks. If we want to change the pattern then let's use a separate issue.

PR #5884, opened 2017-03-01
Remove OpenSUSE13 and Fedora23 since they are end of lifed.

eerhardt:
This is necessary in order to continue getting new builds from the runtime.  See #5717  .NET Core runtime has already dropped OpenSUSE.13, and will soon drop Fedora23.

@livarcocc @jgoshi @piotrpMSFT


eerhardt:
FYI - The Fedora23 and OpenSUSE13 builds are expected to fail, since they are being removed.

PR #5889, opened 2017-03-02
Update the install script to support osname

jgoshi:
@piotrpMSFT @eerhardt @livarcocc @jonsequitur @krwq


piotrpMSFT:
Can you show an example of how this new option will be used?

jgoshi:
Passing in --osname linux allows us to download the stage0 version that has the Linux shared framework. I manually uploaded one to our blob storage to bootstrap this scenario.

But actually thinking about this some more, @eerhardt do I really need to do this? I could just build on whatever OS I happen to be on, but produce an sdk that packages the Linux shared framework right? Do does the dotnet we used to build ourselves (stage0 dotnet) need to use the Linux binaries?

piotrpMSFT:
@jgoshi is this needed, then?

eerhardt:
I think we need to change the install script for the following reasons:

1. Yes, we _could_ use an Ubuntu stage0 to build our `linux-x64` build, but then we would be getting less test coverage on `linux-x64`.  It would be one less test (building ourselves) that would be running on `linux-x64`.
1. When customers want to install the `linux-x64` tarball, how will they be able to do it? We need to change the install script somehow for them to use it to install `linux-x64`.

One thought is instead of having `--osname blah`, what do you think about having a boolean parameter named `--linux-portable`?  This would be similar to our current `--shared-runtime` parameter to just install the shared framework.

jgoshi:
I like that idea. Will change to --linux-portable

jgoshi:
@piotrpMSFT @eerhardt If this looks good I'd like to merge it. Then I can work on using the Linux portable to build our stuff as Eric suggested.

scripts/obtain/dotnet-install.sh
```diff
@@ -648,6 +655,8 @@ do
             echo "  --no-path, -NoPath             Do not set PATH for the current process."
             echo "  --verbose,-Verbose             Display diagnostics information."
             echo "  --azure-feed,-AzureFeed        Azure feed location. Defaults to $azure_feed"
+            echo "  --linux-portable               Installs the Linux portable shared runtime instead of the shared runtime for OS."
```
eerhardt: I wouldn't say "shared runtime" here.  It is confusing when I am trying to install the SDK.

scripts/obtain/dotnet-install.sh
```diff
@@ -648,6 +655,8 @@ do
...
+			echo "      -LinuxPortable"
```
eerhardt: I'm not sure I understand why this line is here.  Do you know?

jgoshi: Which line? -LinuxPortable? I was just copying what was there for -SharedRuntime. I think it's supposed to show it can also be used as the command-line argument. But I'll fix it to appear on the same line like -AzureFeed.

scripts/obtain/dotnet-install.sh
jgoshi: What should I call it?

jgoshi: Followed up with Eric offline. Will use this text:
Installs the Linux portable .NET Tools instead of a distro-specific version.

scripts/obtain/dotnet-install.sh
jgoshi: Followed up with Eric offline. This is fine since we are supporting multiple param names. This follows the convention with the other params.

PR #5915, opened 2017-03-03
Remove BuildPipeline definitions.

ravimeda:
BuildPipeline definitions need not be checked into the repository.  CLI uses and maintains VSTS build definitions only, and does not depend on BuildPipeline.  Hence maintaining a copy of definitions for BuildPipline in this repository is not required.

@piotrpMSFT @MichaelSimons please review.

Thanks,
Ravi


ravimeda:
Thanks @MichaelSimons @livarcocc .

PR #5935, opened 2017-03-06
Replace CommandLineApplication with new command line parser, enable tab completion

jonsequitur:
This PR is for review purposes only at this stage. Tab completion is fully functional here, though the test coverage is still very slim. I'm mainly interested in getting feedback on the naming and coding patterns.

If you want to try it out, build the CLI and then run one of the new scripts (`register-completions.*`), depending on which shell you're using) in your shell. Tab completion will be enabled for the duration of your shell session.

@piotrpMSFT @livarcocc @eerhardt @jgoshi @blackdwarf


NuGet.Config
```diff
@@ -9,5 +9,7 @@
     <add key="roslyn" value="https://dotnet.myget.org/f/roslyn/api/v3/index.json" />
     <add key="xunit" value="https://www.myget.org/F/xunit/api/v3/index.json" />
     <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
+    
+    <add key="dev" value="C:\dev\github\CliCommandLineParser\CommandLine\bin\Debug\" />
```
jonsequitur: I'll back this out.

scripts/register-completions.bash
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Parameter completion for the dotnet CLI
+
+_dotnet_bash_complete()
+{
+  local word=${COMP_WORDS[COMP_CWORD]}
+  local dotnetPath=${COMP_WORDS[1]}
+
+  local completions=("$(./bin/Debug/netcoreapp1.0/osx.10.11-x64/publish/dotnet complete --position ${COMP_POINT} "${COMP_LINE}")")
```
jonsequitur: In the near term I'll replace these paths with an unqualified path which will allow this to work as long as dotnet is on your path. In the future, the script will provide the path to `dotnet` as a parameter so that multiple side-by-side installs will work correctly in the event that the available completions differ among them.

ErikSchierboom:
This would be a fantastic improvement!

src/dotnet/commands/dotnet-build/BuildCommandParser.cs
```diff
@@ -0,0 +1,39 @@
...
+                           Create.Option("-o|--output",
+                                         "Output directory in which to place built artifacts.",
+                                         Accept.ExactlyOneArgument
+                                               .With(name: "OUTPUT_DIR")),
+                           Create.Option("-f|--framework",
+                                         "Target framework to build for. The target framework has to be specified in the project file.",
+                                         Accept.AnyOneOf(Suggest.TargetFrameworksFromProjectFile)),
+                           Create.Option("-r|--runtime",
+                                         "Target runtime to build for. The default is to build a portable application.",
+                                         Accept.AnyOneOf(Suggest.RunTimesFromProjectFile)),
```
eerhardt: FYI - we should be able to accept any RID, not just the ones in the .csproj.  Does this support values that aren't in the .csproj?

src/dotnet/commands/dotnet-add/AddCommandParser.cs
```diff
@@ -0,0 +1,74 @@
...
+        public static IEnumerable<string> QueryNuGet(string match)
+        {
+            var httpClient = new HttpClient();
+
+            string result;
+
+            try
+            {
+                var cancellation = new CancellationTokenSource(TimeSpan.FromSeconds(10));
+                var response = httpClient.GetAsync($"https://api-v2v3search-0.nuget.org/query?q={match}&skip=0&take=100&prerelease=true", cancellation.Token)
```
eerhardt: This should use the NuGet.config and not hard code to nuget.org.

jonsequitur: That's the plan. This was a minimal implementation to prove out the design and let people start playing with it.

@srivatsn, would we potentially be able to leverage the work that your team did recently to supply NuGet completions for VS?

src/dotnet/commands/dotnet-build/BuildCommandParser.cs
jonsequitur: No but we can change it easily. There are two methods (with overloads) you'll see here:

* `Accept.AnyOneOf` will result in a parse error if the provided value is not in the set, and
* `Accept.WithSuggestionsFrom` will provide completions from the provided values, but other input is also allowed.

jonsequitur:

piotrpMSFT:
@dotnet-bot test this please.

test/dotnet-msbuild.Tests/GivenDotnetCacheInvocation.cs
```diff
@@ -46,7 +47,7 @@ public void MsbuildInvocationIsCorrect(string[] args, string expectedAdditionalA
         [InlineData("--output")]
         public void ItAddsOutputPathToMsBuildInvocation(string optionName)
         {
-            string path = "/some/path";
+            string path = Directory.GetCurrentDirectory();
             var args = ArgsPrefix.Concat(new string[] { optionName, path }).ToArray();
```
piotrpMSFT: Why is this necessary? are we confident it's not breaking any scenarios?

test/dotnet-add-reference.Tests/GivenDotnetAddReference.cs
```diff
@@ -114,8 +112,8 @@ public void WhenTooManyArgumentsArePassedItPrintsError()
                     .WithProject("one two three")
                     .Execute("proj.csproj");
             cmd.ExitCode.Should().NotBe(0);
-            cmd.StdErr.Should().Be("Unrecognized command or argument 'two'");
-            cmd.StdOut.Should().Be("Specify --help for a list of available options and commands.");
```
piotrpMSFT: Do we not need this `--help` text anymore?

jonsequitur: Prior to this PR, we're inconsistent. In some cases when there's a parse error, we tell the user to request help. In others, we immediately show help. I've converged on the latter behavior but verifying it here is redundant with a number of other tests, and doing a text comparison to the complete help text is fairly high maintenance, so I opted to go to a single assertion here.

test/dotnet-msbuild.Tests/GivenDotnetCacheInvocation.cs
jonsequitur: The parser configuration had a default set to the current directory at one point, which resulted in a failure on Windows. This is no longer an issue and this change can be reverted, but it seems to me like we should either be testing explicitly that both Windows and *nix paths are handled correctly, or we should at least be writing tests in a way that exercises the variations in behavior we expect to see. 

Thoughts?

jonsequitur:

jonsequitur:

jonsequitur:
@dotnet-bot test this please.

jonsequitur:
@dotnet-bot test this please.

eerhardt:
@jonsequitur - just an FYI, https://github.com/dotnet/cli/pull/6090 got merged, which will conflict (hopefully not much) with your change.

jonsequitur:
@eerhardt I'm merging it now. Thanks for the heads-up.

src/dotnet/commands/dotnet-run/RunCommandParser.cs
```diff
@@ -0,0 +1,41 @@
...
+                    CommonOptions.ConfigurationOption(),
+                    CommonOptions.FrameworkOption(),
+                    Create.Option(
+                        "-p|--project",
+                        LocalizableStrings.CommandOptionProjectDescription,
+                        Accept.ExactlyOneArgument()),
+                    Create.Option(
+                        "--no-build",
+                        LocalizableStrings.CommandOptionNoBuildDescription,
+                        Accept.NoArguments().ForwardAs("/p:NoBuild=true"))
```
eerhardt: I'm not sure `/p:NoBuild` is a real property in MSBuild, is it?

jonsequitur: This looks like a translation mistake from the previous parser code. Removing it.

PR #5938, opened 2017-03-07
Add support to build linux portable

jgoshi:


run-build.sh
```diff
@@ -93,6 +96,14 @@ while [[ $# > 0 ]]; do
             args=( "${args[@]/$2}" )
             shift
             ;;
+        --linux-portable)
+            # Pass in a specific version (the one we set up to bootstrap this experience). Once 
+            # we have an official build we can remove the version parameter and use the latest.
+            LINUX_PORTABLE_INSTALL_ARGS="--linux-portable --version 2.0.0-alpha-005151"
```
eerhardt: IMO, it is better to just pin the stage0 to a specific build for all platforms (windows and unix).  That way our build is consistent across everything.

See https://github.com/dotnet/cli/pull/5839 (the non-README changes) for what unpinning stage0 looks like.

run-build.sh
```diff
@@ -93,6 +96,14 @@ while [[ $# > 0 ]]; do
...
+            # Until we get test support for 2.0 we need to pass in the targets without test.
+            LINUX_PORTABLE_BUILD_ARGS="/p:Rid=\"linux-x64\" /p:OSName=\"linux\" /p:IncludeAdditionalSharedFrameworks=false /p:CLITargets=\"Prepare;Compile;Package;Publish\""
+            args=( "${args[@]/$1}" )
+            ;;
         --help)
```
eerhardt: Should probably update the help for `--linux-portable`.

run-build.sh
```diff
@@ -93,6 +96,14 @@ while [[ $# > 0 ]]; do
             args=( "${args[@]/$2}" )
             shift
             ;;
+        --linux-portable)
+            # Pass in a specific version (the one we set up to bootstrap this experience). Once 
+            # we have an official build we can remove the version parameter and use the latest.
+            LINUX_PORTABLE_INSTALL_ARGS="--linux-portable --version 2.0.0-alpha-005151"
+            # Until we get test support for 2.0 we need to pass in the targets without test.
+            LINUX_PORTABLE_BUILD_ARGS="/p:Rid=\"linux-x64\" /p:OSName=\"linux\" /p:IncludeAdditionalSharedFrameworks=false /p:CLITargets=\"Prepare;Compile;Package;Publish\""
```
eerhardt: I don't think `p:IncludeAdditionalSharedFrameworks` should be required to be passed in on the command line.  It should be calculated from the RID, same as what I did here:

https://github.com/dotnet/cli/commit/916a19c4936d92b92b5fbef2b801dba9c125d0db#diff-d595ac18a42f088f16eefcd37b417b04R136

eerhardt: Can we rename `LINUX_PORTABLE_BUILD_ARGS` to be something more general?  That way if other things need to specify build args, they can use the same variable?

run-build.sh
jgoshi: Ok will do.

run-build.sh
jgoshi: Will do.

run-build.sh
jgoshi: Will do. How about CUSTOM_BUILD_ARGS

eerhardt: sounds like a good name to me

jgoshi:

jgoshi:

jgoshi:

run-build.sh
```diff
@@ -102,6 +111,7 @@ while [[ $# > 0 ]]; do
             echo "  --nopackage                         Skip packaging targets"
             echo "  --nobuild                           Skip building, showing the command that would be used to build"
             echo "  --docker <IMAGENAME>                Build in Docker using the Dockerfile located in scripts/docker/IMAGENAME"
+            echo "  --linux-portable                    Builds with the Linux portable .NET Tools instead of a distro-specific version."
```
eerhardt: Shouldn't say "Builds with".  Instead "Builds the Linux portable .NET Tools..."

jgoshi:

jgoshi:

PR #5972, opened 2017-03-09
Removed the square brackets from the dotnet-test `--logger` parameter help

droyad:
The command line documentation `dotnet help test` currently reads:
```
 -l|--logger <LoggerUri/FriendlyName>  Specify a logger for test results.
                                        Example: --logger "trx[;LogFileName=<Defaults to unique file name>]"
```

However the example does not work. It should read `--logger "trx;LogFileName=<Defaults to unique file name>"`, ie without the `[]`. This correct usage can be seen in this [test](https://github.com/dotnet/cli/blob/70c65160f6163666dbe5f052b2331ec59582fce5/test/dotnet-test.Tests/GivenDotnetTestBuildsAndRunsTestfromCsproj.cs#L180)

### DIFF
--- a/src/dotnet/commands/dotnet-test/xlf/LocalizableStrings.cs.xlf
+++ b/src/dotnet/commands/dotnet-test/xlf/LocalizableStrings.cs.xlf
@@ -73,9 +73,9 @@
       </trans-unit>
       <trans-unit id="CmdLoggerDescription">
         <source>Specify a logger for test results. 
-                                        Example: --logger "trx[;LogFileName=&lt;Defaults to unique file name&gt;]"</source>
+                                        Example: --logger "trx;LogFileName=&lt;Defaults to unique file name&gt;"</source>
         <target state="translated">Zadejte protokolovací nástroj pro výsledky testů. 
-                                        Příklad: --logger "trx[;LogFileName=&lt;Standardně se nastaví jedinečný název souboru&gt;]"</target>
+                                        Příklad: --logger "trx;LogFileName=&lt;Standardně se nastaví jedinečný název souboru&gt;"</target>
         <note />
       </trans-unit>
       <trans-unit id="CmdConfiguration">
--- a/src/dotnet/commands/dotnet-test/xlf/LocalizableStrings.de.xlf
+++ b/src/dotnet/commands/dotnet-test/xlf/LocalizableStrings.de.xlf
@@ -73,9 +73,9 @@
       </trans-unit>
       <trans-unit id="CmdLoggerDescription">
         <source>Specify a logger for test results. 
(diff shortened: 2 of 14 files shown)


eerhardt:
@jgoshi - 2 things

1. We need to rename the "Linux portable" leg so we can disambiguate it from the real Ubuntu16.04 leg.
2. It looks like we need to turn off the '.trx' scanning for the "Linux portable" leg, until we get tests running.

```
01:10:54 [xUnit] [INFO] - [MSTest-Version N/A (default)] - No test report file(s) were found with the pattern '**/*.trx' relative to '/mnt/j/workspace/dotnet_cli/master/release_linux_x64_prtest' for the testing framework 'MSTest-Version N/A (default)'.  Did you enter a pattern relative to the correct directory?  Did you generate the result report(s) for 'MSTest-Version N/A (default)'?
01:10:54 
[xUnit] [ERROR] - No test reports found for the metric 'MSTest' with the resolved pattern '**/*.trx'. Configuration error?.
01:10:54 [xUnit] [INFO] - Failing BUILD.
```

eerhardt:
@droyad - I believe the intention of the square brackets was to signify that this part is optional.  You can just call `dotnet test --logger trx`.

This is consistent with other command line help:

```
git clone [--template=<template_directory>]
	  [-l] [-s] [--no-hardlinks] [-q] [-n] [--bare] [--mirror]
	  [-o <name>] [-b <name>] [-u <upload-pack>] [--reference <repository>]
	  [--dissociate] [--separate-git-dir <git dir>]
	  [--depth <depth>] [--[no-]single-branch]
	  [--recursive | --recurse-submodules] [--] <repository>
	  [<directory>]
```

But you are right, since this is an example, it should work to just copy and paste it.

Maybe we could put 2 examples?  One with just `--logger trx` and another with the log file?

@blackdwarf - thoughts?

/cc @codito @Faizan2304

droyad:
I'll look at adding both examples.

Faizan2304:
The intension of square bracket here is to signify that this part is optional.

droyad:
@Faizan2304 The problem is that it is in the example not the options specification.

Faizan2304:
@pvlakshn

wli3:
Hi @droyad, sorry for the long delay. Do you mind to change the file cli\src\dotnet\commands\dotnet-test\xlf\LocalizableStrings.cs.xlf first, and then run

```
build.cmd /p:UpdateXlfOnBuild=true
```

You will get all the loc file changed automatically, then 

```
git commit --all
```

I can also create a separate updated pull request and close this one if you like

droyad:
Replaced by #7406

PR #5982, opened 2017-03-09
Enable the tests for Linux portable

jgoshi:
@eerhardt @livarcocc 

The tests will fail until we get 2.0 support. But CI was failing anyway because it was looking for .trx files which are generated by running the tests. At least this way we can see we got all the way to tests and consider it good for now.


jgoshi:
Actually I'm going to do something else so it is green.

PR #6004, opened 2017-03-10
Remove unused constants.

eerhardt:
These strings were used in the project.json days, but are no longer used.

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/Constants.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/Constants.cs
@@ -12,40 +12,9 @@ namespace Microsoft.DotNet.Cli.Utils
 
         public static readonly string ProjectFileName = "project.json";
         public static readonly string ExeSuffix = CurrentPlatform == Platform.Windows ? ".exe" : string.Empty;
-        public static readonly string ConfigSuffix = ".config";
-
-        // Priority order of runnable suffixes to look for and run
-        public static readonly string[] RunnableSuffixes = CurrentPlatform == Platform.Windows
-                                                         ? new string[] { ".exe", ".cmd", ".bat" }
-                                                         : new string[] { string.Empty };
 
         public static readonly string BinDirectoryName = "bin";
         public static readonly string ObjDirectoryName = "obj";
-
-        public static readonly string DynamicLibSuffix = CurrentPlatform == Platform.Windows ? ".dll" :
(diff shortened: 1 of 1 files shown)


eerhardt:
@livarcocc - please someone disable this test `.Microsoft.DotNet.Tests.PackagedCommandTests.WhenToolAssetsFileIsInUseThenCLIRetriesLaunchingTheCommandForAtLeastOneSecond`

nguerrera:
@eerhardt This actually broke moving the SDK repos test code forward to latest Cli.Utils package.

eerhardt:
Ok.  So that test code can copy these strings into its own project.

I don't think this assembly should be the place to expose these strings.

nguerrera:
Yes, I did they. Still, this lib shipped RTM on nuget.org.

eerhardt:
And it changed major versions.

nguerrera:
Is that our policy? Ok if so but I thought I saw CR feedback about not breaking things in this lib elsewhere.

eerhardt:
Here's my 2 cents on the subject though:

This library was first created using "ASP.NET public API" policy.  (I'm not sure that is an official term, but that's how I think of it in my head.)  This means that pretty much everything was `public` or `private`.  There isn't very much usage of `internal`.  From my understanding, ASP.NET has no qualms over making public API changes across major version changes.  See https://github.com/aspnet/Announcements/issues?q=is%3Aissue+is%3Aopen+label%3A%22Breaking+change%22.

This "public API style" is vastly different than ".NET CoreFX public API" policy.  Every detail of a public API is scrubbed before it can be added as public, with the very certain understanding that you WILL forever support this API.  It is a much higher bar, with a much higher long term cost.

My opinion is that we shouldn't be taking the 2 conflicting parts of these policies and apply them to this library.  It is unrealistic to expect that APIs that were added with minimal API review should be held to the same support policy as APIs which can absolutely never change.

Going forward, if we want to start adopting the "CoreFX public API" policy to this library, then I think we should do an API scrub/review and ensure each and every API we expose is the correct long-term API that we want to support.  However, my opinion is that these APIs are not even close to the usage of CoreFX.  They are used very little, and should be even less going forward.  So I think keeping the "ASP.NET public API" policy for this library makes a lot of sense.

nguerrera:
That makes sense and I appreciate the added context because I always cringe at "major version changed" without more information. I think it is a case-by-case decision with tradeoffs. I think the semver-based policy is fine here based on the low usage.

PR #6083, opened 2017-03-20
Fix README.md table

kendaleiv:


eerhardt:
Thanks for the help, @kendaleiv. 

I don't think there are supposed to be 3 columns.  Can you instead make the change to remove the extra `:---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|` on the 2nd line?

kendaleiv:
Sure! See #6084.

PR #6097, opened 2017-03-22
Don't write `\b` when output is redirected

### LINKED ISSUE #6096: Microsoft.DotNet.Archive's ConsoleProgressReport is goofy when stdout isn't a tty.
## Steps to reproduce

Run some code which will cause Microsoft.DotNet.Archive to use it's console logger (e.g. do a build of the CLI) but redirect standard output to a file.

## Expected  behavior

The file doesn't contain literal backspace characters.

## Actual behavior

The progress reporter [writes](https://github.com/dotnet/cli/blob/cc80822c16b955a1f664f8a65ed2c76c3c52a824/src/Microsoft.DotNet.Archive/ConsoleProgressReport.cs#L32) a literal `\b` character into the output stream.

I don't know what the behavior on Windows is.  You can see this in the build logs in Jenkins.

It would be better to detect if the output is a tty and if not do something like write "..." between numbers (or just don't do any logging at all?)

I don't know if we have a good way to detect you are in this case without breaking down and pinovking to isatty or something in libc.

ellismg:
Fixes #6096

### DIFF
--- a/src/Microsoft.DotNet.Archive/ConsoleProgressReport.cs
+++ b/src/Microsoft.DotNet.Archive/ConsoleProgressReport.cs
@@ -29,9 +29,16 @@ namespace Microsoft.DotNet.Archive
                 string line = $"{value.Phase} {progress}%";
                 if (value.Phase == _currentPhase)
                 {
-                    Console.Write(new string('\b', _lastLineLength));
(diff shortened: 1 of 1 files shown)


ellismg:
test Debian8.2 x64 Debug Build please
test Windows_NT x64 Release Build please

PR #6115, opened 2017-03-22
Adaptive help spacing

jonsequitur:
The latest Microsoft.DotNet.Cli.CommandLine adds support for adaptive spacing in help between the columns for the options/commands lists and their descriptions. This code change pulls in that version and adjusts a number of test expectations accordingly.

There are also a few code cleanups included.


jonsequitur:
@livarcocc

PR #6208, opened 2017-03-30
Fixing the reference to dotnet-deb-tool to be correct.

eerhardt:
No more using * versions.  Only specify the version in a single place.

Our build is currently broken without this.

@livarcocc @nguerrera

### DIFF
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -23,5 +23,6 @@
   <!-- infrastructure and test only dependencies -->
   <PropertyGroup>
     <VersionToolsVersion>1.0.27-prerelease-01402-01</VersionToolsVersion>
+    <DotnetDebToolVersion>2.0.0-preview1-001877</DotnetDebToolVersion>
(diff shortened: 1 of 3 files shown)


eerhardt:
> Do we know why it broke?

Yes.  Two classic issues came together to hide the issue for a while, only to rear its head later:

1. Using star versions.
2. Duplicating version strings and they get out of sync.

image

There was a time when core-setup produced `beta-beta` packages for a day.  These were the "highest" version packages even though newer ones were being produced.  In the meantime, we added the `chmod` code, which required getting a path to the tool.  To produce the path, we hard-coded the current "highest" version in another piece of code.  Time goes on, all is well.  We also use the latest `beta-beta` package.

Yesterday, dotnet/core-setup changed its version from `beta` to `preview`, making the new packages higher than `beta-beta`.  Thus the star version kicked in and picked up the `preview` packages.  But we still had the hard-coded `beta-beta` path, which caused it to break.

PR #6275, opened 2017-04-08
Replace dotnet help parser with CliCommandLineParser

wli3:
CliCommandLineParserVersion 138 cannot use help as command and --help as
opinion at the same time, update to version 142


wli3:
test not pass

PR #6451, opened 2017-04-28
[WIP] Fix windows build

nguerrera:
Starting with no-op commit to check that Windows build is in fact broken in CI and then I'll rebase this to be the fix.


nguerrera:
It's not broken.

PR #6613, opened 2017-05-16
Update MSBuild to 15.3.0-preview-000234-01

eerhardt:
This build has the new RuntimeInformation feature in MSBuild which allows MSBuild logic to be able to tell if you are building on OSX or Linux.

@livarcocc @nguerrera


eerhardt:
@MattGertz merging since this is just uptaking the new MSBuild version.

PR #6635, opened 2017-05-19
Addresses #6317

EdwardBlair:
skipciplease

Addresses #6317


src/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
```diff
@@ -36,9 +36,22 @@ public void Configure()
...
 
             Reporter.Output.WriteLine();
             Reporter.Output.WriteLine(firstTimeUseWelcomeMessage);
+
+            var telemetryEnabled = 
+                _environmentProvider.GetEnvironmentVariableAsBool("DOTNET_CLI_TELEMETRY_OPTOUT", false);
+            
+            if (telemetryEnabled) {
+                Reporter.Output.WriteLine();
+                Reporter.Output.WriteLine(firstTimeConfiguringMessage);
```
EdwardBlair: This should be `Reporter.Output.WriteLine(firstTimeTelemetryEnabledMessage);`

Oops.

EdwardBlair:
Also incorperates https://github.com/dotnet/cli/pull/6634

src/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
EdwardBlair: Fixed it and rebased. Also renamed variable names so they are `FirstTimeUse*`

src/Microsoft.DotNet.Configurer/DotnetFirstTimeUseConfigurer.cs
```diff
@@ -35,10 +35,23 @@ public void Configure()
...
         {
-            const string firstTimeUseWelcomeMessage = LocalizableStrings.FirstTimeWelcomeMessage;
+            const string firstTimeUseWelcomeMessage = LocalizableStrings.FirstTimeUseWelcomeMessage;
+            const string firstTimeUseTelemetryEnabledMessage = LocalizableStrings.FirstTimeUseTelemetryEnabledMessage;
+            const string firstTimeUseConfiguringMessage = LocalizableStrings.FirstTimeUseConfiguringMessage;
 
             Reporter.Output.WriteLine();
             Reporter.Output.WriteLine(firstTimeUseWelcomeMessage);
+
+            var telemetryEnabled =
```
livarcocc: Could you please add a couple of tests for this?

livarcocc:
@EdwardBlair thanks for the PR. Can you please add tests to cover this change? Also, for functional changes, please, don't skip CI.

EdwardBlair:
Of course, completely missed that. Will do it when back at my terminal

EdwardBlair:
How do you propose one writes test for this without being dumb?

Couple of ideas I had: 
- Put `Reporter`  in the constructor and mock this -- but checking that it printed something to a fake output seems poor practice. Also, as far as I can see `Reporter` isn't unit tested (at least not in this repository -- it may be internally).
- Create an `IDotnetNoticePrinter` for example. The implementation would be practically untestable also (see previous). At least this way you'd have some sort of peace of mind that the process flow was correct.

Any thoughts?

livarcocc:
See here: https://github.com/dotnet/cli/blob/master/test/dotnet.Tests/GivenThatTheUserIsRunningDotNetForTheFirstTime.cs#L64.

I would imagine that you could add some variation where you set the expected environment variables and assert on how the output message changes.

EdwardBlair:
Brilliant, thanks. Will take a look tomorrow.

On 22 May 2017 8:17 p.m., "Livar" <notifications@github.com> wrote:

See here: https://github.com/dotnet/cli/blob/master/test/dotnet.Tests/
GivenThatTheUserIsRunningDotNetForTheFirstTime.cs#L64.

I would imagine that you could add some variation where you set the
expected environment variables and assert on how the output message changes.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/dotnet/cli/pull/6635#issuecomment-303194261>, or mute
the thread

EdwardBlair:
I have no idea how to reset the environment. Someone else can take this up.

PR #6665, opened 2017-05-22
Update dotnet/sdk version

nguerrera:
@livarcocc 

@MattGertz  FYI. Normal code flow. Ingestion of latest 2.0 dotnet/sdk in to dotnet/cli


nguerrera:
The reason for the test failure is that the test sdk package currently sets the output type in targets after we've already cached whether it is exe or winexe. This amounts to a breaking change in the core sdk. I will fix it in the morning.

nguerrera:

PR #6780, opened 2017-06-07
Prepare for non windows builds to MIT Licensed

ellismg:
I want to get a read from CI on this change.  I need to also update the RTF files for the OSX installers.

Note that this will also need changes from Core-Setup (since that's what drops the LICENSE.TXT file in the root).

Please don't merge this yet.


ellismg:

eerhardt:
Do you want this for 2.0?  If yes, then it should go into the `release/2.0.0` branch.

Unless you plan on cherry-picking the change into both branches.

eerhardt:
:shipit:

ellismg:
> Unless you plan on cherry-picking the change into both branches.

Yes, I will do that for preview 3 (since I have to make corresponding core-setup changes and they don't want to respin a build).

PR #6500, opened 2017-05-03
Run dotnet new command as current user in installer

wli3:
If not, the dotnet new will run as root and the cache folders will be in
root ownership

How to get current user
https://apple.stackexchange.com/questions/144159/how-can-i-determine-the-invoking-user-in-an-apple-installer-postinstall-script

PR #6830, opened 2017-06-11
Produce RPM packages

wli3:
**Customer scenario**

RHEL users/internal users install CLI/SDK via RPM package

The package will be located in the same place as Debian packages in blob storage

**Bugs this fixes**

https://github.com/dotnet/cli/issues/6851

**Workarounds, if any**

Install CLI/SDK via install script

**Risk**

RHEL users/internal users cannot test CLI/SDK with RPM packages easily

**Performance impact**

no

**Root cause analysis**

N/A

**How was the bug found?**

N/A

-----------------

- [x] Pending docker config to have a RHEL7 with fpm(3rd party tool to pack RPM)

Reference from https://github.com/dotnet/core-setup/tree/master/src/pkg/packaging/rpm

The goal is to have parity with Debian,
but does not publish to the feed

When run the script with rhel, it will
produce rpm file in the package step
and publish rpm to the blob storage

I tested it on VSTS 
https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build/index?buildId=801873&_a=summary

and it was successfully uploaded to my own blob storage
image


build/BundledRuntimes.props
```diff
@@ -20,7 +20,8 @@
          NOTE: There isn't a 'linux-x64' version of the installers or the additional shared framweork. -->
     <SharedFrameworkRid>$(CoreSetupRid)</SharedFrameworkRid>
     <SharedFrameworkRid Condition=" '$(UsePortableLinuxSharedFramework)' == 'true' ">linux-x64</SharedFrameworkRid>
-    <CombinedFrameworkHostCompressedFileName>dotnet-runtime-$(SharedFrameworkVersion)-$(SharedFrameworkRid)$(ArchiveExtension)</CombinedFrameworkHostCompressedFileName>
+    <CombinedFrameworkHostCompressedFilePrefix>dotnet-runtime</CombinedFrameworkHostCompressedFilePrefix>
+    <CombinedFrameworkHostCompressedFileName>$(CombinedFrameworkHostCompressedFilePrefix)-$(SharedFrameworkVersion)-$(SharedFrameworkRid)$(ArchiveExtension)</CombinedFrameworkHostCompressedFileName>
```
livarcocc: Why did you separate these? Curious.

build/package/Installer.RPM.props
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <rpmConfigJsonName>rpm_config.json</rpmConfigJsonName>
```
livarcocc: NIT: I think we should call this RPMConfigJsonName

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+        public string InputDir { get; set; }
+        [Required]
+        public string OutputDir { get; set; }
+        [Required]
+        public string PackageVersion { get; set; }
+        [Required]
+        public string ConfigJsonFile { get; set; }
+        [Output]
+        public string FPMParameters { get; set; }
+        public override bool Execute()
```
livarcocc: NIT: empty line here.

build/package/Installer.RPM.props
livarcocc: The file you added below is called dotnet_config.json? Does it get transformed into rpm_config.json?

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project='Installer.RPM.props'/>
+
+  <UsingTask TaskName="ReplaceFileContents" AssemblyFile="$(CLIBuildDll)"/>
+  <UsingTask TaskName="BuildFPMToolPreReqs" AssemblyFile="$(CLIBuildDll)"/>
+
+  <PropertyGroup>
+    <BuildRpmPackage Condition=" '$(IsRPMBasedDistro)' == True and '$(Architecture)' == 'x64' ">True</BuildRpmPackage>
```
livarcocc: Why the check on Architecture? All our linux builds are x64.

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
...
+  </PropertyGroup>
+
+  <Target Name="GenerateRpms"
+          DependsOnTargets="TestFPMTool;BuildRpms;TestSdkRpm"
+          Condition=" '$(BuildRpmPackage)'=='True' "
```
livarcocc: NIT: style-wise, we usually but spaces before and after the conditional operator above. so ' == '.

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
...
+          Outputs="@(GeneratedInstallers)"/>
+
+  <Target Name="BuildRpms"
+          DependsOnTargets="GenerateSharedHostRpm"
+          Condition=" '$(BuildRpmPackage)'=='True' and '$(FPMPresent)'=='True' "/>
```
livarcocc: NIT: same here about spacing.

livarcocc: Why check on BuildRpmPackage again here? And in which case will FPM not be present here? Should that be an error when that happens?

build/package/Installer.RPM.props
livarcocc: I see you using dotnet_config.json below. Nvm.

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
...
+      <RpmFile>$(SdkInstallerFile)</RpmFile>
+      <ManPagesDir>$(RepoRoot)/Documentation/manpages</ManPagesDir>
+      <ConfigJsonFile>$(RepoRoot)/packaging/rpm/dotnet-config.json</ConfigJsonFile>
+      <RpmIntermediatesDir>$(IntermediateDirectory)/$(RpmPackageName)/$(RpmPackageVersion)</RpmIntermediatesDir>
+      <RpmTestResultsXmlFile>$(RpmIntermediatesDir)/debian-testResults.xml</RpmTestResultsXmlFile>
+      <RpmInstalledDirectory>/usr/share/dotnet</RpmInstalledDirectory>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <rpmLayoutDirectory>$(RpmIntermediatesDir)/rpmLayoutDirectory/</rpmLayoutDirectory>
```
livarcocc: Let's stay consistent with the naming of these properties. I prefer either Rpm or RPM at the beginning. And I see in some cases you used Rpm, and other just rpm, all lowercase.

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
...
+    <!-- Copy files to rpm layout -->
+    <ItemGroup>
+      <SDKFiles Include="$(InputRoot)/**/*"/>
+      <SDKManpages Include="$(ManPagesDir)/**/*"/>
+      <SDKTemplatesFiles Include="$(TemplatesDir)/**/*"/>
+      <SDKScriptsFiles Include="$(ScriptsDir)/**/*"/>
+      <AspNetRuntimeFilesInput Include="$(AspNetRuntimePackageStorePublishDirectory)/**/*" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(SDKFiles)"
```
livarcocc: You can probably reduce this to a single Copy by creating a single "ToCopy" item and adding metadata to it indicating the Destination folder.

You can do it by doing something like:
```
<ItemGroup>
  <SomeGoodName Include="Include="$(InputRoot)/**/*">
    <DestinationFolder>$(rpmLayoutSDK)</DestinationFolder>
  </SomeGoodName>
</ItemGroup>
```

And then, when copying it, instead of having to set a different base to each one, you just access that metadata as the base folder.

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
...
+  <Target Name="GenerateRpms"
+          DependsOnTargets="TestFPMTool;BuildRpms;TestSdkRpm"
+          Condition=" '$(BuildRpmPackage)'=='True' "
+          Outputs="@(GeneratedInstallers)"/>
+
+  <Target Name="BuildRpms"
+          DependsOnTargets="GenerateSharedHostRpm"
+          Condition=" '$(BuildRpmPackage)'=='True' and '$(FPMPresent)'=='True' "/>
+
+  <Target Name="GenerateSharedHostRpm">
```
livarcocc: Should this be called GenerateSdkRpm? We are not really building the shared host rpm here.

build/package/Installer.RPM.targets
```diff
@@ -0,0 +1,213 @@
...
+          OverwriteReadOnlyFiles="True"
+          SkipUnchangedFiles="False"
+          UseHardlinksIfPossible="False"/>
+
+  </Target>
+
+  <Target Name="TestFPMTool">
+
+    <!-- run FPM  -->
+    <Exec Command="/usr/bin/env fpm --help > /dev/null" ContinueOnError="True">
```
omajid: Sorry if this is a stupid question, but isnt `/usr/bin/env fpm` same as just calling `fpm`?

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+                string changelogFile = Path.Combine(InputDir, "templates", "changelog");
+                if (!File.Exists(changelogFile))
+                {
+                    throw new FileNotFoundException($"Expected file {changelogFile} was not found.");
+                }
+                string str = File.ReadAllText(changelogFile);
+                str = str.Replace("{PACKAGE_NAME}", configJson.Package_Name);
+                str = str.Replace("{PACKAGE_VERSION}", package_version);
+                str = str.Replace("{PACKAGE_REVISION}", configJson.Release.Package_Revision);
+                str = str.Replace("{URGENCY}", configJson.Release.Urgency);
```
omajid: Urgency is not used by RPM

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+                // Update the tokens in the copyright file from the config Json 
+                string copyrightFile = Path.Combine(InputDir, "templates", "copyright");
+                if (!File.Exists(copyrightFile))
+                {
+                    throw new FileNotFoundException($"Expected file {copyrightFile} was not found.");
+                }
+                string str = File.ReadAllText(copyrightFile);
+                str = str.Replace("{COPYRIGHT_TEXT}", configJson.CopyRight);
+                str = str.Replace("{LICENSE_NAME}", configJson.License.Type);
+                str = str.Replace("{LICENSE_NAME}", configJson.License.Type);
```
omajid: Looks like a duplicated line?

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+                str = str.Replace("{LICENSE_TEXT}", configJson.License.Full_Text);
```
omajid: RPM generally has a single field for license. The entire license text is not needed.

packaging/rpm/dotnet-config.json
```diff
@@ -0,0 +1,41 @@
...
+    "homepage": "https://github.com/dotnet/core",
+    "release": {
+        "package_version": "1.0.0.0",
+        "package_revision": "1",
+        "urgency": "low",
+        "changelog_message": "Bootstrap loop package"
+    },
+    "control": {
+        "priority": "standard",
+        "section": "libs",
```
omajid: RPM has a very different naming for libraries. `Development/Languages` and `Development/Libraries` are some valid options.

packaging/rpm/dotnet-config.json
```diff
@@ -0,0 +1,41 @@
...
+    "long_description": ".NET Core is a development platform that you can use to build command-line applications, microservices and modern websites. It is open source, cross-platform and is supported by Microsoft. We hope you enjoy using it! If you do, please consider joining the active community of developers that are contributing to the project on GitHub (https://github.com/dotnet/core). We happily accept issues and PRs.",
+    "homepage": "https://github.com/dotnet/core",
+    "release": {
+        "package_version": "1.0.0.0",
+        "package_revision": "1",
+        "urgency": "low",
+        "changelog_message": "Bootstrap loop package"
+    },
+    "control": {
+        "priority": "standard",
```
omajid: Priority does not exist as a concept in RPM as far as I am aware.

packaging/rpm/templates/copyright
```diff
@@ -0,0 +1,8 @@
+Comment: Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
omajid: RPM doesnt need a copyright file like this. The convention for RPM is to install any LICENSE files shipped by upstream but no extra/custom copyright file like this.

scripts/obtain/uninstall/dotnet-uninstall-rpm-packages.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
omajid: I think a separate uninstall script for packages is a bad idea. I left a more detailed comment here: https://github.com/dotnet/cli/issues/6345#issuecomment-295918725

packaging/rpm/scripts/after_install_host.sh
```diff
@@ -0,0 +1,22 @@
...
+
+.NET Core Tools Telemetry
+--------------
+The .NET Core Tools include a telemetry feature that collects usage information. It is important that the .NET Team understands how the tools are being used so that we can improve them.
+
+The data collected is anonymous and will be published in an aggregated form for use by both Microsoft and community engineers under the Creative Commons Attribution License.
+
+The .NET Core Tools telemetry feature is enabled by default. You can opt-out of the telemetry feature by setting an environment variable DOTNET_CLI_TELEMETRY_OPTOUT (for example, 'export' on macOS/Linux, 'set' on Windows) to true (for example, 'true', 1). You can read more about .NET Core tools telemetry at https://aka.ms/dotnet-cli-telemetry."
+
+su - $SUDO_USER -c "dotnet new > /dev/null 2>&1 || true"
```
omajid: I fundamentally disagree with this approach. I explained why at https://github.com/dotnet/core-setup/issues/2613#issuecomment-306485031

livarcocc: Is the problem here the sudo or the dotnet new invocation? If it is SUDO, we won't need it for long. As we will move the archive into the dotnet hive itself and we won't modify NuGet.Config anymore. This work will probably happen this week. If the problem is the new invocation, there is no way around it today due to the custom nature of how we produce the archive to save space.

build/BundledRuntimes.props
wli3: Thank you, this is an mistake. At one point I thought dotnet runtime rpm package does not have version on the end, I was wrong in the end.

build/package/Installer.RPM.props
wli3: Just realize core-setup's name is not matching ours, there are several similar naming need to fix

build/package/Installer.RPM.targets
wli3: Removing, you are right

build/package/Installer.RPM.targets
wli3: Fixing

build/package/Installer.RPM.targets
wli3: I didn't think too much about this case. This is the same behavior for debian -- we check distro first and later check FPM/debuild, if it doesn't exist, we have silent error and produce tarball only.

It does make sense to error out when there is no FPM/debuild. Do you agree? But I think I should send separate pull request to fix deb

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+            try
+            {
+                if (!File.Exists(ConfigJsonFile))
+                {
+                    throw new FileNotFoundException($"Expected file {ConfigJsonFile} was not found.");
+                }
+
+                // Open the Config Json and read the values into the model
+                TextReader projectFileReader = File.OpenText(ConfigJsonFile);
+                if (projectFileReader != null)
```
nguerrera: File.OpenText doesn't return null on failure, it throws.

build/package/Installer.RPM.targets
wli3: Ahh there is an comment. Jenkins don't have the tools

build/package/Installer.RPM.targets
wli3: Fixing, all to Rpm

build/package/Installer.RPM.targets
wli3: Oops

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+                }
+                else
+                {
+                    throw new IOException($"Could not open the file {ConfigJsonFile} for reading.");
+                }
+            }
+            catch (Exception e)
+            {
+                Log.LogError("Exception while processing RPM paramters: " + e.Message);
+            }
```
nguerrera: It's not helpful to log only e.Message for arbitrary exceptions. If there's a hard to reproduce issue/bug, we'll want the inner exceptions and full stack trace.  I'd just let the let exceptions go and msbuild will log everything and fail the task.

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
```diff
@@ -0,0 +1,319 @@
...
+                str = str.Replace("{DATE}", DateTime.UtcNow.ToString("ddd MMM dd yyyy")); 
+                File.WriteAllText(changelogFile, str);
+            }
+            catch (Exception e)
+            {
+                Log.LogError("Exception while updating the changelog file: " + e.Message);
+            }
+        }
+
+        public void UpdateCopyRight(ConfigJson configJson)
```
nguerrera: Copyright is one word so lowercase the r.

packaging/rpm/scripts/after_install_host.sh
wli3: That is what i did in the beginning and end up being a blocking issue
https://github.com/dotnet/cli/pull/6500

build/package/Installer.RPM.targets
wli3: I didn't think too much about it, this is the same why we detect debuild. But AFAIK they are the same. just use fpm instead

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: Removing

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: This file is a copy from https://github.com/dotnet/core-setup/blob/53771d8bfc8b19afbd45e5dbdcf727049d4f9231/tools-local/tasks/BuildFPMToolPreReqs%20.cs

However I should invest more time in it, since we will maintain this file in our repo. Let me fix it

packaging/rpm/scripts/after_install_host.sh
omajid: @livarcocc My concern is mostly a combination of the two. Package installation is a system-wide operation. It should create a system wide cache, or none at all. This approach creates a per-user cache. Touching user directories from a package installation is a no-no. It is even unpredictable which user the cache is created for depend on whether you do `su` or `sudo` when you trigger package installation. Similarly, each package tells the package manager about all the files it installs/owns. This one does not, because the files are owned by a particular user. Please also see https://github.com/dotnet/core-setup/issues/2613 for additional reasons @ellismg and I both think this line should be removed.

packaging/rpm/scripts/after_install_host.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+echo "This software may collect information about you and your use of the software, and send that to Microsoft."
+echo "Please visit http://aka.ms/dotnet-cli-eula for more information."
+
+# Run 'dotnet new' as the user to trigger the first time experience to initialize the cache
+echo "Welcome to .NET Core!
```
omajid: Compared to debian packages, RPMs are *non-interactive*. Its safe to assume a user may never see this, ever.

build/BundledRuntimes.props
wli3: Fixed

build/package/Installer.RPM.props
wli3: Fixed

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: Fixed

build/package/Installer.RPM.targets
wli3: Fixed

build/package/Installer.RPM.targets
wli3: Fixed

build/package/Installer.RPM.targets
wli3: Fixed

build/package/Installer.RPM.targets
wli3: Pending... Maybe later? I want to get that in today

build/package/Installer.RPM.targets
wli3: fixed

build/package/Installer.RPM.targets
wli3: fixed

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: Removed

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: Removed

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: Need more time to look into it, since it involves legal, also, that is what runtime package does. I don't think it is a blocking issue today

packaging/rpm/dotnet-config.json
wli3: section removed

packaging/rpm/dotnet-config.json
wli3: priority removed

packaging/rpm/templates/copyright
wli3: Again, please give me more time for a separate pull request

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: null check removed

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: catch removed -- I hate this pattern as well

build_projects/dotnet-cli-build/BuildFPMToolPreReqs.cs
wli3: Fixed

scripts/docker/rhel/Dockerfile
```diff
@@ -0,0 +1,27 @@
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# Dockerfile that creates a container suitable to build dotnet-cli
+FROM microsoft/dotnet-buildtools-prereqs:rhel-7-rpmpkg-c982313-20174116044113
+
+# Install from sudo main package TODO This package needs to be mirrored
+RUN yum install -y https://www.sudo.ws/sudo/dist/packages/RHEL/7/sudo-1.8.20-3.el7.x86_64.rpm \
```
wli3: Help needed. For some weird reason I cannot install sudo from yum. I tried adding repo.
Currently this is from sudo's home page

wli3:
@nguerrera @omajid @livarcocc  After the docker change, this pull request is in ready to merge state. I also addressed feedback so far.

packaging/rpm/scripts/after_install_host.sh
livarcocc: I chatted about this with @ellismg, going forward, we will actually move the fallback folder inside the dotnet hive, so, no matter if you install per user (tarball) or global (rpm), the fallback folder will be placed relative to the SDK.

It will be
```
/dotnet
  /fallbackfolder
  /sdk
  /shared
  dotnet.exe
```
Something like above.

packaging/rpm/scripts/after_install_host.sh
livarcocc: I agree we need to fix this. We should separate the condition of first run and telemetry and show the telemetry message in the first run despite the fact that the cache has been expanded. I believe we have a bug tracking this already, but I also believe this should not block the PR.

wli3:
@MattGertz Could you review this change?

wli3:
@livarcocc all good --  there is the first official RPM https://dotnetcli.blob.core.windows.net/dotnet/Sdk/2.0.0-preview2-006445/dotnet-sdk-2.0.0-preview2-006445-rhel-x64.rpm

scripts/docker/rhel/Dockerfile
wli3: @omajid do you know what is the correct yum repository to enable in order to yum install sudo?

omajid: @wli3 Assuming you are building this container on a RHEL 7 box registered with a subscription and receiving updates, the RHEL container should automatically be able to find yum from the right/default repositories. Can you find `sudo` using yum on the RHEL host?

wli3: So the host or the container should be registered? The host is not. I need to double check if container is registered. However, if it is not registered there is no access to the feed?

omajid: @wli3 Only the host needs to be. The container picks up the registrations and the repositories from the host. If the host is not registered, neither the host nor the container gets access to subscriptions and updates.

PR #6867, opened 2017-06-14
Fixing a build pubish to blob storage problem

ghost:
We stopped producing distribution specific assets for Linux. Therefore, removing the check for existence before publishing to 'latest'.

Plainly stated, we are checking for badges that are no longer being produced; this causes the build to fail the check to trigger the publish to 'latest'.

@dotnet/dotnet-cli

@MattGertz


nguerrera (approved):

MattGertz:
Yep, I saw the mail, makes sense.  Approved.

wli3 (approved):

PR #6906, opened 2017-06-16
Stop produce duplicated package for VSTS build definition

wli3:
Currently VSTS is producing duplicate package and they are running into
each other will producing

Today:
Linux x64 build definition: both .deb and .tar.gz
Ubuntu 16.04 build definition: both .deb and .tar.gz (ex.
dotnet-sdk-2.0.0-preview3-006477-linux-x64.tar.gz,
dotnet-sdk-debug-2.0.0-preview3-006477-linux-x64.tar.gz,
dotnet-sdk-internal-2.0.0-preview3-006477-linux-x64.tar.gz)

After this change:
Linux x64: .tar.gz
Ubuntu 16.04: .deb

No change for windows and macOS
Tested on Ubuntu with and without --linux-portable and windows

### DIFF
--- a/build/DerivedHostMachineInfo.props
+++ b/build/DerivedHostMachineInfo.props
@@ -1,0 +1,7 @@
+<Project>
+    <PropertyGroup>
+      <IsDebianBaseDistro Condition=" '$(HostOSName)' == 'ubuntu' OR '$(HostOSName)' == 'debian' ">true</IsDebianBaseDistro>
+      <IsRPMBasedDistro Condition=" $(HostRid.StartsWith('rhel')) ">true</IsRPMBasedDistro>
+      <IsLinuxDistroSpecific Condition=" ('$(IsDebianBaseDistro)' == 'true' OR '$(IsRPMBasedDistro)' == 'true') AND '$(IslinuxPortable)' != 'true' ">true</IsLinuxDistroSpecific>
+    </PropertyGroup>
+</Project>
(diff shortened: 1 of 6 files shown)


build/DerivedHostMachineInfo.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+    <PropertyGroup>
+      <IsDebianBaseDistro Condition=" '$(HostOSName)' == 'ubuntu' OR '$(HostOSName)' == 'debian' ">true</IsDebianBaseDistro>
+      <IsRPMBasedDistro Condition=" $(HostRid.StartsWith('rhel')) ">true</IsRPMBasedDistro>
```
wli3: according to this https://github.com/dotnet/corefx/issues/1022 we should use lower case

build/FileExtensions.props
```diff
@@ -1,20 +1,17 @@
...
-      <IsRPMBasedDistro Condition=" $(HostRid.StartsWith('rhel')) ">True</IsRPMBasedDistro>
-
       <ArchiveExtension Condition=" '$(HostOSName)' == 'win' ">.zip</ArchiveExtension>
       <ArchiveExtension Condition=" '$(HostOSName)' != 'win' ">.tar.gz</ArchiveExtension>
 
       <InstallerExtension Condition=" '$(HostOSName)' == 'win' ">.msi</InstallerExtension>
       <InstallerExtension Condition=" '$(HostOSName)' == 'osx' ">.pkg</InstallerExtension>
-      <InstallerExtension Condition=" '$(IsDebianBaseDistro)' == 'True' ">.deb</InstallerExtension>
-      <InstallerExtension Condition=" '$(IsRPMBasedDistro)' == True ">.rpm</InstallerExtension>
+      <InstallerExtension Condition=" '$(IsDebianBaseDistro)' == 'true' ">.deb</InstallerExtension>
```
wli3: It is used here, so make it lower case

PR #7036, opened 2017-06-29
Fix test failures in local build due to environment variable filter

nguerrera:
Finally got to the bottom of two tests that always fail locally for me, as it turned out to have the same root cause as a test issue with the UI language environment variable that I'm working on.

The culprit is af9a85fd3f3b25a0a371527e3fe611fe184c2b75, which tries to make DOTNET_MULTILEVEL_LOOKUP=0 global to the entire build and test process and removes the setting of DOTNET_MULTILEVEL_LOOKUP=0 on individual tests, but:

* It fails to export the environment variable to sub-process on Unix
* Only an explicit list of environment variables prefixed with DOTNET_ are passed down to our DotnetTool build tasks, and DOTNET_MULTILEVEL_LOOKUP was not in the list.

I debated reverting back to how things were (with individual test opt-outs of multilevel lookup), but decided to instead move forward from af9a85fd3f3b25a0a371527e3fe611fe184c2b75 and fix the two issues above. This makes the entire build free of a dependency on which SDKs are installed globally as af9a85fd3f3b25a0a371527e3fe611fe184c2b75 intended.

@dotnet/dotnet-cli @livarcocc 

@MattGertz FYI, infrastructure-only change

### DIFF
--- a/build_projects/dotnet-cli-build/EnvironmentVariableFilter.cs
+++ b/build_projects/dotnet-cli-build/EnvironmentVariableFilter.cs
@@ -28,6 +28,7 @@ namespace Microsoft.DotNet.Cli.Build
         private IEnumerable<string> _environmentVariablesToKeep = new string []
         {
             "DOTNET_CLI_TELEMETRY_SESSIONID",
+            "DOTNET_MULTILEVEL_LOOKUP",
             "DOTNET_RUNTIME_ID",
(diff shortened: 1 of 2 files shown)


nguerrera:
Test Windows_NT x64 Release Build

It actually already succeeded, but did not update the status here. :(

PR #7076, opened 2017-07-05
Security build definition for CLI

ravimeda:
Similar to https://github.com/dotnet/core-setup/pull/2722
The approach for acquiring official build blobs is different since CLI does not have a dependency dotnet/buildtools that offers reliable means to download Azure blobs. 

Approach:
1. Clone core-setup, and run `init-tools` command to acquire BuildTools.
2. Download blobs with from `dotnetcli/Sdk/<branch>` . For Windows build, include only blobs with extension `.zip` or `.exe`.
3. Run BinSkim and APIScan on all assemblies and executables in the downloaded blobs.
4. Get sources corresponding to the official build. This means checkout at the SHA listed `latest.version`
5. Run CredScan and PoliCheck on the sources.

@morganbr  list of issues found -
https://msazure.visualstudio.com/DefaultCollection/One/_workitems?tempQueryId=740da6e8-15ce-4912-b983-2cf87ed6a530

@livarcocc PTAL.

Post PR, I will update documentation for launching security build for CLI at  https://github.com/dotnet/core-eng/blob/master/Documentation/Project-Docs/security-builds.md

https://github.com/dotnet/core-eng/issues/650


ravimeda:
@livarcocc any feedback, please?

build/buildpipeline/security/dir.props
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\..\core-setup\dir.props" />
```
ellismg: What's up with this import?

ellismg: Does this file even need to be here?

ellismg: Nevermind, I see what's happening.

ravimeda: Glad you figured :)

ellismg:
FWIW, cloning `master` of `dotnet/corefx` to get the tooling seems a little weird.  Now a change in `dotnet/corefx` has the ability to break CLI.  Is that what we want?  Are we just doing this dance to get buildtools?

morganbr:
@ravimeda , it looks like there are lots of missing symbol errors. I'm not sure if all of those are for components we own or not.

ravimeda:
@ellismg `DownloadFromAzure` and `ListAzureBlobs` MSBuild tasks from BuildTools are being used in Core-Setup security build to download packages produced from official builds. Since CLI does not have dependency on BuildTools, I had a couple options to download official build packages -

1. Write a new script that will download CLI packages.
2. Write a new MSBuild task similar to `DownloadFromAzure`.
3. Clone BuildTools. Preferably, at a specific SHA and not the latest.
4. Clone Core-Setup.

Option 1 is simple and straightforward but requires CLI to maintain the script.
Option 2 introduces redundancy in the repos, and requires CLI to maintain the task.
Option 3 I tried this, and it seems to work fine. However, @dagood recommended option 4.

ravimeda:
@morganbr about the missing symbols issue, I will kickoff a fresh build to confirm these issues still exist. If so @livarcocc can clarify which assemblies should be excluded.

dagood:
> Option 3 I tried this, and it seems to work fine. However, @dagood recommended option 4.

The reason is that BuildTools uses an old version of the BuildTools package to build itself, so you're not going to get recent versions of the tasks you're looking for. You could actually *build* BuildTools, but there's no precedent that I'm aware of to use the tooling that way and it's unnecessary overhead. Core-Setup (or CoreFX), on the other hand, fetches and sets up new BuildTools packages.

ravimeda:
Please ignore the noise. I'm trying to rebase and resolve merge conflicts.  If it does not work, I'll send a fresh PR.

ravimeda:
Please use https://github.com/dotnet/cli/pull/7296

PR #7094, opened 2017-07-06
MIT License *NIX Builds.

ellismg:
Remove non MIT licensed components from the CLI when building on non
windows platforms. The shared framework provides the LICENSE file that is
included with the tarballs, so we just need to ensure we don't pull any
windows specific stuff that is not MIT licensed.

Fixes: dotnet/core-setup#676

### DIFF
--- a/src/redist/redist.csproj
+++ b/src/redist/redist.csproj
@@ -142,6 +142,14 @@
                                    AssetPath="%(NuPkgContentForMSBuildExtensionsRelativePaths.Identity)" />
   </Target>
 
+  <!-- The msdia140typelib_clr0200.dll file is not MIT licensed (and it only used on Windows). Remove it, so
+       we can MIT license the published dotnet -->
+  <Target Name="RemoveMSDiaTypeLib"
+          AfterTargets="Publish"
+          Condition="'$(OSName)' != 'win'">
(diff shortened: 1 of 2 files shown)


ellismg:
This ports a change which was already in master which prevents a few windows specific components from being included in the *nix builds.  The shared framework already only contains binaries built from MIT licensed code and with this change, the CLI will not carry any more non MIT licensed binaries on *nix.  The Windows build remain as they were (and in core-setup, the shared framework for windows continues to use the Microsoft library license)

ellismg:
@Petermarcu

Petermarcu:
@livarcocc I don't know what process you have for this branch at this point. The change looks good to me. Is this the resolution to the issue in the other repo?

ellismg:
> I don't know what process you have for this branch at this point.

Ditto, I hope @livarcocc can tell me the right thing to do.

> The change looks good to me. Is this the resolution to the issue in the other repo?

Yes. (Fixes: dotnet/core-setup#676)

ellismg:
Note that a version of this change has been baking in master since before Preview 2.

Petermarcu:
Ok, @livarcocc we need to take this. Can you review? If not resolved today, I'll create a new tracking issue in this repo so the 2.0.0 milestone can be clean in the runtime repos.

Issue #6086, opened 2017-03-21
Revisit Telemetry configuration


vcsjones:
## Problem

There are a few issues around telemetry that are open today, such as #3093 and #2960.

I think it is safe to say that the ship has sailed on Telemetry and its default options, and what it collects. What I want to focus on is making it clear to understand what is going on and how to adjust it.

A current pain point of telemetry is that its opt-out experience is driven by the presence of an environment variable. This has to be done per-machine, and may not always be obvious how to do it in a persistent way for all accounts on a machine.

This can be troublesome in enterprises that need to be able to control these settings.

### User and System Configuration

Across supported platforms, there should be a user level and machine level telemetry option. For example, in Windows this might be a registry key in HKLM and HKCU, where HKCU is checked first, then falls back to HKLM, then falls back to default. For macOS, this could be a plist in `~/Library` and `/Library`.

This allows organizations to configure telemetry via registry, which is a process they should already be familiar with. Same with `plist` files for macOS.

### CLI configuration

There should be a `dotnet telemetry` command that changes the user-level configuration for telemetry. This allows people to make the change without trying to do this in `.bash_profile`, etc.

The environment variable should continue to be respected, and would probably "win" in all situations. So the order of preference is Environment Variable, User Config, Machine Config, Default.

gortok:
I'll caveat this by saying that there are particular industries where this is especially problematic, and I think a minor tweak can be done to improve those issues.

By default, if you're invoking .NET Core for deployed Software, i.e, "`dotnet MyApp.dll`", that command should not collect telemetry.

In my particular use case (and this is a rather common one in industries where they aren't "SaaS" products, but rather Client Hosted solutions, the current setup means we need to ensure our Docker containers have a common line in them to set the environment variable correctly.

I understand that you want to see how developers use your product; but that's not the same as people who aren't developers -- people who should not be opted in (by default) to having metrics collected that can be used by nefarious actors (of course, you're only doing this over encrypted connections, right?)

NickCraver:
I'm happy to provide my own telemetry. If it was opt-in, I would. But that's not the decision gone with here, so instead we'll deploy a GPO that disables it for everyone on our domain as we roll out the .NET Core tooling, simply because that's the simplest thing to safely do.

I sincerely hope the opt-in/out is re-visited. At the very least, this shouldn't require an environmental variable to opt-out, they're a PITA and impractical. I'm a fan of the `dotnet telemetry` command option above.

richlander:
[...]

PR #7096, opened 2017-07-07
Switched telemetry to be OFF by default

rafaelrpinto:
Switched the telemetry option from disable if you don't agree to enable if you agree. The translations need further work.
Addresses issue #3093.
skipciplease


mgravell:
Might as well rename that to "remove all telemetry", in terms of the effect it will have. Fetches popcorn...

vcsjones:
Well, this would be one way to address #6086.

yuhong:
This should be a build option IMO.

rafaelrpinto:
@mgravell 
Just materializing an attempt to address all the different opinions of  #3093. Don't shoot the messenger =]

rafaelrpinto:
Attempted to update the readme locally and messed the whole thing up. Will make a new Pr later.

PR #7121, opened 2017-07-10
Update to SetupCrossgen 277 - Coherence 26182

mlorbetske:
CLI impacting changes:
- Enable loc for dotnet new
- Re-enables auto-restore

Non-impacting changes:
- Adds chmod post-action (not used by default) to unblock Fable templates
- Fixes an issue where URL encoded paths are used in the output names when creating templates from ZIPs (default templates don't hit this)
- Adds a post-action for calling chmod (not used by any of the built in templates)
- Fixes an issue where a directory that gets renamed will prevent files from being renamed in the output


mlorbetske:
@dotnet-bot test OpenSUSE42.1 x64 Debug Build please

PR #7177, opened 2017-07-17
update nuget to 4.3.0-rtm-4294

rohit21agrawal:
looks like CLI needs to be updated before we can insert into SDK (since SDK is on a much older version of nuget 4.3.0-preview3-4168), so doing that first and will update the sdk [PR](https://github.com/dotnet/sdk/pull/1419) once we get a build of CLI out with the update nuget.


nguerrera:
@rohit21agrawal We also need to update the 1.1 branch

nguerrera:
@srivatsn FYI Flow of dependencies to 2.0 CLI

rohit21agrawal:
@nguerrera which is the 1.1 branch?

nguerrera:
rel/1.1.0

PR #7186, opened 2017-07-18
make opensuse more stable

wli3:
Use docker image for opensuse 13.2 from runtime

@srivatsn FYI this is infrastructure change


wli3:
image

PR #7311, opened 2017-07-28
Improve the way Roslyn is pulled in

khyperia:
In particular, use the Microsoft.NETCore.Compilers package, instead of manually patching together various bits.

This is the first part of work that will allow Roslyn to be more independent from CLI, and be able to implement optimizations and improve performance without having to coordinate with the CLI.

This PR may not be ready to merge, as the `RetargetRuntimeConfig` hack is not something that we would like to continue doing into the future (even though I believe we're already doing so today?). The fix for that depends on Microsoft.NETCore.Compilers using netcoreapp2.0, which I believe might not happen until 2.0 ships (@jaredpar likely could explain more on why that is, I'm not 100% sure on the reasoning). However, I wanted to start the code review process on other aspects of this change.

---

A major change in this PR is changing the way crossgen works - before, we were crossgenning all assemblies simultaneously, with pulled-in packages allowing to stomp on each other's dependencies and change versions, etc.. This caused a lot of issues, especially with the new Roslyn bits. I have changed the way crossgenning works to crossgen batches of assemblies - where each batch is in its own isolated world, and cannot see assemblies from other batches. (At runtime, each "batch" is always in a separate process, so this is okay). The currently-defined batches are the F# dir, the Roslyn dir, and "everything else". F# gets a little complicated, since they have an assembly (`FSharp.Build.dll`) that is intended to be used by the process running in the "everything else" dir, and doesn't reference anything in the FSharp dir. (The way Roslyn handles this is having its Build.dll be in the root Roslyn folder, and then the out-of-proc compiler stuff is in the `bincore` folder - so it's simple to just crossgen the bincore folder as a batch, and include the Roslyn dir in the "everything else" batch.)

Ping @livarcocc, as they were included in the planning of this PR.


src/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -23,8 +23,10 @@ internal class MSBuildForwardingAppWithoutLogging
             new Dictionary<string, string>
             {
                 { "MSBuildExtensionsPath", AppContext.BaseDirectory },
```
jaredpar: What does this all control?

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
```diff
@@ -15,6 +16,35 @@ public DotnetCommand()
...
+            var containingDirectory = Path.GetDirectoryName(dotnetUnderTest);
+            if (!string.IsNullOrEmpty(containingDirectory))
+            {
+                // if it's already set, use that
+                string path;
+                if (!Environment.TryGetValue("PATH", out path))
+                {
+                    // otherwise grab it from the environment
+                    path = System.Environment.GetEnvironmentVariable("PATH");
+                }
```
jaredpar: Need a blank line after the closing `}` here.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
```diff
@@ -15,6 +16,35 @@ public DotnetCommand()
...
+                // if it's the only item in the path, set it
+                if (string.IsNullOrEmpty(path))
+                {
+                    path = containingDirectory;
+                }
+                else
+                {
+                    // otherwise, join it with the rest of the path,
+                    // putting it in front so it takes priority
+                    path = containingDirectory + Path.PathSeparator + path;
```
jaredpar: What if `path` already contains `containingDirectory`?

src/redist/redist.csproj
```diff
@@ -200,41 +198,72 @@
           Condition=" '$(DISABLE_CROSSGEN)' == '' "
           AfterTargets="PublishMSBuildExtensions">
     <ItemGroup>
+      <RoslynFiles Include="$(PublishDir)Roslyn\bincore\**\*" />
```
jaredpar: This seems like it will place all of our compiler binaires inside bincore. How do our MSBuild task binaries get published?

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
```diff
@@ -15,6 +16,35 @@ public DotnetCommand()
         public DotnetCommand(string dotnetUnderTest)
             : base(dotnetUnderTest)
         {
+            // Some scripts (e.g. RunCsc/RunVbc) depend on the correct `dotnet`
+            // being in the PATH. When tests use `DotnetUnderTest.WithBackwardsCompatibleRuntimes`,
+            // this means that the up-to-date dotnet command is resolved by the
+            // old-runtime-targeting RunCsc/Roslyn, causing compatability issues.
+            // So, detect if we're running a dotnet that's not just a plain "dotnet",
+            // and if so, make sure it's on the PATH.
```
nguerrera: I think we should arrange in the product without mucking with PATH that a build initiated by a given `dotnet` uses the same `dotnet` to launch the compiler that ships with the CLI. Perhaps we can set something other than PATH that RunCsc/RunVbc can use. It can fallback to PATH if that thing is not set.

This feels like we're patching the test logic without fixing a product issue.

build/DependencyVersions.props
```diff
@@ -4,7 +4,6 @@
     <CLI_SharedFrameworkVersion>2.0.0</CLI_SharedFrameworkVersion>
     <CLI_MSBuild_Version>15.3.409</CLI_MSBuild_Version>
     <CLI_Roslyn_Version>2.3.2-beta1-61921-05</CLI_Roslyn_Version>
-    <CLI_Roslyn_Satellites_Version>2.3.0-pre-20170624-6</CLI_Roslyn_Satellites_Version>
```
nguerrera: Removing this is gated on getting roslyn to build its own satellites assemblies in the open and delivering them with the rest of the roslyn bits via a .nupkg out of the roslyn repo. If we remove this before that happens, then localization of the compiler will regress. The plan is to get roslyn on to the OSS localization process in 15.5, but I don't want to delete the interim process here until that happens in case we need to ship these CLI bits first.

src/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
nguerrera: These are environment variables passed down to the msbuild process created by the CLI whenever it shells out to msbuild.

src/redist/redist.csproj
khyperia: This variable is not for publishing - it's only for crossgen. The variable for what files to include is called `RoslynBits`, and is the first diff above this one. This variable is used for what I describe in the PR as "crossgen batches".

> The way Roslyn handles this is having its Build.dll be in the root Roslyn folder, and then the out-of-proc compiler stuff is in the bincore folder - so it's simple to just crossgen the bincore folder as a batch, and include the Roslyn dir in the "everything else" batch.

(Build.dll, as well as targets, etc.)

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
khyperia: Agreed. I recall talking this over with @agocke, so they can comment more, but - this is a result of a major fundamental issue that we need to figure out of "how do we pass the host down to client code". That requires fairly significant design work, and as of right now I've left it as this cludge. (Setting a special variable goes against the purpose of decoupling the CLI and Roslyn more)

nguerrera: I think if the variable is something like DOTNET_HOST_PATH and the contract is with all child processes of dotnet, then there's no bad coupling. It should probably be set by the host itself (core-setup repo) not here though...

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
khyperia: Nothing harmful - the directory will just be searched twice when looking up an exe. (I just pushed a commit that solves this in the basic case, but ignores all the complex cases that filesystems can introduce, like symlinks)

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
khyperia: Sounds good. I'll submit a PR to Roslyn to honor that contract, then update this PR once that version of Roslyn is on nuget.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
```diff
@@ -15,6 +17,41 @@ public DotnetCommand()
...
+            var containingDirectory = Path.GetDirectoryName(dotnetUnderTest);
+            if (!string.IsNullOrEmpty(containingDirectory))
+            {
+                // if it's already set, use that
+                string path;
+                if (!Environment.TryGetValue("PATH", out path))
+                {
+                    // otherwise grab it from the environment
```
wli3: There is too much comments. And people have to first understand the comments and do the code change. please build up abstraction by chain of responsibility pattern or at least extract several methods and put the comment as their name.

wli3: -

khyperia: I believe that the comments are necessary for understanding this code, as they explain the why, instead of just the how. I don't think a method name is adequate for describing the subtleties going on in this code. And for the feedback that understanding comments is required before changes to the code can be made: Yes, that is correct, and very important, because any changes done without a strong understanding of what this code is doing may introduce subtle bugs.

Additionally, I've just pushed a commit that takes into account feedback yesterday, to use `DOTNET_HOST_PATH` instead of modifying `PATH` here. This cuts out some of the comments - could you let me know if you still think there are too many comments, and if so, what should be done to resolve them?

khyperia:
Note that in the latest commit, I had to bump the Roslyn version to `2.6.0-beta1-62003-02`. This is because commit https://github.com/dotnet/roslyn/commit/6a6e45600902fe2b5cdc0844e171989927e30605 only just merged yesterday (as it was in response to @nguerrera's feedback yesterday to use `DOTNET_HOST_PATH`). I believe 2.6.0 is what CLI `master` should be targeting going forward, but a review on that would be good.

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
wli3: @khyperia  

> as they explain the why, instead of just the how

I disagree, C# is a high level language that is capable explain why in the code. But I know it is very debatable 

>  And for the feedback that understanding comments is required before changes to the code can be made

If it is this case, could you add test coverage for this? Comment will out of sync and people make mistake. I hope at least test will fail due to it and point me to the code. (Even it is a test helper method). I don't want this code to be another legacy part of the code I am afraid to change.

wli3: Since this is test helper, will it fail the build upfront if I make a mistake in the code?

khyperia: > If it is this case, could you add test coverage for this?

As you pointed out, this is part of tests, not shipped code - as far as I know, there isn't a good framework in place for thoroughly testing tests.

> will it fail the build upfront if I make a mistake in the code?

Yes, and this is why it was originally included - omissions or changes will fail the test runs. However, the purpose of comments is to make sure that future people who either read this code or would like to update this code understand the purpose of this code, and so know how to add additional tests for any functionality they may add or change.

Please do also note the updated commit - could I ask for clarification if your above comments were for the commit this comment is on, or the most recent commit?

khyperia:
Also, I think the `OSX10.12 x64` test failure may be legitimate - I think it was happening consistently during various revisions of this PR, and the test that fails (`Microsoft.DotNet.Cli.Utils.Tests.GivenAppThrowingException.ItShowsStackTraceWhenRun`) is a failure I encountered in private testing. I think that's due to it requiring `artifacts/*/stage2WithBackwardsCompatibleRuntimes` (`DotnetUnderTest.WithBackwardsCompatibleRuntime`), which is what the change to `DotnetCommand.cs` in this PR was supposed to fix. I don't have access to my Mac right now (it's at home), so I will investigate and update this PR tomorrow (the Jenkins run logs don't provide enough diagnostics to find the root cause of the issue, especially because it's platform-specific)

test/Microsoft.DotNet.Tools.Tests.Utilities/Commands/DotnetCommand.cs
jaredpar: > I disagree, C# is a high level language that is capable explain why in the code. But I know it is very debatable

C# is very capable of displaying "what". And like practically every other language poor at explaining "why". These comments help explain the "why" here.

khyperia:
> Also, I think the `OSX10.12 x64` test failure may be legitimate

Just tested locally, this is the basic repro of the error. It seems that some of the runtime-swapping behavior the CLI packaging is doing is not valid on OSX. Advice on how to resolve this would be appreciated.

```
khyperia@ashleys-mbp ../stage2WithBackwardsCompatibleRuntimes % pwd
/Users/khyperia/Documents/cli/artifacts/osx.10.12-x64/stage2WithBackwardsCompatibleRuntimes
khyperia@ashleys-mbp ../stage2WithBackwardsCompatibleRuntimes % ./dotnet ./sdk/2.1.0-preview1-007006/Roslyn/bincore/csc.dll

Unhandled Exception: System.BadImageFormatException: Could not load file or assembly '/Users/khyperia/Documents/cli/artifacts/osx.10.12-x64/stage2WithBackwardsCompatibleRuntimes/sdk/2.1.0-preview1-007006/Roslyn/bincore/csc.dll'. An attempt was made to load a program with an incorrect format.

zsh: abort      ./dotnet ./sdk/2.1.0-preview1-007006/Roslyn/bincore/csc.dll
```

---

(for completeness, the same command works fine in the normal `stage2` dir)

```
khyperia@ashleys-mbp ..nts/cli/artifacts/osx.10.12-x64/stage2 % pwd
/Users/khyperia/Documents/cli/artifacts/osx.10.12-x64/stage2
khyperia@ashleys-mbp ..nts/cli/artifacts/osx.10.12-x64/stage2 % ./dotnet ./sdk/2.1.0-preview1-007006/Roslyn/bincore/csc.dll
Microsoft (R) Visual C# Compiler version 2.6.0.62003 (4535642b)
... blah blah, "no sources specified error" ...
```

khyperia:
Update on the `PATH` investigations: it looks like *many* things in the CLI depend on the correct `dotnet` being in the `PATH`. For example, every single subcommand boils down to [this line](https://github.com/dotnet/cli/blob/cc9e942ede3df06c70fc3a25613e85cc071f189d/src/Microsoft.DotNet.Cli.Utils/ForwardingAppImplementation.cs#L105), which invokes an unqualified `dotnet`. Asking Roslyn to be resilient against `dotnet` not being in `PATH` doesn't make sense, we should create a plan to fix up the CLI in general, and submit a separate issue/PR for that.

As such, leaving the setting of `DOTNET_HOST_PATH` in tests only makes sense for now, as tests are broken (they assume we're resilient about `dotnet` not being in `PATH`, but we're not - we just happen to be dancing around any case where it would fail)

An initial sketch for how it could be fixed:

* On the main `dotnet` entry point, we check the environment variable `DOTNET_HOST_PATH`. If it is unset, we do `[magic]` to grab it and set the environment variable. If it's already set, we don't touch it.
* When invoking any sub-`dotnet` command, we grab the path from `DOTNET_HOST_PATH` (and error if it's unset?). This needs to be fixed in multiple places, as a quick search looks like we have an unqualified `dotnet` in quite a few places (not just `ForwardingAppImplementation.cs`).

This solution also allows subcommands that are not the CLI (such as Roslyn) to use `DOTNET_HOST_PATH`, as the environment variable will be implicitly passed down to every spawned process.

Suggestions for what `[magic]` could be:

* Find an existing API that gives us the location of the executable. (`Process.GetCurrentProcess().MainModule.FileName`?)
* Implement a new API that allows us to access it (perhaps native entry points simply set `DOTNET_HOST_PATH` to `argv[0]`? We currently drop `argv[0]` everywhere, as far as I can tell)
* Get the location of `dotnet.dll` (a fairly easy thing to do) and walk up the directory tree until `dotnet.exe` (or `dotnet` on linux) is found.

dsplaisted:
I've filed #7396 for the issue where `dotnet` needs to be in the path.

Microsoft.DotNet.Cli.sln
```diff
@@ -673,30 +673,6 @@ Global
...
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.Release|x64.ActiveCfg = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.Release|x64.Build.0 = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.Release|x86.ActiveCfg = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.Release|x86.Build.0 = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.RelWithDebInfo|Any CPU.ActiveCfg = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.RelWithDebInfo|Any CPU.Build.0 = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.RelWithDebInfo|x64.ActiveCfg = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.RelWithDebInfo|x64.Build.0 = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.RelWithDebInfo|x86.ActiveCfg = Release|Any CPU
-		{A0670C63-BA7A-4C1B-B9A7-1CA26A7F235C}.RelWithDebInfo|x86.Build.0 = Release|Any CPU
```
jaredpar: Are these all projects that are just missing from the repo?

khyperia: I think this was actually just tool_roslyn, and I misread the GUIDs when removing it from the sln the first time around.

nguerrera:
> I've filed #7396 for the issue where dotnet needs to be in the path.

https://github.com/dotnet/cli/issues/7396#issuecomment-321389629

khyperia:
@nguerrera Could you give advice on what you meant by:

> It should probably be set by the host itself (core-setup repo) not here though...

(or more specifically, where exactly should we be setting `DOTNET_HOST_PATH`? - this PR only sets it in tests right now)

jaredpar:
The goal of this PR is meant to make it easier for us to insert Roslyn packages. It seems like we're pushing it to make the underlying scripts more reliable, implement features that don't yet exist, etc ... 

I'd prefer we keep this PR focussed on the goal: simplifying insertion. This is just one step in the larger effort of making the C# experience in CLI better. I agree with the principle behind many of the suggestions around abstracting out host. But that should be tackled in a different PR (in particular because we need a design that takes into account the broader picture like Mono).

nguerrera:
@jardepar I don't disagree with that at all. My only concern is that we don't regress things. If we're having to set something in tests, but don't make the same arrangement in the product, then we risk a backwards step. If I've understood correctly, the transcript at https://github.com/dotnet/cli/issues/7396#issuecomment-321389629 would start to fail if we just merge as-is.

I'm giving some thought to the most tactical fix to keep things moving in the right direction and I will reply back in a bit on it.

nguerrera:
I think the simplest approach would be to set DOTNET_HOST_PATH where we set CscToolExe to the script that we're expecting to honor DOTNET_HOST_PATH if present: https://github.com/dotnet/cli/pull/7311/files#diff-5f48ce6cbfb26992402a6c8044d370c9. 

To obtain it, I think we should use `System.Diagnostics.Process.GetCurrentProcess().MainModule.FileName`. 

My more long term thinking was that the native booting process would simply set DOTNET_HOST_PATH and we wouldn't have to do anything in CLI. I will open a bug on core-setup to discuss, but I'm fine with above to get this unblocked. It may well turn out to be a fine long term solution.

khyperia:
Ping on this, could I have a review? (@nguerrera?)

src/redist/redist.csproj
```diff
@@ -17,15 +17,16 @@
...
     <PackageReference Include="NuGet.Build.Tasks" Version="$(CLI_NuGet_Version)" />
     <PackageReference Include="Microsoft.TestPlatform.CLI" Version="$(CLI_TestPlatform_Version)" />
     <PackageReference Include="Microsoft.TestPlatform.Build" Version="$(CLI_TestPlatform_Version)" />
     <PackageReference Include="NuGet.ProjectModel" Version="$(CLI_NuGet_Version)" />
-    <!-- The project json migration commands depend on an older version of Roslyn.
-         Lift the version here to match what tool_roslyn depends on (otherwise an older version will
-         be added to the TPA when we crossgen and we won't be able to crossgen tool_roslyn -->
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(CLI_Roslyn_Version)" />
+    <PackageReference Include="Microsoft.NETCore.Compilers" Version="$(CLI_Roslyn_Version)">
+      <ExcludeAssets>All</ExcludeAssets>
```
nguerrera: Why are we excluding assets?

khyperia:
> I had imagined this would be full file path including file name.

Pushed a commit to Roslyn to respect this. I'll need to wait until that's merged and in a nuget package before updating this PR to match.

src/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -23,9 +23,12 @@ internal class MSBuildForwardingAppWithoutLogging
...
-                { "CscToolExe", GetRunCscPath() },
-                { "VbcToolExe", GetRunVbcPath() },
-                { "MSBuildSDKsPath", GetMSBuildSDKsPath() }
+                { "CscToolPath", GetRunToolPath() },
```
nguerrera: What is this used for?

khyperia: `CscToolExe` is not supposed to be the full path to csc, but rather just the filename. `CscToolPath` is supposed to be the full containing directory of csc.

src/redist/redist.csproj
nguerrera: I think I understand now. This is just to get files to nuget cache then we pull from there directly. I think this can cause issues with deps file content. Satellites at least are probably dropped. I need to pull this down and check loc impact.

src/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
khyperia: For reference, you can see both being [used here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets#L141), and [this test](https://github.com/Microsoft/msbuild/blob/06323ab096b9efc505c4266f3228a10b9efdac79/src/Utilities.UnitTests/ToolTask_Tests.cs#L366) that shows ToolExe should just be the filename.

nguerrera: Ok. I see.

src/redist/redist.csproj
khyperia: If we did not use `ExcludeAssets` here, the package would be pulled in as an MSBuild extension and override the compiler for building CLI. That's obviously not what we want, we just want to include the bits as content and splat them to the output directory.

nguerrera: Yep, that is what I understood in my second comment. I just need to inspect the differences to deps.json and I *think* it may cause a loc issue, but I'm not sure yet,

src/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
nguerrera: Thanks for fixing that!

khyperia:
Roslyn nuget is updated with that change, so I've added a new commit here that updates that.

Additionally, I have [work in progress](https://github.com/dotnet/roslyn/pull/21673) that allows the CLI to remove `CscToolExe`, which is required for overall progress on the decoupling to advance. I'll submit that PR here once this PR is in.

rainersigwald:
Looping @dotnet/msbuild in here: the changes with `DOTNET_HOST_PATH` are interesting and relate to https://github.com/Microsoft/msbuild/issues/720 and https://github.com/Microsoft/msbuild/issues/1669. If there's a generalized solution to the problem, we should expose or at least document it at the MSBuild layer.

jaredpar:
> If there's a generalized solution to the problem, we should expose or at least document it at the MSBuild layer.

I really think we need to get to a point where `ToolTaks` has the following method

```
bool ExecuteManagedTool(string toolPath, string arguments);
```

The host can control the implementation, via some mechanism, to either:

- Desktop: just forward to `base.ExecuteTool`
- CLI: forward to `base.ExecuteTool(dotnet, toolPath + " " + arguments)`
- Mono: forward to `base.ExecuteTool("mono", toolPath + " " + arguments)`

Essentially let the host control how a managed tool is executed by using the active runtime. Don't force the logic down into the specific tool tasks. I'd be much more comfortable with that solution than the Roslyn code base ever knowing about `%DOTNET_HOST_PATH%`

khyperia:
Merge conflicts resolved. Still waiting on review.

nguerrera:
I still need to check loc impact, but otherwise LGTM. I'll verify loc today. 

Thanks for improving this. I apologize for the long delay in reviewing, we've been flooded by things affecting bits that ship well before master.

livarcocc:

nguerrera:
OK, there is a localization problem. I verified that we no longer pick up the roslyn satellites because we use (csc|vbc).deps.json straight from the package instead of doing a publish on our side.The .NET Core runtime will not load satellite assemblies if they are not listed in the .deps.json.

This is only due to an ordering issue vs. two things that we need/want to happen:

1. Roslyn moves to the XLIFF OSS localization process, and takes over building its own satellite assemblies, and inserting them in to CLI. 
2. This change

Ideally, we'd just do (1) first. I'm still thinking about alternatives.

src/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs
```diff
@@ -78,20 +81,20 @@ private static string GetMSBuildSDKsPath()
...
+            var scriptExtension = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".cmd" : "";
+            return $"Run{compilerName}{scriptExtension}";
         }
 
-        private static string GetRunToolPath(string compilerName)
+        private static string GetDotnetPath()
         {
-            var scriptExtension = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".cmd" : ".sh";
-            return Path.Combine(AppContext.BaseDirectory, "Roslyn", $"Run{compilerName}{scriptExtension}");
+            return Process.GetCurrentProcess().MainModule.FileName;
```
eerhardt: There is an API for this called `new Muxer().MuxerPath`.  We should be consistent in how we find the `dotnet` executable.

tannergooding:
@nguerrera, I believe this fixes https://github.com/dotnet/sdk/issues/1615 (after talking with @khyperia).

After the merge conflict is resolved, is there anything further blocking this?

khyperia:
@tannergooding FYI, that issue is not the same one that you discussed with me (unless your comment here was after further thought process). It sounded like you were talking about an override of `CscToolPath`, not using an entirely separate CLI install. (This PR may still fix that issue, though)

khyperia:
Actually, this PR still does not contain my fix for `CscToolPath`. I just wrote that fix so long ago, I assumed it was this active PR, but apparently this PR is even older than that (the fix depends on this PR to be in first).

nguerrera:
Sorry about yet another long delay on this. I still don't have a quick solution for the satellites vs. deps.json identified above and it slipped my mind to get back to it until another thread pointing to this PR just hit my inbox. Let me pull this down and see if I can come up with something now.

nguerrera:
I've put together a solution which publishes the satellite dependencies independently and then merges in the resulting .deps.json. https://github.com/nguerrera/cli/commit/345a02e6f08c8b7cc159d4ac5c6ef16334672b9a I still need to review my code and test it in the morning.

nguerrera:
@khyperia is the intent for this to ship improvement to be used in the CLI release that ships with VS 15.5? Do the bits in/going in to 15.5 have the corresponding Roslyn changes?

* rebase onto release/15.5
* retarget PR to release/15.5
* set roslyn package version to match what is inserted in to VS 15.5

I'll then update the satellite assemblies to corresponding version and test with my satellite deps.json fixup on a localized setup. If that works, we can add that to your PR and finally merge.

nguerrera:
Note that the whole tool_roslyn_satellites.csproj in my change can go away when we move roslyn to XLIFF localization and building its own satellites, but there won't be time for that in 15.5.

khyperia:
> is the intent for this improvement to ship in the CLI release that will ship alognside with VS 15.5?

I'm honestly not sure. Jared is the one driving the schedule, he's back on Monday, I can loop him in then.

> Do the bits in/going in to 15.5 have the corresponding Roslyn changes?

Yes, the changes for this PR are in https://github.com/dotnet/roslyn/pull/21590 , and the `CscToolPath` changes are in this PR https://github.com/dotnet/roslyn/pull/21673 (iirc roslyn master is 15.5)

---

If we _are_ targeting 15.5, would you like me to append the `CscToolPath` fixes to this PR to speed things up? (That fix is necessary to implement the coreclr compiler server in the CLI - it would be awfully convenient to get that fix in earlier rather than later, as we can always patch in the compiler server implementation later without CLI changes, but not the other way around)

nguerrera:
It turns out that things don't work with 2.6/15.5 bits but not this change. See failures in #7734. So we need to take this in 15.5 indeed. 

@khyperia Can you please go ahead and rebase onto release/15.5?

agocke:
> is the intent for this improvement to ship in the CLI release that will ship alognside with VS 15.5?

I suspect not, but we'll check with Jared. The signing infrastructure issues probably preclude getting in before escrow.

nguerrera:
If not, then we need to fix the issue that was causing #7734 to fail with 2.6 but not this. We have to ship 2.6 in 15.5 cli somehow.

Get Outlook for iOS<https://aka.ms/o0ukef>

khyperia:
See also https://github.com/dotnet/roslyn/issues/22257 - the issue reporter said ["I can confirm that 2.6.0-beta1-62126-01 works."](https://github.com/dotnet/roslyn/issues/22257#issuecomment-332460241), so something weird is going on in https://github.com/dotnet/cli/pull/7734

ektrah:
@khyperia https://github.com/dotnet/cli/pull/7734 doesn't seem to reference the `Microsoft.NETCore.Compilers` package that I was talking about in https://github.com/dotnet/roslyn/issues/22257, but the `Microsoft.Net.Compilers.netcore` package. I don't know if that makes a difference.

khyperia:
@nguerrera Rebased onto release/15.5 and changed the PR target to that.

Obviously, existing comments are going to be broken, due to the rebase (took the opportunity to squash as well). Sorry about that.

khyperia:
Re-asking - If we are targeting 15.5, would you like me to append the `CscToolPath` fixes to this PR to speed things up? It'd be really great to get those fixes in ASAP, as it's a blocker for the compiler server on cli.

nguerrera:
Sure, you can append to this.

Get Outlook for iOS<https://aka.ms/o0ukef>

livarcocc:
@khyperia are you working on porting the CscToolPath change to this PR? We need to get this change in today.

khyperia:
Sorry, @livarcocc (was working on something else, didn't realize this was super high-pri).

Hopefully that cherry-pick succeeds CI, I didn't do local testing at all.

khyperia:
@nguerrera @livarcocc `Ubuntu16.04 x64 Debug Build` is failing with `502 (Bad Gateway)` when restoring `Microsoft.DotNet.Cli.Tests.sln`. Not sure if you just want to retest or what, I'll leave it up to you.

livarcocc:
These failed due to my get, retrying...

livarcocc:

khyperia:
Merging https://github.com/dotnet/cli/pull/7731 caused a merge conflict in `build/DependencyVersions.props`. Should I rebase this PR (and re-run all tests)?

jaredpar:
> We need to get this change in today.

Why the sudden urgency on this?

livarcocc:
Because this carries the roslyn that is inserted into 15.5 and this will be the CLI in 15.5. We don't really want them to mismatch, specially when the VS one will support more than the CLI one.

nguerrera:
@jaredpar, @khyperia FYI, https://github.com/dotnet/roslyn/issues/22478 is why just bumping the roslyn version without taking this PR (as in #7734) didn't work.

khyperia:
Is this change in a nuget package? (It's needed for my work on the cross-platform compiler server)

nguerrera:
@khyperia, it's not in a nuget package (because CLI doesn't ship as one), but you can grab a 15.5-preview build via installer or zip from https://github.com/dotnet/cli/tree/release/15.5#installers-and-binaries and it will have your changes.

khyperia:
I think what I was actually asking for is a version number with this change in it (and I got mixed up with how Roslyn obtains things). [Roslyn uses dotnet-install.ps1/sh](https://github.com/dotnet/roslyn/blob/master/build/scripts/build-utils.ps1#L177) so it'd be nice to see somehow which branches are what versions, or even just a list of versions that _exist_ (I was using [this](https://dotnet.myget.org/feed/dotnet-cli/package/nuget/Microsoft.DotNet.Cli.Utils) as a source of version numbers, but I can't seem to find one with this change)

nguerrera:
15.5.0-preview-007082 is latest 15.5.x and should have this change.

jcouv:
@nguerrera I was able to download the package for 15.5.0-preview-007082. 
Is there a corresponding windows installer? (How do I find it?)
I assume that the corresponding sdk string (for global.json) is "15.5.0". Is that right?

jcouv:
Nevermind, the link above has the 15.5.0 installer: https://github.com/dotnet/cli/tree/release/15.5#installers-and-binaries

Thanks! :-)

PR #7323, opened 2017-07-31
Add firstrun command

wli3:
An empty command to trigger first run experience

_dotnet firstrun_


wli3:
@livarcocc i will add ask mode template after we get ship room approval

PR #7353, opened 2017-08-02
Remove unneeded char[] allocations

davkean:
This was causing 0.2% of allocations in a design-time build trace from a customer.

image

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/EnvironmentProvider.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/EnvironmentProvider.cs
@@ -11,6 +11,8 @@ namespace Microsoft.DotNet.Cli.Utils
 {
     public class EnvironmentProvider : IEnvironmentProvider
     {
+        private static char[] s_pathSeparator = new char[] { Path.PathSeparator };
+        private static char[] s_quote = new char[] { '"' };
         private IEnumerable<string> _searchPaths;
         private IEnumerable<string> _executableExtensions;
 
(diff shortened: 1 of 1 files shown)


wli3:
@nguerrera you may be interested in this topic. Although there is evidence to prove it actually affect the performance and the change is relatively small, 0.2% don't meet the bar for me for the trade off to increase the obscurity of the code. This will also start open the gate of FUD on performance.

nguerrera:
@wil3 When going after the perf of a scenario involving as many components as a design-time build, every little bit counts. If each component author says each individual cheap fix doesn't meet the bar, then we can never get the aggregated benefit of all the cheap fixes, which can add up quickly. I think it's unfair to say that this opens a FUD gate. As long as future fixes along the same lines come from measurements along the same lines, there's no FUD to speak about. I'm fine holding contributors to providing measurements for such changes in the future as well. I don't see any bad precedent here at all.

wil3:
HI there, I think you meant @wli3 :-)

nguerrera:
@wil3 Yes. Sorry for the noise.

davkean:
Look at my last 20 or so PRs, we are getting bit by these paper cuts everywhere across lots of repos; 0.2% here, 0.4% there. Our GC time for opening Roslyn and wait for full intellisense is sitting around 16% of total CPU, it should below 10%. That's 50 seconds of blocking UI just GCing and cleaning up these things. This thing runs both in VS everytime we evaluate, kick off a design-time build and do a full build. We pay for these over and over and over again both in VS and MSBuild processes.

wli3:
I do like the approach with measurement included. And I agree that measurement is a good way to prevent FUD and end up write everything like it is .NET 2.0

PR #7357, opened 2017-08-02
Automate triggering of CLI security build.

ravimeda:
These changes allow automatic queuing of security build. 

`build/buildpipeline/security/Get-LatestVersion.ps1` is invoked by the root build (https://devdiv.visualstudio.com/DevDiv/_build/index?context=allDefinitions&path=%5CDotNet%5CSecurity&definitionId=6698&_a=completed) to retrieve commit SHA and package Id from latest.version. This info is passed onto the build leg i.e. `DotNet-CLI-Security-Windows.json`, which perform the scanning of assemblies and sources for security issues.

Refer (https://github.com/dotnet/core-eng/issues/1401). 

Post merge, I will setup a trigger to run security build once per day, and cleanup documentation at https://github.com/dotnet/core-eng/blob/master/Documentation/Project-Docs/security-builds.md

@livarcocc PTAL.

Cc @MichaelSimons

### DIFF
--- a/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
+++ b/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
@@ -255,7 +255,7 @@
         "scriptType": "filePath",
         "scriptName": "$(Build.SourcesDirectory)\\build\\buildpipeline\\security\\Get-LatestVersion.ps1",
         "arguments": "-Branch \"$(CodeBase)\"",
-        "workingFolder": "$(Build.SourcesDirectory)\\$(PB_Repo)\\build\\buildpipeline\\security\\",
+        "workingFolder": "",
         "failOnStandardError": "true"
       }
     },
--- a/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
+++ b/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
@@ -678,7 +678,7 @@
     "type": "TfsGit",
     "name": "DotNet-Cli-Trusted",
     "url": "https://devdiv.visualstudio.com/DevDiv/_git/DotNet-Cli-Trusted",
-    "defaultBranch": "refs/heads/sec_ext",
+    "defaultBranch": "refs/heads/master",
     "clean": "true",
     "checkoutSubmodules": false
   },
--- a/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
+++ b/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
@@ -238,6 +238,28 @@
       }
     },
     {
+      "environment": {},
+      "enabled": true,
+      "continueOnError": false,
+      "alwaysRun": false,
+      "displayName": "Get latest version info",
+      "timeoutInMinutes": 0,
+      "condition": "succeeded()",
+      "refName": "PowerShell23",
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "filePath",
+        "scriptName": "$(Build.SourcesDirectory)\\build\\buildpipeline\\security\\Get-LatestVersion.ps1",
+        "arguments": "-Branch \"$(CodeBase)\"",
+        "workingFolder": "$(Build.SourcesDirectory)\\$(PB_Repo)\\build\\buildpipeline\\security\\",
+        "failOnStandardError": "true"
+      }
+    },
+    {
       "enabled": true,
       "continueOnError": true,
       "alwaysRun": false,
--- a/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
+++ b/build/buildpipeline/security/DotNet-CLI-Security-Windows.json
@@ -286,8 +286,8 @@
         "softwareFolder": "$(Build.SourcesDirectory)\\security",
         "mpdFolder": "",
         "softwareName": "CLI",
-        "softwareVersionNum": "$(PB_BuildNumber)",
-        "softwareBuildNum": "$(PB_BuildNumber)",
+        "softwareVersionNum": "$(CliLatestPackageId)",
+        "softwareBuildNum": "$(CliLatestPackageId)",
         "modeType": "prerelease",
         "noCopySymbols": "false",
         "noCopyBinaries": "false",
@@ -318,9 +318,9 @@
       "inputs": {
         "scriptType": "inlineScript",
         "scriptName": "",
-        "arguments": "-SrcDir \"$(Build.SourcesDirectory)\" -git \"$(PB_Git)\"",
+        "arguments": "-sha \"$(CliLatestCommitSha)\" -git \"$(PB_Git)\"",
         "workingFolder": "$(Build.SourcesDirectory)",
-        "inlineScript": "param($SrcDir, $git)\n$secDir = Join-Path \"$SrcDir\" \"security\"\n$shaFile= Join-Path \"$secDir\" \"latest.version\"\n$sha = gc \"$shaFile\" -first 1\n\nif ([string]::IsNullOrWhiteSpace($sha))\n{ Write-Error \"Unable to determine latest commit SHA.\" }\n\nStart-Process \"$git\" -ArgumentList \"clean -df\" -Wait -Verbose -ErrorAction Stop\nStart-Process \"$git\" -ArgumentList \"checkout $sha\" -Wait -Verbose -ErrorAction Stop\nWrite-Host \"Checked out at $sha\"\n",
+        "inlineScript": "param($sha, $git)\n\nStart-Process \"$git\" -ArgumentList \"clean -df\" -Wait -Verbose -ErrorAction Stop\nStart-Process \"$git\" -ArgumentList \"checkout $sha\" -Wait -Verbose -ErrorAction Stop\nWrite-Host \"Checked out at $sha\"\n",
         "failOnStandardError": "true"
       }
     },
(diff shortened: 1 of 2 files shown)


ravimeda:
@nguerrera since Livar is OOF, would you PTAL.  This PR is a follow up of https://github.com/dotnet/cli/pull/7296

build/buildpipeline/security/DotNet-CLI-Security-Windows.json
```diff
@@ -678,7 +678,7 @@
     "type": "TfsGit",
     "name": "DotNet-Cli-Trusted",
     "url": "https://devdiv.visualstudio.com/DevDiv/_git/DotNet-Cli-Trusted",
-    "defaultBranch": "refs/heads/master",
+    "defaultBranch": "refs/heads/sec_ext",
```
ravimeda: `sec_ext` should be replaced with `master`.

ravimeda: Fixed through e3bca32

MichaelSimons:
@ravimeda, Does the `Get-LatestVersion.ps1` script have to be invoked by the pipebuild definition or can it be invoked from the `DotNet-CLI-Security-Windows` build leg?  It is generally a best practice to keep as much logic as possible out of the pipebuild definitions since they are not checked in.

ravimeda:
Good question, Michael.  The way I was thinking is, the root build invokes `Get-LatestVersion.ps1` once, and passes the info to build leg(s). One invocation and multiple uses.
But you have a valid point about having most of the logic in build legs i.e. the JSON file, which is versioned, and makes it easier to track changes. I agree. I'll update accordingly.

build/buildpipeline/security/Get-LatestVersion.ps1
```diff
@@ -0,0 +1,83 @@
...
+    $retryCount = 1
+    $oldEap = $ErrorActionPreference
+
+    while ($retryCount -le 3)
+    {
+        $ErrorActionPreference = "Stop"
+
+        try
+        {
+            if(Test-Path "$latestVersionFilePath")
```
MichaelSimons: What do you think about not writing the latest.version file to disk and just grabbing the content from Invoke-WebRequest.  It seems like it would simplify this logic a fair bit.

build/buildpipeline/security/Get-LatestVersion.ps1
```diff
@@ -0,0 +1,83 @@
...
+    [Parameter(Mandatory=$true)]
+    [string]$Branch,
+    [string]$Filename="latest.version",
+    [string]$UrlPrefix="https://dotnetcli.blob.core.windows.net/dotnet/Sdk"
+)
+
+$latestVersionUrl = "$UrlPrefix/$Branch/$Filename"
+$latestVersionFilePath = ".\latest.version"
+$env:CliLatestCommitSha = ""
+$env:CliLatestPackageId = ""
```
MichaelSimons: Why set environment variables?  These don't get used by the VSTS build agent do they?  I am trying to figure out if both an env var is required in addition to the `Write-Host "##vso[task.setvariable variable=CliLatestCommitSha;]$env:CliLatestCommitSha"` logic.

ravimeda: I was trying to mimic the pattern at https://devdiv.visualstudio.com/DevDiv/_git/DotNet-BuildPipeline?path=%2Fsrc%2FPipeBuild%2FPipeBuild%2FGetLatestCommit.ps1&version=GBmaster&_a=contents

I'll investigate if we need both (env + vso).

build/buildpipeline/security/Get-LatestVersion.ps1
ravimeda: Yes, something like this would return an array that contains SHA and package Id - `return (Invoke-WebRequest -Uri "$latestVersionUrl").Content.Split([Environment]::NewLine, [System.StringSplitOptions]::RemoveEmptyEntries)`

I'll run a test build to confirm.

ravimeda:
@MichaelSimons addressed the following through 0d837ca

1. `Get-LatestVersion` is invoked in build leg
2. `CliLatestCommitSha` and `CliLatestPackageId` are VSTS Task Variables and not environment variables
3. Extracting the latest info from the web request response directly as opposed to downloading and reading the content.

PTAL.

build/buildpipeline/security/Get-LatestVersion.ps1
```diff
@@ -0,0 +1,70 @@
...
+)
+
+function Get-VersionInfo
+{
+    Write-Host "Attempting to retrieve latest version info from $latestVersionUrl"
+    $retries = 3
+    $retryCount = 1
+    $oldEap = $ErrorActionPreference
+
+    while ($retryCount -le 3)
```
MichaelSimons: Seems like `$retries` should be used here.

build/buildpipeline/security/Get-LatestVersion.ps1
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $ErrorActionPreference = "Stop"
+
+        try
+        {
+            return (Invoke-WebRequest -Uri "$latestVersionUrl" -UseBasicParsing).Content.Split([Environment]::NewLine, [System.StringSplitOptions]::RemoveEmptyEntries)
```
MichaelSimons: <nit> This line is a little long, could it be broken into 2 lines?

build/buildpipeline/security/Get-LatestVersion.ps1
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        $retryCount++
+    }
+}
+
+$latestVersionUrl = "$UrlPrefix/$Branch/$Filename"
+$latestVersionContent = Get-VersionInfo
+
+if (-not [string]::IsNullOrWhiteSpace($latestVersionContent) -and $latestVersionContent.Length -eq 2)
```
MichaelSimons: The `[string]::IsNullOrWhiteSpace($latestVersionContent)` check seems out of place.  Isn't `$LatestVersionContent` a string array?

build/buildpipeline/security/DotNet-CLI-Security-Windows.json
```diff
@@ -238,6 +238,28 @@
...
+      "task": {
+        "id": "e213ff0f-5d5c-4791-802d-52ea3e7be1f1",
+        "versionSpec": "1.*",
+        "definitionType": "task"
+      },
+      "inputs": {
+        "scriptType": "filePath",
+        "scriptName": "$(Build.SourcesDirectory)\\build\\buildpipeline\\security\\Get-LatestVersion.ps1",
+        "arguments": "-Branch \"$(CodeBase)\"",
+        "workingFolder": "$(Build.SourcesDirectory)\\$(PB_Repo)\\build\\buildpipeline\\security\\",
```
MichaelSimons: Is the `workingFolder` necessary?  How is it used?

build/buildpipeline/security/Get-LatestVersion.ps1
ravimeda: Good catch! Will update.

build/buildpipeline/security/Get-LatestVersion.ps1
ravimeda: How about - 
`$content = (Invoke-WebRequest -Uri "$latestVersionUrl" -UseBasicParsing).Content`
`return $content.Split([Environment]::NewLine, [System.StringSplitOptions]::RemoveEmptyEntries)`

build/buildpipeline/security/Get-LatestVersion.ps1
ravimeda: This tricked me too. In the case where `Get-VersionInfo` hits an exception, `$latestVersionContent` would be null, and accessing an element in that empty array would throw an error. Hence, adding this null/empty check will display cleaner error.

build/buildpipeline/security/DotNet-CLI-Security-Windows.json
ravimeda: Not necessary when `scriptName` is an absolute path as in this case. I will leave it this way.

build/buildpipeline/security/Get-LatestVersion.ps1
MichaelSimons: Seems like a regular null check would be more appropriate.  Using [string]::IsNullOrWhiteSpace is confusing because the variable is an array.

build/buildpipeline/security/Get-LatestVersion.ps1
MichaelSimons: Like it

build/buildpipeline/security/DotNet-CLI-Security-Windows.json
MichaelSimons: The part that was confusing to me was that the paths are similar yet different.  IMO if it isn't necessary it should be removed as it makes you think there is some reason for it.

build/buildpipeline/security/Get-LatestVersion.ps1
ravimeda: Ok. I'll update as `if ($latestVersionContent -ne $null -and ...`

build/buildpipeline/security/DotNet-CLI-Security-Windows.json
ravimeda: Ok. I'll update as -

`"scriptName": "$(Build.SourcesDirectory)\\build\\buildpipeline\\security\\Get-LatestVersion.ps1",`

`"workingFolder": "",`

ravimeda:
Addressed the following through d6bd18e

1. Replaced `3` with `$retries` at line 30 of `Get-LatestVersion.ps1`
2. In the same file, simplified the null check as `if ($latestVersionContent -ne $null -and $latestVersionContent.Length -eq 2)`
3. Removed the value against `workingFolder` in the JSON.

PTAL

Thanks again.

ravimeda:
@nguerrera FYI I will merge as soon as the checks pass.

ravimeda:
@nguerrera please could you merge this PR. The two failures appear unrelated to these changes.

wli3:
@dotnet-bot test OpenSUSE42.1 x64 Debug Build
@dotnet-bot test Ubuntu16.10 x64 Debug Build

wli3:
requeueing these tests, they should be healthy

ravimeda:
Thanks all!

PR #7424, opened 2017-08-12
Ingest new version Aspnet store

wli3:
Change the url and name accordingly for the name naming convention

In new protocol, to insert new store to CLI, only 2 properties are required to change.

AspNetCoreRuntimePackageFolderName
AspNetCoreRuntimePackageVersion
And CLI will construct artifact URL completely rely on these 2 properties, in straight forward
https://dotnetcli.azureedge.net/dotnet/aspnetcore/store/$(AspNetCoreRuntimePackageFolderName)/FILE-$(AspNetCoreRuntimePackageFolderName).zip
And no more timestamp/no timestamp

There are also rename for the achieve to ex:

AspNetCorePackageStoreLib_x64-2.1.0-preview1-26609.wixlib
AspNetCorePackageStoreLib_x64-2.1.0-preview1-final.wixlib
AspNetCorePackageStoreLib_x86-2.1.0-preview1-26609.wixlib
AspNetCorePackageStoreLib_x86-2.1.0-preview1-final.wixlib
AspNetCoreRuntimePackageStore_x64-2.1.0-preview1-26609.exe
AspNetCoreRuntimePackageStore_x64-2.1.0-preview1-26609.msi
AspNetCoreRuntimePackageStore_x64-2.1.0-preview1-final.exe
AspNetCoreRuntimePackageStore_x64-2.1.0-preview1-final.msi
AspNetCoreRuntimePackageStore_x86-2.1.0-preview1-26609.exe
AspNetCoreRuntimePackageStore_x86-2.1.0-preview1-26609.msi
AspNetCoreRuntimePackageStore_x86-2.1.0-preview1-final.exe
AspNetCoreRuntimePackageStore_x86-2.1.0-preview1-final.msi

### DIFF
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -32,18 +32,9 @@
     <MicroBuildVersion>0.2.0</MicroBuildVersion>
     <SpaTemplateVersion>1.0.417</SpaTemplateVersion>
     <XliffTasksVersion>0.2.0-beta-000042</XliffTasksVersion>
-
-    <!-- This should either be timestamped or notimestamp as appropriate -->
-    <AspNetCoreRuntimePackageFlavor>notimestamp</AspNetCoreRuntimePackageFlavor>
-    <!--BranchName and AspNetCoreVersion will not always be the same-->
-    <AspNetCoreBranchName>2.0.0</AspNetCoreBranchName>
-    <AspNetCoreRelease>rtm</AspNetCoreRelease>
-    <AspNetCoreVersion>2.0.0</AspNetCoreVersion>
-    <AspNetCoreRuntimePackageTimestamp>26452</AspNetCoreRuntimePackageTimestamp>
-
     <AspNetCoreRuntimePackageBrandName>aspnetcore-store</AspNetCoreRuntimePackageBrandName>
-    <AspNetCoreVersionAndRelease>$(AspNetCoreVersion)-$(AspNetCoreRelease)</AspNetCoreVersionAndRelease>
-    <AspNetCoreRuntimeAzureblobStoresSubfolderName>$(AspNetCoreBranchName)-$(AspNetCoreRuntimePackageTimestamp)</AspNetCoreRuntimeAzureblobStoresSubfolderName>
+    <AspNetCoreRuntimePackageFolderName>dev-26623</AspNetCoreRuntimePackageFolderName>
+    <AspNetCoreRuntimePackageVersion>2.1.0-preview1-26623</AspNetCoreRuntimePackageVersion>
   </PropertyGroup>
 
   <!-- infrastructure and test only dependencies -->
--- a/build/Test.targets
+++ b/build/Test.targets
@@ -22,7 +22,7 @@
 
   <Target Name="Test"
           Condition=" '$(CLIBUILD_SKIP_TESTS)' != 'true' "
-          DependsOnTargets="BuildTests;">
+          DependsOnTargets="TestsForBuildItself;BuildTests;">
 
     <ItemGroup>
       <ProjectsToTest Include="build/test/RunTest.proj">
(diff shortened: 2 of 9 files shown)


wli3:
@joeloff

test/MsBuildScript.Tests/runtests.target
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="RunMsBuildScriptTests" 
+  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="../../dir.props" />
```
wli3: @nguerrera  @dsplaisted  I am experimenting "unit test" for msbuild script. So far it is pretty weak due to msbuild scope, do you like it or not? I can move this part

build/Test.targets
```diff
@@ -85,6 +85,11 @@
 
   </Target>
 
+  <Target Name="TestsForBuildItself">
+    <DotNetMSBuild Arguments="/v:diag $(RepoRoot)/test/MsBuildScript.Tests/runtests.target"
+                   ToolPath="$(Stage0Directory)" />
```
wli3: complete separate thread, no pollution

build/DependencyVersions.props
```diff
@@ -32,18 +32,9 @@
     <MicroBuildVersion>0.2.0</MicroBuildVersion>
     <SpaTemplateVersion>1.0.417</SpaTemplateVersion>
     <XliffTasksVersion>0.2.0-beta-000042</XliffTasksVersion>
-
-    <!-- This should either be timestamped or notimestamp as appropriate -->
-    <AspNetCoreRuntimePackageFlavor>notimestamp</AspNetCoreRuntimePackageFlavor>
```
joeloff: So with this change your branch will always build only a single specific failure of either timestamp or notimestamp, instead of building both versions?

wli3: We never build both versions. When time close to release, we will consume no timestamp version. However, both our own timestamp and no timetamp build (we call it suffix and no suffix)

test/MsBuildScript.Tests/runtests.target
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="RunMsBuildScriptTests"
```
nguerrera: nit: we shouldn't need toolsversion or xlmns for msbuild 15+, which we can rely on.

build/package/Installer.RPM.targets
```diff
@@ -182,8 +138,52 @@
...
+      <AfterInstallHostScriptDestinationFile>$(RpmLayoutScripts)/$(AfterInstallHostScriptName)</AfterInstallHostScriptDestinationFile>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <EndToEndTestProject>$(RepoRoot)/test/EndToEnd/EndToEnd.csproj</EndToEndTestProject>
+    </PropertyGroup>
+  </Target>
+
   <Target Name="DownloadAspNetCoreRuntimeRpmInstaller"
-        DependsOnTargets="SetupDebProps">
```
dsplaisted: Is the SetupDebProps target used anymore?  If not, should it be deleted?

test/MsBuildScript.Tests/runtests.target
dsplaisted: The `<?xml` tag, and the `ToolsVersion` and `xmlns` attributes on the Project element aren't necessary.

test/MsBuildScript.Tests/runtests.target
dsplaisted: I think this test looks good if there is some risk that the dependency files don't exist in blob storage.

I would adjust the naming though.  I would use names that describe more specifically what it does, instead of calling these generic "MSBuild script tests".  So something like `CheckForBlobStorageDependencies` instead of `TestsForBuildItself`.  Also I'm not sure if there's much reason to follow the pattern of having a separate project file which is invoked via the `DotNetMSBuild` task.  I would probably remove that indirection.

build/package/Installer.RPM.targets
wli3: That is actually a bug, no actual affect but wrong

test/MsBuildScript.Tests/runtests.target
wli3: fixed

wli3: fixed

PR #6302, opened 2017-04-11
Skip versions repo publish if auth not defined

dagood:
We need to disable `UpdateVersionsRepo` when publishing to only blob storage. This adds a condition so that when `GITHUB_PASSWORD` isn't set, versions repo update is skipped.

/cc @crummel @dleeapho

PR #7436, opened 2017-08-14
Skip versions repo publish if auth not defined

dagood:
See https://github.com/dotnet/cli/pull/6302 (same change on rel/1.0.1):

> We need to disable `UpdateVersionsRepo` when publishing to only blob storage. This adds a condition so that when `GITHUB_PASSWORD` isn't set, versions repo update is skipped.

I'm submitting this to `master` for the future, but what's important now is the port to `release/2.0.0` for servicing.

@nguerrera @Chrisboh

### DIFF
--- a/build/publish/FinishBuild.targets
+++ b/build/publish/FinishBuild.targets
@@ -34,6 +34,7 @@
 
     <UpdateVersionsRepo BranchName="$(BranchName)"
                         PackagesDirectory="$(PackagesDirectory)"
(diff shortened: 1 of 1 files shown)


dagood:
Thanks @nguerrera! What's the path to get the port merged into release/2.0.0?

dagood:
@dotnet-bot test CentOS7.1 x64 Debug Build (known flakiness)

PR #7437, opened 2017-08-14
[release/2.0.0] Skip versions repo publish if auth not defined

dagood:
(cherry picked from commit b4c1a809b883927e6103569feb8c4ac976d6131f)

Ports https://github.com/dotnet/cli/pull/7436 to release/2.0.0.

> We need to disable UpdateVersionsRepo when publishing to only blob storage. This adds a condition so that when GITHUB_PASSWORD isn't set, versions repo update is skipped.

@nguerrera @Chrisboh

### DIFF
--- a/build/publish/FinishBuild.targets
+++ b/build/publish/FinishBuild.targets
@@ -34,6 +34,7 @@
 
     <UpdateVersionsRepo BranchName="$(BranchName)"
                         PackagesDirectory="$(PackagesDirectory)"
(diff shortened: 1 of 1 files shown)


dagood:
@nguerrera Is this ok to merge?

nguerrera:
@MattGertz FYI, infrastructure only change for 2.x servicing.

nguerrera:
@dagood Yes.

dagood:
Thanks! Looks like I don't have permission to hit the merge button, though.

PR #7467, opened 2017-08-17
Test performance impact of v:diag

TyOverby:


TyOverby:
CC: https://github.com/dotnet/cli/issues/7349

TyOverby:
Size goes from 9MB to 40kb, but no change in build speed.

PR #7514, opened 2017-08-24
Pin Stage0 to a version with .NET Core 2.0 instead of 2.1

dsplaisted:
This will avoid https://github.com/dotnet/corefx/issues/23496, which https://github.com/dotnet/cli/pull/7506 will work around

PR #7515, opened 2017-08-24
Add -SkipNonVersionedFiles to dotnet-install.{sh,ps1}

natemcmaster:
In some of our build scripts, we want to install new versions of the .NET Core runtime, but this can cause write errors if dotnet.exe is currently executing. There was already an internal non-clobber option. This exposes a command-line switch for that option so we can avoid clobbering dotnet.exe.


nguerrera:
@johnbeisner

nguerrera:
CI failure will be fixed by #7514

johnbeisner:
@natemcmaster   merging...

PR #7527, opened 2017-08-25
Update SDK version 2.0.2-vspre-20170825-1

wli3:
@MattGertz  FYI this is a SDK version flow to CLI to align nuget version

### DIFF
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -12,7 +12,7 @@
     <!-- We'll usually want to keep these versions in sync, but we may want to diverge in some
          cases, so use separate properties but derive one from the other unless we want to
          explicitly use different versions. -->
-    <CLI_NETSDK_Version>2.0.2-vspre-20170817-4</CLI_NETSDK_Version>
+    <CLI_NETSDK_Version>2.0.2-vspre-20170825-1</CLI_NETSDK_Version>
     <CLI_MSBuildExtensions_Version>$(CLI_NETSDK_Version)</CLI_MSBuildExtensions_Version>
 
     <CLI_NuGet_Version>4.4.0-preview1-4434</CLI_NuGet_Version>


build/DependencyVersions.props
```diff
@@ -17,7 +17,7 @@
 
     <CLI_NuGet_Version>4.4.0-preview1-4434</CLI_NuGet_Version>
     <CLI_NETStandardLibraryNETFrameworkVersion>2.0.0-preview3-25514-04</CLI_NETStandardLibraryNETFrameworkVersion>
-    <CLI_WEBSDK_Version>2.0.0-rel-20170629-588</CLI_WEBSDK_Version>
```
wli3: SDK version looks odd, but most of the time the version looks like https://github.com/dotnet/cli/commit/53f6fe597f305913bfa691fb5355b3b06326e300

build/DependencyVersions.props
```diff
@@ -17,7 +17,7 @@
...
+    <CLI_WEBSDK_Version>2.0.2-vspre-20170825-1</CLI_WEBSDK_Version>
```
nguerrera: This is not right. You should be updating CLI_NETSDK_Version

wli3: So only stage0, but the sdk used in produce will not be updated?

build/DependencyVersions.props
nguerrera: See below. That's because this is a different component. It's what puts the contents of https://github.com/aspnet/websdk in to the CLI not https://github.com/dotnet/sdk

build/DependencyVersions.props
nguerrera: No, it will update the product. This is updating a different component in the product. The version numbering scheme is different.

wli3:

PR #7552, opened 2017-08-29
Sign additional files

nguerrera:
Some files that get their signature invalidated by crossgen were missing
from the Signing.proj list.

@dsplaisted for review

@MattGertz for approval -- needed for 15.4 P2.


MattGertz:
Please update the VSO bugs appropriately if you haven't already, including a VSO PR, and I'll loop in DoRon.

nguerrera:
@MattGertz Would it be OK to merge this to GitHub proactively so that we can validate the signing fix? (I need it to go through the real signing infrastructure to be sure.) We would still have the option to insert the previous build to get just the nuget update.

MattGertz:
@nguerrera Absolutely.

PR #7564, opened 2017-08-30
Push the 'VS.Redist.Common.Net.Core.*.nupkg' to blob storage

ghost:
The Microsoft.DotNet...nupkg packages are being pushed to blob storage; we should include the VS.Redist.Common.Net.Core...nupkg package as well.


nguerrera (approved):

nguerrera (changes requested):
Please retarget this at the release/2.0-vs branch

Issue #7367, opened 2017-08-03
Wget does not resolve 'specific_version' when out_path is undefined


johnbeisner:
This is a proposed fix to 'dotnet-install.sh'

IM thread:
John Beisner 3:42 PM: 
so the basic problem is that one side of the conditional for wget does not dump the stdout
John Beisner 3:42 PM: 
ok I understand
Tomas Weinfurt 3:43 PM: 
I think the code to parse the version expects function to return it as string on stdout
Tomas Weinfurt 3:43 PM: 
get_latest_version_info()
John Beisner 3:43 PM: 
I believe that to be true
Tomas Weinfurt 3:43 PM: 
ends with download $version_file_url
    return $?
}
John Beisner 3:43 PM: 
yup
Tomas Weinfurt 3:44 PM: 
now the second part:
Tomas Weinfurt 3:44 PM: 
wget -q --tries 10 -O - $remote_path || failed=true
Tomas Weinfurt 3:44 PM: 
when wget fails it should return error code/errno
Tomas Weinfurt 3:44 PM: 
failed than should be set to true
Tomas Weinfurt 3:44 PM: 
and than it is explicitly checked
Tomas Weinfurt 3:45 PM: 
so I think downloaded problem would be detected properly
John Beisner 3:46 PM: 
so for your case, the download succeeded
Tomas Weinfurt 3:46 PM: 
yes.
John Beisner 3:46 PM: 
but the harvesting of the version failed
Tomas Weinfurt 3:46 PM: 
but it saves output to latest.version file
Tomas Weinfurt 3:47 PM: 
and that leaves the variable empty
Tomas Weinfurt 3:47 PM: 
++ get_version_from_version_info
++ eval say_verbose '"Calling:' '${yellow:-}${FUNCNAME[0]}' '${green:-}$*${normal:-}"'
+++ say_verbose 'Calling: get_version_from_version_info '
+++ '[' false = true ']'
++ cat
++ tail -n 1
++ return 0
++ return 0
+ specific_version=
+ say_verbose specific_version=

Issue #6995, opened 2017-06-23
Consider replacing GetTempFileName with GetTempPath and GetRandomFileName


jbagga:
`dotnet-install.ps1` uses [GetTempFileName](https://github.com/dotnet/cli/blob/master/scripts/obtain/dotnet-install.ps1#L480), [however](https://msdn.microsoft.com/en-us/library/system.io.path.gettempfilename(v=vs.110).aspx) 

> The GetTempFileName method will raise an IOException if it is used to create more than 65535 files without deleting previous temporary files.

See https://github.com/aspnet/Coherence-Signed/issues/510 for more information.

cc @Eilon

PR #7572, opened 2017-08-31
Install-script changes

ghost:
Miscellaneous script fixes...

Addresses:
https://github.com/dotnet/cli/issues/6995
    scripts/obtain/dotnet-install.ps1 - lines 483 & 492
https://github.com/dotnet/cli/issues/7367
    scripts/obtain/dotnet-install.sh  - line 605

Other minor changes...

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -264,7 +264,7 @@ function Get-Specific-Version-From-Version([string]$AzureFeed, [string]$Channel,
     }
 }
 
-function Get-Download-Link([string]$AzureFeed, [string]$Channel, [string]$SpecificVersion, [string]$CLIArchitecture) {
+function Get-Download-Link([string]$AzureFeed, [string]$SpecificVersion, [string]$CLIArchitecture) {
     Say-Invocation $MyInvocation
 
     if ($SharedRuntime) {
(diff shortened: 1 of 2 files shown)


scripts/obtain/dotnet-install.sh
```diff
@@ -583,7 +584,7 @@ downloadcurl() {
 
     local failed=false
     if [ -z "$out_path" ]; then
-        curl --retry 10 -sSL -f --create-dirs $remote_path || failed=true
+        curl --retry 10 -sSL -f --create-dirs -o $remote_path || failed=true
```
wfurt: I don't think this is right. -o $remore_path will try to output of $remote_path. 
As far as I could tell, curl path was always right as curl would output to stdout by default.

scripts/obtain/dotnet-install.sh
```diff
@@ -601,7 +602,7 @@ downloadwget() {
 
     local failed=false
     if [ -z "$out_path" ]; then
-        wget -q --tries 10 $remote_path || failed=true
+        wget -q --tries 10 -O $remote_path || failed=true
```
wfurt: I think the change should be '-O - ' 
('-' stands for stdout)

scripts/obtain/dotnet-install.ps1
```diff
@@ -449,14 +449,14 @@ function Prepend-Sdk-InstallRoot-To-Path([string]$InstallRoot, [string]$BinFolde
 
 $CLIArchitecture = Get-CLIArchitecture-From-Architecture $Architecture
 $SpecificVersion = Get-Specific-Version-From-Version -AzureFeed $AzureFeed -Channel $Channel -Version $Version
-$DownloadLink = Get-Download-Link -AzureFeed $AzureFeed -Channel $Channel -SpecificVersion $SpecificVersion -CLIArchitecture $CLIArchitecture
```
wli3: So there is no channel anymore?

johnbeisner: Not for this call - the channel has been used to determine the version, therefore is no longer needed at this point.

wli3 (approved):

PR #7573, opened 2017-08-31
Update shared framework dependency versions for 1.1.x

nguerrera:


nguerrera:
> why not parameterize CoreCLRVersion and JitVersion in the DependencyVersion.props too

I opportunistically chose the versions that were already defined instead of pasting, I do plan to clean them up, but want to get the build green again first.

nguerrera:
Ubuntu 16.10 failure can be ignored. It is no longer supported and needs to be removed.

nguerrera:
@dotnet-bot Test Ubuntu x64 Release Build please

PR #7594, opened 2017-09-05
Avoid importing incompatible .props files for signing build

dsplaisted:
Together with #7593, this should fix the signing build, by making sure that all files imported during the signing build are compatible with MSBuild 14.


dsplaisted:
I've included the changes from #7593 and #7595 in this PR so we can merge them all as one

dsplaisted:

PR #7659, opened 2017-09-14
Checkin corefx editorconfig

wli3:
https://github.com/dotnet/corefx/blob/master/.editorconfig

The suggestion from VS can be much better. And I removed the rule of static type need s_ in the beginning


wli3:

PR #7694, opened 2017-09-21
Fix issue and minor cleanup of the RHEL 6 support

janvorli:
This change fixes extraction of the unprocessed args and also implements
few minor changes that @eerhardt has asked for, like renaming the new argument
and changing usage of HostRid to Rid in build/BuildDefaults.props.


janvorli:

PR #7702, opened 2017-09-22
Fix issues caused by adding RHEL6 support

janvorli:
There were two issues. One was that additional arguments that are not
recognized by the build.sh script were incorrectly passed as a single
argument in single-quotes to msbuild.
The second issue was that when LD_LIBRARY_PATH was not set, the
dotnet-install.sh script was failing due to its using of the "set -u"
option that prevents usage of unset variables.

Fix https://github.com/dotnet/cli/issues/7698


johnbeisner:
@SteveL-MSFT
Addresses: https://github.com/dotnet/cli/issues/7698

PR #7708, opened 2017-09-26
Update ASP.NET Core to 2.1.0-preview1-27131

aspnetbot:


SteveSandersonMS:
FYI: @mlorbetske @seancpeters @livarcocc @phenning @Eilon

As requested, the ASP.NET build process will now automatically submit PRs like this after successful coherence builds. If there is already a pending PR of this kind, the process will update it in place rather than creating a new one.

We're hoping that these PRs will be merged regularly so that the flow of CLI builds reflects the latest ASP.NET builds.

Shortly I'll also make these PRs update the ASP.NET templates version injected into the CLI (but first, we need to change the CLI repo so it does get the ASP.NET templates from external packages rather than from dotnet/templating).

If anyone thinks anything is missing/wrong with this, please let me know!

SteveSandersonMS:
It looks like the build is failing because it's trying to get the ASP.NET build not from our drop share, but instead from some blob store location such as [this](https://dotnetcli.azureedge.net/dotnet/aspnetcore/store/dev-27131/aspnetcore-store-2.1.0-preview1-27131-linux-x64.tar.gz).

@mlorbetske, @livarcocc: how do the ASP.NET builds get on to that blob store? Is this a manual process or is there some automation we can tie into? Or would it be better to change the CLI build process to fetch the ASP.NET builds from the ASP.NET CI drop share?

mlorbetske:
The SetupCrossgen CI build pushes the runtime store to blob.

SteveSandersonMS:
@mlorbetske Thanks. I'll try having the template build depend on SetupCrossgen so the runtime store should be updated before it issues the PR.

SteveSandersonMS:
OK, this is blocked while code signing is down. I'll close this PR for now, and @aspnetbot will automatically create a new one next time we get a signed build out of SetupCrossgen.

PR #7711, opened 2017-09-26
Add Dockerfile for RHEL 6 builds

janvorli:
This change enables building CLI for RHEL 6 in docker container. I also needed to fix one more place where the parameters were being parsed in a way that was causing a problem if the argument value string matched a substring of another argument. Like ` --runtime-id rhel.6-x64 --docker rhel.6` in the RHEL 6 build.

The Dockerfile builds and installs more recent git that in necessary for building CLI. I will later move it to the original docker image that the image we are building is based on. It takes about a minute only though, so it is not causing a problem.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -24,7 +24,7 @@ if [ -z "$HOME" ]; then
     mkdir -p $HOME
 fi
 
-args=( "$@" )
+args=
 
 while [[ $# > 0 ]]; do
     lowerI="$(echo $1 | awk '{print tolower($0)}')"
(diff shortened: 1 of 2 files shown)


janvorli:

build.sh
```diff
@@ -24,39 +24,31 @@ if [ -z "$HOME" ]; then
...
-
 dockerbuild()
 {
     BUILD_COMMAND=/opt/code/run-build.sh $DIR/scripts/dockerrun.sh --non-interactive "$@"
 }
 
 # Check if we need to build in docker
 if [ ! -z "$BUILD_IN_DOCKER" ]; then
-    dockerbuild "${args[@]}"
+    dockerbuild $args
```
wli3: this is doing the same thing as before but with less code?

scripts/docker/rhel.6/Dockerfile
```diff
@@ -0,0 +1,42 @@
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# Dockerfile that creates a container suitable to build dotnet-cli
+FROM microsoft/dotnet-buildtools-prereqs:centos-6-783abde-20171304101322
```
wli3: based on centos? Not rhel 6?

janvorli: Yes, CentOS 6 and RHEL 6 are equivalent w.r.t. to the versions of packages available for these.

PR #7735, opened 2017-09-29
Update FSharp compiler

brettfo:


nguerrera:
The package isn't being found. Do the other digits need to change or are we missing a feed?

PR #7751, opened 2017-10-03
Update satellite assemblies and fix up roslyn satellite handling

nguerrera:
1. Publish satellites with a separate project and merge the resulting .deps.json into (csc|vbc).deps.json

2. Move the build tasks satellites to the correct location next to main assembly (one level higher than bincore\)

3. Update roslyn satellite package to 2.6.x to get matching assembly version and updated strings

4. Update F# satellite package

Marked *no merge* because the CliDeps.Satellites being ingested isn't real-signed.


nguerrera:
@livarcocc @dsplaisted We'll need to amend cd68283 with the right build number when we have the signed build

nguerrera:
Removing *NO MERGE*. We will send follow up PR with signed bits.

PR #7756, opened 2017-10-03
Quote variable expansions in dotnet-install.sh

### LINKED ISSUE #5281: dotnet-install.sh can't install to path containing spaces
Non-blocking bug, but I noticed this when making an upstream shell script handle spaces properly.

## Steps to reproduce
Run `scripts/obtain/dotnet-install.sh --install-dir a\ b\ c`

## Expected  behavior
The .NET CLI is installed into a folder in the current directory called "a b c".

## Actual behavior
The .NET CLI is installed into a folder in the current directory called "a".

Adding `--verbose` shows a step where the path is trimmed:
```
dotnet-install: Calling: resolve_installation_path a b c
dotnet-install: install_root=a
```

## Environment data
OSX bash `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin15)`
Ubuntu Trusty bash `GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)`

benknoble:
Fix #5281 to prevent paths with spaces from causing problems due to word
splitting.

Some of them aren't likely to have issues, but safety first.

See also dotnet/roslyn#22320 for something similar

### DIFF
--- a/scripts/obtain/dotnet-install.sh
+++ b/scripts/obtain/dotnet-install.sh
@@ -240,7 +240,7 @@ to_lowercase() {
 remove_trailing_slash() {
     #eval $invocation
     
-    local input=${1:-}
+    local input="${1:-}"
     echo "${input%/}"
     return 0
 }
@@ -250,7 +250,7 @@ remove_trailing_slash() {
 remove_beginning_slash() {
     #eval $invocation
     
-    local input=${1:-}
+    local input="${1:-}"
     echo "${input#/}"
     return 0
 }
@@ -267,8 +267,8 @@ combine_paths() {
         return 1
     fi
     
-    local root_path=$(remove_trailing_slash $1)
-    local child_path=$(remove_beginning_slash ${2:-})
+    local root_path="$(remove_trailing_slash "$1")"
+    local child_path="$(remove_beginning_slash "${2:-}")"
     say_verbose "combine_paths: root_path=$root_path"
     say_verbose "combine_paths: child_path=$child_path"
     echo "$root_path/$child_path"
@@ -288,10 +288,10 @@ get_machine_architecture() {
 get_normalized_architecture_from_architecture() {
     eval $invocation
     
-    local architecture=$(to_lowercase $1)
-    case $architecture in
+    local architecture="$(to_lowercase "$1")"
+    case "$architecture" in
         \<auto\>)
-            echo "$(get_normalized_architecture_from_architecture $(get_machine_architecture))"
+            echo "$(get_normalized_architecture_from_architecture "$(get_machine_architecture)")"
             return 0
             ;;
         amd64|x64)
@@ -338,11 +338,11 @@ get_commit_hash_from_version_info() {
 is_dotnet_package_installed() {
     eval $invocation
     
-    local install_root=$1
-    local relative_path_to_package=$2
-    local specific_version=${3//[$'\t\r\n']}
+    local install_root="$1"
+    local relative_path_to_package="$2"
+    local specific_version="${3//[$'\t\r\n']}"
     
-    local dotnet_package_path=$(combine_paths $(combine_paths $install_root $relative_path_to_package) $specific_version)
+    local dotnet_package_path="$(combine_paths "$(combine_paths "$install_root" "$relative_path_to_package")" "$specific_version")"
     say_verbose "is_dotnet_package_installed: dotnet_package_path=$dotnet_package_path"
     
     if [ -d "$dotnet_package_path" ]; then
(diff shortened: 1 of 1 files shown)


scripts/obtain/dotnet-install.sh
```diff
@@ -534,17 +534,18 @@ get_absolute_path() {
...
+    local root_path="$(remove_trailing_slash "$1")"
+    local out_path="$(remove_trailing_slash "$2")"
+    local override="$3"
+    local override_switch="$(if [ "$override" = false ]; then printf -- "-n"; fi)"
     
     cat | uniq | while read -r file_path; do
-        local path=$(remove_beginning_slash ${file_path#$root_path})
-        local target=$out_path/$path
+        local path="$(remove_beginning_slash "${file_path#$root_path}")"
+        local target="$out_path"/"$path"
```
johnbeisner: Why quote each part? Doesn't this work in a way that's expected: local target="$out_path/$path"?

scripts/obtain/dotnet-install.sh
```diff
@@ -534,17 +534,18 @@ get_absolute_path() {
...
         if [ "$override" = true ] || (! ([ -d "$target" ] || [ -e "$target" ])); then
-            mkdir -p $out_path/$(dirname $path)
-            cp -R $override_switch $root_path/$path $target
+            mkdir -p "$out_path"/"$(dirname "$path")"
+            say "cp -R $override_switch $root_path/$path $target"
+            cp -R "$override_switch" "$root_path"/"$path" "$target"
```
johnbeisner: This shouldn't be quoted: "$override_switch"

scripts/obtain/dotnet-install.sh
benknoble: It does--I think I did that because I was searching around for $word patterns and quote-surrounding them. Can definitely be changed.

scripts/obtain/dotnet-install.sh
benknoble: Yes, discovered that just now

scripts/obtain/dotnet-install.sh
```diff
@@ -765,7 +766,7 @@ do
             override_non_versioned_files=false
             ;;
         -?|--?|-h|--help|-[Hh]elp)
-            script_name="$(basename $0)"
+            script_name="$(basename "$0")"
```
johnbeisner: Is this necessary: "$0"? What is the breaking scenario?

scripts/obtain/dotnet-install.sh
```diff
@@ -534,17 +534,18 @@ get_absolute_path() {
...
     eval $invocation
 
-    local root_path=$(remove_trailing_slash $1)
-    local out_path=$(remove_trailing_slash $2)
-    local override=$3
-    local override_switch=$(if [ "$override" = false ]; then printf -- "-n"; fi)
+    local root_path="$(remove_trailing_slash "$1")"
+    local out_path="$(remove_trailing_slash "$2")"
+    local override="$3"
+    local override_switch="$(if [ "$override" = false ]; then printf -- "-n"; fi)"
```
johnbeisner: This line looked fine as it was; the resultant "-n" was quoted, I think that's enough...

scripts/obtain/dotnet-install.sh
benknoble: Hypothetically:
```
$ basename music/imagine dragons
imagine
$ basename "music/imagine dragons"
imagine dragons
$ basename a dir/file
a
$ basename "a dir/file"
file
```

If the script name **never** contains spaces, then there wouldn't be any issues. If the full path to the script **never** contains spaces, then there wouldn't be any issues. The goal was to fix potential space-containing paths breaking scripts.

johnbeisner: So it's the path that could potentially contains spaces; okay - good.

scripts/obtain/dotnet-install.sh
benknoble: I agree, but I figured presenting quoted "$()" as a good style might also be warranted. Would you prefer the quotes here removed?

johnbeisner: Yes, please.

scripts/obtain/dotnet-install.sh
johnbeisner: Yes, please.

benknoble:
@johnbeisner Looks like it passes.

johnbeisner:
@benknoble 
Thank you for your contribution.

PR #7768, opened 2017-10-05
Append to existing PATH in build env scripts/cli-test-env.bat

nguerrera:
We were clobbering the path down to just one directory


nguerrera:
(Disregard last comment, scripts aren't ported to dev/release/2.0.0 yet)

PR #7795, opened 2017-10-09
Remove 15.5 from 2.0

nguerrera:
There were incorrect merges from release/15.5 to release/2.0.0. This reverts them.


nguerrera:

A test is checking that we can do something in under 1 second and it took 1.3 seconds. Jenkins machine probably running slowly. That's not a good test pattern.

PR #7797, opened 2017-10-10
Fix RHEL 6 Dockerfile

janvorli:
The new base docker image sets the LD_LIBRARY_PATH to /usr/local/lib
and the yum is not compatible with the special CURL version we build.
This change fixes that by clearing the LD_LIBRARY_PATH for the yum
command invocation.

### DIFF
--- a/scripts/docker/rhel.6/Dockerfile
+++ b/scripts/docker/rhel.6/Dockerfile
@@ -6,7 +6,9 @@
 # Dockerfile that creates a container suitable to build dotnet-cli
 FROM microsoft/dotnet-buildtools-prereqs:centos-6-d485f41-20173404063424
 
-RUN yum -q -y install sudo
+# yum doesn't work with the special curl version we have in the base docker image,
+# so we remove /usr/local/lib from the library path for this command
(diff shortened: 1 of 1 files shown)


janvorli:
@johnbeisner I am planning to try to find out which feature is our CURL version missing and the yum needs and fix it in the base docker image. But for now, this fix works just fine.

scripts/docker/rhel.6/Dockerfile
```diff
@@ -6,7 +6,9 @@
 # Dockerfile that creates a container suitable to build dotnet-cli
 FROM microsoft/dotnet-buildtools-prereqs:centos-6-d485f41-20173404063424
 
-RUN yum -q -y install sudo
+# yum doesn't work with the special curl version we have in the base docker image,
+# so we remove /usr/local/lib from the library path for this command
+RUN LD_LIBRARY_PATH= yum -q -y install sudo
```
johnbeisner: Is the extra space after the '=' a problem here? In a Windows command shell, right or wrong, the extra space would be preserved.

janvorli: The space after the `=` means we are assigning empty string to the variable.

PR #7807, opened 2017-10-11
Update CentOS 6 base docker image

janvorli:
This update fixes problem with git not supporting https protocol (which
wasn't affecting cli, but the other repos). The image now has a build of
CURL that supports file:// protocol and so I am removing the workaround
from the Dockerfile.

### DIFF
--- a/scripts/docker/rhel.6/Dockerfile
+++ b/scripts/docker/rhel.6/Dockerfile
@@ -4,11 +4,11 @@
 #
 
 # Dockerfile that creates a container suitable to build dotnet-cli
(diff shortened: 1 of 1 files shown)


johnbeisner:
@janvorli 
CRITICAL:yum.cli:Config Error: Error accessing file for config file:///etc/yum.conf
The command '/bin/sh -c yum -q -y install sudo' returned a non-zero code: 1

janvorli:
Oops, I am sorry, I have forgotten to update the base image in the Dockerfile. Let me fix that.

PR #7814, opened 2017-10-12
MSBuild 15.5.154

AndyGerlicher:
Contains Double Evaluation Fix: https://github.com/Microsoft/msbuild/pull/2595

RPS with NuGet not complete. To take this change you will need NuGet 4527 or greater.


AndyGerlicher:
@livarcocc This has been merged into VS.

PR #7818, opened 2017-10-12
Separating the URL construction methods: dotnet-install.sh

ghost:
Separating the 'legacy' URL construction from the 'current' URL construction methods and logic.

Why?
RedHat6 was not supported in the legacy time space, therefore the script should not attempt resolve to a RedHat6 distro specific URL when constructing the legacy download link. In that way, we will not infer that 1.0 supported RedHat6; et al. The 2 logical methods for URL construction should not share execution paths.

Addresses: https://github.com/dotnet/cli/issues/7752


johnbeisner:
This PR is running the 'master' Jenkins legs, not the release/2.0.0 - clearly incorrect - replacing this PR

PR #7890, opened 2017-10-23
update fsharp compiler for preview release

KevinRansom:
Update dotnet cli to latest F# compiler.


KevinRansom:
@livarcocc what do we need to do to ensure this is in preview 3?

KevinRansom:
@livarcocc here you go mate.

PR #7891, opened 2017-10-23
MSBuild 15.5.165

AndyGerlicher:
Approved by Shiproom. PR pending Val build pass:
https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/90779?_a=overview

Will update here when that passes.


AndyGerlicher:
This is merged into VS d15rel.

PR #7967, opened 2017-11-06
Changes in the Transport feed DLL requires our response

ghost:
Changes in the Transport feed DLL requires our response; change DLL version and base URL.


jcagme (approved):
LGTM

PR #7976, opened 2017-11-07
Fixing the 'PublishNupkgToTransportFeed' target

ghost:
Isolating the 'PublishNupkgToTransportFeed' into as few files as possible; calling the DLL directly as the 'dotnet-cli-build.dll' no longer pulls this referenced assembly into its bin directory.

### DIFF
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -44,5 +44,6 @@
   <PropertyGroup>
     <VersionToolsVersion>1.0.27-prerelease-01723-01</VersionToolsVersion>
     <DotnetDebToolVersion>2.0.0-preview2-25331-01</DotnetDebToolVersion>
+    <BuildTasksFeedDllVersion>1.0.0-prerelease-02203-01</BuildTasksFeedDllVersion>
   </PropertyGroup>
 </Project>
--- a/build/publish/PublishNupkgToTransportFeed.targets
+++ b/build/publish/PublishNupkgToTransportFeed.targets
@@ -6,6 +6,12 @@
   </ItemGroup>
 
   <PropertyGroup>
+    <BuildTasksFeedDll>$(NuGetPackagesDir)\microsoft.dotnet.build.tasks.feed\$(BuildTasksFeedDllVersion)\build\netstandard1.5\Microsoft.DotNet.Build.Tasks.Feed.dll</BuildTasksFeedDll>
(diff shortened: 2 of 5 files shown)


build/publish/PublishNupkgToTransportFeed.targets
```diff
@@ -6,6 +6,12 @@
   </ItemGroup>
 
   <PropertyGroup>
+    <BuildTasksFeedDll>$(NuGetPackagesDir)\microsoft.dotnet.build.tasks.feed\$(BuildTasksFeedDllVersion)\build\netstandard1.5\Microsoft.DotNet.Build.Tasks.Feed.dll</BuildTasksFeedDll>
```
livarcocc: Why did we switch to look at the nuget cache versus the dotnet-cli-build output?

build/DependencyVersions.props
```diff
@@ -44,5 +44,6 @@
   <PropertyGroup>
     <VersionToolsVersion>1.0.27-prerelease-01723-01</VersionToolsVersion>
     <DotnetDebToolVersion>2.0.0-preview2-25331-01</DotnetDebToolVersion>
+    <BuildTasksFeedDllVersion>1.0.0-prerelease-02203-01</BuildTasksFeedDllVersion>
```
livarcocc: Can we call this BuildTasksTransportFeedDllVersion?

build/publish/PublishNupkgToTransportFeed.targets
livarcocc: BuildTasksTransportFeedDll?

jcagme (approved):
LGTM

build/publish/PublishNupkgToTransportFeed.targets
```diff
@@ -6,6 +6,12 @@
...
+  </PropertyGroup>
+
+  <UsingTask TaskName="PushToBlobFeed" AssemblyFile="$(BuildTasksFeedDll)" />
```
livarcocc: Why did we move this from the Task.targets file?

johnbeisner:
The 'lib' directory is missing in the microsoft.dotnet.build.tasks.feed NuGet package. This prevents the DLL from being copied to the 'bin' directory of the referring project - instead of the above workaround, we will ask that the 'lib' directory be reinstated.

PR #7981, opened 2017-11-08
Correcting 'PUBLISH_TO_AZURE_BLOB' and 'PUBLISH_NUPKG_TO_AZURE_BLOB'

ghost:
Correcting 'PUBLISH_TO_AZURE_BLOB' and 'PUBLISH_NUPKG_TO_AZURE_BLOB' Booleans - 'false' should not be conveyed by the empty string.

To be clear the conditionals are: “true”/anything other than “true”

### DIFF
--- a/build/Publish.targets
+++ b/build/Publish.targets
@@ -10,7 +10,7 @@
 
   <!-- PUBLISH_TO_AZURE_BLOB env variable set by CI -->
   <Target Name="Publish"
-          Condition=" '$(PUBLISH_TO_AZURE_BLOB)' != '' "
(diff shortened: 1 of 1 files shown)


livarcocc:
I would recommend making this change in release/2.0.0

johnbeisner:
@livarcocc 
The Publish.targets is different enough because master has an 'AllLinuxDistros' leg - I will prepare a separate PR for release/2.0.0

johnbeisner:
@livarcocc 
The Publish.targets is different enough because master has an 'AllLinuxDistros' leg - I will prepare a separate PR for release/2.0.0

wli3 (approved):

PR #7984, opened 2017-11-09
Changes in the Transport feed DLL requires our response; release/2.0.0

ghost:
Changes in the Transport feed DLL requires our response; update Transport feed DLL version and base URL.

Defining "BuildTasksFeedDll" & relevent 'UsingTask' no longer needed as PublishNupkgToTransportFeed.targets uses the imported Microsoft.DotNet.Build.Tasks.Feed.targets to locate the Feed DLL

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -52,7 +52,7 @@
   <Import Project="build/Compile.targets" />
   <Import Project="build/Package.targets" />
   <Import Project="build/Test.targets" />
-  <Import Project="build/Publish.targets" />
+  <Import Condition=" '$(GeneratePropsFile)' != 'true' " Project="build/Publish.targets" />
 
   <Import Project="build/InitRepo.targets" />
 </Project>
--- a/build.proj
+++ b/build.proj
@@ -52,7 +52,7 @@
   <Import Project="build/Compile.targets" />
   <Import Project="build/Package.targets" />
   <Import Project="build/Test.targets" />
-  <Import Condition=" '$(GeneratePropsFile)' != 'true' " Project="build/Publish.targets" />
+  <Import Project="build/Publish.targets" />
 
   <Import Project="build/InitRepo.targets" />
 </Project>
--- a/build/publish/PublishNupkgToTransportFeed.targets
+++ b/build/publish/PublishNupkgToTransportFeed.targets
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(NuGetPackagesDir)/microsoft.dotnet.build.tasks.feed/$(BuildTasksFeedToolVersion)/build/Microsoft.DotNet.Build.Tasks.Feed.targets" />
-  
+  <PropertyGroup>
+    <BuildTasksFeedToolTargetsFileName>"$(NuGetPackagesDir)/microsoft.dotnet.build.tasks.feed/$(BuildTasksFeedToolVersion)/build/Microsoft.DotNet.Build.Tasks.Feed.targets"</BuildTasksFeedToolTargetsFileName>
+  </PropertyGroup>
+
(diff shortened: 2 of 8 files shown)


build.proj
```diff
@@ -52,7 +52,7 @@
   <Import Project="build/Compile.targets" />
   <Import Project="build/Package.targets" />
   <Import Project="build/Test.targets" />
-  <Import Project="build/Publish.targets" />
+  <Import Condition=" '$(GeneratePropsFile)' != 'true' " Project="build/Publish.targets" />
```
weshaggard: Instead of this I would suggest doing the condition in publish.targets on the existence of the targets file you are trying to import there.

build_projects/dotnet-cli-build/dotnet-cli-build.csproj
```diff
@@ -29,6 +29,6 @@
     <PackageReference Include="Microsoft.Build.Framework" Version="$(CLI_MSBuild_Version)" />
     <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="$(PlatformAbstractionsVersion)" />
     <PackageReference Include="Microsoft.DotNet.VersionTools" Version="$(VersionToolsVersion)" />
-    <PackageReference Include="Microsoft.DotNet.Build.Tasks.Feed" Version="1.0.0-prerelease-01929-02" />
+    <PackageReference Include="Microsoft.DotNet.Build.Tasks.Feed" Version="$(BuildTasksFeedToolVersion)" />
```
weshaggard: Don't you still need to do the restore in another context?

johnbeisner: Line 109 runs some prep work with including a restore: https://github.com/dotnet/cli/blob/08e594b8605f1e60ed17735ec96dba3ea6936ad3/run-build.ps1#L109

We just have to exclude the importation of 'build/Publish.targets' during this call.

build.proj
johnbeisner: Good idea; done...

build/publish/PublishNupkgToTransportFeed.targets
```diff
@@ -1,28 +1,27 @@
...
+  <Import Condition="Exists('$(BuildTasksFeedToolTargetsFileName)')" Project="$(BuildTasksFeedToolTargetsFileName)" />
+
   <ItemGroup>
     <NupkgsForPublishing Include="$(PackagesDirectory)/Microsoft.DotNet.Cli.Utils.*.nupkg" />
     <NupkgsForPublishing Include="$(PackagesDirectory)/Microsoft.DotNet.MSBuildSdkResolver.*.nupkg" />
   </ItemGroup>
 
   <PropertyGroup>
-    <RelativePath>packages</RelativePath>
+    <ExpectedFeedUrl>https://dotnetfeed.blob.core.windows.net/$(TransportFeedContainerName)/index.json</ExpectedFeedUrl>
```
livarcocc: Should this be configurable? Like, for MSRCs? I would probably add a condition here to only set this if the property is empty already. This way we could overwrite this in build definitions if necessary.

weshaggard: Yes it will need to be configurable.

johnbeisner: The "TransportFeedContainerName" is configurable; for testing I've replaced it with "jbeisner".
See: https://github.com/johnbeisner/cli/blob/PublishToTFMethodUpdate/build/AzureInfo.props#L15

"TransportFeedContainerName" has a default of "dotnet-core" and only sets this if the property is empty; however if we change "TRANSPORTFEED_STORAGE_CONTAINER" in build launch we can publish to an 'MSRC' container.

dsplaisted (approved):

build/publish/PublishNupkgToTransportFeed.targets
mmitche: I believe the entire URL will need to be configurable.

johnbeisner: I've made the "ExpectedFeedUrl" fully configurable.

PR #8058, opened 2017-11-18
WIP Add Shim maker and Env:PATH handling for Global tools

wli3:
THIS PR WON'T BUILD

It relates to https://github.com/dotnet/cli/pull/8035 and I need to rebase that change to make it build after it is merged. Since it won't build anyway, I kept the important part.

This is another 2 component. Create a shim for the command and handing the PATH. There are still one left, which is the installCommand. It is basically just call Shim maker and package obtainer


wli3:
#8062 #8063 #8064 #8065

PR #8090, opened 2017-11-23
Update dependencies and fix cli update dependencies build

safern:
Updated the sdk download to be from release/2.0.0 channel and ran the script locally. 

cc: @livarcocc @dagood


wli3:
@safern build failed

PR #8104, opened 2017-11-27
Making the 'RestoreSources' configurable: CLI

ghost:


test/NuGet.config
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
livarcocc: Why is this here? I imagine that the tests projects should use the same RestoreSources as the source projects. Otherwise, we will end up with packages outside the build feed.

TestAssets/NuGet.Config
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
livarcocc: Same here, not sure if this is correct. That means that when we set the PackageSources from the build from sources, the TestAssets are not going to respect it.

johnbeisner:
Replaced by: https://github.com/dotnet/cli/pull/8123

PR #8113, opened 2017-11-28
WIP Add install tool command

wli3:
WIP because I am still doing end to end test. I've tested on mac.

* compose all the parts
* Fix on obtain for better end to end experience

### DIFF
--- a/src/dotnet/commands/dotnet-install/InstallCommand.cs
+++ b/src/dotnet/commands/dotnet-install/InstallCommand.cs
@@ -1,0 +1,137 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Cli.CommandLine;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Configurer;
+using Microsoft.DotNet.ShellShimMaker;
+using Microsoft.DotNet.ToolPackageObtainer;
+using Microsoft.Extensions.EnvironmentAbstractions;
+using NuGet.Build.Tasks;
+using NuGet.Configuration;
+
+namespace Microsoft.DotNet.Cli
+{
+    public class InstallCommand
+    {
+        public static int Run(string[] args)
+        {
+            CommandLine.Parser parser = Parser.Instance;
+            ParseResult result = parser.ParseFrom("dotnet install", args);
+
+            result.ShowHelpOrErrorIfAppropriate();
+
+            AppliedOption parseResult = result["dotnet"]["install"]["tool"];
+
+            var packageId = parseResult.Arguments.Single();
+            var packageVersion = parseResult.ValueOrDefault<string>("version");
+
+            FilePath? configFile;
+
+            var configFilePath = parseResult.ValueOrDefault<string>("configfile");
+            configFile = configFilePath != null
+                ? new FilePath(configFilePath)
+                : GetConfigFileProbingCurrentDirectory();
+
+            var framework = parseResult.ValueOrDefault<string>("framework");
+
+            var executablePackagePath = new DirectoryPath(new CliFolderPathCalculator().ExecutablePackagesPath);
+
+            var toolConfigurationAndExecutableDirectory = ObtainPackage(
+                packageId,
+                packageVersion,
+                configFile,
+                framework,
+                executablePackagePath);
+
+            DirectoryPath executable = toolConfigurationAndExecutableDirectory
+                .ExecutableDirectory
+                .WithSubDirectories(
+                    toolConfigurationAndExecutableDirectory
+                        .Configuration
+                        .ToolAssemblyEntryPoint);
+
+            var shellShimMaker = new ShellShimMaker.ShellShimMaker(executablePackagePath.Value);
+            var commandName = toolConfigurationAndExecutableDirectory.Configuration.CommandName;
+            shellShimMaker.EnsureCommandNameUniqueness(commandName);
+
+            shellShimMaker.CreateShim(
+                executable.Value,
+                commandName);
+
+            EnvironmentPathFactory
+                .CreateEnvironmentPathInstruction()
+                .PrintAddPathInstructionIfPathDoesNotExist();
+
+            Reporter.Output.WriteLine(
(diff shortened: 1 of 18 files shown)


src/dotnet/commands/dotnet-install/InstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+                .CreateEnvironmentPathInstruction()
+                .PrintAddPathInstructionIfPathDoesNotExist();
+
+            Reporter.Output.WriteLine(
+                $"{Environment.NewLine}The installation succeeded. If there is no other instruction. You can type the following command in shell directly to invoke: {commandName}");
+
+            return 0;
+        }
+
+        private static FilePath? GetConfigFileProbingCurrentDirectory()
```
wli3: @emgarten Am I doing it right?

src/dotnet/dotnet.csproj
```diff
@@ -68,9 +68,10 @@
     <PackageReference Include="Microsoft.TemplateEngine.Utils" Version="$(TemplateEngineVersion)" />
     <PackageReference Include="Microsoft.DotNet.Archive" Version="$(DotNetCliArchiverVersion)" />
     <PackageReference Include="XliffTasks" Version="$(XliffTasksVersion)" PrivateAssets="All" />
+    <PackageReference Include="NuGet.Build.Tasks" Version="$(CLI_NuGet_Version)" />
```
wli3: Walk around for not Nuget.Config prob folder setting

src/dotnet/commands/dotnet-install/InstallCommand.cs
emgarten: I would use Settings.LoadDefaultSettings instead of RestoreSettingsUtils. It is less likely to have changes in the future and is more direct.

```cs
Configuration.Settings.LoadDefaultSettings(
                    Directory.GetCurrentDirectory(),
                    configFileName: null,
                    machineWideSettings: new XPlatMachineWideSetting())
```

src/dotnet/commands/dotnet-install/InstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+            // Pending NuGet change
+
+            var settings = RestoreSettingsUtils.ReadSettings(
+                solutionDirectory: Directory
+                    .GetCurrentDirectory(), // current directory instead of fake project as start
+                restoreDirectory: Directory.GetCurrentDirectory(), // current directory instead of fake project as start
+                restoreConfigFile: null, // no explicit config file that why we prob it
+                machineWideSettings: new Lazy<IMachineWideSettings>(() => new XPlatMachineWideSetting()));
+
+            var nugetconfigfilePath = SettingsUtility.GetConfigFilePaths(settings).FirstOrDefault(); // it is ordered
```
emgarten: What is the reason for getting only the first nuget.config file? Typically the entire hierarchy is used through ISettings

wli3: The problem is I cannot pass ISettings down. The interface for restore is the temp project. I need to find a way to serialize ISettings into the temp project or passing it via MSBuild props. However, restore target takes only one config file as parameter, so I will pick the first one(it will be the config file in current directory, or something without explicit configfile path passed in anyway) or null (equal to no explicit --configfile).

src/dotnet/commands/dotnet-install/InstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+            shellShimMaker.CreateShim(
+                executable.Value,
+                commandName);
+
+            EnvironmentPathFactory
+                .CreateEnvironmentPathInstruction()
+                .PrintAddPathInstructionIfPathDoesNotExist();
+
+            Reporter.Output.WriteLine(
+                $"{Environment.NewLine}The installation succeeded. If there is no other instruction. You can type the following command in shell directly to invoke: {commandName}");
```
wli3: @KathleenDollard customer facing string -- this is shown after the installation. If tools path does not exist on env:path. It will still show, but after the instruction. It will be something like:

tools path does not exist on, please run echo.....

The installation succeeded. If there is no other instruction. You can type the following command in shell directly to invoke demo

src/dotnet/commands/dotnet-install/InstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+                    packageId: packageId,
+                    packageVersion: packageVersion,
+                    nugetconfig: configFile,
+                    targetframework: framework);
+            }
+            catch (PackageObtainException ex)
+            {
+                throw new GracefulException(
+                    message:
+                    $"Install failed. Failed to download package:{Environment.NewLine}{ex.Message}",
```
wli3: @KathleenDollard customer facing string, failed to restore package

src/dotnet/commands/dotnet-install/InstallCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+                    innerException: ex);
+            }
+            catch (ToolConfigurationException ex)
+            {
+                throw new GracefulException(
+                    message:
+                    $"Install failed. Failed to apply the configuration of the tool:{Environment.NewLine}{ex.Message}",
```
wli3: @KathleenDollard customer facing string. failed to restore package (package config is wrong)

src/dotnet/commands/dotnet-install/InstallCommandParser.cs
```diff
@@ -0,0 +1,39 @@
...
+                Accept.NoArguments(), CommonOptions.HelpOption(), InstallTool());
+        }
+
+        private static Command InstallTool()
+        {
+            return Create.Command("tool",
+                "Install tool",
+                Accept.ExactlyOneArgument(o => "packageId")
+                    .With(name: "packageId",
+                        description: "Package Id in NuGet"),
```
wli3: @KathleenDollard all the `name` and `description` are customer facing string. They are shown in --help

wli3:
https://github.com/dotnet/cli/issues/8066

src/dotnet/commands/dotnet-install/InstallCommand.cs
emgarten: How about setting `<RestoreSolutionDirectory>` in the project? 

Passing a single config as if it were the only config used is likely to cause problems. Often times users have nuget.org defined in the their user wide config and other settings in the working directory config file. If you were to select just the working directory one it would miss sources.

wli3:
Will repoen later

PR #8119, opened 2017-11-28
Add parameter to dotnet-install to support pulling from private blob feeds

natemcmaster:
Typically this will only be used for internal customers when publishing dotnet-cli assets to private Azure blob feeds.

skip ci please

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -203,10 +203,9 @@ function GetHTTPResponse([Uri] $Uri)
             # Default timeout for HttpClient is 100s.  For a 50 MB download this assumes 500 KB/s average, any less will time out
             # 10 minutes allows it to work over much slower connections.
             $HttpClient.Timeout = New-TimeSpan -Minutes 10
-            $ActualUri = if (($Uri -like "$AzureFeed*") -or ($Uri -like "$UncachedFeed*")) { "${Uri}${FeedCredential}" } else { $Uri }
-            $Response = $HttpClient.GetAsync($ActualUri).Result
+            $Response = $HttpClient.GetAsync("${Uri}${FeedCredential}").Result
             if (($Response -eq $null) -or (-not ($Response.IsSuccessStatusCode))) {
-                 # The feed credential is potential sensitive info. Do not log ActualUri to console output.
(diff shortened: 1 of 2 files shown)


johnbeisner:
@natemcmaster 
All changes of this type are merged into CLI:release/2.00 - please rebase to release/2.0.0

scripts/obtain/dotnet-install.ps1
```diff
@@ -200,8 +204,10 @@ function GetHTTPResponse([Uri] $Uri)
             # Default timeout for HttpClient is 100s.  For a 50 MB download this assumes 500 KB/s average, any less will time out
             # 10 minutes allows it to work over much slower connections.
             $HttpClient.Timeout = New-TimeSpan -Minutes 10
-            $Response = $HttpClient.GetAsync($Uri).Result
+            $ActualUri = if (($Uri -like "$AzureFeed*") -or ($Uri -like "$UncachedFeed*")) { "${Uri}${FeedCredential}" } else { $Uri }
```
johnbeisner: Why are we checking the $Uri here? Why would we not append the FeedCredential?

natemcmaster: Today, it is always appended. My thinking was that may not always be true. Since `function GetHTTPResponse` is a general-purpose method, I wanted to avoid the possibility of appending a query string to a URL that shouldn't have the token added.

But I'd be okay to drop this if you're not worried about it.

natemcmaster:
Rebased on release/2.0.0

scripts/obtain/dotnet-install.ps1
johnbeisner: I think we drop the check now; even though it 'may not always be true'. When that situation arises, then the dev should put in an appropriate check.

natemcmaster: Sounds good. Updated the PR to drop this check.

johnbeisner:
@dotnet-bot test please

johnbeisner:

johnbeisner:

johnbeisner:
@natemcmaster 
Are you merging this or is it proper that I merge this?

natemcmaster:
@johnbeisner you'll have to do it. I don't have push permissions to this repo. Thanks!

PR #8123, opened 2017-11-28
Writing a NuGet.config as part of the preparatory steps

ghost:


build/NugetConfigFile.targets
```diff
@@ -0,0 +1,58 @@
...
+      <NugetConfigHeader>
+        &lt;?xml version=&quot;1.0&quot; encoding=&quot;utf-8&quot;?&gt;
+        &lt;configuration&gt;
+        &lt;packageSources&gt;
+        &lt;!--To inherit the global NuGet package sources remove the &lt;clear/&gt; line below --&gt;
+        &lt;clear /&gt;
+      </NugetConfigHeader>
+
+      <NugetConfigFeedsPrivate>
+        &lt;add key=&quot;PrivateBlobFeed&quot; value=&quot;$(ExternalRestoreSources)&quot; /&gt;
```
dsplaisted: Can there be more than one external restore source?  If so, then this probably won't work.  If not, then the variable seems named incorrectly.

run-build.ps1
```diff
@@ -106,7 +106,7 @@ if ($NoBuild)
 }
 else
 {
-    dotnet msbuild build.proj /p:Architecture=$Architecture /p:GeneratePropsFile=true /t:WriteDynamicPropsToStaticPropsFiles
+    dotnet msbuild build.proj /p:Architecture=$Architecture /p:GeneratePropsFile=true /t:WriteDynamicPropsToStaticPropsFiles $ExtraParameters
```
dsplaisted: I think it was probably intentional that the extra parameters weren't passed to this invocation.  I'd suggest asking Livar about this.

build/NugetConfigFile.targets
```diff
@@ -0,0 +1,58 @@
...
+      </NugetConfigFeedsPrivate>
+      
+      <NugetConfigFeedsNormal>
+        &lt;add key=&quot;BlobFeed&quot; value=&quot;https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json&quot; /&gt;
```
dsplaisted: You can probably make this prettier using `CDATA`, so you don't have to escape all the brackets.  See an example here: https://stackoverflow.com/a/2784200

mmitche (changes requested):
Seems reasonable, assuming the multiple restore sources issue is fixed.

johnbeisner:
@mmitche
@dsplaisted 
This should be finished now, please confirm.

mmitche (approved):

run-build.ps1
livarcocc: I don't remember why. But why do we need the extra parameters here? For the Extra restore sources? Should we make it explicit then?

run-build.ps1
```diff
@@ -80,16 +80,16 @@ $env:VSTEST_TRACE_BUILD=1
 # install a stage0
 $dotnetInstallPath = Join-Path $RepoRoot "scripts\obtain\dotnet-install.ps1"
 
-Write-Output "$dotnetInstallPath -Channel ""release/2.0.0"" -InstallDir $env:DOTNET_INSTALL_DIR -Architecture ""$Architecture"""
-Invoke-Expression "$dotnetInstallPath -Channel ""release/2.0.0"" -InstallDir $env:DOTNET_INSTALL_DIR -Architecture ""$Architecture"""
+Write-Output "$dotnetInstallPath -InstallDir $env:DOTNET_INSTALL_DIR -Architecture ""$Architecture"" -Channel ""release/2.0.0"""
```
livarcocc: Why the change here?

run-build.ps1
johnbeisner: The extra parameters needed in the 'GeneratePropsFile' step are: /p:ExternalRestoreSources & /p:DotNetBuildOffline - they are needed to control the content of the NuGet.config
When reviewing the properties now available to the 'GeneratePropsFile' step; I do not believe that the behavior of this pre-step would change.

run-build.ps1
johnbeisner: It's easier to read.

livarcocc (approved):

livarcocc (approved):

PR #8152, opened 2017-12-01
Dependency Uptake: CLI - part 2

ghost:
**Dependency Uptake: download and import a package version props file.**

_dir.props_ 
Import the package version props file only if it exists.

_Follow-up: run-build.sh_ 
The first pass will need the extra args [same as run-build.ps1].

Please note: 
A rename of dependency version properties in alignment with sample file at: https://dotnetfeed.blob.core.windows.net/orchestrated/20171129-03/orchestration-metadata/PackageVersions.props to follow - 12/1/2017

### DIFF
--- a/build/BundledSdks.props
+++ b/build/BundledSdks.props
@@ -1,10 +1,10 @@
 <Project>
   <ItemGroup>
     <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(CLI_NuGet_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk" Version="$(CLI_NETSDK_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk.Web" Version="$(CLI_WEBSDK_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk.Publish" Version="$(CLI_WEBSDK_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk.Web.ProjectSystem" Version="$(CLI_WEBSDK_Version)" />
+    <BundledSdk Include="Microsoft.NET.Sdk" Version="$(MicrosoftNETSdkPackageVersion)" />
+    <BundledSdk Include="Microsoft.NET.Sdk.Web" Version="$(MicrosoftNETSdkWebPackageVersion)" />
+    <BundledSdk Include="Microsoft.NET.Sdk.Publish" Version="$(MicrosoftNETSdkPublishPackageVersion)" />
+    <BundledSdk Include="Microsoft.NET.Sdk.Web.ProjectSystem" Version="$(MicrosoftNETSdkWebProjectSystemPackageVersion)" />
     <BundledSdk Include="FSharp.NET.Sdk" Version="1.0.4-bundled-0100" />
   </ItemGroup>
 </Project>
--- a/build/BundledSdks.props
+++ b/build/BundledSdks.props
@@ -1,6 +1,6 @@
 <Project>
   <ItemGroup>
-    <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(CLI_NuGet_Version)" />
+    <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(NuGetBuildTasksPackageVersion)" />
     <BundledSdk Include="Microsoft.NET.Sdk" Version="$(MicrosoftNETSdkPackageVersion)" />
     <BundledSdk Include="Microsoft.NET.Sdk.Web" Version="$(MicrosoftNETSdkWebPackageVersion)" />
     <BundledSdk Include="Microsoft.NET.Sdk.Publish" Version="$(MicrosoftNETSdkPublishPackageVersion)" />
--- a/build/BundledSdks.props
+++ b/build/BundledSdks.props
@@ -1,6 +1,6 @@
 <Project>
   <ItemGroup>
-    <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(NuGetBuildTasksPackageVersion)" />
+    <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(NuGetBuildTasksPackPackageVersion)" />
     <BundledSdk Include="Microsoft.NET.Sdk" Version="$(MicrosoftNETSdkPackageVersion)" />
     <BundledSdk Include="Microsoft.NET.Sdk.Web" Version="$(MicrosoftNETSdkWebPackageVersion)" />
     <BundledSdk Include="Microsoft.NET.Sdk.Publish" Version="$(MicrosoftNETSdkPublishPackageVersion)" />
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -29,6 +29,7 @@
     <MicrosoftExtensionsDependencyModelPackageVersion>2.0.0</MicrosoftExtensionsDependencyModelPackageVersion>
     <MicrosoftDotNetCliCommandLinePackageVersion>0.1.1-alpha-167</MicrosoftDotNetCliCommandLinePackageVersion>
     <MicrosoftDotNetProjectJsonMigrationPackageVersion>1.2.1-alpha-002133</MicrosoftDotNetProjectJsonMigrationPackageVersion>
+    <MicrosoftDotNetToolsMigrateCommandPackageVersion>$(MicrosoftDotNetProjectJsonMigrationPackageVersion)</MicrosoftDotNetToolsMigrateCommandPackageVersion>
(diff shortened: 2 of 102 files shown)


build/DownloadedPackageVersions.targets
```diff
@@ -0,0 +1,11 @@
+<Project ToolsVersion="15.0">
+  <Target Name="DownloadPackageVersionsProps">
+    <PropertyGroup>
+      <PB_AssetRootAccessTokenSuffixParam Condition=" '$(PB_AssetRootAccessTokenSuffix)' != '' ">?$(PB_AssetRootAccessTokenSuffix)</PB_AssetRootAccessTokenSuffixParam>
+    </PropertyGroup>
+
+    <DownloadFile Uri="$(PB_PackageVersionPropsUrl)?$(PB_AssetRootAccessTokenSuffixParam)"
```
livarcocc: If a token is not provided, your url will end up with a question mark at the end of it.

build/DownloadedPackageVersions.targets
```diff
@@ -0,0 +1,11 @@
+<Project ToolsVersion="15.0">
+  <Target Name="DownloadPackageVersionsProps">
+    <PropertyGroup>
+      <PB_AssetRootAccessTokenSuffixParam Condition=" '$(PB_AssetRootAccessTokenSuffix)' != '' ">?$(PB_AssetRootAccessTokenSuffix)</PB_AssetRootAccessTokenSuffixParam>
```
livarcocc: Are these access tokens going to be different? Why not use the same one we are using for the blob storage?

build/DownloadedPackageVersions.targets
johnbeisner: Fixed...

build/DownloadedPackageVersions.targets
johnbeisner: Fixed...

build/DependencyVersions.props
```diff
@@ -1,34 +1,40 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CLI_SharedFrameworkVersion>2.0.3</CLI_SharedFrameworkVersion>
-    <CLI_MSBuild_Version>15.4.8</CLI_MSBuild_Version>
-    <CLI_Roslyn_Version>2.3.2-beta1-61921-05</CLI_Roslyn_Version>
+    <MicrosoftNETCoreAppPackageVersion>2.0.3</MicrosoftNETCoreAppPackageVersion>
```
livarcocc: Shouldn't these be set conditionally? How will the downloaded version file overwrite this?

build/DependencyVersions.props
```diff
@@ -1,34 +1,40 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <CLI_SharedFrameworkVersion>2.0.3</CLI_SharedFrameworkVersion>
```
livarcocc: This change will require an update on dotnet/versions, so that the right things are automatically update here.

src/tool_nuget/tool_nuget.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.App" Version="$(CLI_SharedFrameworkVersion)" />
+    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppPackageVersion)" />
     <PackageReference Include="NuGet.CommandLine.XPlat" Version="$(CLI_NuGet_Version)" />
```
livarcocc: NuGet version not being update to the same new format?

build/DependencyVersions.props
johnbeisner: The conditionals are in the downloads version prop file. Example: 
`<SystemMemoryPackageVersion Condition="$(SystemMemoryPackageVersion.Contains('-'))">4.5.0-preview1-25930-01</SystemMemoryPackageVersion>`

src/tool_nuget/tool_nuget.csproj
johnbeisner: NuGet is not included in the versions that can be overridden; maybe in the future?

build/DependencyVersions.props
livarcocc: I don't follow then. I thought the goal of this file was that if it was passed in, then its values would win. The way it is coded up right now, if will only win if the original value of the dependency does not contain a suffix in its version. Is this really the correct semantics for this?

mmitche: @weshaggard @dagood Did we decide to remove the conditionals from the imported dependency props and spread them into the repo-specific version props?  I think we did, to allow for more flexibility on the repo side.

weshaggard: Yes we are removing the conditions in the generated package list, so repo's need to correctly condition or name their properties.

nguerrera: Wasn't the idea to import the downloaded file after our property group? I don't get why we need conditions.

dir.props
```diff
@@ -25,6 +25,7 @@
   <Import Project="build/BundledSdks.props" />
   <Import Project="build/BundledTemplates.props" />
   <Import Project="build/DependencyVersions.props" />
+  <Import Condition=" Exists('$(DownloadedPackageVersionsProps)') " Project="$(DownloadedPackageVersionsProps)" />
```
nguerrera: Yes, here it comes after DependencyVersion.props, so DependencyVErsions.props doesn't need conditions.

I would prefer if this import were put inside DependencyVersions.props because I suspect there will be scenarios where we need to override the flow. E.g. sometimes you need to reference a prior stable version and not the version from current orchestrated build.

nguerrera: I would also prefer OrchestratedPackageVersionsProps over DownloadedPackageVersionsProps as the name.  I'd rather the name say what importing the file does rather than how the file was acquired (implementation detail).

build/BundledSdks.props
```diff
@@ -1,10 +1,10 @@
 <Project>
   <ItemGroup>
-    <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(CLI_NuGet_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk" Version="$(CLI_NETSDK_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk.Web" Version="$(CLI_WEBSDK_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk.Publish" Version="$(CLI_WEBSDK_Version)" />
-    <BundledSdk Include="Microsoft.NET.Sdk.Web.ProjectSystem" Version="$(CLI_WEBSDK_Version)" />
+    <BundledSdk Include="NuGet.Build.Tasks.Pack" Version="$(NuGetBuildTasksPackageVersion)" />
```
livarcocc: Seems like you are missing Pack in the package name. Intentional?

build_projects/dotnet-cli-build/dotnet-cli-build.csproj
```diff
@@ -15,19 +15,19 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build" Version="$(CLI_MSBuild_Version)" />
+    <PackageReference Include="Microsoft.Build" Version="15.4.8" />
```
livarcocc: Why are we pinning these versions? In theory, we don't really need to keep them moving forward, as this is our build project, but I am curious anyways.

livarcocc (approved):

livarcocc:

livarcocc:
@johnbeisner Is there anything else we need for this to go in?

build/BundledSdks.props
johnbeisner: It was not intentional. We were using one variable for many packages here, like we did in a few places.

build_projects/dotnet-cli-build/dotnet-cli-build.csproj
johnbeisner: @livarcocc 
I was caught in a small paradox. To wit: we need the dotnet-cli-build dll built before using the 'download' task to download the props file & we need the props file before we build the dotnet-cli-build dll. 

1) since the move to repo toolset is likely to replace the download task
2) since I do not want anyone to think that these versions would be overridden
3) since this is how we build the product [tooling] and not the product itself
therefore, it seems right to pin them explicitly to a particular version.

PR #8125, opened 2017-11-28
Insert NuGet Build 4.6.0-preview1-4688 into cli

nugetlurker:
Insert NuGet Build 4.6.0-preview1-4688 into cli master branch

CC: @livarcocc

PR #8170, opened 2017-12-05
Adding a 'nuget-build' feed.

ghost:
Adding a 'nuget-build' feed; fixing the feed line-up: 'dotnet-core' and 'AspNetMaster'.

Reference: https://github.com/dotnet/cli/pull/8125


rohit21agrawal (approved):

rohit21agrawal:
with this change in place, can we stop publishing NuGet.Build.Tasks.Pack.nupkg to dotnet-core feed and you guys can just use it from nuget-build instead?

johnbeisner:
@dotnet-bot test this please.

livarcocc:
Yeah, that should be ok, though, eventually you need to start publishing to the blob feed.

cc @mmitche so that he is aware that this package is not in the blob feed yet.

PR #8189, opened 2017-12-08
Global tools: treat manifest EntryPoint value as a file path within the package

natemcmaster:
Currently dotnet install tools assumes that EntryPoint is the filename inside `tools/$(BundledTargetFramework)/any/` in the package. This restriction seems unnecessary. This also helps avoid the next-version problem when BundledTargetFramework changes to netcoreapp2.2. Without this, netcoreapp2.1 tool packages would no longer install when CLI is upgraded to a new version.

As a side note, if I've missed something and the `tools/$(BundledTargetFramework)/any/` restriction is important, it seems like you could easily add some kind of assert during install, such as:
```c#
if (Path.GetDirectoryName(toolConfig.EntryPoint) != $"tools/{BundledTargetFramework}/any/")
    throw new Exception("Global tools must put the entry point in tools/{BundledTargetFramework}/any/ for (reasons)");
```

### DIFF
--- a/src/dotnet/ToolPackage/ToolPackageObtainer.cs
+++ b/src/dotnet/ToolPackage/ToolPackageObtainer.cs
@@ -97,10 +97,7 @@ namespace Microsoft.DotNet.ToolPackage
                 toolConfiguration,
                 toolDirectory.WithSubDirectories(
                     packageId,
-                    packageVersion,
-                    "tools",
-                    targetframework,
-                    "any"));
+                    packageVersion));
         }
 
         private static void MoveToVersionedDirectory(
--- a/test/Microsoft.DotNet.ToolPackage.Tests/SampleGlobalTool/DotnetToolSettings.xml
+++ b/test/Microsoft.DotNet.ToolPackage.Tests/SampleGlobalTool/DotnetToolSettings.xml
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <DotNetCliTool>
   <Commands>
-    <Command Name="demo" EntryPoint="consoledemo.dll" Runner="dotnet" />
+    <Command Name="demo" EntryPoint="tools/netcoreapp2.1/any/consoledemo.dll" Runner="dotnet" />
   </Commands>
 </DotNetCliTool>
--- a/test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
+++ b/test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
@@ -59,9 +59,6 @@ namespace Microsoft.DotNet.ToolPackage.Tests
                 .ExecutableDirectory
                 .GetParentPath()
                 .GetParentPath()
-                .GetParentPath()
-                .GetParentPath()
-                .GetParentPath()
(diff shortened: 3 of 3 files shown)


natemcmaster:
@livarcocc @wli3 after our conversation yesterday, I'm assuming this PR won't be merged.

As a side note, are you ready to share publicly design notes on this feature so we can get feedback from others in the community?

wli3:
@natemcmaster we are working on the document. However, a lot of people are oof this month, it hard to give an ETA

wli3:
More on what we discussed: 

Note, most of the following are still in design

The existing master format can avoid the duplication of NuGet logic. When there is multiple TFM in the same package, the tools folder will be filtered by NuGet. The closest TFM tool will be picked.

CLI will read asset.json that have TFM already filtered. Only one TFM is present. At this point, "tools/netcoreapp2.1/any" is handled by NuGet, only "mytool.dll" is needed.

```json
"tools": {
    "my.tool/1.0.1": {
      "path": "my.tool/1.0.1",
      "files": [
        "my.tool.1.0.1.nupkg.sha512",
        "my.tool.nuspec",
        "tools/DotnetToolSettings.xml",
        "tools/netcoreapp2.1/any/mytool.deps.json",
        "tools/netcoreapp2.1/any/mytool.dll",
        "tools/netcoreapp2.1/any/mytool.pdb",
        "tools/netcoreapp2.1/any/mytool.runtimeconfig.json"
      ]
    }
  }
```

We can still extend this format in the future if we want to support "anything, no guard rail". We can use the folder structure `tools/nocheck-TBD/any/random.exe`. And the settings will be just `random.exe`

dasMulli:
Will there be a way to create a tool that can be used both as `<DotNetCliToolReference>`? per-project tools currently go into the lib folder so it would be good if there was a way to have the files in one place only to use it for per-project and global tools

wli3:
add @KathleenDollard for more

There are some technical constrains I can think of. 

People are using package reference to DotNetCliToolReference which end up in very confusing experience, so we try to block that for global tools. Which indicate they will not be compatible.

@nkolev92 FYI
From NuGet folder layout perspective, tools is not "lib", and tools requires different rules to filter than lib. For example, the RID problem and the requirement of a settings file.

nkolev92:
@dasMulli 
From NuGet side, we'd strongly like to discourage packages authored that way. 
It creates a lot of confusion because there's no enforceable contract on the package authoring for dotnetcli tool references. 
As @wli3 said, the tools(both DotNetCliToolReference  and global ones) are both in the tools folder, with a different structure. 
NuGet currently "ignores" the tools folder. The filtering rules that we'll add are a bit different than the lib ones. 

The overlapping folder structure makes it an unmaintainable hell for both authors and NuGet/CLI teams.

dasMulli:
I do (mostly) understand the problems this causes, esp since `dotnet add` and NuGet UI just add package references on install. The current guidance and examples for tools all use the lib folder though (e.g. https://github.com/dotnet/cli/issues/8183#issuecomment-351173417).

I just hoped there'd be a (simple, somewhat-easy-to-explan / msbuild-hack-less) way to create a NuGet package that can be used both as per-project and as global tool, even if it means including the entry point DLLs multiple times.

For example, `dotnet-xunit` would be a good example as it already includes a console tlest runner and a CLI verb.

PR #8199, opened 2017-12-08
Write UTF-8 BOM for solution files.

### LINKED ISSUE #8184: dotnet sln needs to encode UTF8 w/ BOM
VS and MSBuild seem to treat the sln file with a default ANSI encoding if there is no BOM, but we're writing it as UTF-8 without BOM.

## Steps to reproduce

```
> mkdir é
> cd é
> dotnet new console
> dotnet new sln
> dotnet sln add é.csproj
> dotnet build
```
## Expected  behavior
 Build succeeds
Opens fine in VS

## Actual behavior
Build fails
Projects are not found when opened in VS

```
C:\Program Files\dotnet\sdk\2.1.2\NuGet.targets(227,5): warning MSB3202: The project file "D:\Temp\é\Ã©.csproj" was not found. [D:\Temp\é\é.sln]
D:\Temp\é\Ã©.csproj : warning NU1503: Skipping restore for project 'D:\Temp\é\Ã©.csproj'. The project file may be invalid or missing targets required for restore. [D:\Temp\é\é.sln]
C:\Program Files\dotnet\sdk\2.1.2\NuGet.targets(103,5): warning : Unable to find a project to restore! [D:\Temp\é\é.sln]
D:\Temp\é\é.sln.metaproj : error MSB3202: The project file "D:\Temp\é\Ã©.csproj" was not found. [D:\Temp\é\é.sln]
```

## Environment data
```
.NET Command Line Tools (2.1.2)

Product Information:
 Version:            2.1.2
 Commit SHA-1 hash:  5695315371

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.2\

  Version  : 2.0.3
```

peterhuene:
Currently the solution file written out by the `sln` command uses a
UTF-8 encoding without a BOM.  This causes problems when the solution
file contains non-ASCII code points because Visual Studio and MSBuild
will not use a UTF-8 encoding when reading the solution file if the
BOM is omitted.

This commit causes the BOM to always be written when writing the
solution files.

Fixes #8184.

### DIFF
--- a/src/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs
+++ b/src/Microsoft.DotNet.Cli.Sln.Internal/SlnFile.cs
@@ -33,6 +33,7 @@ using System.IO;
 using System.Collections;
 using System.Globalization;
 using System.Reflection;
+using System.Text;
 using Microsoft.DotNet.Cli.Sln.Internal.FileManipulation;
 using Microsoft.DotNet.Tools.Common;
 
@@ -211,7 +212,7 @@ namespace Microsoft.DotNet.Cli.Sln.Internal
             }
             var sw = new StringWriter();
             Write(sw);
-            File.WriteAllText(FullPath, sw.ToString());
+            File.WriteAllText(FullPath, sw.ToString(), Encoding.UTF8);
         }
 
         private void Write(TextWriter writer)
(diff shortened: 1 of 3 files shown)


aolszowka:
FWIW this appears to break the Icon Overlay:

Without BOM
image

With BOM
image

In both cases the project still opens in Visual Studio. The IDE itself does not insert a BOM when creating a new solution file (or when modifying it), but there is no standard that defines the solution file format (that I can find). Historically when we've written internal tools around it we've left the encoding alone. However tools such as `dotnet sln` will write the BOM when performing operations such as `remove` as part of this commit.

Unclear if this was the intent of the change but this is ripple caused by it.

PR #8204, opened 2017-12-11
Fixing CLI:master - parameterizing 'GenerateInstallersAndCopyOutOfSandBox'

ghost:
The target: 'GenerateInstallersAndCopyOutOfSandBox' should be parametrized in run-build.sh.

Testing:
private branch VSO build of "AllLinuxDistrosNativeInstallers": https://devdiv.visualstudio.com/DevDiv/Default/_build/index?buildId=1211734&_a=summary
publishing to private blob storage: https://dotnetcli.blob.core.windows.net/jbeisner/Sdk/2.2.0-preview1-007786/all_linux_distros_native_installer_Release_version_badge.svg

### DIFF
--- a/run-build.sh
+++ b/run-build.sh
@@ -59,6 +59,7 @@ BUILD=1
 
 LINUX_PORTABLE_INSTALL_ARGS=
 ALL_LINUX_INSTALLERS_TARGET=
+GENERATE_INSTALLERS_TARGET=
 CUSTOM_BUILD_ARGS=
 
 # Set nuget package cache under the repo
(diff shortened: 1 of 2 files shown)


run-build.sh
```diff
@@ -116,7 +120,10 @@ while [[ $# > 0 ]]; do
             echo "  --nobuild                           Skip building, showing the command that would be used to build"
             echo "  --docker <IMAGENAME>                Build in Docker using the Dockerfile located in scripts/docker/IMAGENAME"
             echo "  --linux-portable                    Builds the Linux portable .NET Tools instead of a distro-specific version."
-            echo "  --all-linux-installers              Builds and publishes all the Linux distros' native installers."
+            echo "  --all-linux-installers              Builds and publishes all the Linux distros' native installers; outer call"
```
livarcocc: Can't we just use the all-linux-installers parameters for this? Also, why isn't passing the target directly not working anymore?

johnbeisner: 'Why isn't passing the target directly not working anymore?' - because defining 2 targets on the first 'dotnet msbuild' call in 'run-build.sh' is where the break is occurring. 

'Can't we just use the all-linux-installers parameters for this?' - There is an 'outer' or entry point call to 'build.sh' with a "BuildAndPublishAllLinuxDistrosNativeInstallers" target in addition to an 'inner' or a recursive call to 'build.sh' [see: SandBoxAndPackage.proj ] with a "GenerateInstallersAndCopyOutOfSandBox" target - they appear to need separate treatment.

wli3 (approved):

PR #8221, opened 2017-12-12
[WIP] **NO MERGE** Testing if analyzers are causing the hangs we're seeing

nguerrera:


nguerrera:
This won't work as UseRoslynAnalyzers is not something supported by the product.

PR #8224, opened 2017-12-12
MSBuild 15.6.22 in release/2.1

AndyGerlicher:


nguerrera:
After talking to @livarcocc We should not be merging 15.6 bits of anything into the 2.1 branch.

PR #8225, opened 2017-12-12
Bump stage0 to 2.2.0-preview1-007799.

eerhardt:
Need to remove some unnecessary dependencies in dotnet-cli-build.csproj because the NuGet.CommandLine.XPlat reference was bringing in an old MSBuild version, which was breaking restore. I didn't see any usages of this reference, so I'm removing it since it is not needed.

```JSON
 "code": "NU1608",
      "level": "Error",
      "message": "Detected package version outside of dependency constraint: Microsoft.Build.Runtime 15.1.1012 requires Microsoft.Build (= 15.1.1012) but version Microsoft.Build 15.4.8 was resolved.",
      "libraryId": "Microsoft.Build",
      "targetGraphs": [
        ".NETCoreApp,Version=v2.1"
      ] 
```

This was caused because

```JSON
"NuGet.CommandLine.XPlat/4.4.0-preview3-4475": {
       "dependencies": {
          "Microsoft.Build.Runtime": "15.1.1012", 
```

and

```JSON
"Microsoft.Build.Runtime/15.1.1012": {
        "type": "package",
        "dependencies": {
          "Microsoft.Build": "[15.1.1012]", 
```

Notice the `[15.1.1012]` version there, meaning that exact version - no higher, no lower.

This unblocks @janvorli's work to bootstrap alpine, since he is putting his locally built alpine SDK in Azure for that version.

### DIFF
--- a/build_projects/dotnet-cli-build/dotnet-cli-build.csproj
+++ b/build_projects/dotnet-cli-build/dotnet-cli-build.csproj
@@ -16,17 +16,13 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppPackageVersion)" />
     <PackageReference Include="Microsoft.Build" Version="15.4.8" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.4.8" />
     <PackageReference Include="Microsoft.CSharp" Version="4.0.1" />
     <PackageReference Include="System.Dynamic.Runtime" Version="4.0.11" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.4.1-beta-24410-02" />
     <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.1.1" />
     <PackageReference Include="System.Xml.XmlSerializer" Version="4.0.11" />
     <PackageReference Include="WindowsAzure.Storage" Version="7.2.1" />
-    <PackageReference Include="NuGet.CommandLine.XPlat" Version="4.4.0-preview3-4475" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.4.8" />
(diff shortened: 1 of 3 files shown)


build_projects/dotnet-cli-build/dotnet-cli-build.csproj
```diff
@@ -16,17 +16,13 @@
...
     <PackageReference Include="Microsoft.CSharp" Version="4.0.1" />
     <PackageReference Include="System.Dynamic.Runtime" Version="4.0.11" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.4.1-beta-24410-02" />
     <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.1.1" />
     <PackageReference Include="System.Xml.XmlSerializer" Version="4.0.11" />
     <PackageReference Include="WindowsAzure.Storage" Version="7.2.1" />
-    <PackageReference Include="NuGet.CommandLine.XPlat" Version="4.4.0-preview3-4475" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.4.8" />
-    <!-- This dependency was added due to an issue in restore where a lower version of this package coming from nuget.commandline.xplat
-    led to an error. This is tracked as NuGet issue : https://github.com/NuGet/Home/issues/4213 -->
```
nguerrera: I could swear I already deleted these once. Perhaps it never made it in or came back in a merge.

eerhardt: I looked in the history, and this dates back to when @jgoshi first converted from `project.json` to `.csproj`.  I'm thinking it was needed a long time ago and isn't needed anymore.

I checked all the `release` branches I know about, and it is in all of them.  So maybe you never made the PR?  or didn't merge it?  Either way I hope we can get rid of it here.

nguerrera: I must not have sent a PR. I forget the context. I don't see any reason why these need to be kept if things are green without them.

PR #8227, opened 2017-12-13
Enable Alpine 3.6 CLI build

janvorli:
This change enables CLI build for Alpine 3.6 and also adds CI job for
it. It is mostly based on changes that were necessary to
enable building of rhel.6 CLI

### DIFF
--- a/build/BuildDefaults.props
+++ b/build/BuildDefaults.props
@@ -3,7 +3,7 @@
     <CLITargets Condition=" '$(CLITargets)' == '' ">Prepare;Compile;Test;Package;Publish</CLITargets>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <IncludeNuGetPackageArchive Condition=" '$(IncludeNuGetPackageArchive)' == '' ">true</IncludeNuGetPackageArchive>
-    <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' AND $(Rid.StartsWith('rhel.6')) AND $(Rid.StartsWith('alpine.3.6'))">true</SkipBuildingInstallers>
+    <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' AND ($(Rid.StartsWith('rhel.6')) OR $(Rid.StartsWith('alpine.3.6')))">true</SkipBuildingInstallers>
     <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' ">false</SkipBuildingInstallers>
     <UsePortableLinuxSharedFramework Condition=" '$(UsePortableLinuxSharedFramework)' == '' AND '$(OSPlatform)' == 'linux' AND '$(Rid)' != 'rhel.6-x64' AND '$(Rid)' != 'alpine.3.6-x64' ">true</UsePortableLinuxSharedFramework>
     <IncludeSharedFrameworksForBackwardsCompatibilityTests Condition=" $(IncludeSharedFrameworksForBackwardsCompatibilityTests) == '' AND '$(Rid)' != 'linux-x64' AND '$(Rid)' != 'rhel.6-x64' AND '$(Rid)' != 'alpine.3.6-x64'">true</IncludeSharedFrameworksForBackwardsCompatibilityTests>
--- a/build/BuildDefaults.props
+++ b/build/BuildDefaults.props
@@ -3,10 +3,10 @@
     <CLITargets Condition=" '$(CLITargets)' == '' ">Prepare;Compile;Test;Package;Publish</CLITargets>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <IncludeNuGetPackageArchive Condition=" '$(IncludeNuGetPackageArchive)' == '' ">true</IncludeNuGetPackageArchive>
-    <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' AND $(Rid.StartsWith('rhel.6')) ">true</SkipBuildingInstallers>
+    <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' AND $(Rid.StartsWith('rhel.6')) AND $(Rid.StartsWith('alpine.3.6'))">true</SkipBuildingInstallers>
     <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' ">false</SkipBuildingInstallers>
-    <UsePortableLinuxSharedFramework Condition=" '$(UsePortableLinuxSharedFramework)' == '' AND '$(OSPlatform)' == 'linux' AND '$(Rid)' != 'rhel.6-x64' ">true</UsePortableLinuxSharedFramework>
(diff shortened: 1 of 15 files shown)


build/DerivedHostMachineInfo.props
```diff
@@ -2,8 +2,7 @@
   <PropertyGroup>
     <IsDebianBaseDistro Condition=" '$(HostOSName)' == 'ubuntu' OR '$(HostOSName)' == 'debian' ">true</IsDebianBaseDistro>
     <IsRPMBasedDistro Condition=" $(HostRid.StartsWith('rhel')) AND '$(HostRid)' != 'rhel.6-x64' ">true</IsRPMBasedDistro>
-
-    <PublishNativeInstallers Condition=" '$(IslinuxPortable)' != 'true' AND '$(HostRid)' != 'rhel.6-x64' ">true</PublishNativeInstallers>
+    <PublishNativeInstallers Condition=" '$(IslinuxPortable)' != 'true' AND '$(HostRid)' != 'rhel.6-x64' AND '$(HostRid)' != 'alpine.3.6-x64'">true</PublishNativeInstallers>
```
eerhardt: It seems like this property would be better to default to if we are creating native installers at all. If we aren't even creating native installers, then there is no reason to check the Rid.

So it seems that this property should be based on `$(SkipBuildingInstallers)`.  If `$(SkipBuildingInstallers) == true`, this property should be false.

scripts/obtain/dotnet-install.sh
```diff
@@ -146,8 +150,8 @@ get_current_os_name() {
         local linux_platform_name
         linux_platform_name="$(get_linux_platform_name)" || { echo "linux" && return 0 ; }
 
-        if [[ $linux_platform_name == "rhel.6"* ]]; then
-            echo "rhel.6"
+        if [[ $linux_platform_name == "rhel.6" || $linux_platform_name == "alpine.3.6" ]]; then
```
eerhardt: Did you remove a `*` here on purpose?  Did that change this check?

scripts/docker/alpine.3.6/Dockerfile
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+# Dockerfile that creates a container suitable to build dotnet-cli
+FROM microsoft/dotnet-buildtools-prereqs:alpine-3.6-3148f11-20171119021156
+
+RUN apk add --no-cache sudo ncurses
```
eerhardt: What's `ncurses` for?

build/DerivedHostMachineInfo.props
janvorli: @johnbeisner made it this way few months ago. John, what do you think?

scripts/docker/alpine.3.6/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Setup User to match Host User, and give superuser permissions
```
eerhardt: Do we need this?  I thought we needed this to test our `.deb` packages in the docker container on Ubuntu, but I don't see why we need superuser permissions to do a "normal" build.

scripts/docker/alpine.3.6/Dockerfile
janvorli: I don't actually remember, but it was needed for the build to work. I can remove it locally and try what was it if you are interested.

scripts/docker/alpine.3.6/Dockerfile
janvorli: I don't know, we do it in all Dockerfiles so I followed the suite. But you may be right. I'll try to see what happens if I remove it.

scripts/obtain/dotnet-install.sh
janvorli: I have removed it since it is not needed. We always generate rhel.6 for RHEL / CentOS 6 in the code above.

scripts/docker/alpine.3.6/Dockerfile
janvorli: So, without ncurses, I get this error:
`/opt/code/scripts/obtain/dotnet-install.sh: line 27: tput: command not found`

scripts/docker/alpine.3.6/Dockerfile
janvorli: I've verified it is not needed. The build completed ok with all of the user handling stuff disabled.

scripts/docker/alpine.3.6/Dockerfile
eerhardt: IMO - this is a bug in the `dotnet-install.sh` script.  The only reason we are using `tput` is because we are trying to print colors on the console.  Printing colors on the console shouldn't be required in order to install use `dotnet-install.sh`.

This is going to mean that anyone who either a) wants to build the dotnet/cli repo or b) wants to use `dotnet-install.sh` is going to have to install this utility on their machine

I don't feel this should block your work.  We can get the initial alpine work in by installing `ncurses` on this docker image.  But I think we should log a bug for `dotnet-install.sh` to not require `tput`, and only use it if it is available.

cc @livarcocc @johnbeisner

scripts/obtain/dotnet-install.sh
```diff
@@ -646,7 +650,7 @@ downloadwget() {
     if [ -z "$out_path" ]; then
         wget -q --tries 10 -O - "$remote_path" || failed=true
     else
-        wget -v --tries 10 -O "$out_path" "$remote_path" || failed=true
+        wget --tries 10 -O "$out_path" "$remote_path" || failed=true
```
johnbeisner: Why are we changing the wget call?

janvorli: Alpine's wget doesn't accept the -v flag.

build/BuildDefaults.props
```diff
@@ -3,10 +3,10 @@
     <CLITargets Condition=" '$(CLITargets)' == '' ">Prepare;Compile;Test;Package;Publish</CLITargets>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <IncludeNuGetPackageArchive Condition=" '$(IncludeNuGetPackageArchive)' == '' ">true</IncludeNuGetPackageArchive>
-    <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' AND $(Rid.StartsWith('rhel.6')) ">true</SkipBuildingInstallers>
+    <SkipBuildingInstallers Condition=" '$(SkipBuildingInstallers)' == '' AND $(Rid.StartsWith('rhel.6')) AND $(Rid.StartsWith('alpine.3.6'))">true</SkipBuildingInstallers>
```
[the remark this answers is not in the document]
janvorli: Oh, right. Thank you for spotting this mistake.

scripts/obtain/dotnet-install.sh
janvorli: Btw, the `-v` sets a verbose mode that's on by default anyways.

janvorli:
@johnbeisner I've added a commit that fixes the incorrect condition you've spotted.

Issue #8247, opened 2017-12-14
First run experience adds $USERPROFILE/.dotnet/tools to PATH many times on CI machines


natemcmaster:
This is happening most often on CI where we install DOTNET to different paths and clear $USERPROFILE/.dotnet between each build.

## Steps to reproduce
Download and extract .NET Core CLI to a folder.
Run this script
```ps1
$tmpDir = "$PSScriptRoot/tmp"
$env:USERPROFILE = $tmpDir
for ($i = 0; $i -lt 100; $i++) {
    rm -r $tmpDir
    # triggers the first-run experience
    & ./dotnet.exe new console --no-restore -o $tmpDir/
}
```

## Expected  behavior
CLI should only add $USERPROFILE/.dotnet/tools to path if it is not already present

## Actual behavior
It gets add many times.

[...]

PR #8248, opened 2017-12-14
Fix windows keep adding tools path to env:PATH

wli3:
Environment.GetEnvironmentVariable(PathName) means

Environment.GetEnvironmentVariable(PathName,
EnvironmentVariableTarget.Process)

However, I have added to .User. So the detection of path existence
failed. And it ends up adding the path again and again

Fix https://github.com/dotnet/cli/issues/8247

why no tests: this is a place hard to test, since it depends on a Windows API. To test it, I would need to play around CI's env:PATH, which is the problem in the beginning.

### DIFF
--- a/src/dotnet/ShellShim/WindowsEnvironmentPath.cs
+++ b/src/dotnet/ShellShim/WindowsEnvironmentPath.cs
@@ -11,18 +11,14 @@ namespace Microsoft.DotNet.ShellShim
     internal class WindowsEnvironmentPath : IEnvironmentPath
     {
         private readonly IReporter _reporter;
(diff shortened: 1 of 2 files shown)


src/dotnet/ShellShim/WindowsEnvironmentPath.cs
```diff
@@ -44,7 +40,9 @@ public void AddPackageExecutablePathToUserPath()
 
         private bool PackageExecutablePathExists()
         {
-            return _environmentProvider.GetEnvironmentVariable(PathName).Split(';').Contains(_packageExecutablePath);
```
wli3: no longer needed. it is windows specific.

PR #8286, opened 2017-12-22
Check null for getenv

wli3:
As you see, the implementation of the 3 OS to check PATH is different.
There is not too much to extract. And I tried to move the
"contains a folder path in env:path?" to a method. But that
method requires to pass in env:path, targetapath and path separator for
different OS. This method has low value. And people still must
remember to use this method. So I just added simple null check to the
place I used GetEnvPath.


wli3:
https://github.com/dotnet/corefxlab/pull/1986

wli3:
Get a repo using (don't try on your main machine) `REG delete "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /F /V PATH`

PR #8314, opened 2018-01-02
Check if tput command exists before using it

### LINKED ISSUE #8311: dotnet-install.sh shouldn't require tput
## Steps to reproduce
On the base Alpine3.6 image, run `dotnet-install.sh`.

## Expected  behavior
The script should install `dotnet` as usual, just like on any other Linux distro.

## Actual behavior

`/opt/code/scripts/obtain/dotnet-install.sh: line 27: tput: command not found`

## Environment data
`dotnet --info` output: N/A

## Notes

From my inspection, we are only using `tput` to change the console color during the install.  Printing colors on the console shouldn't be required in order to install use dotnet-install.sh.

This is going to mean that anyone who either a) wants to build the dotnet/cli repo or b) wants to use dotnet-install.sh is going to have to install this utility on their machine.

See https://github.com/dotnet/cli/pull/8227#discussion_r156692376

kasper3:
Fix #8311

@eerhardt

### DIFF
--- a/scripts/docker/alpine.3.6/Dockerfile
+++ b/scripts/docker/alpine.3.6/Dockerfile
@@ -6,8 +6,6 @@
 # Dockerfile that creates a container suitable to build dotnet-cli
 FROM microsoft/dotnet-buildtools-prereqs:alpine-3.6-3148f11-20171119021156
 
-RUN apk add --no-cache sudo ncurses
-
 # This Dockerfile doesn't use the USER_ID, but the parameter needs to be declared to prevent docker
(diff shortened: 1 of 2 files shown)


eerhardt:
Thanks for the contribution, @kasper3.

Do you think you can remove the `ncurses` entry from here as well?

https://github.com/dotnet/cli/blob/master/scripts/docker/alpine.3.6/Dockerfile#L9

It will ensure this change is working correctly.

kasper3:
👍  Removed

PR #8326, opened 2018-01-04
[ci skip] Get more Open Source Helpers

schneems:
[CodeTriage](https://www.codetriage.com/) is an app I have maintained
for the past 4-5 years with the goal of getting people involved in
Open Source projects like this one. The app sends subscribers a random
open issue for them to help "triage". For some languages you can also
suggested areas to add documentation.

The initial approach was inspired by seeing the work of the small
core team spending countless hours asking "what version was
this in" and "can you give us an example app". The idea is to
outsource these small interactions to a huge team of volunteers
and let the core team focus on their work.

I want to add a badge to the README of this project. The idea is to
provide an easy link for people to get started contributing to this
project. A badge indicates the number of people currently subscribed
to help the repo. The color is based off of open issues in the project.

Here are some examples of other projects that have a badge in their
README:

- https://github.com/crystal-lang/crystal
- https://github.com/rails/rails
- https://github.com/codetriage/codetriage

Thanks for building open source software, I would love to help you find some helpers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # .NET Command Line Interface
 
 [![.NET Slack Status](https://aspnetcoreslack.herokuapp.com/badge.svg?2)](http://tattoocoder.com/aspnet-slack-sign-up/) [![Join the chat at https://gitter.im/dotnet/cli](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/dotnet/cli?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![Help Contribute to Open Source](https://www.codetriage.com/dotnet/cli/badges/users.svg)](https://www.codetriage.com/dotnet/cli)
 
 This repo contains the source code for cross-platform [.NET Core](http://github.com/dotnet/core) command line toolchain. It contains the implementation of each command, the native packages for various supported platforms and the documentation.
 


richlander:
Thanks for taking an interest in the dotnet org. I took a 20min look into this and signed up for the service. I now have two days of triage mails that I've looked at.

We have very large repos with many open issues, some of which are very stale. I was sent some of these in the triage mails, some of which had long threads that were left with no obvious conclusion but lots of open questions. Looking at them, I think it is very hard for arbitrary users to know what to do with some of these old issues. My belief is that people who would already feel confident culling old issues are the only ones who could do that effectively/efficiently.

Also, we have many repos. Adding this badge on just one repo is only so useful. We'd need to reason about adding this badge to all of them to make this conversation meaningful.

I would find this service more useful if it could do some additional things:

- Sign up for a set of repos in an org, all at once, per user choice ("these are the three dotnet repos I work on / lurk at").
- Limit the age of issues that you see. I'd want to set that default as an org owner. I'd go for 1 month.
- Limit issues to # of comments. If an issue has over n comments, skip it. I'd set that at something like 5 or maybe even lower.

My meta point is that I'd find this service interesting if I knew that users could be effective w/o any help from maintainers. I'm not confident that this is the case. I see that the service has existed for a long time so I'm not trying to be critical of it. I'm sure it has many happy users. I've not yet seen how this service is going to make our lives easier, hence why I want to pass. Thanks.

As an aside, my comment about "w/o any help from maintainers" might sound like dotnet maintainers don't care about helping new users. We have lots of programs (for lack of a better term) in place to help new users develop confidence and context in the codebase/product to the end of making effective contributing. We're pretty happy with how those programs are working out. So, anything that is a pure add-on (like this) has to be self-sustaining in order for us to accept it.

schneems:
Thanks a ton for this in-depth reply. I want to reply to some of your suggestions

> Sign up for a set of repos in an org, all at once, per user choice ("these are the three dotnet repos I work on / lurk at").

That's a really interesting idea that's not come up before. As written CodeTriage mimics the structure of GitHub, but this could be doable. If I were to do this my first pass would be to have org pages where it lists all repos  in that org then people could subscribe directly to each repo, maybe with a button to subscribe to all.

I think there are cases such as the `rails` org where lots of people are interested in `rails/rails` but not necessarily `rails/arel` or `rails/sprockets`. Giving a page to allow someone to subscribe to all but one or two could be useful, with an easy shortcut to subscribing to all of them.

> Limit the age of issues that you see. I'd want to set that default as an org owner. I'd go for 1 month.

My personal philosophy is that open source is like a shark. If it stops moving then it dies. Issues are no different, if an issue isn't being actively worked on or interacted with, it will sap focus and energy from other issues. Even commenting "is this still an issue on master" can be valuable to keep an issue top of mind or to drive it towards completion. The original idea of CodeTriage isn't that the people getting the emails will have enough technical acumen to solve all issues, but rather they can find ways to move progress forwards. 

For such old issues, maybe there's other problems keeping them from getting solved, was an example app provided that demonstrates the issue? Could other people reproduce the problem? Did someone say they would do research and never reported back?

There are cases where issues become a discussion rather than a bug report, and then it's hard to drive things forwards. In that case I would argue that an issue inactive for one year is doing more harm open than closed.

As a maintainer I also use CodeTriage for my own repos, which helps me when I've forgotten about an issue that went dormant. I can certainly understand the desire to want to limit to recent issues (I also subscribe to rails). I'll have to think about a generalized way that maybe I could implement this request.

> Limit issues to # of comments. If an issue has over n comments, skip it. I'd set that at something like 5 or maybe even lower.

This is also an interesting idea. I think the combination "issues younger than a month and with fewer than 5 comments" should be a likely candidate for needing help. It makes more sense to work on issues when they're young and keep momentum moving. 

> We have lots of programs (for lack of a better term) in place to help new users develop confidence and context in the codebase/product to the end of making effective contributing. We're pretty happy with how those programs are working out. 

Do you have any resources I could look at? Anything you think might be generally applicable to a larger open source community? 

Again thanks for the detailed response. This type of feedback is invaluable. Also thanks for taking a look at the service.

nguerrera:
Closing this and I'll let @richlander decide if it's something we're going to want to enable on dotnet org as a whole. If so, we can reopen. Feel free to continue discussing here.

schneems:
Thanks for letting me know.

PR #8346, opened 2018-01-08
MSBuild 15.6.44

rainersigwald:
Current build of MSBuild, already inserted into VS 15.6 internal builds.

cc @AndyGerlicher

### DIFF
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -3,9 +3,9 @@
   <PropertyGroup>
     <MicrosoftNETCoreAppPackageVersion>2.0.4</MicrosoftNETCoreAppPackageVersion>
     <MicrosoftNETCoreDotNetHostResolverPackageVersion>$(MicrosoftNETCoreAppPackageVersion)</MicrosoftNETCoreDotNetHostResolverPackageVersion>
-    <MicrosoftBuildPackageVersion>15.5.180</MicrosoftBuildPackageVersion>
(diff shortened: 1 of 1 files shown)


build/DependencyVersions.props
```diff
@@ -2,8 +2,8 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <CLI_SharedFrameworkVersion>2.0.4</CLI_SharedFrameworkVersion>
-    <CLI_MSBuild_Version>15.5.180</CLI_MSBuild_Version>
-    <ClI_MSBUILD_Runtime_Version>15.5.180.1</ClI_MSBUILD_Runtime_Version>
+    <CLI_MSBuild_Version>15.6.0-preview-000044-1264514</CLI_MSBuild_Version>
+    <CLI_MSBUILD_Runtime_Version>$(CLI_MSBuild_Version)</CLI_MSBUILD_Runtime_Version>
```
rainersigwald: I'm not sure I understand the divergence here; why was this (possibly?) different?

It looks like the versions of the runtime package differ between our MyGet private feed and nuget.org.

AndyGerlicher: @rainersigwald 15.5.180.1 was published because I broke 15.5.180 saving it with NuGet package explorer. It lost some metadata so I had to re-publish just that package and de-list 15.5.180.

rainersigwald: I see, thanks.

rainersigwald:
Rebased, which changed the format a bit, but was still easy.

PR #8347, opened 2018-01-09
CLI produce 'DEB' distro agnostic installer packages; CLI:master

ghost:
_CLI to produce 'DEB' distro agnostic installer packages..._

Directory.Build.props
Minor reordering: Branding.props uses 'IsDebianBaseDistro' defined in DerivedHostMachineInfo.props

README.md
Updated - see: https://github.com/johnbeisner/cli/blob/CoalesceDebInstaller/README.md

Branding.props 
DEB filename token "HostMonikerRid" = "x64"

LinuxDistrosNativeInstaller.props 
Only build the 'DEB' on debian.8-x64; remove the two other 'DEB' entries.

SandBoxAndPackageInParallel.proj 
No change - formatting.

**Please note:**
Previously produced set of DEB files:
dotnet-sdk-version-ubuntu-x64.deb for Ubuntu 14.04 
dotnet-sdk-version-debian-x64.deb for Debian 8
dotnet-sdk-version-ubuntu.16.04-x64.deb for Ubuntu 16.04

Proposed set of DEB files:
dotnet-sdk-version-x64.deb

**Testing:**
Built a private VSO definition:
https://devdiv.visualstudio.com/DevDiv/_build/index?buildId=1272947&_a=summary
Published to private blob storage:
https://dotnetcli.blob.core.windows.net/jbeisner/Sdk/2.2.0-preview1-007881/

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This project has adopted the code of conduct defined by the [Contributor Covenan
 Build Status
 ------------
 
-|Windows x64 |Windows x86 |macOS |Linux x64 |Linux Native Installers |RHEL 6 Archive|Alpine 3.6 Archive|
+|Windows x64|Windows x86|macOS|Linux x64 Archive|Linux Native Installers|RHEL 6 Archive|Alpine 3.6 Archive|
 |:------:|:------:|:------:|:------:|:------:|:------:|:------:|
 |[![][win-x64-build-badge]][win-x64-build]|[![][win-x86-build-badge]][win-x86-build]|[![][osx-build-badge]][osx-build]|[![][linux-build-badge]][linux-build]|[![][linuxnative-build-badge]][linuxnative-build]|[![][rhel6-build-badge]][rhel6-build]|[![][alpine3.6-build-badge]][alpine3.6-build]|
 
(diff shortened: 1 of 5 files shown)


johnbeisner:
@leecow 
Please review the file name changes for alignment with the Runtime.

leecow (approved):
lgtm

README.md
```diff
@@ -64,11 +64,8 @@ To download the .NET Core runtime **without** the SDK, visit https://github.com/
...
-| **Linux x64** | [tar.gz][linux-targz] - [Checksum][linux-targz-checksum] |
-| **Ubuntu 14.04 / Linux Mint 17** | [Installer][ubuntu-14.04-installer] - [Checksum][ubuntu-14.04-installer-checksum]<br>_see installer note below_<sup>1</sup><br>tar.gz - See **Linux x64** |
-| **Ubuntu 16.04** | [Installer][ubuntu-16.04-installer] - [Checksum][ubuntu-16.04-installer-checksum]<br>_see installer note below_<sup>1</sup><br>tar.gz - See **Linux x64** |
-| **Debian 8**  | [Installer][debian-8-installer] - [Checksum][debian-8-installer-checksum]<br>_see installer note below_<sup>1</sup><br>tar.gz - See **Linux x64** |
-| **RHEL 7.2** | [Installer][rhel-7-installer] - [Checksum][rhel-7-installer-checksum]<br>_see installer note below_<sup>1</sup><br>tar.gz - See **Linux x64** |
+| **Linux x64** | [DEB Installer][linux-DEB-installer] - [Checksum][linux-DEB-installer-checksum]<br>_see installer note below_<sup>1</sup><br>[tar.gz][linux-targz] - [Checksum][linux-targz-checksum] |
+| **RHEL 7.2** | [Installer][rhel-7-installer] - [Checksum][rhel-7-installer-checksum]<br>tar.gz - See **Linux x64** |
```
eerhardt: It's a little weird (in my opinion) to split RHEL and Linux out like this.  They are both Linux, why does the `deb` installer get put under "Linux"?

What if we just had a `Linux` row and both the `deb` and `rpm` installers were in that same row?  Then you don't need `See Linux x64` under `RHEL 7.2`.

build/LinuxDistrosNativeInstaller.props
```diff
@@ -1,22 +1,20 @@
 <Project>
   <ItemGroup>
     <LinuxDistrosNeedNativeInstaller Include="ubuntu.16.04-x64">
```
eerhardt: Why do we still need these `ubuntu.XX.04-x64` entries?

livarcocc: We should need only a single ubuntu leg to produce the .deb file.

eerhardt: > however, continue to publish to all 3 Debian 'feeds'

I thought we were going to stop publishing to the Debian feeds.  We had a discussion about this a few weeks ago.

/cc @weshaggard

johnbeisner: @weshaggard 
@eerhardt 
@livarcocc 
To be clear, we are saying that I should completely remove the publishing to the debian feeds for all variants.

README.md
johnbeisner: The RPM installer package is distro specific where the DEB installer package is not.
See if the latest edit looks better to you: https://github.com/johnbeisner/cli/blob/CoalesceDebInstaller/README.md

eerhardt: Looks better, thanks for the update.

build/LinuxDistrosNativeInstaller.props
johnbeisner: @weshaggard
@eerhardt
@livarcocc
To be clear, we are saying that I should completely remove the publishing to debian feeds for all variants. Please advise.

eerhardt: I think that's the consensus from the discussion.

I don't think you'd need to completely remove all the debian publish feed stuff in this change.  I would just remove the extra `LinuxDistrosNeedNativeInstaller` entries.  And then in a separate change we can remove the debian publish feed stuff.

build/SandBoxAndPackageInParallel.proj
```diff
@@ -3,15 +3,15 @@
...
-      Targets="SandBoxAndPackage"
-            Properties="RepoRoot=$(RepoRoot);
-            LinuxDistrosNeedNativeInstaller=%(LinuxDistrosNeedNativeInstaller.Identity);
-            DockerFolder=%(LinuxDistrosNeedNativeInstaller.DockerFolder);
-            BuildConfiguration=$(BuildConfiguration);
-            InstallerOutputDirectory=$(InstallerOutputDirectory);
-            BuildInParallel='true'
-            " />
+  <Target Name="RunInSandBoxAndPackage">
+    <MSBuild Condition=" '%(LinuxDistrosNeedNativeInstaller.DockerFolder)' != '' "
```
eerhardt: (minor) This check might not be needed anymore.

eerhardt (approved):
Looks good.  Thanks, @johnbeisner

build/SandBoxAndPackageInParallel.proj
johnbeisner: Fixed...

PR #8361, opened 2018-01-10
Ignore incoherent rpm dependency on dotnet-runtime when testing aspnetcore

natemcmaster:
It looks like netci isn't configured correctly to actually run the TestSdkRpm target on PRs, but it fails on official builds.

### DIFF
--- a/build/package/Installer.RPM.targets
+++ b/build/package/Installer.RPM.targets
@@ -219,7 +219,8 @@
     <Exec Command="sudo yum -y install $(DownloadedSharedHostInstallerFile)" />
     <Exec Command="sudo yum -y install $(DownloadedHostFxrInstallerFile)" />
     <Exec Command="sudo yum -y install $(DownloadedSharedFrameworkInstallerFile)" />
-    <Exec Command="sudo yum -y install $(DownloadedAspNetCoreSharedFxInstallerFile)" />
+    <!-- Ignore dependencies, which may have an incoherent dependency on dotnet-runtime -->
+    <Exec Command="sudo rpm --install --nodeps $(DownloadedAspNetCoreSharedFxInstallerFile)" />
 
     <Exec Command="sudo yum -y install $(SdkInstallerFile)" />
 


natemcmaster:
@johnbeisner ideally, this target should be running for the RHEL7 and RHEL6 test builds on PRs, right? How do we enable that to run?

natemcmaster:

johnbeisner:
@natemcmaster 
"ideally, this target should be running for the RHEL7 and RHEL6 test builds on PRs, right? How do we enable that to run?" - yes the RHEL7 CI leg should create the  native RPM installer package; RHEL6 does not create RPM installers

natemcmaster:
So, this passes in PR checks, but I tested this manually and this doesn't fix the issue. Doesn't look like `--exclude` actually work the way I was expecting. Neither does `--nodeps`. 

I think what this means is that we *must* only build a dotnet-sdk RPM package that has coherent versions of dotnet-runtime and aspnetcore-runtime.

johnbeisner:
"I think what this means is that we must only build a dotnet-sdk RPM package that has coherent versions of dotnet-runtime and aspnetcore-runtime." - however, we have been creating dotnet-sdk RPM packages from non-coherent versions in the older methodology correct?

natemcmaster:
When moving from aspnetcore-store to aspnetcore-runtime in #8341 , the dependency graph of RPM packages changed. Unlike aspnetcore-store, the new aspnetcore-runtime package depends on dotnet-runtime directly.

Before:
```
dotnet-sdk  -> aspnetcore-store
   \-> dotnet-runtime
```

After
```
dotnet-sdk  -> aspnetcore-runtime -> dotnet-runtime
   \-> dotnet-runtime
```

natemcmaster:
I think we found the magic incantation. `yum`'s guardrails don't let us put our machine in the bad state (even intentionally), but `rpm` will.

johnbeisner:
@natemcmaster 
Private build launched: https://devdiv.visualstudio.com/DevDiv/Default/_build/index?buildId=1279390&_a=summary

johnbeisner:
@natemcmaster
Private build succeeded; private blob publish looks good: https://dotnetcli.blob.core.windows.net/jbeisner/Sdk/2.2.0-preview1-007922/dotnet-sdk-2.2.0-preview1-007922-rhel-x64.rpm

natemcmaster:
Thanks @johnbeisner

Issue #8363, opened 2018-01-10
Creation of RPM packages should happen on CI runs


johnbeisner:
The creation of RPM packages does not happen on Jenkins CI runs; it should so we can prevent build breaks in these method.

https://github.com/dotnet/cli/blob/master/build/package/Installer.RPM.targets#L202

```
    Workaround for Jenkins machines that don't have the necessary packages https://github.com/dotnet/core-setup/issues/2260
    <Message Condition=" '$(FPMPresent)' != 'True' "
             Text="FPM tool Not found, RPM packages will not be built."
             Importance="High"/>
  </Target>

  <Target Name="TestSdkRpm"
          DependsOnTargets="RestoreTests"
          Condition=" '$(IsRPMBasedDistro)' == 'True' and '$(FPMPresent)' == 'True' "
```

johnbeisner:
Steps to solve:
1) Ensure the DEB CI images have 'debuild'
2) Solve the non-execution of the RPM packaging on the Centos7.1 CI leg
3) Resubmit: https://github.com/dotnet/cli/issues/8363

johnbeisner:
RPM packaging & installation testing indeed occurs on the "Linux Build_Rhel_7_2_Release_x64" CI leg of dotnet/core-sdk

PR #8364, opened 2018-01-10
Remove the 'fpm' and 'debuild' existence checks

ghost:
Remove the 'fpm' and 'debuild' existence checks

Creation of RPM/DEB packages should happen on CI runs. The creation of RPM/DEB packages does not happen on Jenkins CI runs; it should so we can prevent build breaks in these methods.

Addresses: https://github.com/dotnet/cli/issues/8363

### DIFF
--- a/build/package/Installer.DEB.proj
+++ b/build/package/Installer.DEB.proj
@@ -7,7 +7,6 @@
 
   <Target Name="GenerateDebs"
           DependsOnTargets="SetupDebProps;
-                            TestDebuild;
                             BuildSdkDeb;
                             TestSdkDeb;"
           Condition=" '$(IsDebianBaseDistro)' == 'True' "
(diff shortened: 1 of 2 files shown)


wli3:
It should not be removed. This is used to build rpm package. Like debtool on deb. If we do not check it there, it will fail with 127 later if the docker do not have it pre installed

johnbeisner:
@wli3 
We need this to fail - we need to try the RPM packaging so we can confirm that it works before checkin.

wli3:
Make be let BuildRpms depends on TestFPMTool? Only check it if we need to build RPM

johnbeisner:
@wli3
We should let the generation of RPM packages execute without condition - this will catch 2 categories of errors: 1) problems with the merge with respect to RPM package generation 2) Jenkins infrastructure issues

johnbeisner:
@dotnet/dotnet-cli
Can I get an approve on this if applicable.

livarcocc:
It seems like you and will are not in agreement on this change just yet. I would like for Will to approve this change.

johnbeisner:
This change is not ready for merge.

PR #8373, opened 2018-01-11
Insert Testplatform 15.6.0-preview-20180109-01

smadala:


smadala:
@livarcocc Master is right branch to insert payload for VS 15.6?

PR #8380, opened 2018-01-11
Run-Build.sh/ps1 does not behave correctly when passing in a target.

ghost:
When passing in a 'target' on the command line from Build.sh/cmd; the target is attempted on the first pass call to "dotnet msbuild build.proj". This causes unpredictable results.

https://github.com/dotnet/cli/issues/8375

SandBoxAndPackage.proj 
Using the original implementation "/t:GenerateInstallersAndCopyOutOfSandBox"

run-build.ps1 
Removing the targets into "ExtraParametersNoTargets"

run-build.sh 
Removing the targets into "ExtraParametersNoTargets"
Parameterized targets: 'ALL_LINUX_INSTALLERS_TARGET' and 'GENERATE_INSTALLERS_TARGET' are no longer necessary.
Making the 'help' more consistent with run-build.ps1 

Follow-up actions:
Change "--all-linux-installers" to "/t:BuildAndPublishAllLinuxDistrosNativeInstallers" in the 'BuildArguments' of the 'AllLinuxDistrosNativeInstallers' VSO build definition: https://devdiv.visualstudio.com/DevDiv/Default/_build/index?context=allDefinitions&path=%5CDotNet%5CCLI%5Cmaster&definitionId=7254

### DIFF
--- a/run-build.sh
+++ b/run-build.sh
@@ -103,7 +103,7 @@ while [[ $# > 0 ]]; do
             shift
             ;;
         --help)
-            echo "Usage: $0 [--configuration <CONFIGURATION>] [--architecture <ARCHITECTURE>] [--skip-prereqs] [--nopackage] [--nobuild ] [--help]"
+            echo "Usage: $0 [--configuration <CONFIGURATION>] [--architecture <ARCHITECTURE>] [--skip-prereqs] [--nopackage] [--nobuild ] [--docker <IMAGENAME>] [--stage0 <DIRECTORY>] [--help]"
             echo ""
             echo "Options:"
             echo "  --configuration <CONFIGURATION>     Build the specified Configuration (Debug or Release, default: Debug)"
@@ -111,6 +111,8 @@ while [[ $# > 0 ]]; do
             echo "  --skip-prereqs                      Skip checks for pre-reqs in dotnet_install"
             echo "  --nopackage                         Skip packaging targets"
             echo "  --nobuild                           Skip building, showing the command that would be used to build"
+            echo "  --docker <IMAGENAME>                Build in Docker using the Dockerfile located in scripts/docker/IMAGENAME"
+            echo "  --stage0 <DIRECTORY>                Set the stage0 source directory. The default is to download it from Azure."
             echo "  --help                              Display this help message"
             exit 0
             ;;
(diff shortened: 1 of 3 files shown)


run-build.sh
```diff
@@ -137,6 +123,15 @@ while [[ $# > 0 ]]; do
     shift
 done
 
+# The first 'pass' call to "dotnet msbuild build.proj" has a hard-coded "WriteDynamicPropsToStaticPropsFiles" target
+#    therefore, this call should not have other targets defined. Remove all targets passed in as 'extra parameters'.
+argsnotargets=( )
+for element in ${args[@]} do
+  if [[ $element != /t:* ]]; then
```
nguerrera: I know we only use `/t: but should we defensively check `/target` as well? Also, I think msbuild switches are case insensitive.

johnbeisner:
@dotnet/dotnet-cli
Can I get an approve on this if applicable.

livarcocc:
Have you considered instead of sanitizing the arguments by removing things the first build invocation has problems with, to only pass through it the things that matters to it?

johnbeisner:
@livarcocc 
I have considered this; however, the first pass potentially needs a secret property from VSO - I do not believe it would survive the parameterization in the scripts.

johnbeisner:
@dotnet/dotnet-cli
Can I get an approve on this if applicable.

run-build.sh
```diff
@@ -100,31 +98,19 @@ while [[ $# > 0 ]]; do
...
-            echo "Usage: $0 [--configuration <CONFIGURATION>] [--skip-prereqs] [--nopackage] [--docker <IMAGENAME>] [--help]"
+            echo "Usage: $0 [--configuration <CONFIGURATION>] [--architecture <ARCHITECTURE>] [--skip-prereqs] [--nopackage] [--nobuild ] [--help]"
             echo ""
             echo "Options:"
             echo "  --configuration <CONFIGURATION>     Build the specified Configuration (Debug or Release, default: Debug)"
+            echo "  --architecture <ARCHITECTURE>       Build the specified architecture (x64 or x86 (supported only on Windows), default: x64)"
             echo "  --skip-prereqs                      Skip checks for pre-reqs in dotnet_install"
             echo "  --nopackage                         Skip packaging targets"
             echo "  --nobuild                           Skip building, showing the command that would be used to build"
-            echo "  --docker <IMAGENAME>                Build in Docker using the Dockerfile located in scripts/docker/IMAGENAME"
```
janvorli: Could you please explain why have you removed the usage documentation for the options for building in docker container and for the `--linux-portable` `--stage0` option? All of these options are still supported by the script and they are needed e.g. for Alpine, RHEL 6 or for bootstrapping new Unixes in the future.

johnbeisner: "linux-portable" is used only by our Linux VSO build and as such should be an unadvertised parameter.
@janvorli 
Do you feel that our customers are using the "docker" and "stage0" parameters?

janvorli: It seems strange to not to report all parameters that the script can get. In all of our other repos, we list all parameters, no matter how obscure they are. 
Anyone from the community can attempt to bring up the CLI for a Linux repo or Unix flavor of their preference and in such case, they would need the stage0 parameter and some would need the docker one.

johnbeisner: @janvorli 
I have put back the 'docker' and 'stage0' parameters in the usage.

run-build.sh
johnbeisner: @nguerrera 
For PS1: if((-not $param.StartsWith("/t")) -and (-not $param.StartsWith("/T"))) 
For SH: if [[ $arg != '/t'* ]] && [[ $arg != '/T'* ]]; then  

That should do it...

nguerrera: I see. Feels a little clumsy that it's StartsWith /t. It would match /toolsversion as well. Also, msbuild switches can start with - instead of /.

johnbeisner:
@dotnet/dotnet-cli
@janvorli 
Can I get an approve on this if applicable.

janvorli (approved):

PR #8402, opened 2018-01-16
Use rich mock to test InstallCommand

wli3:
To ensure the mock has the same behavior the component has, run mock under the same tests the adapter (production code) has.
It is a common problem that moq has -- "everything is mocked out, you are not test anything"

To test InstallCommand independent form real NuGet seems has a lot of over head. However, in practical speaking, if we test from command line level, all the logic about env:path will be executed as well. And that will mass up CI machines. It won't be better if we find some way to avoid running that logic. More high level, this makes `Microsoft.DotNet.ToolPackage` an adapter to NuGet, so as a result, other part of the system do not need to run integration test against NuGet. An in-memory mock over adapter interface is good enough.


wli3:
Fix 2 problem from the first time sending this PR

1. Use xunit theory to run mock and production code
2. Change the names to mock.
3. Moved all mocks to a new project. I tried to move it to test.utility, but that dll need to reference dotnet.dll which is netcoreapp2.1 due to interfaces are there. If I change test.utility to netcoreapp2.1, nearly all projects need to move to netcoreapp2.1, since it is been referenced in most of the tests including back-compact tests.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
```diff
@@ -37,12 +47,29 @@ public class InstallToolCommand : CommandBase
...
+                                              .CreateEnvironmentPathInstruction();
+
+            _shellShimMaker = shellShimMaker ?? new ShellShimMaker(_executablePackagePath.Value);
+
+            _reporter = reporter ?? Reporter.Output;
         }
 
         public override int Execute()
         {
             var executablePackagePath = new DirectoryPath(new CliFolderPathCalculator().ExecutablePackagesPath);
```
peterhuene: Do we still need this local vs. `_executablePackagePath` and, subsequently, the argument to `ObtainPackage`?

wli3: you are right, should use _executablePackagePath  instead

peterhuene: Seems like `_executablePackagePath` could be made a local to the constructor, though.

wli3:
fyi, due to time I will work on this pr after https://github.com/dotnet/cli/pull/8414 is merge

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
wli3: Fixed, change _executablePackagePath  local

test/Microsoft.DotNet.Tools.Tests.ComponentMocks/ToolPackageObtainerMock.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            if (packageVersion == null)
+            {
+                return p.PackageId == packageId;
+            }
+            return p.PackageId == packageId && p.Version == packageVersion;
+        }
+
+
+        public string GetFakeEntryPointPath()
```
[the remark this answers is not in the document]
wli3: fixed, method removed, not used

wli3:
@dotnet-bot test this please

wli3:
@dotnet-bot Test this please

wli3:
@dotnet-bot test this please

PR #8408, opened 2018-01-16
Fix missing package id error message for install tool command. 

### LINKED ISSUE #8381: Incorrect error message when package name is missing for install tool command
## Steps to reproduce
```
$ dotnet install tool
```

## Expected  behavior
An error message is printed about the missing tool package name.

## Actual behavior
```
$ dotnet install tool
packageId
Usage: dotnet install tool [options] <packageId>

Arguments:
  <packageId>   Package Id in NuGet

Options:
  --version           Version of the package in NuGet
  --configfile        NuGet configuration file
  -f, --framework     Target framework to publish for. The target framework has to be specified in the project file.
  -h, --help          Show help information.
```
## Environment data
```
$ dotnet --info
.NET Command Line Tools (2.2.0-preview1-007928)

Product Information:
 Version:            2.2.0-preview1-007928
 Commit SHA-1 hash:  0d6f8f9d59

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx-x64
 Base Path:   /Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/sdk/2.2.0-preview1-007928/

  Version  : 2.1.0-preview1-26110-02
  Build    : 77152705cd56e7d89f52585959ec3559d91084c4
```

peterhuene:
This PR fixes the error message that is displayed when the `install tool` command is not given a package id to install.

Previously, only `packageId` was output, which was confusing.

Fixes #8381.

This PR also refactors the string resources of the `install` command so that strings related to the subcommands are separated.

### DIFF
--- a/src/dotnet/commands/dotnet-help/LocalizableStrings.resx
+++ b/src/dotnet/commands/dotnet-help/LocalizableStrings.resx
@@ -268,6 +268,6 @@
     <value>Path to additional deps.json file.</value>
   </data>
   <data name="InstallDefinition" xml:space="preserve">
-    <value>Add item to development environment.</value>
+    <value>Installs an item into the development environment.</value>
   </data>
 </root>
--- a/src/dotnet/commands/dotnet-help/xlf/LocalizableStrings.zh-Hant.xlf
+++ b/src/dotnet/commands/dotnet-help/xlf/LocalizableStrings.zh-Hant.xlf
@@ -253,8 +253,8 @@
         <note />
       </trans-unit>
       <trans-unit id="InstallDefinition">
-        <source>Add item to development environment.</source>
-        <target state="new">Add item to development environment.</target>
+        <source>Installs an item into the development environment.</source>
+        <target state="new">Installs an item into the development environment.</target>
         <note />
       </trans-unit>
     </body>
--- a/test/dotnet-help.Tests/GivenThatIWantToShowHelpForDotnetCommand.cs
+++ b/test/dotnet-help.Tests/GivenThatIWantToShowHelpForDotnetCommand.cs
@@ -39,7 +39,7 @@ SDK commands:
   msbuild          Runs Microsoft Build Engine (MSBuild).
   vstest           Runs Microsoft Test Execution Command Line Tool.
   store            Stores the specified assemblies in the runtime store.
-  install          Add item to development environment.
+  install          Installs an item into the development environment.
   help             Show help.
 
 Common options:
--- a/scripts/cli-build-env.sh
+++ b/scripts/cli-build-env.sh
@@ -13,7 +13,7 @@ done
 
 REPO_ROOT="$( cd -P "$( dirname "$SOURCE" )/../" && pwd )"
 
-STAGE0_DIR=$REPO_ROOT.dotnet_stage0/x64
+STAGE0_DIR=$REPO_ROOT/.dotnet_stage0/x64
 export PATH=$STAGE0_DIR:$PATH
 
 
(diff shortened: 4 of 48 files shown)


src/dotnet/commands/dotnet-help/LocalizableStrings.resx
```diff
@@ -268,6 +268,6 @@
     <value>Path to additional deps.json file.</value>
   </data>
   <data name="InstallDefinition" xml:space="preserve">
-    <value>Add item to development environment.</value>
+    <value>Installs an item into the development environment.</value>
```
wli3: @KathleenDollard do you think this is a good change? I just copy pasted what you give me :)

peterhuene: My thinking behind the change: match the verbs as to not confuse with the `add command`.  Also, the tense of the verb matches what most of the command descriptions use.

Can easily remove the commit if we don't think it is a valuable change.

wli3:
@peterhuene please check with Kathleen about the install description. Everything else is good. Btw, did you use any tools to move string around? I had a hard time

peterhuene:
Unfortunately not.  I just moved everything around by hand.

peterhuene:
Holding off merging until @KathleenDollard provides feedback regarding the strings.

src/dotnet/commands/dotnet-help/LocalizableStrings.resx
KathleenDollard: I think Peter is right on this...

To help future thinking on similar issues, my logic was initially that we don't want descriptions that repeat the name of the feature, because they add nothing. 

However, he is correct, that "add" had a different meaning in context, and is not really adding much beyond install. 

or put another way, his pragmatism trumps my theory.

src/dotnet/commands/dotnet-help/xlf/LocalizableStrings.zh-Hant.xlf
```diff
@@ -253,8 +253,8 @@
         <note />
       </trans-unit>
       <trans-unit id="InstallDefinition">
-        <source>Add item to development environment.</source>
-        <target state="new">Add item to development environment.</target>
+        <source>Installs an item into the development environment.</source>
+        <target state="new">Installs an item into the development environment.</target>
```
KathleenDollard: First, I don't know what I did to turn this yellow. No change or concern. 

Second, somebody tell me we have a tool that creates these and no one had to copy this change.

test/dotnet-help.Tests/GivenThatIWantToShowHelpForDotnetCommand.cs
```diff
@@ -39,7 +39,7 @@ public class GivenThatIWantToShowHelpForDotnetCommand : TestBase
   msbuild          Runs Microsoft Build Engine (MSBuild).
   vstest           Runs Microsoft Test Execution Command Line Tool.
   store            Stores the specified assemblies in the runtime store.
-  install          Add item to development environment.
+  install          Installs an item into the development environment.
```
KathleenDollard: This was a long PR. If I didn't comment on something, point it out as I think I left comments where I focused. If I missed something let me know.

src/dotnet/commands/dotnet-help/xlf/LocalizableStrings.zh-Hant.xlf
peterhuene: We thankfully have a build task that generates the xlfs.

KathleenDollard:
If the description was the only change, I agree with Peter, made a comment, and will approve as soon as you let me know I didn't overlook anything on my scan of the remainder of the PR.

(Reason for my question: message below says "strings" and I perceive I reviewed only one "string" used multiple times that I trust you copied correctly).

peterhuene:
@KathleenDollard that was the string in question.  Thanks for the feedback!

Issue #8395, opened 2018-01-14
[tools] Require -g to install global tools


wli3:
-g, --global : Install user wide

PR #8414, opened 2018-01-17
[tools] Integrate NuGet

wli3:
Please review this change **by commit**.

The CI won't pass since I am pointing to a private NuGet build

The only WIP is to insert correct NuGet version after https://github.com/NuGet/NuGet.Client/pull/1900/ is merged. I hope to get this PR approved and merge it once NuGet is available. So I didn't make separate PRs.

It has 

* Update NuGet version. Rely on NuGet to filter TFM. And use asset.json to find entrypoint

* Update XML file to per TFM

* Add extra property to the fake project according to nuget

* Treat nuget fallback folder as offline cache for tool

* Require -g to install global tool #8395


wli3:
cc @nkolev92 @emgarten @danroth27 @natemcmaster

src/dotnet/commands/dotnet-install/xlf/LocalizableStrings.cs.xlf
```diff
@@ -98,6 +98,16 @@ Arguments: {1}
...
         <note />
       </trans-unit>
+      <trans-unit id="GlobalOptionDescription">
+        <source>Install user wide.</source>
+        <target state="new">Install user wide.</target>
+        <note />
+      </trans-unit>
+      <trans-unit id="InstallToolCommandOnlySupportGlobal">
+        <source>Install tool only support install user wide (-g).</source>
+        <target state="new">Install tool only support install user wide (-g).</target>
```
wli3: @KathleenDollard this need your review. This is the message when the consumer forget to add `-g` to `dotnet install tool my.tool`. Do you think it is clear enough to guide them adding `-g` ?

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace Microsoft.DotNet.ToolPackage
+{
+    internal class LockFileMatcher
+    {
+        private static string[] _pathInLockFilePathInArray;
+        private static string[] _entryPointPathInArray;
+
+        internal static bool MatchesFile(string pathInLockFile, string entryPoint)
+        {
+            _pathInLockFilePathInArray = SplitPathByDirectorySeparator(pathInLockFile);
```
peterhuene: Why the use of static fields here instead of locals?

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (_pathInLockFilePathInArray[0] != "tools")
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        private static bool EntryPointHasFileAtLeast()
```
livarcocc: This method name sounds weird. EntryPointHasFile seems enough to me.

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+
+namespace Microsoft.DotNet.ToolPackage
+{
+    internal class LockFileMatcher
+    {
+        private static string[] _pathInLockFilePathInArray;
```
livarcocc: Why is this whole class static?

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        private static bool RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories()
+        {
+            string[] pathAfterToolsTfmRid = _pathInLockFilePathInArray.Skip(3).ToArray();
+            return !pathAfterToolsTfmRid
+                .Where((directoryOnEveryLevel, i) => directoryOnEveryLevel != _entryPointPathInArray[i])
+                .Any();
+        }
+
+        private static bool PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories()
```
livarcocc: This method name has a typo in it. `Excatly` instead of `Exactly`.

Also, it is not really doing what it claims to be doing. It says it will check for ToolsTfmRid and it only checks that it has three directories there and that it starts with tools.

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            _pathInLockFilePathInArray = SplitPathByDirectorySeparator(pathInLockFile);
+            _entryPointPathInArray = SplitPathByDirectorySeparator(entryPoint);
+
+            return EntryPointHasFileAtLeast()
+                   && PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories()
+                   && RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories();
+        }
+
+        private static bool RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories()
```
livarcocc: This is also a very hard name to follow.

peterhuene: Nit: I think these method names are a bit too descriptive (I know they're private, but still); they read like test names.  I'd hate to have to type these out without IntelliSense.

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        private static bool EntryPointHasFileAtLeast()
+        {
+            return _entryPointPathInArray.Length >= 1;
```
peterhuene: Do we really need a method for checking if `_entryPointPathInArray` is empty?

src/dotnet/ToolPackage/ToolConfigurationAndExecutablePath.cs
```diff
@@ -5,17 +5,18 @@
 
 namespace Microsoft.DotNet.ToolPackage
 {
-    internal class ToolConfigurationAndExecutableDirectory
+    internal class ToolConfigurationAndExecutablePath
```
peterhuene: Nit: I wonder if this type could simply be replaced with a value tuple in the future (note: not important for this PR).

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: @peterhuene including why static field. It simply look nicer in linq. `Where(p => LockFileMatcher.MatchesFile(p, entry))`. I want to test it separately, and it is not very related to ToolPackageObtainer.cs, so it is a separate class. I don't want to pass _pathInLockFilePathInArray and _entryPointPathInArray all over the place. So I used field.

Do you think make it a non static class it fine? Where(p => new LockFileMatcher(p, entry).MatchesFile()) another minor downside is, there will be extra allocation.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -95,16 +100,49 @@ IProjectRestorer projectRestorer
...
+            LockFile lockFile = new LockFileFormat()
+                .ReadWithLock(toolDirectory.WithFile("project.assets.json").Value)
+                .Result;
 
-            return new ToolConfigurationAndExecutableDirectory(
+            LockFileItem dotnetToolSettings = FindAssetInLockFile(lockFile, "DotnetToolSettings.xml", packageId);
+
+            if (dotnetToolSettings == null)
+            {
+                throw new PackageObtainException("Cannot find DotnetToolSettings");
```
peterhuene: Will it be possible to install a tool with a missing settings file (despite a package type check)?  If so, I think this message should be localized and a little bit friendlier, something along the lines of `Package '{packageId}' is missing tool settings file DotNetToolSettings.xml.`

src/dotnet/commands/dotnet-install/LocalizableStrings.resx
```diff
@@ -172,4 +172,10 @@ The error was:
   <data name="InstallToolVersionDefinition" xml:space="preserve">
     <value>Version of the package in NuGet</value>
   </data>
+  <data name="GlobalOptionDescription" xml:space="preserve">
+    <value>Install user wide.</value>
```
peterhuene: Unless I was a user coming from node.js, I'm going to find this option perplexing.  Installing "globally" means to install "user wide"?  Given that we have no "local" install to default to, a required boolean option strikes me as bad UX.  I get that we're trying to future-proof a default of local install (is that something that even makes sense for tools? it does for npm which also manages runtime package dependencies), but this seems premature to me.

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: `return EntryPointHasFileAtLeast()
                  && PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories()
                   && RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories(); `

If we do not check it there, PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories will throw. This 3 check is more and more specific

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: TFM and RID is checked by NuGet already, since it is in asset.json. Maybe a good place to add comment

src/dotnet/ToolPackage/LockFileMatcher.cs
peterhuene: Sorry, I didn't mean to imply that the check itself is unneeded, only refactoring the check out into its own method when it is only checked from exactly one place.  Every time a method call is introduced I now have to stop and jump to the method's definition to see what it does.  It would be far more concise to simply do a check for empty where needed.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: > Will it be possible to install a tool with a missing settings file (despite a package type check)?

Yes, NuGet will not check the setting file.

And yes, I am missing loc. And your error message is better

build/MSBuildExtensions.targets
```diff
@@ -131,6 +131,7 @@ Copyright (c) .NET Foundation. All rights reserved.
     <BundledNETCoreAppPackageVersion>$(_NETCoreAppPackageVersion)</BundledNETCoreAppPackageVersion>
     <BundledNETStandardTargetFrameworkVersion>$(_NETStandardTargetFrameworkVersion)</BundledNETStandardTargetFrameworkVersion>
     <BundledNETStandardPackageVersion>$(_NETStandardPackageVersion)</BundledNETStandardPackageVersion>
+    <NETCorePlatformsImplicitPackageVersion>$(MicrosoftNETCoreAppPackageVersion)</NETCorePlatformsImplicitPackageVersion>
```
dsplaisted: I don't think the Platforms package will always have the same version number as the NETCoreApp package.  Did you verify whether this will be the case with the runtime team?

src/dotnet/commands/dotnet-install/LocalizableStrings.resx
wli3: @KathleenDollard on -g

src/dotnet/ToolPackage/LockFileMatcher.cs
dsplaisted: I agree that you shouldn't use static state to keep track of the internal execution state of a method.

What I would do is move the static fields inside the method, and move the private methods inside the internal method as [local functions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions).

peterhuene:
Other than my comments above, the rest of the changes are 👍

src/dotnet/ToolPackage/LockFileMatcher.cs
dsplaisted: I think Peter is not saying don't check it, he's just saying it doesn't seem worth it to factor the length check out into a separate method.

src/dotnet/ToolPackage/LockFileMatcher.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Linq;
+
+namespace Microsoft.DotNet.ToolPackage
+{
+    internal class LockFileMatcher
+    {
+        private static string[] _pathInLockFilePathInArray;
+        private static string[] _entryPointPathInArray;
+
+        internal static bool MatchesFile(string pathInLockFile, string entryPoint)
```
dsplaisted: From just reading the code without knowing more context, this class seems confusing.  `MatchesFile` is a pretty generic name and doesn't help me understand the code.

I would recommend adding comments describing what the MatchesFile method does, and the context in which it does it.

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: nit: this will be a philosophical debate. But I think method name length should be low priority here. Extreme case:
 [objective c naming](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html). And most editor has auto complete on words (by simply scan the existing doc and find words). Most "starter-pack" of vim and emacs will include such function as well.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -95,16 +96,47 @@ IProjectRestorer projectRestorer
...
                     packageId,
-                    packageVersion,
-                    "tools",
-                    targetframework,
-                    "any"));
+                    packageVersion)
+                .WithFile(entryPointFromLockFile.Path));
+        }
+
+        private static LockFileItem FindAssetInLockFile(LockFile lockFile, string toolConfigurationToolAssemblyEntryPoint, string packageId)
```
dsplaisted: `toolConfigurationToolAssemblyEntryPoint` seems like the wrong name here since you're passing `"DotnetToolSettings.xml"` as the argument value in one instance.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -95,16 +96,47 @@ IProjectRestorer projectRestorer
...
+
+            FilePath toolConfigurationPath =
+                toolDirectory
+                    .WithSubDirectories(packageId, packageVersion)
+                    .WithFile(dotnetToolSettings.Path);
+
+            ToolConfiguration toolConfiguration =
+                ToolConfigurationDeserializer.Deserialize(toolConfigurationPath.Value);
+
+            var entryPointFromLockFile = FindAssetInLockFile(lockFile, toolConfiguration.ToolAssemblyEntryPoint, packageId);
```
dsplaisted: Is there a reason we need to specifically look for the entry point in the lockfile?  Could we just combine the path of the directory the settings file is in with the relative path to the entry point in that settings file?

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: I see, I think make it non static class can satisfy all

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -95,16 +96,47 @@ IProjectRestorer projectRestorer
...
+
+            if (entryPointFromLockFile == null)
+            {
+                throw new PackageObtainException("Cannot find tool entry point the package");
```
dsplaisted: Error messages need to be localized and more user friendly (ie specify that this is probably a package authoring error, and that the entry point {0} specified in {toolsettingspath} could not be found).

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: will in line this method

test/Microsoft.DotNet.ToolPackage.Tests/LockFileMatcherTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.DotNet.Tools.Install.Tool;
+using Xunit;
+
+namespace Microsoft.DotNet.ToolPackage.Tests
+{
+    public class LockFileMatcherTests : TestBase
+    {
+
+        [Fact]
+        public void MatchesEntryPointTests()
```
dsplaisted: I'd suggest making this a `Theory` instead of testing multiple things in one `Fact`.  As it stands, if the first assertion fails, you'll get no information about the subsequent ones.

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
```diff
@@ -49,15 +45,16 @@ public void GivenNugetConfigAndPackageNameAndVersionAndTargetFrameworkWhenCallIt
...
+            ToolConfigurationAndExecutablePath toolConfigurationAndExecutablePath = packageObtainer.ObtainAndReturnExecutablePath(
+                packageId: TestPackageId,
+                packageVersion: TestPackageVersion,
+                nugetconfig: nugetConfigPath,
+                targetframework: _testTargetframework);
+
+            var assetJsonPath = toolConfigurationAndExecutablePath
+                .Executable
+                .GetDirectoryPath()
                 .GetParentPath()
```
dsplaisted: If you're going five levels up the directory tree, I'd suggest adding comments noting what each level is.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -154,8 +176,15 @@ IProjectRestorer projectRestorer
                     new XElement("PropertyGroup",
                         new XElement("TargetFramework", targetframework),
                         new XElement("RestorePackagesPath", individualToolVersion.Value),
-                        new XElement("RestoreSolutionDirectory", Directory.GetCurrentDirectory()), // https://github.com/NuGet/Home/issues/6199
-                        new XElement("DisableImplicitFrameworkReferences", "true")
+                        new XElement("RestoreProjectStyle", "DotnetToolReference"),
```
nkolev92: It was looked over in the design doc you've been usuing as a reference. 
How does the runtimeidentifier get set? 
does the SDK set it?

src/dotnet/commands/dotnet-install/LocalizableStrings.resx
KathleenDollard: @peterhuene 

Yes, --user would be more correct than --global, but...

It's not coming from node.js, it's coming from npm experience, which happens to be deployed by node. I make that distinction because everybody (even us - Azure CLI as example) has the install node to run npm experience. So paralleling that is important. 

We plan to have a "session" based tool that will be the default. I anticipate that in this release, but I don't think Preview 1. That will install into the user's session. 

"Local" doesn't make sense to me. There is my whole experience (global), this session (planned) and this repo (in future thinking, not designed, not this version). 

If global was default, we could not change later. 

So, yes, this is odd, but I think correct.

src/dotnet/commands/dotnet-install/xlf/LocalizableStrings.cs.xlf
KathleenDollard: We'll check this at standup. I anticipate this is temporary as I think we're still planning session tools after Preview 1. 

The --global switch (-g) is currently required because only user wide tools supported.

@peterhuene can you look at that as well

src/dotnet/commands/dotnet-install/LocalizableStrings.resx
peterhuene: I totally get what you're saying, but I don't think I share the sentiment that we need to be consistent with a package manager from another ecosystem, even if we've been instructing users to use it for our own tools prior to this feature.  If it truly makes sense for `dotnet` for the global option to be non-default, then I'm all for this.

src/dotnet/ToolPackage/LockFileMatcher.cs
peterhuene: To be clear, I do think this should remain a static class; making these instance fields does not address the reentrancy problem.

src/dotnet/ToolPackage/LockFileMatcher.cs
peterhuene: Agreed that naming can be a complete bikeshedding exercise (not to mention that good naming is difficult).  I simply prefer concise names that convey enough meaning; they don't need to be fully self-describing as that's what either documentation or the code itself is for.

src/dotnet/commands/dotnet-install/xlf/LocalizableStrings.cs.xlf
peterhuene: @wli3's PR properly checks for the presence of the global flag and aborts if not present.

Agree with @KathleenDollard's suggestion of making it clear that this is "currently required" because only user-wide is supported.  As a user encountering a message like that, I would think "oh okay, perhaps one day the -g option won't be always required".

I'd just amend the suggestion to: `The --global switch (-g) is currently required because only user wide tools are supported.` (added `are`).

KathleenDollard: Yes, agree with Peter

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: by passing via command line https://github.com/wli3/cli/blob/a1bcedb87b885e52d1fcc862f06ef9fda09ccd99/src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs#L41

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: what in nuget is "tools/tfm/rid/my/tools.dll" what we need from nuget is "tools/tfm/rid" in your case, however, we need to match tfm and rid pattern to get that. I think it is easier to match entrypoint directly

build/MSBuildExtensions.targets
wli3: This is a problem. I think we should remove this property in CLI. And we should get this property in SDK derived from its dependency 
image

src/dotnet/ToolPackage/LockFileMatcher.cs
livarcocc: It seems to me like both the fields could simply be passed as parameters to the private static methods, no?

wli3: This is my plan, it is good?

``` C#
internal class LockFileMatcher
    {
        private readonly string[] _pathInLockFilePathInArray;
        private readonly string[] _entryPointPathInArray;

        public LockFileMatcher(LockFileItem lockFileItem, string targetRelativeFilePath)
        {
            _pathInLockFilePathInArray = SplitPathByDirectorySeparator(lockFileItem.Path);
            _entryPointPathInArray = SplitPathByDirectorySeparator(targetRelativeFilePath);
        }

        internal bool MatchesFile()
        {
            return EntryPointHasFileAtLeast()
                   && PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories()
                   && RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories();
        }
```

wli3: Or as Livar suggested, make it static method, but downside is passing pathInLockFilePathInArray entryPointPathInArray down to PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories and RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories.

src/dotnet/ToolPackage/LockFileMatcher.cs
livarcocc: If TFM and RID are already being checked, than don't mention it in this method's name? Seems more confusing to do so, then when someone looks at the implementation, than it does not match the method name.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
```diff
@@ -43,21 +43,14 @@ public override int Execute()
...
+            var toolConfigurationAndExecutablePath = ObtainPackage(executablePackagePath);
 
             var shellShimMaker = new ShellShimMaker(executablePackagePath.Value);
-            var commandName = toolConfigurationAndExecutableDirectory.Configuration.CommandName;
+            var commandName = toolConfigurationAndExecutablePath.Configuration.CommandName;
             shellShimMaker.EnsureCommandNameUniqueness(commandName);
 
             shellShimMaker.CreateShim(
-                executable.Value,
+                toolConfigurationAndExecutablePath.Executable.Value,
```
livarcocc: NIT: Should ToolConfigurationAndExecutablePath have a ExecutablePath property, instead of this ugly Executable.Value? I think it would read better.

build/MSBuildExtensions.targets
wli3: PackageDefinitions won't work, This need to be added before PackageDefinitions is computed

nguerrera: We are already getting the NETStandard.Library 2.0.1 from PackageDefinitions here in this file. Why does this need to be computed before PackageDefinitions in this file?

wli3: "computed before PackageDefinitions" in during pack, ideally this should computed according to consumer project's runtime. Instead of predefined in CLI. the problem is if the user set _RuntimeFrameworkVersion_, and we will add the package version according to the runtime CLI carries. If RuntimeFrameworkVersion > CLI's runtime, there will be a package downgrade.

wli3: Talked to Nick. For preview 1, we should get the version from PackageDefinitions and set it in build/MSBuildExtensions.targets.

After that, we need to design whether we want to keep `Microsoft.NETCore.Platforms` on Pack side or install side (add the reference in fake project). Or maybe we can ask NuGet for pack target  "add extra dependency" feature.

Also, the property name need to change. This require change on both SDK and CLI, I will send PR once this is merged. https://github.com/dotnet/cli/issues/8421

build/NugetConfigFile.targets
```diff
@@ -18,9 +18,12 @@
 
       <NugetConfigCLIFeeds>
         <![CDATA[
+    <add key="localtestfeed" value="C:\work\testfeed" />
```
wli3: Will revert before merge

src/dotnet/ToolPackage/LockFileMatcher.cs
dsplaisted: I would just do it as a static method with local functions.  Something like this:

```C#
internal class LockFileMatcher
{
    internal static bool MatchesFile(string pathInLockFile, string entryPoint)
    {
        string[] pathInLockFilePathInArray = SplitPathByDirectorySeparator(pathInLockFile);
        string[] entryPointPathInArray = SplitPathByDirectorySeparator(entryPoint);

        return EntryPointHasFileAtLeast()
                && PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories()
                && RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories();

        //  Local functions
        bool RestMatchsOtherThanPathInLockFileHasToolsTfmRidAsParentDirectories()
        {
            string[] pathAfterToolsTfmRid = pathInLockFilePathInArray.Skip(3).ToArray();
            return !pathAfterToolsTfmRid
                .Where((directoryOnEveryLevel, i) => directoryOnEveryLevel != entryPointPathInArray[i])
                .Any();
        }

        bool PathInLockFileHasExcatlyToolsTfmRidAsParentDirectories()
        {
            if (pathInLockFilePathInArray.Length - entryPointPathInArray.Length != 3)
            {
                return false;
            }

            if (pathInLockFilePathInArray[0] != "tools")
            {
                return false;
            }

            return true;
        }

        bool EntryPointHasFileAtLeast()
        {
            return entryPointPathInArray.Length >= 1;
        }

        string[] SplitPathByDirectorySeparator(string path)
        {
            return path.Split('\\', '/');
        }
    }    
}
```

src/dotnet/ToolPackage/LockFileMatcher.cs
dsplaisted: To me even the long name doesn't really help me understand what the method does.  So more than changing the name I would add comments to explain it.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -172,7 +174,11 @@ private static LockFileItem FindAssetInLockFile(LockFile lockFile, string toolCo
                         new XElement("RestorePackagesPath", individualToolVersion.Value),
                         new XElement("RestoreProjectStyle", "DotnetToolReference"),
                         new XElement("RestoreRootConfigDirectory", Directory.GetCurrentDirectory()),
-                        new XElement("DisableImplicitFrameworkReferences", "true")
+                        new XElement("DisableImplicitFrameworkReferences", "true"),
```
dsplaisted: Can you add comments explaining why these properties are set to these values?

test/Microsoft.DotNet.ToolPackage.Tests/Microsoft.DotNet.ToolPackage.Tests.csproj
```diff
@@ -52,5 +57,7 @@
     </MSBuild>
     <MSBuild BuildInParallel="False" Projects="SampleGlobalTool/consoledemo.csproj" Targets="pack" Properties="Configuration=Release;NuspecFile=includepublish.nuspec;NuspecBasePath=$(testAssetSourceRoot);PackageOutputPath=$(OutputPath)/TestAssetLocalNugetFeed">
     </MSBuild>
+    <Copy SourceFiles="ExtraTestFeedPackages/microsoft.netcore.platforms.2.0.1.nupkg" DestinationFolder="$(OutputPath)/TestAssetLocalNugetFeed" />
```
dsplaisted: Instead of explicitly copying this here, I'd suggest adding the file to the `Content` item.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
wli3: CreateShim should take in a strong type FilePath. However this will conflict with Nate's work. So I will change it later.

src/dotnet/commands/dotnet-install/xlf/LocalizableStrings.cs.xlf
wli3: Fixed, change to above text

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: Please see if the local function + strong type input + comments is good enough

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: Please see if the local function + strong type input + comments is good enough

src/dotnet/ToolPackage/LockFileMatcher.cs
wli3: Please see if the local function + strong type input + comments is good enough

build/MSBuildExtensions.targets
wli3: Changed as discussed above

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Fixed

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Fixed

test/Microsoft.DotNet.ToolPackage.Tests/LockFileMatcherTests.cs
wli3: Fixed

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
wli3: Fixed, added comment

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Fixed, added comment

test/Microsoft.DotNet.ToolPackage.Tests/Microsoft.DotNet.ToolPackage.Tests.csproj
wli3: Fixed, use Content

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
```diff
@@ -99,6 +99,13 @@ public void GivenNugetConfigAndPackageNameAndVersionAndTargetFrameworkWhenCallIt
                 nugetconfig: nugetConfigPath,
                 targetframework: _testTargetframework);
 
+            /*
+              From mytool.dll to project.assets.json
+               .dotnet/.tools/packageid/version/packageid/version/mytool.dll
```
dsplaisted: Shouldn't the path also include `tools/tfm/rid`?

wli3:
@dsplaisted  > Is there any way we can avoid checking in a .nupkg file (microsoft.netcore.platforms.2.0.1.nupkg)?

when testing `--source`, the local feed need to have all the packages. Do you think it is ok to download this package from myget during build? Since some of tests have myget as feed and it eventually downloads it from the internet

dsplaisted:
@wli3 Yes, I would try to download the package as part of the build or test before it tests the `--source` option.

wli3:
@dsplaisted changed to download package during build

wli3:
Rebased including shim change. Find a bug in manual test, DisableImplicitNuGetFallbackFolder need to set in temp project.

src/dotnet/commands/dotnet-install/InstallCommandParser.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.DotNet.Cli.CommandLine;
 using LocalizableStrings = Microsoft.DotNet.Tools.Install.LocalizableStrings;
 
+
```
peterhuene: Looks like the change to `InstallCommandParser.cs` can be reverted.

peterhuene:
Should we squash the PR?

wli3:
> Should we squash the PR?

Yes, I retry to make it nicer user rebase -i, but there is too much xlf merge conflict

wli3:
@dotnet-bot test Alpine3.6 x64 Debug Build

wli3:
@dotnet-bot test Alpine3.6 x64 Debug Build

wli3:
Updated test asset's platforms package version to a preview version that has alpine support

PR #8422, opened 2018-01-18
Fix source option for install command.

peterhuene:
The original PR that implemented the source option was updated incorrectly
during review and the source option was accidentally not passed into the
package obtainer.  This resulted in the source option not being respected from
the install command.

The tests passed because the only test coverage is at the package obtainer
level; tests of the install command itself were waiting on additional changes
to enable (still not yet merged).

The fix is to properly pass the source option through when obtaining the
package.


peterhuene:
@wil3 I accidentally broke the source option from the install command itself during PR review.  This corrects that mistake.

PR #8433, opened 2018-01-19
Simplify the tool_launcher to build for just one version of .NET Framework

natemcmaster:
As @nguerrera pointed out, we don't actually need to cross-target. A .NET Framework app can be configured to automatically roll forward from 3.5 to 4.5 by listing multiple supported runtime versions in app.config.


natemcmaster:

natemcmaster:
@peterhuene squashed and updated

wli3:
I will merge it once the build is done. I happen to need a new build to trigger ci

Issue #8436, opened 2018-01-20
Cannot install tool on macOS high sierra(temp fix merged)


dasMulli:
## Steps to reproduce

* Using `2.2.0-preview1-008000` on high sierra
* Create a console app targeting `netcoreapp2.1` and set `PackAsTool` property to `true`. (`sayhi.csproj`)
* Create a nuget.config file to add `dotnet-core` feed and a local folder for the produced nupkg file.
* Use `dotnet pack` to create a nupkg file from the tool app.
* Use `dotnet install tool -g --configfile $(pwd)/nuget.config sayhi`

## Expected  behavior

Installs the tool

## Actual behavior

Error that the package is not compatible with `netcoreapp2.1 (.NETCoreApp,Version=v2.1) / osx.10.13-x64` but only supports `netcoreapp2.1 (.NETCoreApp,Version=v2.1) / any`:

```
$ dotnet install tool -g --configfile $(pwd)/nuget.config sayhi
Install failed. Failed to download package:
NuGet returned:

Failed to restore package. 
WorkingDirectory: 
Arguments: restore /var/folders/b1/p_sdf_v932sfdvr0ccxt2gd80000gn/T/4zstq4rl.ya5/jaj1vnsm.wf1.csproj --configfile /Users/martin/tmp/toolstest/sayhi/nuget.config --runtime osx.10.13-x64 /p:BaseIntermediateOutputPath=\"/Users/martin/.dotnet/tools/sayhi/tn02uk3w.ykb\"
Output:   Restoring packages for /var/folders/b1/p_sdf_v932sfdvr0ccxt2gd80000gn/T/4zstq4rl.ya5/jaj1vnsm.wf1.csproj...
  Installing Microsoft.NETCore.Platforms 2.1.0-preview1-26116-01.
  Installing sayhi 1.0.0.
/var/folders/b1/p_sdf_v932sfdvr0ccxt2gd80000gn/T/4zstq4rl.ya5/jaj1vnsm.wf1.csproj : error NU1202: Package sayhi 1.0.0 is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1) / osx.10.13-x64. Package sayhi 1.0.0 supports: netcoreapp2.1 (.NETCoreApp,Version=v2.1) / any
  Restore failed in 1.26 sec for /var/folders/b1/p_sdf_v932sfdvr0ccxt2gd80000gn/T/4zstq4rl.ya5/jaj1vnsm.wf1.csproj.
```

[...]

PR #8437, opened 2018-01-21
[tools]MacOs HighSierra Fallback

wli3:
https://github.com/dotnet/cli/issues/8436

### DIFF
--- a/src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
+++ b/src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
@@ -38,7 +38,7 @@ namespace Microsoft.DotNet.Tools.Install.Tool
             argsToPassToRestore.AddRange(new List<string>
             {
                 "--runtime",
-                RuntimeEnvironment.GetRuntimeIdentifier(),
+                GetRuntimeIdentifierWithMacOsHighSierraFallback(),
                 $"/p:BaseIntermediateOutputPath={assetJsonOutput.ToQuotedString()}"
             });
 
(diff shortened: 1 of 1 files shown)


src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
```diff
@@ -56,5 +56,17 @@ internal class ProjectRestorer : IProjectRestorer
                         result.StartInfo.WorkingDirectory, result.StartInfo.Arguments, result.StdErr, result.StdOut));
             }
         }
+
+        // walk around for https://github.com/dotnet/corefx/issues/26488
+        // fallback osx.10.13 to osx
+        private static string GetRuntimeIdentifierWithMacOsHighSierraFallback()
+        {
+            if (RuntimeEnvironment.GetRuntimeIdentifier() == "osx.10.13-x64")
```
dasMulli: is this maybe a little too specific? when 10.14 comes out, we're probably going to have the same problems again..
What about testing for the Mac platform specifically?

wli3: This is a walkaround. The right fix should be wait until runtime.json have 10.13 in corefx, the same as when 10.14 come out in the future. Also I don’t think restore graph should not be a CLI concern, it should be in corefx. And tests are running in 10.12 machines. I do need to discuss with more people. Abort this walkaround and focus on runtime.json might be the right approach.

dasMulli: A problem here would be that the tool would need to be re-published with a new `Microsoft.NETCore.Platforms` package dependency for every new macOS version :(
Or is this dependency in the temp project? That would help.

tannergooding: I agree. With OSX in particular, there isn't a good way to downgrade if your backups are gone (that is, after 10.13 was published, all the 10.12 ISOs and install channels were removed).

We need a better story for handling it so that people don't wind up being unable to work after every minor revision change. (CC @KathleenDollard)

PR #8443, opened 2018-01-22
Generate a Nuget.config once

ghost:
If the Nuget.config has already been generated, do not re-generate the file every time 'Run-Build' or 'Build' is called.
If 'OrchestratedPackageVersionsProps.props' is present, do not re-download the file every time 'Run-Build' or 'Build' is called.

This is consistent with the treatment for: 
BuildInfo.props
GitCommitInfo.props
HostInfo.props

This change should repair the errors we are seeing in the 'AllLinuxDistrosNativeInstallers' leg of orchestrated build; to wit: "Failure at Linux leg because native installer tests are not honoring the package sources." 

The 'AllLinuxDistrosNativeInstallers' leg is markedly different in that there is another call to 'build.sh' here: https://github.com/dotnet/cli/blob/master/build/SandBoxAndPackage.proj#L50 
This call generates a new Nuget.config without the benefit of the 'ExternalRestoreSources' property.

### DIFF
--- a/build/DownloadPackageVersions.targets
+++ b/build/DownloadPackageVersions.targets
@@ -1,6 +1,7 @@
 <Project ToolsVersion="15.0">
   <Target Name="DownloadPackageVersionsProps"
-    DependsOnTargets="BuildDotnetCliBuildFramework">
+          Condition=" !Exists('$(OrchestratedPackageVersionsProps)') "
+          DependsOnTargets="BuildDotnetCliBuildFramework">
 
     <DownloadFile Uri="$(PB_PackageVersionPropsUrl)$(CoreSetupBlobAccessTokenParam)"
                   DestinationPath="$(OrchestratedPackageVersionsProps)"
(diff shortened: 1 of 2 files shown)


johnbeisner:
@mmitche 
@dagood 
@dotnet/dotnet-cli 
Can I get a code review on this.

dagood (approved):
Looks reasonable to me.

mmitche:
@johnbeisner Looks good, only one question...does this mean that if you had a repo that already ran the orchestrated build (e.g. reproing a failure locally) then a subsequent build attempting to use a different restore source or package version props file wouldn't pick up those new changes, because a generated nuget.config file was already sitting there?

dagood:
My expectation is that builds are always assumed to be clean when you want to change these things. For example, even now, you have to know which files to delete to go back to doing a non-orchestrated build (or just clean everything).

johnbeisner:
@mmitche
"if you had a repo that already ran the orchestrated build" - No, the preservation of a previous Nuget.config, et al is a build over build thing. When a new build on a clean or new local repo instance is started, a fresh set of property files, including a fresh Nuget.config, is created.

Issue #8445, opened 2018-01-22
[tool] tools install failed when the packageid and the command name are the same on unix


wli3:
repo: create tool package that has the same packageid and command name. Run dotnet install tool, it will fail with access denied.

The cause is since shim and package content are stored in the same folder. They will end up with same folder/file name. And unix treats folder as file. They conflict.

PR #8447, opened 2018-01-22
Separate tool package and shim file location

wli3:
**Customer scenario**

Consumer install a global tool on CLI that has the same command name as the package id, for example dotnet-watch. The install will fail with access denied error on macOS and Linux.

**Bugs this fixes**

https://github.com/dotnet/cli/issues/8445

**Workarounds, if any**

Republish the global tools with a different packageId

**Risk**

low

**Performance impact**

no

**Root cause analysis**

Since tools shim and package are stored in the same folder. If the tool command name and the package name is the same, a folder and a file with the same name will be created. And this is not allow in unix file system. More integration test coverage is needed.

**How was the bug found?**

Manual testing

### DIFF
--- a/src/Microsoft.DotNet.Configurer/CliFolderPathCalculator.cs
+++ b/src/Microsoft.DotNet.Configurer/CliFolderPathCalculator.cs
@@ -12,20 +12,23 @@ namespace Microsoft.DotNet.Configurer
 {
     public class CliFolderPathCalculator
     {
-        private const string ToolsFolderName = "tools";
+        // ToolsShimFolderName ToolPackageFolderName cannot be the same
+        // or if the PackageId is the same as CommandName, they will conflict on unix.
+        private const string ToolsShimFolderName = "tools";
+        private const string ToolPackageFolderName = "toolspkgs";
         private const string DotnetProfileDirectoryName = ".dotnet";
 
         public string CliFallbackFolderPath => Environment.GetEnvironmentVariable("DOTNET_CLI_TEST_FALLBACKFOLDER") ??
                                                Path.Combine(new DirectoryInfo(AppContext.BaseDirectory).Parent.FullName, "NuGetFallbackFolder");
         
-        public string ExecutablePackagesPath => Path.Combine(DotnetUserProfileFolderPath, ToolsFolderName);
-
-        public BashPathUnderHomeDirectory ExecutablePackagesPathInUnix
+        public string ToolsShimPath => Path.Combine(DotnetUserProfileFolderPath, ToolsShimFolderName);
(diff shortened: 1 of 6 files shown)


src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
```diff
@@ -49,12 +49,13 @@ public override int Execute()
             }
 
             var cliFolderPathCalculator = new CliFolderPathCalculator();
-            var executablePackagePath = new DirectoryPath(cliFolderPathCalculator.ExecutablePackagesPath);
             var offlineFeedPath = new DirectoryPath(cliFolderPathCalculator.CliFallbackFolderPath);
 
-            var toolConfigurationAndExecutablePath = ObtainPackage(executablePackagePath, offlineFeedPath);
+            var toolConfigurationAndExecutablePath = ObtainPackage(
+                executablePackagePath: new DirectoryPath(cliFolderPathCalculator.ToolsPackagePath),
```
wli3: Package obtainer use ToolsPackagePath, shell shim maker use ToolsShimPath

wli3:
@MattGertz could you review this change?

src/Microsoft.DotNet.Configurer/CliFolderPathCalculator.cs
```diff
@@ -12,20 +12,23 @@ namespace Microsoft.DotNet.Configurer
 {
     public class CliFolderPathCalculator
     {
-        private const string ToolsFolderName = "tools";
+        // ToolsShimFolderName ToolPackageFolderName cannot be the same
+        // or if the PackageId is the same as CommandName, they will conflict on unix.
+        private const string ToolsShimFolderName = "tools";
+        private const string ToolPackageFolderName = "toolspkg";
```
peterhuene: Nit: how about `toolpkgs`?  `toolspkg` seems a singular package to me.

wli3: i feel it is better for them to have the same prefix. and "tools" are from nuget's folder /tools (beside /lib)

wli3: also "pkg" as afraid of long path

peterhuene: Compromise on `toolspkgs`? 😄 I can also live with this as is.

wli3: done :)

wli3:

PR #8448, opened 2018-01-23
Use a unchanging version for 'Microsoft.NETCore.App'

ghost:
The orchestrated "PackageVersion" type variables should not be used in any of the 'proj' files of the built build tools under: "build_projects"

Why? The build-tools restore & compile first and are used in the download of the orchestrated package version props file - version variables can be used, however, they cannot change.

### DIFF
--- a/build_projects/dotnet-cli-build/dotnet-cli-build.csproj
+++ b/build_projects/dotnet-cli-build/dotnet-cli-build.csproj
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppPackageVersion)" />
+    <PackageReference Include="Microsoft.NETCore.App" Version="2.1.0-preview1-26116-04" />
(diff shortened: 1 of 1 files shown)


build_projects/dotnet-cli-build/dotnet-cli-build.csproj
```diff
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppPackageVersion)" />
+    <PackageReference Include="Microsoft.NETCore.App" Version="2.1.0-preview1-26116-04" />
```
dagood: I'm surprised to see so many inline package versions here--even when these are static, shouldn't they still be centralized in DependencyVersions.props like VersionTools and the Feed package?

mmitche: Yeah I agree with @dagood, but I think that doesn't block this check-in (since there are a bunch of other non-centralized properties here).

mmitche (approved):

PR #8452, opened 2018-01-23
Nuget.config has to be re-generated if we have downloaded a versions props file.

ghost:


mmitche (approved):

dagood (approved):
Seems to go back on the idea that these artifacts should be generated only once on a fresh clone build, but I'm guessing this is a pragmatic change.

PR #8460, opened 2018-01-24
Fix non-fatal null exception when no extra parameters are passed.

ghost:
$ExtraParameters.split() throws a non-fatal null exception when build.cmd is called with no 'extra' parameters.


livarcocc (approved):

nguerrera (approved):

nguerrera:
Was about to nag about this. Thanks for fixing. :)

PR #8480, opened 2018-01-25
Fix test folder cannot delete and missing test

wli3:
There is a bug that prevent folder with space in the end to be deleted on corefx. I will file it later.

The space in the test is needed due to shim used to cannot handle space in the be PATH.

### DIFF
--- a/test/Microsoft.DotNet.ShellShim.Tests/ShellShimMakerTests.cs
+++ b/test/Microsoft.DotNet.ShellShim.Tests/ShellShimMakerTests.cs
@@ -93,6 +93,7 @@ namespace Microsoft.DotNet.ShellShim.Tests
             }
         }
 
+        [Theory]
         [InlineData(false)]
         [InlineData(true)]
         public void GivenAnExecutablePathWithExistingSameNameShimItThrows(bool testMockBehaviorIsInSync)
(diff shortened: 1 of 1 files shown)


test/Microsoft.DotNet.ShellShim.Tests/ShellShimMakerTests.cs
```diff
@@ -184,7 +185,7 @@ private static FileInfo MakeHelloWorldExecutableDll()
             const string testAppName = "TestAppSimple";
             const string emptySpaceToTestSpaceInPath = " ";
             TestAssetInstance testInstance = TestAssets.Get(testAppName)
-                .CreateInstance(testAppName + emptySpaceToTestSpaceInPath)
+                .CreateInstance(testAppName + emptySpaceToTestSpaceInPath + "test")
```
wli3: just put something at the end, don't make a folder with space at the end.

test/Microsoft.DotNet.ShellShim.Tests/ShellShimMakerTests.cs
```diff
@@ -93,6 +93,7 @@ public void GivenAShimItPassesThroughArguments(string arguments, string[] expect
             }
         }
 
+        [Theory]
```
wli3: A bad merge.

johnbeisner:
@wil3
Does this need to go in release/2.1.3xx?

wli3:
@johnbeisner I don't think so, it is problem on test, the product is fine. And the missing theory is not in 2.1.3xx at all. It is a fact in that branch

wil3:
Hi, you got the wong username

PR #8484, opened 2018-01-26
Merge install scripts from master to release/2.0.0

ghost:
Typically, CLI:release/2.0.0 holds the latest version of the install scripts. Recently, with the introduction of Alpine, et al, CLI:master holds the latest versions. This will update release/2.0.0

### DIFF
--- a/scripts/obtain/dotnet-install.ps1
+++ b/scripts/obtain/dotnet-install.ps1
@@ -34,8 +34,17 @@
     Architecture of dotnet binaries to be installed.
     Possible values are: <auto>, x64 and x86
 .PARAMETER SharedRuntime
+    This parameter is obsolete and may be removed in a future version of this script.
+    The recommended alternative is '-Runtime dotnet'.
+
     Default: false
-    Installs just the shared runtime bits, not the entire SDK
+    Installs just the shared runtime bits, not the entire SDK.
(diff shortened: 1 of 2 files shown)


scripts/obtain/dotnet-install.ps1
```diff
@@ -34,8 +34,17 @@
...
+
     Default: false
-    Installs just the shared runtime bits, not the entire SDK
+    Installs just the shared runtime bits, not the entire SDK.
+    This is equivalent to specifying `-Runtime dotnet`.
+.PARAMETER Runtime
+    Installs just a shared runtime, not the entire SDK.
+    Possible values:
+        - dotnet
+        - aspnetcore
```
natemcmaster: Just to be clear, there is no ASP.NET core shared runtime for 2.0.x, so this doesn't really apply to this branch.

johnbeisner: @natemcmaster
The script is intended to work in ALL branches - is it your stance that the recent additions for ASP.NET are branch specific?

natemcmaster: The script should work just fine on this branch, but users will have to specify `-channel master` or `-version 2.1.0-something` for installing aspnetcore to work.

natemcmaster (approved):

PR #8488, opened 2018-01-26
Repairing the first pass call to 'dotnet msbuild'; CLI:master

ghost:
'$ExtraParametersNoTargets', which is used on the first pass call to 'dotnet msbuild', currently is of type 'string' not 'List'1' as is '$ExtraParameters'. This results in the non-honoring of any parameter other than parameter one [after the removal of targets and pure parameters]. 

Solution: Make a copy of '$ExtraParameters' to '$ExtraParametersNoTargets' of type 'List'1' and remove the targets from the list.

### DIFF
--- a/run-build.ps1
+++ b/run-build.ps1
@@ -101,6 +101,6 @@ if ($NoBuild)
 else
 {
     dotnet msbuild build.proj /p:Architecture=$Architecture /p:GeneratePropsFile=true /t:WriteDynamicPropsToStaticPropsFiles $ExtraParametersNoTargets
-
+    dotnet msbuild build.proj /m /v:normal /fl /flp:v=diag /p:Architecture=$Architecture $ExtraParameters
     if($LASTEXITCODE -ne 0) { throw "Failed to build" } 
 }
--- a/run-build.ps1
+++ b/run-build.ps1
@@ -36,7 +36,7 @@ if ($ExtraParameters)
     {
         if(($param.StartsWith("/t")) -or ($param.StartsWith("/T")))
         {
-            $ExtraParametersNoTargets.Remove("$param")
+            $remove = $ExtraParametersNoTargets.Remove("$param")
         }
     }
 }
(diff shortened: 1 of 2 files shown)


johnbeisner:
@dotnet/dotnet-cli 
@mmitche 
@dagood 
Can I get a code review please.

run-build.ps1
```diff
@@ -29,12 +29,15 @@ if($Help)
...
+if ($ExtraParameters)
 {
-    if((-not $param.StartsWith("/t")) -and (-not $param.StartsWith("/T")))
+    $ExtraParametersNoTargets = $ExtraParameters.GetRange(0,$ExtraParameters.Count)
+    foreach ($param in $ExtraParameters)
     {
-        $ExtraParametersNoTargets += "{0} " -f $param
+        if(($param.StartsWith("/t")) -or ($param.StartsWith("/T")))
+        {
+            $remove = $ExtraParametersNoTargets.Remove("$param")
```
mmitche: Where does this $remove variable go to?

johnbeisner: Nowhere right now - if it is not there, the Boolean is echoed to the console/logs.

run-build.ps1
```diff
@@ -29,12 +29,15 @@ if($Help)
...
-$ExtraParametersNoTargets = ""
-foreach ($param in $ExtraParameters.split())
+if ($ExtraParameters)
 {
-    if((-not $param.StartsWith("/t")) -and (-not $param.StartsWith("/T")))
+    $ExtraParametersNoTargets = $ExtraParameters.GetRange(0,$ExtraParameters.Count)
+    foreach ($param in $ExtraParameters)
     {
-        $ExtraParametersNoTargets += "{0} " -f $param
+        if(($param.StartsWith("/t")) -or ($param.StartsWith("/T")))
```
dagood: `$param.StartsWith("/t", [StringComparison]::OrdinalIgnoreCase)`?

run-build.ps1
mmitche: @johnbeisner you can also accomplish this by:

`$ExtraParametersNoTargets.Remove("$param") | Out-Null`

run-build.ps1
dagood: MSBuild only accepts `/t:` and `/target:` (as far as I can tell it isn't doing fancy partial matches), so even though nobody will probably use `/toolsversion:` (`/tv:`) you could avoid the ambiguity by comparing both exactly.

dagood (approved):

PR #8504, opened 2018-01-30
Orchestrated final publish; CLI:master

ghost:
Reference: https://github.com/dotnet/core-eng/issues/2407

### DIFF
--- a/build/Publish.targets
+++ b/build/Publish.targets
@@ -84,7 +84,7 @@
                     ManifestBuildId="$(FullNugetVersion)"
                     ManifestCommit="$(CommitHash)"
                     ManifestName="$(BuildName)"
-                    ManifestBuildData="$(Product)=$(FullNugetVersion)"
+                    ManifestBuildData="ProductVersion=$(FullNugetVersion)"
                     Overwrite="false" 
                     PublishFlatContainer="true" 
                     SkipCreateManifest="$(IsNotOrchestratedPublish)" />
--- a/build/Publish.targets
+++ b/build/Publish.targets
@@ -84,6 +84,7 @@
                     ManifestBuildId="$(FullNugetVersion)"
                     ManifestCommit="$(CommitHash)"
                     ManifestName="$(BuildName)"
+                    ManifestBuildData="$(Product)=$(FullNugetVersion)"
                     Overwrite="false" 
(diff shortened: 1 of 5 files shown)


johnbeisner:
@dagood 
Please take a look at this please.

build/Publish.targets
```diff
@@ -84,6 +84,7 @@
                     ManifestBuildId="$(FullNugetVersion)"
                     ManifestCommit="$(CommitHash)"
                     ManifestName="$(BuildName)"
+                    ManifestBuildData="$(Product)=$(FullNugetVersion)"
```
dagood: These should just be `"ProductVersion=$(FullNugetVersion)"`.

The final publish step actually looks for the build name `cli`, grabs the `ProductVersion` it specifies, and knows that `SDK` is the right product name for `cli`. (I didn't think it was worthwhile to make it more dynamic than that.)

johnbeisner: What is the difference between "ManifestBuildId" vs. "ProductVersion"? Is it just happenstance that this is the same value in the CLI?

dagood: Yep, usually BuildId is something like 20180102-03, but it's up to you what makes sense to see for it in the manifest. ProductVersion must be that second virtual dir segment.

You can actually remove ManifestBuildId now that ProductVersion will carry the info.

I could have changed Core-Setup and ASP.NET to pass in their product version as BuildId and called it done, but that would overload the BuildId term which is already pretty well-known in Core repos IMO. I also didn't want to get *too* specialized for each repo by grabbing BuildId for CLI and ProductVersion for core-setup and aspnet.

dagood (approved):
LGTM, up to you if you want to remove ManifestBuildId.

johnbeisner:
@dagood 
"up to you if you want to remove ManifestBuildId" - why would I remove it? Is it not necessary anymore?

dagood:
Correct, `ManifestBuildId` is just informational, and since the information will be passed in `ProductVersion` you don't need to also pass it using `ManifestBuildId`. (But it also doesn't hurt anything to pass it in both ways.)

dagood:
@johnbeisner When will this change be in the 2.1 prodcon builds (release/2.1.3xx)?

/cc @MichaelSimons

PR #8518, opened 2018-02-01
Atomic install tool

wli3:
https://github.com/dotnet/cli/issues/8496

Apply TransactionScope to tool install. It can handle the correct timing
of roll back and commit.

Convert existing ToolPackageObtainer and ShellShimMaker by passing logic
via lambda to an object that has IEnlistmentNotification interface. It
turns out the very clean.

Use .stage as staging place to verify of package content, and shim. It
should roll back when something is wrong. When there is ctrl-c, there
will be garbage in .stage folder but not the root of the package folder.


test/Microsoft.DotNet.ShellShim.Tests/ShellShimMakerTests.cs
```diff
@@ -38,15 +39,16 @@ public void GivenAnRunnerOrEntryPointItCanCreateConfig(string entryPointPath, st
             if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 return;
 
-            var shellShimMaker = new ShellShimMaker(TempRoot.Root);
+            var cleanFolderUnderTempRoot = GetNewCleanFolderUnderTempRoot();
+            var shellShimMaker = new ShellShimMaker(cleanFolderUnderTempRoot);
 
-            var tmpFile = Path.Combine(TempRoot.Root, Path.GetRandomFileName());
+            var tmpFile = new FilePath(Path.Combine(cleanFolderUnderTempRoot, Path.GetRandomFileName()));
```
wli3: All these small lines are due to I need a cleanFolderUnderTempRoot, so I can asset "there is nothing left on disk if install failed"

test/Microsoft.DotNet.ToolPackage.Tests/Microsoft.DotNet.ToolPackage.Tests.csproj
```diff
@@ -41,8 +41,6 @@
     <None Remove="SampleGlobalTool/**" />
   </ItemGroup>
 
-  <UsingTask TaskName="DownloadFile" AssemblyFile="$(CLIBuildDll)" />
```
wli3: Not important for this change, move download to test setup instead of msbuild. But there are several merge conflict so I keep it together

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
```diff
@@ -215,30 +227,6 @@ public void GivenAllButNoPackageVersionItCanDownloadThePackage(bool testMockBeha
             File.Delete(executable.Value);
         }
 
-        [Theory]
-        [InlineData(false)]
-        [InlineData(true)]
-        public void GivenAllButNoPackageVersionAndInvokeTwiceItShouldNotThrow(bool testMockBehaviorIsInSync)
```
wli3: No longer true

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
```diff
@@ -387,7 +469,7 @@ public void GivenASourceItCanObtainThePackageFromThatSource(bool testMockBehavio
                         {
                             new MockFeed
                             {
-                                Type = MockFeedType.ExplicitNugetConfig,
+                                Type = MockFeedType.Source,
```
wli3: This is a test bug.

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageObtainerTests.cs
```diff
@@ -429,10 +512,47 @@ private static FilePath WriteNugetConfigFileToPointToTheFeed()
             return new FilePath(Path.GetFullPath(Path.Combine(tempPathForNugetConfigWithWhiteSpace, nugetConfigName)));
         }
 
+        private static void DownloadPlatformsPackage()
```
wli3: Not relate, download from set up

wli3:
@dotnet/dotnet-cli it is good to review. There is some weird merge conflict.

src/dotnet/xlf/CommonLocalizableStrings.cs.xlf
```diff
@@ -797,6 +797,11 @@ setx PATH "%PATH%;{1}"</target>
         <target state="new">Package '{0}' is missing tool settings file DotnetToolSettings.xml.</target>
         <note />
       </trans-unit>
+      <trans-unit id="ToolPackageConflictPackageId">
+        <source>A tool with the same PackageId {0} existed.</source>
+        <target state="new">A tool with the same PackageId {0} existed.</target>
```
peterhuene: Perhaps this should be `Tool '{0}' is already installed.`?

Also, do we want `dotnet install tool` to be idempotent with regards to installing a currently installed tool package? That is to say, should it return a successful exit code if the tool is already installed rather than an error?

src/dotnet/ShellShim/ShimCreationTransaction.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Transactions;
+
+namespace Microsoft.DotNet.ShellShim
+{
+    public class ShimCreationTransaction : IEnlistmentNotification
```
peterhuene: Could `ShimCreationTransaction` and `ObtainTransaction` perhaps be generalized?  They seem to be near duplicates of one another, except that the latter has a  rollback action, which I think is missing from shim creation (rollback should delete the shim files; for example, what happens if we create the `.exe` but fail to create the `.config` file, or vice versa?).

On a side note: I like this use of the `TransactionScope` and `IEnlistmentNotification`.

src/dotnet/xlf/CommonLocalizableStrings.cs.xlf
wli3: @KathleenDollard for this string -- >"A tool with the same PackageId {0} existed."

I just remember one talk on the internet saying passive voice is considered harmful.. but no strong opinion or opinion at all

src/dotnet/ShellShim/ShimCreationTransaction.cs
wli3: Generalize `ShimCreationTransaction`, `ObtainTransaction` -- they are very likely could,  it is almost java's anonymous class. but i cannot find a good place to put the shared code. Any suggestion?

ShimCreationTransaction is treaty. In the beginning, i put createshim as a separate method, and roll back is delete. And then i realize rollback is deleting the existing shim (the conflicted one). Then I realize that is wrong, it need 2 stage. What about put in .stage and Directory.move.? -- the whole shim creation is a Directory.Move. So I just move them to commit and add verification in prepare.

At the same time, it fits this guideline very well. https://stackoverflow.com/questions/21094406/where-should-i-perform-action-when-implementing-ienlistmentnotification

peterhuene:
So the more I read about implementing two-phase transactional resource managers using `IEnlistmentNotification`, it seems like this PR is implementing it differently than what I would expect.

I think we need to view `ToolPackageObtainer` and `ShellShimMaker` as resource managers; in fact, in my uninstall command work, they have been refactored to be managers so that they can both install and uninstall tool packages / shell shims.  To this effect, I would expect these types to implement `IEnlistNotification` and internally keep a log of completed actions (such as "created a top-level package directory", "created a shim file", etc.) that can be "undone" if a rollback is requested.

I don't think work should be done in the `Prepare` and `Commit` methods.  The `Prepare` method exists to notify the transaction manager on our "vote" to either commit the transaction or abandon it.  The `Commit` method would then simply clear the action log since all actions have completed successfully.  The `Rollback` method would then "undo" the action log and clear it.

Thus, I'd expect the transaction scope to look something like this:

```c#
using (var scope = new TransactionScope())
{
    var packageObtainer = new ToolPackageObtainer(...);
    var shellShimMaker = new ShellShimMaker(...);

    // These could even move to the constructors (`if (Transaction.Current != null) { ...EnlistVolatile(this, ...) }`
    Transaction.Current.EnlistVolatile(packageObtainer, EnlistmentOptions.None);
    Transaction.Current.EnlistVolatile(shellShimMaker, EnlistmentOptions.None);

    // Internally, the obtainer would record the fact the tool package directory was created
    var result = packageObtainer.ObtainAndReturnExecutablePath(...);

    ...

   // Internally, the maker would record an action in its log for each file it creates successfully
   shellShimMaker.CreateShim(...);

    ...
}
```

If an exception is thrown inside the transaction scope, both the obtainer and the shell shim maker would have their `Rollback` methods called, resulting in the action logs being "undone" (i.e. "delete a top-level package directory", "delete a shell shim file", etc).

Does this make sense?

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -31,9 +32,54 @@ IProjectRestorer projectRestorer
             _projectRestorer = projectRestorer ?? throw new ArgumentNullException(nameof(projectRestorer));
             _toolsPath = toolsPath;
             _offlineFeedPath = offlineFeedPath;
+
+            _stageDirectory = _toolsPath.WithSubDirectories(".stage", Path.GetRandomFileName());
```
wli3: Need to move down to CreateObtainTransaction method, make it a local, and let closure capture it

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -31,9 +32,54 @@ IProjectRestorer projectRestorer
...
 
-        public ToolConfigurationAndExecutablePath ObtainAndReturnExecutablePath(
+        public ObtainTransaction CreateObtainTransaction(
+            string packageId,
+            string packageVersion = null,
+            FilePath? nugetconfig = null,
+            string targetframework = null,
+            string source = null)
+        {
+            return new ObtainTransaction(
```
wli3: + stageDirectory = _toolsPath.WithSubDirectories(".stage", Path.GetRandomFileName());

peterhuene:
After talking with Will offline, I thought it'd be clearer to show a code example of what I was describing above as it was hard to put into words alone: https://gist.github.com/peterhuene/a95b370f8bf84ee4b68be7b94de628fe

This is not to say we have to do something like this, but it simply demonstrates my preference for leaving the `IPackageObtainer` and `IShellShimMaker` interfaces alone and let the implementations handle the transactional nature of their operations.

src/dotnet/xlf/CommonLocalizableStrings.cs.xlf
wli3: Changed to Tool '{0}' is already installed.

wli3:
The conclusion of the discussion is to keep the original interface. And detect whether it is in transaction or not. If it is in transaction, add resource manager object to enlistment. If not create a transaction and do work in transaction of one. The reason to create a transaction anyway is we don't want to diverge the code and maintain 2 version of it. Maybe there is downside of perf, however, compared with downloading a package, it is trivial.

wli3:
@peterhuene

src/dotnet/ShellShim/ShellShimMaker.cs
```diff
@@ -23,10 +25,37 @@ public class ShellShimMaker : IShellShimMaker
...
+            var resourceManager = new TransactionResourceManager(
+                commit: () => { PlaceShim(packageExecutable, shellCommandName); },
+                prepare: preparingEnlistment =>
+                {
+                    EnsureCommandNameUniqueness(shellCommandName);
+
+                    preparingEnlistment.Prepared();
+                });
+
+            if (Transaction.Current != null)
```
peterhuene: So I think we should just always create a transaction scope rather than check to see if there's currently a transaction.  Based on the documentation and some experimentation I did, the default mode of the transaction scope will be to inherit an existing transaction or create a new one.  If it inherits the existing one, `Complete()` is a no-op.  If it creates the transaction, `Complete()` attempts to commit the transaction.

I think this would just be cleaner code than having a check and duplicate `EnlistVolatile` calls.

src/dotnet/ShellShim/ShellShimMaker.cs
```diff
@@ -23,10 +25,37 @@ public class ShellShimMaker : IShellShimMaker
...
             _pathToPlaceShim =
-                pathToPlaceShim ?? throw new ArgumentNullException(nameof(pathToPlaceShim));
+                pathToPlaceShim ??
+                throw new ArgumentNullException(nameof(pathToPlaceShim));
         }
 
         public void CreateShim(FilePath packageExecutable, string shellCommandName)
+        {
+            var resourceManager = new TransactionResourceManager(
+                commit: () => { PlaceShim(packageExecutable, shellCommandName); },
```
peterhuene: So in my experimentation, a transaction which throws from `Commit` does not rollback any enlistment; instead the exception goes uncaught by the transaction manager.

I think the proper place to do the work is in `Prepare`. `Commit` should be treated as a notification that the transaction is committed; it should not fail, only do whatever internal bookmarking is needed when the transaction is committed.

src/dotnet/ShellShim/ShellShimMaker.cs
```diff
@@ -23,10 +25,37 @@ public class ShellShimMaker : IShellShimMaker
         public ShellShimMaker(string pathToPlaceShim)
         {
             _pathToPlaceShim =
-                pathToPlaceShim ?? throw new ArgumentNullException(nameof(pathToPlaceShim));
+                pathToPlaceShim ??
+                throw new ArgumentNullException(nameof(pathToPlaceShim));
```
peterhuene: Nit: indentation level for line continuation.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -42,6 +41,79 @@ IProjectRestorer projectRestorer
...
+                        stageDirectory.WithSubDirectories(packageId).Value,
+                        _toolsPath.WithSubDirectories(packageId).Value);
+
+                    Directory.Delete(stageDirectory.Value, true);
+                },
+                prepare: preparingEnlistment =>
+                {
+                    if (Directory.Exists(_toolsPath.WithSubDirectories(packageId).Value))
+                    {
+                        preparingEnlistment.ForceRollback();
```
peterhuene: From my experimentation, this call isn't necessary; the transaction manager will rollback all enlistments from a throwing `Prepare` call.

src/dotnet/TransactionResourceManager.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Transactions;
+
+namespace Microsoft.DotNet.Cli
+{
+    public class TransactionResourceManager : IEnlistmentNotification
```
peterhuene: This name is confusing to me as I don't think it is a "manager" since it isn't managing resources; fundamentally, it's a generic transaction enlistment mechanism.

src/dotnet/ShellShim/ShellShimMaker.cs
peterhuene: Also, think about what happens when a caller calls `CreateShim` twice with the same `shellCommandName` parameter.  We enlist two notifications that check for the existence of the same file; both checks will succeed, however, the second actual commit will fail because the files exist.

Either the work needs be done in `CreateShim` and it enlists only undo actions (my preference), or the work needs to be done in `Prepare`.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -42,6 +41,79 @@ IProjectRestorer projectRestorer
...
+            var stageDirectory = _toolsPath.WithSubDirectories(".stage", Path.GetRandomFileName());
+
+            var resourceManager = new TransactionResourceManager(
+                commit: () =>
+                {
+                    Directory.Move(
+                        stageDirectory.WithSubDirectories(packageId).Value,
+                        _toolsPath.WithSubDirectories(packageId).Value);
+
+                    Directory.Delete(stageDirectory.Value, true);
```
peterhuene: This shouldn't be needed following a successful move, right?

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -42,6 +41,79 @@ IProjectRestorer projectRestorer
             string targetframework = null,
             string source = null,
             string verbosity = null)
+        {
+            var stageDirectory = _toolsPath.WithSubDirectories(".stage", Path.GetRandomFileName());
+
+            var resourceManager = new TransactionResourceManager(
+                commit: () =>
+                {
+                    Directory.Move(
```
peterhuene: Same general comments as above.  If this move or delete operation fails, the transaction will not be rolled back and we'll be left in an inconsistent state.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
peterhuene: Oh never mind, we are restoring into a sub directory of this directory.  Is there a reason we can't just restore into the randomly-named stage directory itself?

src/dotnet/TransactionResourceManager.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public void InDoubt(Enlistment enlistment)
+        {
+            Rollback(enlistment);
+        }
+
+        public void Prepare(PreparingEnlistment preparingEnlistment)
+        {
+            _prepare(preparingEnlistment);
```
peterhuene: I think it's a little strange that for `Prepare` we expect the delegated action to call `Prepared()`, but for the other actions, we implicitly handle calling `Done()`.

src/dotnet/dotnet.csproj
```diff
@@ -77,4 +77,4 @@
     <Folder Include="dotnet-complete\commands\" />
   </ItemGroup>
   <Import Project="dotnet.win.targets" Condition="'$(OS)' == 'Windows_NT'" />
-</Project>
```
peterhuene: Do we need to commit this change?

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: We did, move aasjdhal/my.tool/1.1 to .toolpkg/my.tool/1.1, and delete aasjdhal. We cannot just rename the random folder, there is more under .toolpkg

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Let me try and see if my test passes

src/dotnet/TransactionResourceManager.cs
wli3: I was wonder about that for a well. I hate "XXmanager" as well. But "ResourceManger" is a specific term that means object implements IEnlistmentNotification. Maybe I will just call it "TransactionEnlistment" ?

src/dotnet/ShellShim/ShellShimMaker.cs
wli3: Agree, and this step will not fail. All is checked out during prepare type. This step is just copy the embedded file and write config.

If this a step, the roll back could delete the conflicted file(what used to be there). To avoid it, we need write and rename. This is really no too much different than copy embedded file

src/dotnet/TransactionResourceManager.cs
peterhuene: Possible suggestion of `TransactionNotification`?  I'd like to see the `EnlistVolatile` call encapsulated, but that's just my personal preference; I'm also fine with having the users of this type having to explicitly call `EnlistVolatile`.

src/dotnet/ShellShim/ShellShimMaker.cs
wli3: Let me try to remove this check, tests will catch if it doesn't work

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Yes and it should not fail. If there is uncontrollable case like control-c, nothing can prevent this inconsistency.  We can only rely on `install --force` to correct it.

src/dotnet/TransactionResourceManager.cs
wli3: Adding enlistment to all signature

src/dotnet/dotnet.csproj
wli3: reverting

src/dotnet/TransactionResourceManager.cs
peterhuene: I'm good with `TransactionEnlistment`.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: I think it is common pattern to use Move as a confirm for file system, there is no real commit as in database.

Also, it is way more complicated if we don't trust move. Rollback need to be different before move and after move. Also, how do we trust Delete then?

src/dotnet/ToolPackage/ToolPackageObtainer.cs
peterhuene: Why restore to a sub-directory based on the package id at all if we're operating in a randomly named root directory in the first place?  I guess I'm not seeing why `.staged/<random>` -> `.toolspkg/<package-id>` wouldn't work.  It's no different from restoring into `.toolspkg/<package-id>` in the first place, right?

src/dotnet/ToolPackage/ToolPackageObtainer.cs
peterhuene: But unlike the "process killed" inconsistency (one we can never protect ourselves from), this is a preventable one; if we failed to prepare the transaction (i.e. because we couldn't move from staging), then the transaction should be rolled back.  The way this is implemented now won't roll back the transaction.

I like to also speak towards the "correctness" of the implementation.  Not that we'll ever do so, but if we create a transaction and perform the same operation twice (such as install the same package or create the same shim), then I'd expect the second operation to fail and rollback the entire transaction.  Right now this won't happen because both `Prepare` calls will succeed, but both `Commit` calls will not.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
peterhuene: Actually, I am wrong here.  I was testing my gist code which does handle this correctly; it is not inherent in the transaction manager logic.  If the `Prepare` call fails, we should call `ForceRollback` with an exception that will be raised as an inner exception to `TransactionAbortedException` by the transaction manager after the rollback completes.  The enlistment that aborts the transaction will not be rolled back, so it needs to be invoked manually.

This is what I had in the gist:

```c#
try
{
    Prepare();
}
catch (Exception ex)
{
    Undo();
    preparingEnlistment.ForceRollback(ex);
    return;
}

preparingEnlistment.Prepared();
```

wli3: should we keep throwing it? instead of retrun

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: what you suggested also works

src/dotnet/ToolPackage/ToolPackageObtainer.cs
peterhuene: With my experimentation, it appears that the transaction manager doesn't expect `Prepare` to throw; instead it expects a call to `ForceRollback` (without giving it an exception, it'll throw a `TransactionAbortedException` with no inner exception).  If `Prepare` throws the transaction is not rolled back.

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: > this is a preventable one

In prepare, the existence of a folder with the same name already checked. I don't know what else can cause this to fail. Other than issue like process killed, permission denied (it is a folder dotnet created)

Again, how do you prevent this happen when Move does not considered commit.

There is an existing package
Download package to staging
Diectory.Move failed, due to some reason
Rollback, delete .dotnet/packageid, so the existing proper package get deleted.

> Right now this won't happen because both Prepare calls will succeed, but both Commit calls will not.

Prepare **will** fail, it checks the folder existence or not

peterhuene: Right, but what about:

```c#
using (var scope = new TransactionScope())
{
    var obtainer = new ToolPackageObtainer(...);
    obtainer.ObtainAndReturnExecutablePath("foo");
    obtainer.ObtainAndReturnExecutablePath("foo");
    scope.Complete();
}
```

The order of operations is:

1. Prepare: check for existence of "foo" package directory...succeeds.
2. Prepare: check for existence of "foo" package directory...succeeds.
3. Commit: move directory...succeeds.
4. Commit: move directory...fails and transaction is not rolled back.

Now I'm not saying we need to support this, I'm just arguing from a correctness standpoint, I'd expect this to fail and rollback as if I had inserted two rows into a database with the same primary key in the same transaction.  Neither insert should remain.

If the `Prepare` call did the move, then the subsequent Prepare call would detect it, fail, and rollback the entire transaction.

wli3: Decide on empty commit and empty prepare. Use a non generic IEnlistmentNotification to track all the state of transaction, including it is before move or after move

src/dotnet/ShellShim/ShellShimMaker.cs
wli3: Removed the check, create Transaction anyway

src/dotnet/ShellShim/ShellShimMaker.cs
wli3: Commit is empty now

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Rollback is empty now

src/dotnet/TransactionResourceManager.cs
wli3: ToolPackageObtainTransaction and CreateShimTransaction

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Changed to what you suggested

src/dotnet/ToolPackage/ToolPackageObtainer.cs
wli3: Changed to what we decide on, please review

src/dotnet/TransactionResourceManager.cs
wli3: no longer needed, since they are not generic anymore

src/dotnet/dotnet.csproj
wli3: fixed

wli3:

src/dotnet/ToolPackage/ToolPackageObtainer.cs
```diff
@@ -65,34 +127,34 @@ IProjectRestorer projectRestorer
 
             var packageVersionOrPlaceHolder = new PackageVersion(packageVersion);
 
-            DirectoryPath toolDirectory =
-                CreateIndividualToolVersionDirectory(packageId, packageVersionOrPlaceHolder);
+            DirectoryPath nugetSandboxDirector =
```
peterhuene: Nit: missing `y` on `nugetSandboxDirector`.

wli3:

Issue #8421, opened 2018-01-18
Change the name of NETCorePlatformsImplicitPackageVersion


wli3:
to bundled

- [x] change on SDK
- [x] change on CLI

wli3:
@nguerrera

nguerrera:
Thanks

PR #8538, opened 2018-02-05
Update SDK version and Rename to BundledNETCorePlatformsPackageVersion

wli3:
This is a continuation of https://github.com/dotnet/sdk/pull/1928 and https://github.com/dotnet/cli/issues/8421

I need to update the sdk version (has the renaming change) and rename it on CLI side in one PR

### DIFF
--- a/build/MSBuildExtensions.targets
+++ b/build/MSBuildExtensions.targets
@@ -97,23 +97,25 @@
       of our build against Microsoft.NETCore.App to find the correct NETStandard.Library version
     -->
     <ItemGroup>
-      <_NETStandardLibraryVersions Include="@(PackageDefinitions->'%(Version)')"
+      <_NETStandardLibraryPackageVersions Include="@(PackageDefinitions->'%(Version)')"
                                    Condition="%(PackageDefinitions.Name) == 'NetStandard.Library'" />
-      <_NETCorePlatformsImplicitPackageVersion Include="@(PackageDefinitions->'%(Version)')"
+      <_NETCorePlatformsPackageVersions Include="@(PackageDefinitions->'%(Version)')"
                                    Condition="%(PackageDefinitions.Name) == 'Microsoft.NETCore.Platforms'" />
     </ItemGroup>
 
-    <Error Condition="@(_NETStandardLibraryVersions->Distinct()->Count()) != 1"
+    <Error Condition="@(_NETStandardLibraryPackageVersions->Distinct()->Count()) != 1"
            Text="Failed to determine the NETStandard.Library version pulled in Microsoft.NETCore.App" />
+    <Error Condition="@(_NETCorePlatformsPackageVersions->Distinct()->Count()) != 1"
+           Text="Failed to determine the Microsoft.NETCore.Platforms version pulled in Microsoft.NETCore.App" />
 
     <PropertyGroup>
       <_NETCoreAppPackageVersion>$(MicrosoftNETCoreAppPackageVersion)</_NETCoreAppPackageVersion>
-      <_NETStandardPackageVersion>@(_NETStandardLibraryVersions->Distinct())</_NETStandardPackageVersion>
-      <_NETCorePlatformsImplicitPackageVersion>@(_NETCorePlatformsImplicitPackageVersion->Distinct())</_NETCorePlatformsImplicitPackageVersion>
+      <_NETStandardLibraryPackageVersion>@(_NETStandardLibraryPackageVersions->Distinct())</_NETStandardLibraryPackageVersion>
+      <_NETCorePlatformsPackageVersion>@(_NETCorePlatformsPackageVersions->Distinct())</_NETCorePlatformsPackageVersion>
 
       <!-- Use only major and minor in target framework version -->
       <_NETCoreAppTargetFrameworkVersion>$(_NETCoreAppPackageVersion.Split('.')[0]).$(_NETCoreAppPackageVersion.Split('.')[1])</_NETCoreAppTargetFrameworkVersion>
-      <_NETStandardTargetFrameworkVersion>$(_NETStandardPackageVersion.Split('.')[0]).$(_NETStandardPackageVersion.Split('.')[1])</_NETStandardTargetFrameworkVersion>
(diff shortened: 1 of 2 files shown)


build/MSBuildExtensions.targets
```diff
@@ -99,7 +99,7 @@
     <ItemGroup>
       <_NETStandardLibraryVersions Include="@(PackageDefinitions->'%(Version)')"
                                    Condition="%(PackageDefinitions.Name) == 'NetStandard.Library'" />
-      <_NETCorePlatformsImplicitPackageVersion Include="@(PackageDefinitions->'%(Version)')"
+      <_BundledNETCorePlatformsVersion Include="@(PackageDefinitions->'%(Version)')"
```
nguerrera: Suggest that same naming convention as nearby: _NETCorePlatformsVersions

build/MSBuildExtensions.targets
```diff
@@ -109,7 +109,7 @@
     <PropertyGroup>
       <_NETCoreAppPackageVersion>$(MicrosoftNETCoreAppPackageVersion)</_NETCoreAppPackageVersion>
       <_NETStandardPackageVersion>@(_NETStandardLibraryVersions->Distinct())</_NETStandardPackageVersion>
-      <_NETCorePlatformsImplicitPackageVersion>@(_NETCorePlatformsImplicitPackageVersion->Distinct())</_NETCorePlatformsImplicitPackageVersion>
+      <_BundledNETCorePlatformsVersion>@(_BundledNETCorePlatformsVersion->Distinct())</_BundledNETCorePlatformsVersion>
```
nguerrera: Ditto: _NETCorePlatformsPackageVersion

build/MSBuildExtensions.targets
wli3: Sorry I was not careful of the naming pattern. I corrected it. It should be good now. However, I realize I am missing "Package" in "BundledNETCorePlatformsPackageVersion" so I need to sync up in SDK. Here is the SDK change https://github.com/dotnet/sdk/pull/1935/files

Once, there is a SDK build, I will remove the WIP and update the SDK version

wli3: There is a pattern of XLibraryVersions -> XPackageVersion -> BundledXPackageVersion

I think I caught all of them

build/MSBuildExtensions.targets
```diff
@@ -99,7 +99,7 @@
     <ItemGroup>
       <_NETStandardLibraryVersions Include="@(PackageDefinitions->'%(Version)')"
                                    Condition="%(PackageDefinitions.Name) == 'NetStandard.Library'" />
-      <_NETCorePlatformsImplicitPackageVersion Include="@(PackageDefinitions->'%(Version)')"
+      <_NETCorePlatformsLibraryVersions Include="@(PackageDefinitions->'%(Version)')"
```
nguerrera: Library was part of the package name for the other one.

wli3: But there is the following `<_NETStandardPackageVersion>@(_NETStandardLibraryVersions->Distinct())</_NETStandardPackageVersion>`

wli3: I am lost, if you vote for _NETCorePlatformsLibraryVersion**s**  -> _NETCorePlatformsPackageVersion**s** i will just go this way, both way make sense

nguerrera: OK, we were already inconsistent, but we should fix it.

NETStandard.**Library** -> _NETStandard**Library**PackageVersion**s**
(Microsoft.)NETCore.Platforms -> _NETCorePlatformsPackageVersion**s**

We were missing "Package" in the singular case. Library was not a synonym for package, it was coming from the package name.

build/MSBuildExtensions.targets
```diff
@@ -99,7 +99,7 @@
...
                                    Condition="%(PackageDefinitions.Name) == 'Microsoft.NETCore.Platforms'" />
     </ItemGroup>
```
nguerrera: Below here: https://github.com/dotnet/cli/blob/02a98d4e63e02928b8d359e4a0fff7d78530b67f/build/MSBuildExtensions.targets#L106-L107

There should also be an error if we do not find exactly one version of Microsoft.NETCore.Platforms

build/MSBuildExtensions.targets
```diff
@@ -133,8 +135,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <BundledNETCoreAppTargetFrameworkVersion>$(_NETCoreAppTargetFrameworkVersion)</BundledNETCoreAppTargetFrameworkVersion>
     <BundledNETCoreAppPackageVersion>$(_NETCoreAppPackageVersion)</BundledNETCoreAppPackageVersion>
     <BundledNETStandardTargetFrameworkVersion>$(_NETStandardTargetFrameworkVersion)</BundledNETStandardTargetFrameworkVersion>
-    <BundledNETStandardPackageVersion>$(_NETStandardPackageVersion)</BundledNETStandardPackageVersion>
-    <NETCorePlatformsImplicitPackageVersion>$(_NETCorePlatformsImplicitPackageVersion)</NETCorePlatformsImplicitPackageVersion>
+    <BundledNETStandardPackageVersion>$(_NETStandardLibraryPackageVersion)</BundledNETStandardPackageVersion>
```
wli3: I don't think we should unify the convention here, or it will be a breaking change. It should be BundledNETStandard**Library**PackageVersion instead of BundledNETStandardPackageVersion

nguerrera: Good point.

build/MSBuildExtensions.targets
wli3: Fixed

build/MSBuildExtensions.targets
wli3: I see, I corrected _NETStandardLibraryPackageVersions as well. However there is BundledNETStandardPackageVersion (no _Library_) below, i won't change it since it will introduce a breaking change.

wli3:
cc @nguerrera  I update the SDK version (that has the match property name) and removed the WIP. I will merge once CI finish

PR #8553, opened 2018-02-07
Fixing recent CLI test failures around ASP.NETcore in the orchestrated build

ghost:
Test: "RestoreDoesNotUseAnyCliProducedPackagesOnItsTemplates" should not use the 'patch' NuGet.tempaspnetpatch.config; but use the Nuget.config at the enlistment root as it contains the private feed entry.

Testing [compile and test]:
Using orchestrated build 20180205-02 inputs; all tests passed.


natemcmaster (approved):

test/dotnet-new.Tests/GivenThatIWantANewApp.cs
```diff
@@ -44,11 +44,7 @@ public void RestoreDoesNotUseAnyCliProducedPackagesOnItsTemplates()
             var rootPath = TestAssets.CreateTestDirectory().FullName;
             var packagesDirectory = Path.Combine(rootPath, "packages");
 
-            // For testing the 2.1 templates - some of their packages are currently only in private feeds.
-            var configFile = Path.Combine(rootPath, "NuGet.Config");
-            AspNetNuGetConfiguration.WriteNuGetConfigWithAspNetPrivateFeeds(configFile);
```
dagood: Is there a reason this config is ok in some kinds of [GivenThatIWantANewAppWithSpecifiedType.cs#L54](https://github.com/dotnet/cli/blob/e9b6381dcbb830e487560ce790875277ab27edf5/test/dotnet-new.Tests/GivenThatIWantANewAppWithSpecifiedType.cs#L54)? (I don't have a real grasp of the error or fix.)

johnbeisner: "GivenThatIWantANewAppWithSpecifiedType.cs" is not included in this change; I'm looking into removing the 'aspnetpatch' Nuget.config in two other tests.

@natemcmaster said:
"Microsoft.AspNetCore.App/2.1.0-preview1-28241 is in the LZMA and on the prodcon transport feed" - this particular test cannot access the prodcon transport feed if it is using a static NuGet.tempaspnetpatch.config; therefore, the test should use the Nuget.Config at the enlistment root as it has all the feeds in the static NuGet.tempaspnetpatch.config plus the prodcon transport feed.

dagood (approved):

PR #8660, opened 2018-02-22
Trigger build with unambiguous commit height

nguerrera:
The release/2.1.2xx had a few builds where it still had 2.1.100 branding. This causes 2.1.1xx builds to fail because they conflict with what 2.1.2xx published. Worse, the 2.1.2xx-produced builds look like they are 2.1.100 builds but have an inappropriate payload -- for example, a 4.7/15.7 daily build of nuget. 

We are going to be discussing moving away from commit height versioning tomorrow, but we need a build tonight that is suitable for inserting into Visual Studio tomorrow morning. So, as a workaround, this adds enough commits to get a version that is greater than any of the phony 2.1.100 builds that were produced out of the wrong branch. We also need to have a better process for updating branding whenever a release branch is created.

For obvious reasons, do not squash this when merging.


nguerrera:
@MattGertz FYI. Infrastucture only. Changes build number to get around the conflict I mentioned in email, but does not change a single line of code or the versions of any components.

MattGertz:
Groovy, understood, thanks.

PR #8692, opened 2018-02-27
Move some projects to netstandard2.0

jeffkl:
These were the changes I needed to make when inserting a new MSBuild that targets NETCoreApp2.1.  I'll follow this up with an MSBuild insertion PR.

* Use version agnostic $(TargetFrameworkIdentifier) property to make changing versions easier since we only care about .NET Framework vs .NET Standard
* Add missing project to solution file


jeffkl:
There are a few projects that still target .NETStandard 1.5.  I didn't need to update them to get MSBuild working so I didn't bother.  Should I update those as well?

jeffkl:
FYI @rainersigwald @AndyGerlicher

src/Microsoft.DotNet.Cli.Utils/Microsoft.DotNet.Cli.Utils.csproj
```diff
@@ -27,11 +27,12 @@
...
 </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.5' ">
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETStandard' ">
+    <PackageReference Include="NETStandard.Library" Version="2.0.0" />
     <PackageReference Include="System.Diagnostics.Process" Version="4.1.0" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net46' ">
+  <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' == '.NETFramework' ">
```
nguerrera: I'm not crazy about this pattern because it works for item groups but not property groups.

jeffkl: But is it worth it if you don't have to update every condition when you move from `netstandard2.0` to `netstandard2.1` or from `net46` to `net461`?  That's the reason I prefer it...

nguerrera: The subtlety of msbuild evaluation order is something I'd rather not think about when reading these conditions. If there's a mismatch, the bug is obvious. I worry that cargo cult will consider this "better", then scream "bug" when it fails on a property group. We should document and use the pattern that works for properties and items.

PR #8700, opened 2018-02-28
WIP Rpm agnstic without rename

wli3:
Do not use current directory in test. It is changed


danmosemsft:
Do the other test uses need fixes also?
```c#
C:\git\cli\test\dotnet.Tests\PackagedCommandTests.cs:
  404                  .Should().Pass();
  405  
  406:             var currentDirectory = Directory.GetCurrentDirectory();
  407  
  408              CommandResult result = new HelloCommand()

C:\git\cli\test\dotnet.Tests\ParserTests\AddReferenceParserTests.cs:
   34                  .Should()
   35                  .BeEquivalentTo(
   36:                     PathUtility.EnsureTrailingSlash(Directory.GetCurrentDirectory()));
   37          }
   38  

C:\git\cli\test\Microsoft.DotNet.Cli.Utils.Tests\BuiltInCommandTests.cs:
  119              public string GetWorkingDirectory()
  120              {
  121:                 return Directory.GetCurrentDirectory();
  122              }
```

If your tests run concurrently in one process, Set is problematic also:
```c#
C:\git\cli\test\Microsoft.DotNet.ToolPackage.Tests\ToolPackageInstallerTests.cs:
  163               * In test, we don't want NuGet to keep look up, so we point current directory to nugetconfig.
  164               */
  165:             Directory.SetCurrentDirectory(nugetConfigPath.GetDirectoryPath().Value);
  166  
  167              var package = installer.InstallPackage(

C:\git\cli\test\Microsoft.DotNet.Tools.Tests.Utilities\TempFileSystem\TempDirectory.cs:
  109          }
  110  
  111:         public void SetCurrentDirectory()
  112          {
  113:             Directory.SetCurrentDirectory(_path);
  114          }
```

wli3:
This error is different. Let me fix it.

per
C:\git\cli\test\Microsoft.DotNet.ToolPackage.Tests\ToolPackageInstallerTests.cs: 

that is part of the scenario. So I am not sure I should change it. And other test, I am hesitate to change them in this PR. I will log an issue and correct it

danmosemsft:
We have some tests (~1%) that do need to modify process wide state (like current directory) and we have a [mechanism ](https://github.com/dotnet/corefx/blob/master/src/CoreFx.Private.TestUtilities/src/System/Diagnostics/RemoteExecutorTestBase.cs) to launch child processes in the test. It is easy and you are welcome to copy it if that is helpful.

As a stop gap possibly you can disable parallelism for the assembly containing such tests.

wli3:
The latest error is even more weird. It happens in different place, but it uses the similar directory enumerate. Also no repo on my windows machine. I try to rebase master and see if i can get a repo

PR #8710, opened 2018-03-01
WIP Do not use GetCurrentDirectory in test

wli3:


peterhuene:
Looks like the Windows ones failed do to the mock filesystem issue I mentioned in the corefx issue; something with using a TempRoot is probably breaking the "fake" directory enumeration in the mock filesystem (it's pretty fragile; something like not normalizing the path is probably the culprit).

The other failures look transiet.

@dotnet-bot test this please

wli3:
As i stated before, tests do not use fake file system

wli3:
Maybe it is my memory. I remember these tests are not using in memory file system. But seems it does today

peterhuene:
I changed the implementation as part of the uninstall command work since the mock objects were already taking an `IFileSystem`; the whole point of using this interface is to allow the file system interaction to be mocked.  If we don't intend to use a mock file system, then I see even less of a point for the existence of the tool package mocks.

wli3:
I see, it should not be corefx's problem then

peterhuene:
I apologize for not making that change clear to you when we were discussing this with corefx and it was easy to miss during the major refactoring for the uninstall implementation.  That's my fault.

PR #8716, opened 2018-03-02
tool-path option -- "Session tool"

wli3:
Previous discussion https://github.com/dotnet/cli/pull/8670

From the **beginning design, shim and packageInstaller take _package location_ from constructor** and don't have assumption anymore. From previous discussion, `tool-path `will simply change global location to the one user want, and everything else is the same.

However, this "override" need to happen during the call, that means _InstallToolCommand_ will create different shim and packageInstaller object according to the tool-path during the call instead of constructor DI.

new a shim repo or a packageInstaller during the call will break the testability. So I added factories.

As for tests - it is testing DI at this point. I think a manual test should be enough, we don't have automate tests for package are stored in `.dotnet/tools` either.

-------------

Includes global package location change. And block of leading dot as command name

------------

User's CI sample

```bat
SET toolpath=c:\work\cli\bin\toolpath%RANDOM%
SET PATH=%PATH%;%toolpath%
dotnet install tool dotnet-sql-cache --tool-path %toolpath% --version 2.1.0-preview1-28181
%toolpath%\dotnet-sql-cache -v
```


test/dotnet.Tests/CommandTests/PassThroughToolPackageFactory.cs
```diff
@@ -0,0 +1,27 @@
...
+        private readonly IToolPackageInstaller _toolPackageInstaller;
+
+        public PassThroughToolPackageFactory(IToolPackageStore toolPackageStore,
+            IToolPackageInstaller toolPackageInstaller)
+        {
+            _toolPackageStore = toolPackageStore;
+            _toolPackageInstaller = toolPackageInstaller;
+        }
+
+        public (IToolPackageStore, IToolPackageInstaller)
```
wli3: ToolPackageStore is referenced byToolPackageInstall, so they need to come together. Or I have to DI 2 factories

wli3:
also, all this factories are getting too industrial. Do you think just Func<(IToolPackageStore, IToolPackageInstaller), DirectoryPath?> would be better?

wli3:
But at the same time it is really confusing to see  `DirectoryPath?` what is that about at all? By using an interface, it is clearer

nguerrera:
Without having studied the code in detail, Func<>` and interface/abstract class aren't the only options. Consider a named delegate type where you can document the semantics of parameters and return value such in doc comments that show up when you invoke it in IntelliSense. I find that Func is overused.

src/dotnet/ShellShim/ShellShimRepositoryFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.DotNet.Configurer;
+using Microsoft.Extensions.EnvironmentAbstractions;
+
+namespace Microsoft.DotNet.ShellShim
+{
+    internal class ShellShimRepositoryFactory : IShellShimRepositoryFactory
+    {
+        public IShellShimRepository CreateShellShimRepository(DirectoryPath? nonGlobalLocation = null)
+        {
+            return new ShellShimRepository(GetShimLocation(nonGlobalLocation));
```
peterhuene: I'd prefer this to be `nonGlobalLocation ?? GetGlobalShimLocation()` rather than passing it through to a function whose job it is to return it back if not null.

src/dotnet/ToolPackage/ToolPackageFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    internal class ToolPackageFactory : IToolPackageFactory
+    {
+        private const string NameOfNestedPackageDirectory = ".pkg";
+
+        public (IToolPackageStore, IToolPackageInstaller) CreateToolPackageStoreAndInstaller(
+            DirectoryPath? nonGlobalLocation = null)
+        {
+            var toolPackageStore =
+                new ToolPackageStore(GetPackageLocation(nonGlobalLocation));
```
peterhuene: Same comment as above, perhaps `nonGlobalLocation ?? GetGlobalPackageLocation()`?

src/dotnet/ToolPackage/ToolPackageFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Configurer;
+using Microsoft.DotNet.Tools.Install.Tool;
+using Microsoft.Extensions.EnvironmentAbstractions;
+
+namespace Microsoft.DotNet.ToolPackage
+{
+    internal class ToolPackageFactory : IToolPackageFactory
+    {
+        private const string NameOfNestedPackageDirectory = ".pkg";
```
peterhuene: Nit: perhaps name this `PackageRootDirectory` or something similiar? The `NameOfNested` prefix is unnecessary imo.

src/dotnet/ToolPackage/ToolPackageFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+            DirectoryPath packageLocation;
+
+            if (nonGlobalLocation.HasValue)
+            {
+                packageLocation = nonGlobalLocation.Value.WithSubDirectories(NameOfNestedPackageDirectory);
+            }
+            else
+            {
+                var cliFolderPathCalculator = new CliFolderPathCalculator();
+                packageLocation = new DirectoryPath(cliFolderPathCalculator.ToolsPackagePath);
```
peterhuene: If we get buy off to re-do the package directory layout for preview2, then I think `ToolsPackagePath` and `ToolsShimPath` should maybe combined to a single `ToolsPath` (i.e. the global "tools path") and we append on the `.pkg` subdirectory when calculating the global location here.

src/dotnet/ShellShim/ShellShimRepositoryFactory.cs
wli3: no that function need to call CliFolderPathCalculator,

var cliFolderPathCalculator = new CliFolderPathCalculator();
packageLocation = new DirectoryPath(cliFolderPathCalculator.ToolsShimPath);

wli3: ah, i see, nwm, no strong opinion on that

src/dotnet/ToolPackage/ToolPackageFactory.cs
wli3: PackageRootDirectory  = ".pkg" may confused with say , c:\.pkg ?

and `name` is to tell you that is only the directory name, not the full path. And nested it indicate it will nest one more layer insterad of  tools.pkg vs tools/.pkg

src/dotnet/ShellShim/ShellShimRepositoryFactory.cs
peterhuene: Thinking about this more, though, this is a lot of abstraction for what really boils down to getting the global shim location.  I'm not sure the factory pattern is useful here because all this is doing is calculating the global shim location and constructing only one possible type to give back (i.e. it's simply abstracting away a constructor call).  This is only really saving the users of this factory from doing something like `new ShellShimRepository(myLocation ?? new DirectoryPath(new CliFolderPathCalculator().ToolsShimPath)`.

src/dotnet/ToolPackage/ToolPackageFactory.cs
wli3: I don't like couple 2 things that don't need to. we moved these 2 locations several times. And we once considered, shim stored in .tools and package storage in .nupkg, who knows if we need to change it again. And I don't think there is problem having one place cliFolderPathCalculator to set the location.

src/dotnet/ShellShim/ShellShimRepositoryFactory.cs
peterhuene: Aside: why isn't `CliFolderPathCalculator` static? It has no fields :/

wli3: See my description top comment. Once you `new ShellShimRepository` the location is fixed to .dotnet, and we cannot swap implementation anymore

peterhuene: Ok, now I'm on the same page.  I retract my concern over the complexity (the original comment on this thread, though, I leave in place).

src/dotnet/ToolPackage/IToolPackageFactory.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.EnvironmentAbstractions;
+
+namespace Microsoft.DotNet.ToolPackage
+{
+    internal interface IToolPackageFactory
+    {
+        (IToolPackageStore, IToolPackageInstaller) CreateToolPackageStoreAndInstaller(
```
peterhuene: I think we may need to separate this out into two different factories, unfortunately.  I'm just thinking about the future when the uninstall command accepts a `--tool-path` option.  It doesn't need an installer.

src/dotnet/ToolPackage/ToolPackageFactory.cs
peterhuene: `NestedPackageDirectoryName` perhaps?  I'm very weakly opinioned on this, so if none of those suggested names speak to you, then :+1: on leaving it as is.

src/dotnet/ToolPackage/ToolPackageFactory.cs
peterhuene: In `CliFolderPathCalculator.ToolsPackagesPath`, then, are we going to have another `".pkg"` constant we append onto `ToolShimsPath`?  I just like to minimize places we have to keep in sync.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
```diff
@@ -47,40 +46,32 @@ internal class InstallToolCommand : CommandBase
...
         }
 
         public override int Execute()
         {
-            if (!_global)
+            if (_toolPath != null && _global)
             {
-                throw new GracefulException(LocalizableStrings.InstallToolCommandOnlySupportGlobal);
+                throw new GracefulException("Cannot have global and tool-path as opinion at the same time."); // TODO wul no checkin loc
             }
```
peterhuene: Until we come up with a reasonable default for `--tool-path`, should we need a check for `!_global && string.IsNullOrEmpty(_toolPath)` and error that the `--tool-path` option is required?

wli3: good point

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandParser.cs
```diff
@@ -19,6 +19,10 @@ public static Command InstallTool()
                     "-g|--global",
                     LocalizableStrings.GlobalOptionDescription,
                     Accept.NoArguments()),
+                Create.Option(
+                    "--tool-path",
+                    "Location of shim to access tool",
```
peterhuene: Perhaps `"Location to install the tool to."` (loc'd of course).

src/dotnet/ToolPackage/IToolPackageFactory.cs
wli3: The problem is ToolPackageInstaller need ToolPackageStore object in constructor. So ToolPackageInstaller factory need to take a ToolPackageStore as input, it is even more odd.

In uninstall, i do need to have a separate CreateToolPackageStore method.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandParser.cs
wli3: Yours are better. But we haven't found a good message yet. Need to continue 

KathleenDollard 3 days ago  Member
More like:

**"Location of shim to access tool, allowing unique path for each shell session."**

I don't love this, I'm still working on it because I hate that we are not explaining that it is a temporary as in vulnerable installation.

  @nguerrera
nguerrera 3 days ago  •  Member
But why is it temporary and vulnerable? I think we need to stop thinking in time dimension ("sessions") and just worry about space (global location vs. explicit location vs. default local location). If I can control where, I can control when to cleanup separately. "sessions" becomes a pattern (install to local location, use, uninstall from local location), but not a primitive concept to the CLI.

   @wli3
wli3 23 hours ago  Member
@KathleenDollard @nguerrera In implementation it does have nothing to do with sessions. But the goal of CI author is to achieve isolation from different run. Maybe Location of shim to access tool. Different path can isolate different tool sets.

   @wli3
wli3 23 hours ago  Member
Do you think it is ok to give a paragraph here? This is suitable for automation scenario, recommend each automation run to have a unique PATH so there is no interact with different CI run.... The good part is people don't need to look up for document. And bad part is if we are wrong, we stuck with it

  @nguerrera
nguerrera 23 hours ago  Member
I'd rather we just say what it does, which is much more general than these scenarios. It seems obvious to me why one might pick a unique location. To me, this is like saying man cp should say **"pick a unique destination if you don't want to share the copied file with other things."**

src/dotnet/ToolPackage/IToolPackageFactory.cs
peterhuene: I'm :+1: with just adding a `CreateToolPackageStore` method to this type for the uninstaller in the future.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandParser.cs
wli3: "Override default location"? 

Override??

Also we need to find a way to tell user can call your tools with `%toolpath%mytool`

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
```diff
@@ -18,12 +18,14 @@ namespace Microsoft.DotNet.Tools.Install.Tool
 {
     internal class InstallToolCommand : CommandBase
     {
-        private readonly IToolPackageStore _toolPackageStore;
-        private readonly IToolPackageInstaller _toolPackageInstaller;
-        private readonly IShellShimRepository _shellShimRepository;
+        public delegate IShellShimRepository CreateShellShimRepository(DirectoryPath? nonGlobalLocation = null);
```
wli3: @nguerrera use named delegate, am I doing it right? It does removed a lot of interfaces. It also has nonGlobalLocation in intellisense.

nguerrera: That is fine. Usually I suggest avoiding public nested types though so I'd put this at namespace level.

test_non_location_tool.bat
```diff
@@ -0,0 +1,4 @@
+SET toolpath=c:\work\cli\bin\toolpath%RANDOM%
```
wli3: I will remove this later. This is the "user's experience"

src/dotnet/ToolPackage/ToolPackageFactory.cs
wli3: Fixed

src/dotnet/ToolPackage/ToolPackageFactory.cs
wli3: I think a little dup of a string structure (we could even have a static class to provide nested relation logic). It better than have them coupled in class level. Later toolpackage start to assume shim will just be a directory above. it would horrible to do the directory move again.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
wli3: fixed

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
wli3: Moved to namespace level

src/dotnet/ToolPackage/ToolConfiguration.cs
```diff
@@ -46,6 +47,16 @@ private void EnsureNoInvalidFilenameCharacters(string commandName)
             }
         }
 
+        private void EnsureNoLeadingDot(string commandName)
+        {
+            if (commandName.StartsWith(".", StringComparison.OrdinalIgnoreCase))
+            {
+                throw new ToolConfigurationException(
+                    "Command Name cannot have a leading dot" // TODO wul no checkin loc
```
wli3: @KathleenDollard user facing string review

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommand.cs
```diff
@@ -47,40 +49,37 @@ internal class InstallToolCommand : CommandBase
...
-            if (!_global)
+            if (string.IsNullOrWhiteSpace(_toolPath) && !_global)
+            {
+                throw new GracefulException("Need either global or tool-path provided."); // TODO wul no checkin loc
+            }
+
+            if (!string.IsNullOrWhiteSpace(_toolPath) && _global)
             {
-                throw new GracefulException(LocalizableStrings.InstallToolCommandOnlySupportGlobal);
+                throw new GracefulException("Cannot have global and tool-path as opinion at the same time."); // TODO wul no checkin loc
```
wli3: @KathleenDollard for user facing string review

src/dotnet/commands/dotnet-uninstall/tool/UninstallToolCommand.cs
```diff
@@ -15,46 +15,64 @@
...
 
         public override int Execute()
         {
-            if (!_options.ValueOrDefault<bool>("global"))
+            var global = _options.ValueOrDefault<bool>("global");
+            var toolPath = _options.SingleArgumentOrDefault("tool-path");
+
+            if (string.IsNullOrWhiteSpace(toolPath) && !global)
+            {
+                throw new GracefulException("Need either global or tool-path provided."); // TODO wul no checkin loc
```
wli3: @KathleenDollard for user facing string review

src/dotnet/CommonLocalizableStrings.resx
```diff
@@ -619,4 +619,7 @@ setx PATH "%PATH%;{0}"
   <data name="ToolPackageConflictPackageId" xml:space="preserve">
     <value>Tool '{0}' (version '{1}') is already installed.</value>
   </data>
-</root>
+  <data name="ToolSettingsInvalidLeadingDotCommandName" xml:space="preserve">
+    <value>Command '{0}' has a leading dot.</value>
```
wli3: @KathleenDollard user facing string review

src/dotnet/commands/dotnet-install/dotnet-install-tool/LocalizableStrings.resx
```diff
@@ -178,4 +175,13 @@ Tool '{1}' (version '{2}') was successfully installed.</value>
   <data name="FailedToCreateToolShim" xml:space="preserve">
     <value>Failed to create shell shim for tool '{0}': {1}</value>
   </data>
+  <data name="InstallToolCommandNeedGlobalOrToolPath" xml:space="preserve">
+    <value>Need either global or tool-path provided.</value>
```
wli3: @KathleenDollard user facing string review

wli3:
@KathleenDollard @peterhuene this is ready to go (I think) please review

wli3:
This week we have MVP summit. I created an issue for the user facing strings. So we don't have to block on this https://github.com/dotnet/cli/issues/8728

src/Microsoft.DotNet.Configurer/ToolPackageFolderPathCalculator.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+
+namespace Microsoft.DotNet.Configurer
+{
+    public static class ToolPackageFolderPathCalculator
+    {
+        private const string NestedToolPackageFolderName = ".store";
+        public static string GetToolPackageFolderPath(string ToolsShimPath)
```
peterhuene: Nit: should be caming-casing on argument name.

wli3:
@peterhuene thanks! and osx is good on my machine

wli3:

wli3:
@dotnet-bot  test Ubuntu16.04 x64 Debug Build

wli3:

wli3:
VSO osx is good

PR #8734, opened 2018-03-07
Remove temp project path from tool install warnings and errors.

### LINKED ISSUE #8707: Tool install with nuget error should not display the temp project in output
## Steps to reproduce

See related issue #8698.

```
dotnet install tool Microsoft.Dotnet.Watcher.Tools -g
```

## Expected  behavior

The temp project path should not appear in the output by default:

```
$ dotnet install tool Microsoft.Dotnet.Watcher.Tools -g
error NU1212: Invalid project-package combination for Microsoft.DotNet.Watcher.Tools 2.0.0. DotnetToolReference project style can only contain references of the DotnetTool type
The tool package could not be restored.
Tool 'Microsoft.Dotnet.Watcher.Tools' failed to install.
```

## Actual behavior

The temp project path appears in the output:

```
$ dotnet install tool Microsoft.Dotnet.Watcher.Tools -g
/var/folders/rl/m1rv1rxd643gpwspqrbxy76r0000gn/T/w1heen5l.wc5/yh505uu5.u4i.csproj : error NU1212: Invalid project-package combination for Microsoft.DotNet.Watcher.Tools 2.0.0. DotnetToolReference project style can only contain references of the DotnetTool type
The tool package could not be restored.
Tool 'Microsoft.Dotnet.Watcher.Tools' failed to install.
```

## Environment data
```
$ dotnet --info
.NET Command Line Tools (2.1.300-preview2-008266)

Product Information:
 Version:            2.1.300-preview2-008266
 Commit SHA-1 hash:  6fb919f433

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx-x64
 Base Path:   /Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/sdk/2.1.300-preview2-008266/

  Version  : 2.1.0-preview2-26131-06
  Build    : b13a0d5c331f374afd35ded57b9a4b4ab128864c
```

peterhuene:
This commit attempts to filter the diagnostic messages emitted during tool
installation.  The diagnostic messages may be prefixed with the temporary
project; since this is an implementation detail that only causes confusion and
clutter in the diagnostic messages, the prefix is removed if present.

Fixes #8707.

### DIFF
--- a/src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
+++ b/src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
@@ -16,10 +16,14 @@ namespace Microsoft.DotNet.Tools.Install.Tool
     {
         private const string AnyRid = "any";
         private readonly IReporter _reporter;
+        private readonly IReporter _errorReporter;
+        private readonly bool _forceOutputRedirection;
 
         public ProjectRestorer(IReporter reporter = null)
         {
-            _reporter = reporter;
+            _reporter = reporter ?? Reporter.Output;
+            _errorReporter = reporter ?? Reporter.Error;
+            _forceOutputRedirection = reporter != null;
         }
 
         public void Restore(
@@ -56,11 +60,11 @@ namespace Microsoft.DotNet.Tools.Install.Tool
             var command = new DotNetCommandFactory(alwaysRunOutOfProc: true)
                 .Create("restore", argsToPassToRestore);
 
-            if (_reporter != null)
+            if (verbosity == null || _forceOutputRedirection)
             {
                 command = command
-                    .OnOutputLine((line) => _reporter.WriteLine(line))
-                    .OnErrorLine((line) => _reporter.WriteLine(line));
+                    .OnOutputLine(line => WriteLine(_reporter, line, project))
+                    .OnErrorLine(line => WriteLine(_errorReporter, line, project));
             }
 
             var result = command.Execute();
@@ -68,6 +72,29 @@ namespace Microsoft.DotNet.Tools.Install.Tool
             {
                 throw new ToolPackageException(LocalizableStrings.ToolInstallationRestoreFailed);
             }
+        }
+
+        private static void WriteLine(IReporter reporter, string line, FilePath project)
+        {
+            line = line ?? "";
+
+            // Remove the temp project prefix if present
+            if (line.StartsWith($"{project.Value} : ", StringComparison.OrdinalIgnoreCase))
+            {
+                line = line.Substring(project.Value.Length + 3);
+            }
+
+            // Note: MSBuild intentionally does not localize "warning" and "error" for diagnostic messages
+            if (line.StartsWith("warning ", StringComparison.Ordinal))
+            {
+                line = line.Yellow();
+            }
+            else if (line.StartsWith("error ", StringComparison.Ordinal))
+            {
+                line = line.Red();
+            }
+
+            reporter.WriteLine(line);
         }
     }
 }
--- a/src/dotnet/ToolPackage/IToolPackageInstaller.cs
+++ b/src/dotnet/ToolPackage/IToolPackageInstaller.cs
@@ -15,6 +15,7 @@ namespace Microsoft.DotNet.ToolPackage
             VersionRange versionRange = null,
             string targetFramework = null,
             FilePath? nugetConfig = null,
+            DirectoryPath? rootConfigDirectory = null,
             string source = null,
             string verbosity = null);
     }
--- a/src/dotnet/ToolPackage/ToolPackageFactory.cs
+++ b/src/dotnet/ToolPackage/ToolPackageFactory.cs
@@ -16,7 +16,7 @@ namespace Microsoft.DotNet.ToolPackage
             IToolPackageStore toolPackageStore = CreateToolPackageStore(nonGlobalLocation);
             var toolPackageInstaller = new ToolPackageInstaller(
                 toolPackageStore,
-                new ProjectRestorer(Reporter.Output));
+                new ProjectRestorer());
 
             return (toolPackageStore, toolPackageInstaller);
         }
--- a/src/dotnet/ToolPackage/ToolPackageInstaller.cs
+++ b/src/dotnet/ToolPackage/ToolPackageInstaller.cs
@@ -119,7 +119,7 @@ namespace Microsoft.DotNet.ToolPackage
         {
             var tempProject = _tempProject ?? new DirectoryPath(Path.GetTempPath())
                 .WithSubDirectories(Path.GetRandomFileName())
-                .WithFile(Path.GetRandomFileName() + ".csproj");
+                .WithFile("restore.csproj");
 
             if (Path.GetExtension(tempProject.Value) != "csproj")
             {
--- a/src/dotnet/ToolPackage/ToolPackageInstaller.cs
+++ b/src/dotnet/ToolPackage/ToolPackageInstaller.cs
@@ -35,6 +35,7 @@ namespace Microsoft.DotNet.ToolPackage
             VersionRange versionRange = null,
             string targetFramework = null,
             FilePath? nugetConfig = null,
+            DirectoryPath? rootConfigDirectory = null,
             string source = null,
             string verbosity = null)
         {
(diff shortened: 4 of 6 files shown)


src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
```diff
@@ -69,5 +69,28 @@ public ProjectRestorer(IReporter reporter = null)
                 throw new ToolPackageException(LocalizableStrings.ToolInstallationRestoreFailed);
             }
         }
+
+        private static void WriteLine(IReporter reporter, string line, FilePath project)
+        {
+            line = line ?? "";
+
+            // Remove the temp project prefix if present
+            if (line.StartsWith($"{project.Value} : ", StringComparison.CurrentCultureIgnoreCase))
```
nguerrera: Why CurrentCulture?

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
```diff
@@ -69,5 +69,28 @@ public ProjectRestorer(IReporter reporter = null)
...
+            {
+                line = line.Substring(project.Value.Length + 3);
```
nguerrera: This math may be wrong for some cultures.

peterhuene: I assume you mean the `project.Value.Length` and this should use a `StringInfo` to arrive at the correct length?

nguerrera: Well, I mean if we need to compare to project.Value by CurrentCulture. If we can check for project.Value at start with Ordinal, then this is fine. It is related to my other comment.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
peterhuene: I wanted to guard against MSBuild outputting a different case for the project file path (I don't think it will given how we construct the path, but for case-insensitive file systems it wouldn't be "wrong" to do so).

However, I don't think an ordinal or invariant comparison would give the desired result.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
nguerrera: Do we have full control over project.Value? Will it have characters outside the ASCII range. Who puts that ' : ' is it also not localized like 'error' and 'warning'?

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
nguerrera: The correct way to compare paths like the Windows file system is OrdinalIgnoreCase.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
peterhuene: We do have control over the project path, but it is based on `Path.GetTempPath`.

The ` : ` comes from the format MSBuild uses for its console diagnostic messages.  Likewise, `warning` and `error` are intentionally not localized, presumably so that the level can be parsed in a manner like what is being done here.

nguerrera: I think we can just use OrdinalIgnoreCase, then.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
peterhuene: And is that also true of case-insensitive APFS?

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
peterhuene: I'm fine with `OrdinalIgnoreCase` as a safe-guard (I suspect `Ordinal` would work too provided MSBuild doesn't try to normalize the path given).

peterhuene: Changed to `OrdinalIgnoreCase`.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
```diff
@@ -56,11 +56,11 @@ public ProjectRestorer(IReporter reporter = null)
             var command = new DotNetCommandFactory(alwaysRunOutOfProc: true)
                 .Create("restore", argsToPassToRestore);
 
-            if (_reporter != null)
+            if (verbosity == null || _reporter != null)
             {
                 command = command
-                    .OnOutputLine((line) => _reporter.WriteLine(line))
-                    .OnErrorLine((line) => _reporter.WriteLine(line));
+                    .OnOutputLine(line => WriteLine(_reporter ?? Reporter.Output, line, project))
```
wli3: _reporter should not be null after constructor, if so could you fix it?

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
nguerrera: 👍

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
peterhuene: `_reporter` can't default to a single value because output from `STDERR` should go to `Reporter.Error` and not `Reporter.Output`.  So we would need two fields.  I'd prefer one that says "direct all output to this reporter".

Actually, this just reminded me that your `tool-path` refactoring is now passing in `Reporter.Output` in the constructor where previously it was only being used for tests.  This breaks stderr output.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
nguerrera: I don't know.

src/dotnet/commands/dotnet-install/dotnet-install-tool/ProjectRestorer.cs
peterhuene: Additionally, we'd need another flag field to indicate whether or not the original reporter passed in the constructor was null or not (if it's non-null then output must be redirected).  It's much simpler to allow the field to be null, imo.  But if you feel strongly about it, I could change to these fields:

```
        private readonly IReporter _reporter;
        private readonly IReporter _errorReporter;
        private readonly bool _forceOutputRedirection;
```

peterhuene: I decided to fix it as describe.  I've pushed up the change.

peterhuene:
The RHEL failure occurred again (from the `list tool` command changes, this now manifests as a "failure to locate the staged package" following the restore).  As a theory, I've pushed up a commit to remove the reliance of setting the current working directory from the tool package installer tests in the hopes of improving the stability of the tests.

I still need to setup a RHEL VM to see if I can manually reproduce the failure with what I suspect the issue to be.

peterhuene:
Oh right, the broken much implementation on Windows when using TempRoot. Let me fix good that.

peterhuene:
I tracked down the Windows failure to `TempRoot.Root` containing both `/` (including a double `//`) and `\` in the path.  This caused the mock file system to enumerate incorrectly because no path normalization was being performed and certain `Path` functions were returning differing results from what was stored in the internal `Dictionary`.

I first attempted to fix the mock file system to try to normalize all the path interactions, but this was a more invasive change that impacted other tests.  I decided to just normalize the use of `TempRoot.Root` to fix this particular issue.

@nguerrera @wli3 can you review the second commit that was pushed?  Hopefully it will bring some stability to the tool package installer tests.

nguerrera:
Second commit looks good. Thanks.

wli3:
ack, let me check

wli3:
Good to go, but not perfect.

https://github.com/dotnet/cli/pull/8734/commits/52478e8947d12953119ebf6175adaa0166a7558f#diff-514fdb6bf1ab21b1a21adfd2ccf82145R121  is a test induced design damage. And the test is just testing the nuget flag. I do think use a test to lock nuget's behavior that we depend on is important since they are in rapid development. A regression is very likely. But adding another optional just for test making it further away from production.

Let me also see if there is other alternative. One way is to make is closer to real. Find a better mock library. Or make is less dependent on file system state.

wli3:
I mean please checkin. I'll keep digging on my own

peterhuene:
Agreed, an option just for testing purposes is not ideal. Let's work out a proper solution after this change.

PR #8751, opened 2018-03-08
Convert to graceful exception

wli3:


src/dotnet/DotNetTopLevelCommandBase.cs
```diff
@@ -42,19 +42,26 @@ public int RunCommand(string[] args)
             }
             catch (KeyNotFoundException)
```
nguerrera: Unrelated (pre-existing), but it looks bogus to assume that any KeyNotFoundException in command.Execute() means RequiredCommandNotPassed. Do you agree or am I missing something? If you agree, I'll file a bug.

peterhuene: Agreed, the `Execute` call should be outside of this try-catch.

nguerrera: Filed #8752

PR #8794, opened 2018-03-14
Update tool

wli3:
Need extra check:

1. most of the parser logic is duplicated with install because we cannot reuse description
2. no `--version` in update command because if version is lower than current it is just confusing.
3. the way i resuse try catch logic from install and uninstall, not idea. But I believe they need to be shared since if we dup and forget to update UpdateCommand, there will be serious bug

Other than string.replace "install" to "update" in most of the opinions. Here is several strings need to review.

Update success with no version change
`Tool '{0}' was successfully updated with no version change.`

Update Succeeded
`Tool '{0}' (from version '{1}' to version'{2}') was successfully updated.`

@KathleenDollard


src/dotnet/commands/dotnet-update/UpdateCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        protected override string CommandName => "update";
+        protected override string FullCommandNameLocalized => "TODO"; //TODO wul
+        protected override string ArgumentName => Constants.ToolPackageArgumentName;
+        protected override string ArgumentDescriptionLocalized => "TODO"; // TODO wul
+
+        internal override Dictionary<string, Func<AppliedOption, CommandBase>> SubCommands =>
+            new Dictionary<string, Func<AppliedOption, CommandBase>>
+            {
+                ["tool"] = options => new UninstallToolCommand(options["tool"], ParseResult)
```
wli3: this is wrong... will change

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandLowLevelErrorConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            string[] userFacingMessages = null;
+            if (ex is ToolPackageException)
+            {
+                userFacingMessages = new[]
+                {
+                    string.Format(
+                        LocalizableStrings.InvalidToolConfiguration,
+                        ex.Message),
+                    string.Format(LocalizableStrings.ToolInstallationFailedContactAuthor, packageId)
```
peterhuene: This was `LocalizableStrings.ToolInstallationFailed` before this refactoring.  The "contact the tool author" message should only be used if there's a problem with the tool settings.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandLowLevelErrorConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+    internal static class InstallToolCommandLowLevelErrorConverter
+    {
+        public static IEnumerable<string> GetUserFacingMessages(Exception ex, PackageId packageId)
+        {
+            string[] userFacingMessages = null;
+            if (ex is ToolPackageException)
+            {
+                userFacingMessages = new[]
+                {
+                    string.Format(
```
peterhuene: Likewise, this was just `ex.Message` prior to the refactoring.

src/dotnet/commands/dotnet-update/LocalizableStrings.resx
```diff
@@ -0,0 +1,129 @@
...
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="UpdateFullCommandName" xml:space="preserve">
+    <value>.NET Update Command</value>
+  </data>
+  <data name="CommandDescription" xml:space="preserve">
+    <value>Updates an item into the development environment.</value>
```
peterhuene: Should probably be `Updates an item in the development environment.` (`in` vs. `into`).

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+  <data name="UpdateSucceeded" xml:space="preserve">
+    <value>Tool '{0}' (from version '{1}' to version'{2}') was successfully updated.</value>
+  </data>
+  <data name="GlobalOptionDescription" xml:space="preserve">
+    <value>Update user wide.</value>
+  </data>
+  <data name="UpdateFullCommandNameLocalized" xml:space="preserve">
+    <value>.NET update Command</value>
+  </data>
+  <data name="ToolAlreadyInstalled" xml:space="preserve">
```
peterhuene: I don't think this string resource is needed.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+    <value>Tool '{0}' is already installed.</value>
+  </data>
+  <data name="UpdateToolCommandNeedGlobalOrToolPath" xml:space="preserve">
+    <value>Please specify either the global option (--global) or the tool path option (--tool-path).</value>
+  </data>
+  <data name="UpdateToolCommandInvalidGlobalAndToolPath" xml:space="preserve">
+    <value>(--global) conflicts with the tool path option (--tool-path). Please specify only one of the options.</value>
+  </data>
+  <data name="ToolPathDescription" xml:space="preserve">
+    <value>Location of shim to access tool</value>
```
peterhuene: Should we use the string we're going to change the install/uninstall command descriptions to (i.e. the one we had the meeting with Kathleen about)?  Thus we won't forget to update this location as well.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+  </data>
+  <data name="ToolHasMultipleVersionsInstalled" xml:space="preserve">
```
peterhuene: Should be `...and cannot be updated.`

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+    <value>Tool '{0}' has multiple versions installed and cannot be uninstalled.</value>
+  </data>
+  <data name="ToolNotInstalled" xml:space="preserve">
+    <value>Tool '{0}' is not currently installed.</value>
+  </data>
+  <data name="UpdateSucceededVersionNoChange" xml:space="preserve">
+    <value>Tool '{0}' was successfully updated with no version change.</value>
```
peterhuene: Perhaps this should be `Tool '{0}' (version '{1}') is already the latest stable version.` or something?  A message implying that it was updated but the version didn't change is confusing, at least to me.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+  </data>
+  <data name="UpdateToolFailed" xml:space="preserve">
```
peterhuene: Perhaps `Tool '{0}' failed to update due to the following errors:` (this should be a single sentence that ends with a colon).

src/dotnet/commands/dotnet-update/tool/UpdateToolCommand.cs
```diff
@@ -0,0 +1,246 @@
...
+                    PrintSuccessMessage(oldPackage, newInstalledPackage);
+                });
+
+                scope.Complete();
+            }
+
+            return 0;
+        }
+
+        private void ValidateArgument()
```
peterhuene: nit: `ValidateArguments` because it validates more than a singular argument.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: I forget. I will correct it

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: I am not sure about it as well. But it could be a different feed. So it could be a different package with the same version. And we actually did a uninstall install, something might changed, at least files timestamp.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+  </data>
+  <data name="SourceOptionDescription" xml:space="preserve">
+    <value>Specifies a NuGet package source to use during update.</value>
+  </data>
+  <data name="SourceOptionName" xml:space="preserve">
+    <value>SOURCE</value>
+  </data>
+  <data name="CommandDescription" xml:space="preserve">
+    <value>Updates a tool for use on the command line.</value>
+  </data>
```
KathleenDollard: I wonder if "on the command line" is helpful here. Perhaps drop.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
```diff
@@ -0,0 +1,183 @@
...
+    <value>The NuGet configuration file to use.</value>
+  </data>
+  <data name="FrameworkOptionDescription" xml:space="preserve">
+    <value>The target framework to update the tool for.</value>
+  </data>
+  <data name="NuGetConfigurationFileDoesNotExist" xml:space="preserve">
+    <value>NuGet configuration file '{0}' does not exist.</value>
+  </data>
+  <data name="UpdateSucceeded" xml:space="preserve">
+    <value>Tool '{0}' (from version '{1}' to version'{2}') was successfully updated.</value>
```
KathleenDollard: Suggest:

Tool '{0}' was successfully updated from version '{1}' to version'{2}' .

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
KathleenDollard: Can we give the user a clue what to do next? (Delete directly?)

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
KathleenDollard: I actually think this is going to be super common until we get a method to check for a new update, do we really want to uninstall and reinstall in this case? Assuming no change to that, I would like a friendlier message. Maybe:

Tool '{0}' was reinstalled with the latest stable version (version '{1}')

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
KathleenDollard: Agree

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: @KathleenDollard what do you think about this issue?

peterhuene: I'm good with Kathleen's string.  One thing we didn't consider, though, is that `update` can function as a downgrade when the user has a higher versioned preview package installed.  It should probably call out in the help somewhere that this upgrades to the latest **stable** version.

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
peterhuene: This is just a safe-guard string at the moment since users can't install multiple versions of the same tool (but our underlying API supports it).

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: So the description should be `Updates a tool to the latest stable version for use.`( used to be `Updates a tool for use on the command line.`and Kathleen suggested to drop on the command line in another comment)

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandLowLevelErrorConverter.cs
peterhuene: recent push 👍

wli3: Fixed. Code change made this comment moved. It should be on `ToolPackageException`

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandLowLevelErrorConverter.cs
wli3: fixed

src/dotnet/commands/dotnet-update/LocalizableStrings.resx
wli3: Changed as suggested

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: Fixed, removed

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: Updated to what we deicded

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: Changed as suggested and also changed the command description to "Updates a tool to the latest stable version for use."

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: Changed as suggested

src/dotnet/commands/dotnet-update/tool/UpdateToolCommand.cs
wli3: Changed as suggested

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: Changed as suggested

src/dotnet/commands/dotnet-update/tool/LocalizableStrings.resx
wli3: Changed as suggested

src/dotnet/commands/dotnet-help/LocalizableStrings.resx
```diff
@@ -268,7 +268,7 @@
     <value>Path to additional deps.json file.</value>
   </data>
   <data name="InstallDefinition" xml:space="preserve">
-    <value>Installs an item into the development environment.</value>
+    <value>Installs an item in the development environment.</value>
```
peterhuene: Looks like `UpdateDefinition` was removed and the `InstallDefinition` was changed.

src/dotnet/commands/dotnet-install/dotnet-install-tool/InstallToolCommandLowLevelErrorConverter.cs
peterhuene: I wish GitHub had a way to mark this as "resolved" or otherwise collapse it without deleting it :/

src/dotnet/commands/dotnet-help/LocalizableStrings.resx
peterhuene: Which would mean we need to check that `update` is showing up in `dotnet --help` too.

wli3: I changed both https://github.com/dotnet/cli/pull/8794/files#diff-fb9c7630de9fa4ddaf4b225bed103f6dR124

It is called CommandDescription in the beginning (align with the name in uninstall)

peterhuene: Right, but these are the help strings for the root help text coming from `dotnet help`.  This one is used here: https://github.com/dotnet/cli/blob/master/src/dotnet/commands/dotnet-help/HelpUsageText.cs#L31.

I don't see a corresponding change to add `update {LocalizableStrings.UpdateDefinition}`.

peterhuene: On an aside: these commands really should be in alphabetical order :/

wli3: I think the problem is I didn't update help text at all. Adding it

wli3: I am not sure about the alphabetical order tho. The current order is almost a flow of an app.

peterhuene: 👍 I'd revert this particular change, though. "into" makes sense for install and "in" makes sense for update (the tool is already *in* the environment).

peterhuene: Yeah, don't need to solve that with these changes; I just think it's something we should discuss with Kathleen when and if we do the verb/noun reversal.  There's a lot of commands that aren't in an everyday use-flow and having them in order of potential usage is hard for newly added commands to be positioned properly.  Right now the ordering just seems very haphazard.

Most command line tools that have subcommands tend to list them alphabetically.

peterhuene:
I'm ready to 👍 this if we revert the change to the install command's help text and potentially squash all these commits (lots of fix-up commits here).

src/dotnet/commands/dotnet-help/LocalizableStrings.resx
wli3: Now it should be good

wli3: install into is reverted. And help text for update is added

wli3: Sorry still missing one

Issue #8561, opened 2018-02-07
Change --source to additional instead of exclusive


wli3:
So if Microsoft.NETCore.Platforms is not in the source folder. CLI can still find it in nugetfallbackfolder

Need to reconsider the opinion name

wli3:
This problem is not critical anymore after https://github.com/dotnet/cli/pull/8677 is merged. Since as of today in master there is no requirement for Platform packages. But we need to finish it as part of work https://github.com/dotnet/cli/issues/8618

wli3:
@livarcocc I will move it to backlog

wli3:
Conclusion of Monday's (3/12/18) meeting. 

1.Replace `--source` to `--source-feed`
2. make `--source-feed` additional instead of exclusive
3. only on `tool`, no change on `add package`

PR #8833, opened 2018-03-17
Change --source to --source-feed and make it additional

wli3:
Fix https://github.com/dotnet/cli/issues/8561

Other than change source to source-feed and make it additional instead of exclusive. I changed source to be multiple. Because restore support multiple source https://github.com/Microsoft/dotnet/issues/361

As for mock. The offline feed and source feed is considered the same, so remove the category of “source”. 

I renamed source to “AdditionalFeed” because that is more accurate on implementation level.

Note:
NuGet feed don’t have order. Whichever responses the fastest, is the first.
No change on restore.

scripts/cli-test-env.sh change is due to mac 10.13 is finally added to RID graph. And it is “considered” one of the CLI supported RID

### DIFF
--- a/src/dotnet/ToolPackage/ToolPackageInstaller.cs
+++ b/src/dotnet/ToolPackage/ToolPackageInstaller.cs
@@ -30,13 +30,12 @@ namespace Microsoft.DotNet.ToolPackage
             _offlineFeed = offlineFeed ?? new DirectoryPath(new CliFolderPathCalculator().CliFallbackFolderPath);
         }
 
-        public IToolPackage InstallPackage(
-            PackageId packageId,
+        public IToolPackage InstallPackage(PackageId packageId,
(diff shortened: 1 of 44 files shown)


src/dotnet/ToolPackage/ToolPackageInstaller.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.DotNet.ToolPackage
 {
     internal class ToolPackageInstaller : IToolPackageInstaller
     {
-        private readonly IToolPackageStore _store;
+        private readonly IToolPackageStore _additionalFeed;
```
peterhuene: Was this refactoring intentional?

wli3: No,let me fix it

peterhuene:
👍

Issue #8805, opened 2018-03-15
Add tool package format version metadata to packed tool in settings file


wli3:
To support better error message when future version come.

@KathleenDollard  should it be `SettingsVersion`? it is really not about settings file. It is the format or structure(folder structure + setting file content). Maybe `ToolPackageFormatVersion` ?

example:
``` xml
<?xml version="1.0" encoding="utf-8"?>
<DotNetCliTool>
  <Commands>
    <Command Name="mytool" EntryPoint="mytool.dll" Runner="dotnet" />
  </Commands>

  <SettingsVersion>1.0.0</SettingsVersion>
</DotNetCliTool>
```

And also, warning when it is not compatible (start with `1.`)

cc @dotnet/dotnet-cli 

- [x] change on CLI to warning when major version is different
- [x] change on SDK to put version on packtool

peterhuene:
Perhaps a `Version` attribute on `DotNetCliTool` rather than a sub-element?

wli3:
That's also good. A little vague. But better than settingsVersion since settingsVersion won't cover all

KathleenDollard:
I like that.
________________________________

wli3:
Do you think we should warning the user as long as version is bigger?

If we use semver, when tool package format 1.2.0 come out in the future. It will require more feature, say support self contain, but back compat. That means, if the consumer has a SDK support 1.0.0 format, that SDK cannot support the self contain tools.

wli3:
I think either way are fine, since it won't change that dramatic. However, i think semver is well known enough, and it fits here:

If the package format version is 1.2, it may not work on SDK  that only support 1.0 - 1.1  since it asks for more feature. It is fine for SDK Support 1.2 - 1.9999. But it should warning again for SDK supporting 2.0+

wli3:
Maybe the problem is the hammer is too big?

nguerrera:
> If the package format version is 1.2, it may not work on SDK that only support 1.0 - 1.1 since it asks for more feature. It is fine for SDK Support 1.2 - 1.9999. But it should warning again for SDK supporting 2.0+

I'm having trouble interpreting this. Really the only meaningful thing is to poison the package so that old consumers break cleanly instead of unexpectedly. I think we only need one dimension for that. 

```
int formatVersion = 1; // bump whenever the format changes such that it will break old consumers
```

Right now, when do we change it? Do we need to evaluate if every single change to the format is major, minor, or patch?

[...]

PR #8860, opened 2018-03-21
Add xml setting version warning

wli3:
one part of https://github.com/dotnet/cli/issues/8805

### DIFF
--- a/src/dotnet/CommonLocalizableStrings.resx
+++ b/src/dotnet/CommonLocalizableStrings.resx
@@ -629,7 +629,7 @@ setx PATH "%PATH%;{0}"
     <value>The command name '{0}' cannot begin with a leading dot (.).</value>
   </data>
   <data name="FormatVersionIsHigher" xml:space="preserve">
-    <value>Format version is higher than supported. This tool may not be supported in this SDK version. Please update SDK.</value>
+    <value>Format version is higher than supported. This tool may not be supported in this SDK version. Please update your SDK.</value>
   </data>
   <data name="FormatVersionIsMalformed" xml:space="preserve">
     <value>Format version is malformed. This tool may not be supported in this SDK version. Please contact the author of the tool.</value>
--- a/src/dotnet/CommonLocalizableStrings.resx
+++ b/src/dotnet/CommonLocalizableStrings.resx
@@ -628,4 +628,13 @@ setx PATH "%PATH%;{0}"
   <data name="ToolSettingsInvalidLeadingDotCommandName" xml:space="preserve">
     <value>The command name '{0}' cannot begin with a leading dot (.).</value>
   </data>
+  <data name="FormatVersionIsHigher" xml:space="preserve">
+    <value>Format version is higher than supported. This tool may not be supported in this SDK version. Please update SDK.</value>
+  </data>
+  <data name="FormatVersionIsMalformed" xml:space="preserve">
+    <value>Format version is malformed. This tool may not be supported in this SDK version. Please contact the author of the tool.</value>
+  </data>
+  <data name="FormatVersionIsMissing" xml:space="preserve">
+    <value>Format version is missing. This tool may not be supported in this SDK version. Please contact the author of the tool.</value>
+  </data>
 </root>
--- a/src/dotnet/ToolPackage/ToolConfigurationDeserializer.cs
+++ b/src/dotnet/ToolPackage/ToolConfigurationDeserializer.cs
@@ -13,7 +13,8 @@ namespace Microsoft.DotNet.ToolPackage
 {
     internal static class ToolConfigurationDeserializer
     {
-        // match the number on SDK when the package format change is handled
+        // The supported tool configuration schema version.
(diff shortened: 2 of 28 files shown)


wli3:
@KathleenDollard

KathleenDollard:
This needs to be two sentences grammatically and lets add which author

`Format version is missing. This tool may not be supported in this SDK version. Please contact the author of the tool.`

@peterhuene do you have thoughts on these strings?

peterhuene:
I think that string is good and the other two strings can follow that structure.

wli3:
Finished loc and changed version to int

src/dotnet/CommonLocalizableStrings.resx
```diff
@@ -628,4 +628,13 @@ setx PATH "%PATH%;{0}"
   <data name="ToolSettingsInvalidLeadingDotCommandName" xml:space="preserve">
     <value>The command name '{0}' cannot begin with a leading dot (.).</value>
   </data>
-</root>
+  <data name="FormatVersionIsHigher" xml:space="preserve">
+    <value>Format version is higher than supported. This tool may not be supported in this SDK version. Please update SDK.</value>
```
peterhuene: suggestion:`Format version is higher than supported. This tool may not be supported in this SDK version. Please update your SDK`. (added *your*)

src/dotnet/ToolPackage/ToolConfigurationDeserializer.cs
```diff
@@ -12,6 +13,9 @@ namespace Microsoft.DotNet.ToolPackage
 {
     internal static class ToolConfigurationDeserializer
     {
+        // match the number on SDK when the package format change is handled
```
peterhuene: Comment could be a little more descriptive, such as:

```
// The supported tool configuration schema version.
// This should match the version in the GenerateToolsSettingsFile task from the SDK.
```

src/dotnet/ToolPackage/ToolPackageInstance.cs
```diff
@@ -166,5 +172,28 @@ private static LockFileItem FindItemInTargetLibrary(LockFileTargetLibrary librar
                 ?.ToolsAssemblies
                 ?.SingleOrDefault(t => LockFileMatcher.MatchesFile(t, targetRelativeFilePath));
         }
+
+        public IEnumerable<string> Warnings
```
peterhuene: This property does non-trivial work.  I think it should either use a `Lazy<IEnumerable<string>>` under the covers or this should really be a method called `GetWarnings`.

peterhuene: Also, we already deserialized the configuraiton when we accessed the commands.  Perhaps this and the `Commands` property should share an underlying `Lazy<ToolConfiguration>`?

wli3: I tried to scope the access of this property as explicit as possible. Since `warnings` is not a concern for rest of the component (unless we can predict the future).

peterhuene:
```c#
private Lazy<ToolConfiguration> _toolConfiguration = new Lazy<ToolConfiguration>(GetToolConfiguration);

public IEnumerable<string> Warnings => _toolConfiguration.Value.Warnings;

private ToolConfiguration GetToolConfiguration()
{
    try
    {
        var lockFile = new LockFileFormat().Read(PackageDirectory.WithFile(AssetsFileName).Value);
        var library = FindLibraryInLockFile(lockFile);
        return DeserializeToolConfiguration(ToolSettingsFileName, library);
    }
    catch (Exception ex) when (ex is UnauthorizedAccessException || ex is IOException)
    {
        throw new ToolConfigurationException(
            string.Format(
                CommonLocalizableStrings.FailedToRetrieveToolConfiguration,
                ex.Message),
            ex);
    }
}
```

And then use `_toolConfiguration.Value` in `GetCommands`.

wli3: So i will go GetWarnings. it is pretty much only ever accessed once. I would like Lazy if the syntax is less noisy

wli3: so who ever access ToolConfiguration or Warnings first will trigger GetToolConfiguration()?

My concern is access Warnings will trigger other unrelated field get populated. No strong feeling, if you think it is good, let's go this route. It can reduce a lot of code

peterhuene: Accessing `Warnings` first won't trigger anything to be populated in `Commands`, just save it the time it would have spent on deserializing the tool configuration a second time, is all.

wli3: it needs to deserialize the whole thing anyway. I guess i will just go as you suggested

wli3:

wli3:
@MattGertz for tell mode

PR #8895, opened 2018-03-25
Minimizing collsions in global tool names and future CLI commands

wli3:


src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
```diff
@@ -0,0 +1,161 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.DotNet.ToolPackage
+{
+    public static class CommandNameAllReservedValidator
+    {
+        private static CommandNameValidator[] _explicitReservedStrings = new CommandNameValidator[] {
```
wli3: whole unit test for validation take about 200ms. And it is one off during dotnet install (which usually has networking call to download package). I don't think this affect perf. To have code matches the spec make it easier to edit

nguerrera: 200ms sounds like a lot to me. What's the percentage of dotnet tool install?

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
```diff
@@ -0,0 +1,161 @@
...
+            new CommandNameValidator(false, false, true, "add"),
```
dasMulli: super-super-nit: could this possibly be made more readable by using either named parameters or a flag enum instead of a lot of true/false literals?

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
wli3: the whole test is 200ms. with it around 200ms, without it around 200ms. So the actual increase is minimal.

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
wli3: the spec is like that :(

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
nguerrera: Oh, sorry I misread the original comment. 7ms is much more reasonable.

I withdraw the perf concern, but I'm still worried about the complexity here and how it will be easy to trigger false positives. That's feedback for the proposal not the implementation here, though.

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
nguerrera: > the spec is like that :(

I agree the proposal is somewhat complicated and I hope we simplify it.

But that's not relevant to @dasMulli's comment. You can keep the exact same spec and logic and use either a flags enum or named arguments. I also find unnamed bool arguments hard to read and 3 consecutive bools are especially easy to get confused.

src/dotnet/ToolPackage/CommandNameValidator.cs
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.DotNet.Tools;
+
+namespace Microsoft.DotNet.ToolPackage
+{
+    public class CommandNameValidator
+    {
+        private string _agaist;
```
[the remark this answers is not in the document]
wli3: ya, against

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
wli3: Maybe add param name can help. Or just add comment some that mimic the header of the spreadsheet.

The current form is very easy to create from the spreadsheet use sublime ctrl+shift+l and alt+3, so it is less likely to "just copy it wrong". Also, the shape is the same as the spreadsheet. It is easy to edit accordingly if the sheet is change.

src/dotnet/ToolPackage/CommandNameValidator.cs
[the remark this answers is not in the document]
wli3: let me change to _reserved

src/dotnet/ToolPackage/CommandNameAllReservedValidator.cs
wli3: Add comment, is that better?

PR #8908, opened 2018-03-27
Template updates for April 2018

JunTaoLuo:
These are the latest templates built for 1.1.6.

Note this is not going to pass until we push our template package to the aspnetcore-master feed. @pranavkm I still don't have push access to that feed. I'm requesting it for future updates from @Eilon but can you push it in the mean time?


Eilon:
@JunTaoLuo I just gave you access to push to the feed.

JunTaoLuo:
We decided not to take the System.Console dependency update so this is no longer required.

PR #8919, opened 2018-03-28
Add latest patch versions to BundledVersions.props, and tests to make sure that they are up to date

dsplaisted:
- Add properties to BundledVersions.props for the latest patch version of each minor release of .NET Core.  This will be consumed by dotnet/sdk#2085, which automatically rolls forward self-contained apps to the latest patch version
- Add tests to make sure that the latest patch version properties are up-to-date.  These tests will fail until dotnet/sdk#2085 is merged and inserted into the CLI

I think that these tests will mean that the only thing we will need to do to keep the latest patch numbers up-to-date is to update them when test failures tell us to do so.  The `ItRollsForwardToTheLatestVersion` test compares the resolved version of `Microsoft.NETCore.App` for a self-contained app with the version that is resolved when the `RuntimeFrameworkVersion` floats the patch version (eg `2.0.*`).  If they don't match, then it will fail with a message like the following:

> Expected string to be equivalent to "2.0.6" because the latest patch version properties in Microsoft.NETCoreSdk.BundledVersions.props need to be updated (see MSBuildExtensions.targets in this repo), but "2.0.0" differs near "0" (index 4).

We also need to make sure that the tests cover each minor release of .NET Core.  That's covered by the `WeCoverLatestNetCoreAppRollForward` test, which creates a project via `dotnet new console` and makes sure that the `TargetFramework` from that project has a corresponding entry in the `MemberData` used by the previous test.  So when we update the templates to target a new version of .NET Core, we'll be reminded to update these tests via the following failure message:

> Expected collection {"netcoreapp1.0", "netcoreapp1.1", "netcoreapp2.0"} to contain "netcoreapp2.1" because the SupportedNetCoreAppVersions property should include the default version of .NET Core created by "dotnet new".

### DIFF
--- a/Microsoft.DotNet.Cli.sln
+++ b/Microsoft.DotNet.Cli.sln
@@ -33,7 +33,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{89905EC4
 		build\BundledTemplates.proj = build\BundledTemplates.proj
 		build\BundledTemplates.props = build\BundledTemplates.props
 		build\BundledTools.props = build\BundledTools.props
-		build\BundledVersions.targets = build\BundledVersions.targets
 		build\Compile.targets = build\Compile.targets
 		build\CrossGen.props = build\CrossGen.props
 		build\DependencyVersions.props = build\DependencyVersions.props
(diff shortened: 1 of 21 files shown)


nguerrera:
💯

dsplaisted:
@MattGertz FYI for tell mode

dsplaisted:
@dotnet-bot test this please

dsplaisted:

PR #8943, opened 2018-03-30
Remove empty PATH in path resolver

wli3:
Path adding logic on Windows is wrong. Windows require `;` in the end of the path.

Also filter out empty PATH. So it won't resolve to current directory

https://stackoverflow.com/questions/11391390/placing-the-semicolon-in-the-windows-path-environment-variable

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/EnvironmentProvider.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/EnvironmentProvider.cs
@@ -47,6 +47,7 @@ namespace Microsoft.DotNet.Cli.Utils
                         .GetEnvironmentVariable("PATH")
                         .Split(s_pathSeparator)
                         .Select(p => p.Trim(s_quote))
+                        .Where(p => !string.IsNullOrWhiteSpace(p))
                         .Select(p => ExpandTildeSlash(p)));
 
                     _searchPaths = searchPaths;
--- a/src/dotnet/ShellShim/WindowsEnvironmentPath.cs
+++ b/src/dotnet/ShellShim/WindowsEnvironmentPath.cs
@@ -40,7 +40,7 @@ namespace Microsoft.DotNet.ShellShim
 
             Environment.SetEnvironmentVariable(
                 PathName,
-                $"{existingUserEnvPath};{_packageExecutablePath};",
+                $"{existingUserEnvPath};{_packageExecutablePath}",
                 EnvironmentVariableTarget.User);
         }
 
--- a/src/dotnet/ShellShim/WindowsEnvironmentPath.cs
+++ b/src/dotnet/ShellShim/WindowsEnvironmentPath.cs
@@ -33,9 +33,14 @@ namespace Microsoft.DotNet.ShellShim
 
             var existingUserEnvPath = Environment.GetEnvironmentVariable(PathName, EnvironmentVariableTarget.User);
 
+            if (existingUserEnvPath.EndsWith(';'))
+            {
+                existingUserEnvPath = existingUserEnvPath.Substring(0, (existingUserEnvPath.Length - 1));
+            }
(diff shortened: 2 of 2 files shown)


wli3:
@dotnet-bot test Alpine3.6 x64 Debug Build 

wli3:

wli3:

wli3:

wli3:

src/dotnet/ShellShim/WindowsEnvironmentPath.cs
```diff
@@ -33,9 +33,14 @@ public void AddPackageExecutablePathToUserPath()
...
 
+            if (existingUserEnvPath.EndsWith(';'))
+            {
+                existingUserEnvPath = existingUserEnvPath.Substring(0, (existingUserEnvPath.Length - 1));
+            }
+
             Environment.SetEnvironmentVariable(
                 PathName,
-                $"{existingUserEnvPath};{_packageExecutablePath}",
+                $"{existingUserEnvPath};{_packageExecutablePath};",
```
nguerrera: Where did you get the information that final ; is required? It's not set on my Windows machine.

src/Microsoft.DotNet.Cli.Utils/EnvironmentProvider.cs
```diff
@@ -47,6 +47,7 @@ private IEnumerable<string> SearchPaths
                         .GetEnvironmentVariable("PATH")
                         .Split(s_pathSeparator)
```
nguerrera: Consider StringSplitOptions.RemoveEmptyEntries, though you still need to filter on whitespace I guess for e.g."path1;   ;path2" I guess.

src/dotnet/ShellShim/WindowsEnvironmentPath.cs
wli3: https://stackoverflow.com/questions/11391390/placing-the-semicolon-in-the-windows-path-environment-variable

wli3: in a clean vm, there is a ; at the end

nguerrera: That link doesn't say it's required nor is it normative. 

It's extremely common to not have a ; at the end.

Does clean VM have empty user PATH? Does it remain that way after you add to user path?

nguerrera: I think the earlier line being defensive against ; at the end already there is good, but this part is unnecessary.

wli3: > It's extremely common to not have a ; at the end. 

Agree. Although the code checks that.

> Does clean VM have empty user PATH? Does it remain that way after you add to user path?

No, the empty PATH is due to no check on the extra ; at the end. ";;" after split causes "" in the array of paths. Basically CLI's non perfect way adding path causes problem for CLI's non perfect way reading the path.

It is fine for Windows to read it. Since Windows will skip ";;"

nguerrera: I understand the where the empty PATH was from. As I said, I think the fix a few lines above is good.

I am only wondering if we really need to add this ; on this line at the end. It seems like it only makes ;; more likely for someone else to trip on, and I can find nothing that says it's required.

My question was trying to figure out when a ; at the end of PATH goes away on a clean VM. I never recall seeing it that way before. I was just guessing that once user PATH is set then there is no semicolon at the end.

wli3:
@nguerrera reverted the end semicolon

wli3:

wli3:
screen shot 2018-03-30 at 2 57 59 pm

PR #8950, opened 2018-03-30
Implement `buildserver shutdown` command.

### LINKED ISSUE #8185: Friendly way to shutdown VBCSCompiler and msbuild nodes
There are scenarios where it's important to shutdown the compiler and msbuild processes that can be reused to speed up subsequent builds. (Note these processes don't exist on Core yet, but are coming very soon).

We typically recommend shutting down these proceeses before and after CI. 

VBCSCompiler supports a ~~~`stop`~~~ `-shutdown` argument, but the path to VBCSCompiler.dll within the CLI should remain an implementation detail. I don't know what the equivalent would be for msbuild.

Perhaps we should have a single verb, like `dotnet stop-build-servers` (actual name TBD), that shuts down all of these.

cc @jaredpar @AndyGerlicher @khyperia @livarcocc

peterhuene:
This PR implements the `buildserver shutdown` command that can be used to
shutdown MSBuild, VB/C# compiler, and Razor build servers.

By default, all three build servers are shut down.  Options can be passed to
shut down a subset of the build servers.

Fixes #8185.

### DIFF
--- a/src/dotnet/BuildServer/IBuildServerManager.cs
+++ b/src/dotnet/BuildServer/IBuildServerManager.cs
@@ -1,0 +1,15 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.DotNet.BuildServer
+{
+    internal interface IBuildServerManager
+    {
+        string ServerName { get; }
+
+        Task<Result> ShutdownServerAsync();
+    }
+}
--- a/src/dotnet/BuildServer/RazorAssemblyResolver.cs
+++ b/src/dotnet/BuildServer/RazorAssemblyResolver.cs
@@ -1,0 +1,52 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.Build.Execution;
+using Microsoft.Extensions.EnvironmentAbstractions;
+
+namespace Microsoft.DotNet.BuildServer
+{
+    internal class RazorAssemblyResolver : IRazorAssemblyResolver
+    {
+        private readonly IDirectory _directory;
+
+        public RazorAssemblyResolver(IDirectory directory = null)
+        {
+            _directory = directory ?? FileSystemWrapper.Default.Directory;
+        }
+
+        public IEnumerable<FilePath> EnumerateRazorToolAssemblies()
+        {
+            HashSet<string> seen = new HashSet<string>();
+
+            var globalProperties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                // This property disables default item globbing to improve performance
+                // This should be safe because we are not evaluating items, only properties
+                { Constants.EnableDefaultItems, "false" }
+            };
+
+            foreach (var projectFile in _directory.EnumerateFiles(Directory.GetCurrentDirectory(), "*.*proj"))
+            {
+                var project = new ProjectInstance(projectFile, globalProperties, null);
+                var path = project.GetPropertyValue("_RazorToolAssembly");
+                if (string.IsNullOrEmpty(path))
+                {
+                    continue;
+                }
+
+                if (!seen.Add(path))
+                {
+                    continue;
+                }
+
+                yield return new FilePath(path);
+            }
+        }
+    }
+}
--- a/src/dotnet/BuildServer/RazorServerManager.cs
+++ b/src/dotnet/BuildServer/RazorServerManager.cs
@@ -1,0 +1,65 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Build.Exceptions;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.Utils;
+using Microsoft.DotNet.Tools;
+using Microsoft.Extensions.EnvironmentAbstractions;
+
+namespace Microsoft.DotNet.BuildServer
+{
+    internal class RazorServerManager : IBuildServerManager
+    {
+        private readonly IRazorAssemblyResolver _resolver;
+        private readonly ICommandFactory _commandFactory;
+
+        public RazorServerManager(IRazorAssemblyResolver resolver = null, ICommandFactory commandFactory = null)
+        {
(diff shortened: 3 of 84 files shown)


peterhuene:
For the MSBuild and Razor commands to work correctly, we need to ingest the latest builds of MSBuild and Razor.

src/dotnet/BuildServer/RazorAssemblyResolver.cs
```diff
@@ -0,0 +1,52 @@
...
+            HashSet<string> seen = new HashSet<string>();
+
+            var globalProperties = new Dictionary<string, string>
+            {
+                // This property disables default item globbing to improve performance
+                // This should be safe because we are not evaluating items, only properties
+                { Constants.EnableDefaultItems, "false" }
+            };
+
+            foreach (var projectFile in _directory.EnumerateFiles(Directory.GetCurrentDirectory(), "*.*proj"))
```
nguerrera: Hmm. I wasn't expecting CWD nor the state of projects therein to impact this. Can we not use a well-known path? Is this tool in the nuget cache or something?

peterhuene: It is in the NuGet cache.

peterhuene: e.g. `.nuget/packages/microsoft.aspnetcore.razor.design/2.1.0-preview2-30475/tools/rzc.dll`

src/dotnet/commands/dotnet-buildserver/shutdown/BuildServerShutdownCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+                    _errorReporter.WriteLine(string.Format(LocalizableStrings.ShutDownFailed, result.Message).Red());
+
+                    if (Reporter.IsVerbose)
+                    {
+                        Reporter.Verbose.WriteLine(result.Exception.ToString().Red());
+                    }
+                    return false;
+
+                default:
+                    throw new InvalidOperationException();
```
jaredpar: >throw new InvalidOperationException(); [](start = 20, length = 38)

Don't think this is the right exception. The operation isn't invalid the operation returned an illegal enum value. Should consider adding the value of the enum and name of the enum type to the error so it's diagnosable.

src/dotnet/BuildServer/IBuildServerManager.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.DotNet.BuildServer
+{
+    internal interface IBuildServerManager
+    {
+        Result Shutdown();
```
jaredpar: >Result Shutdown(); [](start = 8, length = 18)

Is there value in making this async? VBCScompiler and likely Razor can respond to this quite promptly as we're just going to kill the process. The MSBuild nodes though have to do cooperative shutdown which may take some amount of time.

src/dotnet/BuildServer/RazorServerManager.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    internal class RazorServerManager : IBuildServerManager
+    {
+        private readonly IRazorAssemblyResolver _resolver;
+        private readonly ICommandFactory _commandFactory;
+
+        public RazorServerManager(IRazorAssemblyResolver resolver = null, ICommandFactory commandFactory = null)
+        {
+            _resolver = resolver ?? new RazorAssemblyResolver();
+            _commandFactory = commandFactory ?? new DotNetCommandFactory(true);
```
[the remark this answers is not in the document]
peterhuene: Will fix.

src/dotnet/BuildServer/RazorAssemblyResolver.cs
```diff
@@ -0,0 +1,52 @@
...
+        public RazorAssemblyResolver(IDirectory directory = null)
+        {
+            _directory = directory ?? FileSystemWrapper.Default.Directory;
+        }
+
+        public IEnumerable<FilePath> EnumerateRazorToolAssemblies()
+        {
+            HashSet<string> seen = new HashSet<string>();
+
+            var globalProperties = new Dictionary<string, string>
```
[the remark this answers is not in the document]
peterhuene: Will fix.

src/dotnet/BuildServer/IBuildServerManager.cs
peterhuene: Sure, that should be easy enough to do.  I'll rethink the command output.

src/dotnet/BuildServer/RazorAssemblyResolver.cs
nguerrera: I don't know the right fix here, but I'm thinking this doesn't actually work with the general CI workflow:

```
dotnet build [sln]
dotnet buildserver shutdown
```

The use of Razor would be buried in the build of the solution, not in a *.*proj in current directory.

peterhuene: @rynowak thoughts?  We could have shutdown take a path to a project file.

nguerrera: Hmm, there's a related issue here for vbcscompiler if user is using a custom version of Microsoft.NETCore.Compilers, which I hear will be supported officially soon.

I think in all of the discussion leading up to this there was an assumption that the command is global to an SDK version, not a function of what's in your project files.

It almost feels like we need a protocol to traverse sln and find all tools to shutdown, etc. But that is going to add a bunch of evaluation time just to shutdown, and is a heck of a lot more complicated than what we were imagining when the issue was filed.

cc @agocke @jaredpar

rynowak: I might be late to this discussion, but what if we used something a little more cooperative. The idea of finding all servers running and telling them each to shut down seems a little fraught.

This could be something like all listening to a specific pipe in servers, or all watching for a specific sentinel file to be created.

> It almost feels like we need a protocol to traverse sln and find all tools to shutdown, etc. But that is going to add a bunch of evaluation time just to shutdown, and is a heck of a lot more complicated than what we were imagining when the issue was filed.

Agree and agree. sln/project are concepts that users already understand and are a good hook for more msbuild magic.

peterhuene:
@jaredpar I've amended the commit to address your feedback.  The command now asynchronously shuts down the servers as well.

src/dotnet/BuildServer/RazorAssemblyResolver.cs
peterhuene: I think this approach is acceptable for preview2, but I would like to discuss (offline) with everyone how to better discover build-related servers so we can shut them down in a more unified manner.

nguerrera: I'm OK with that.

peterhuene:
@MattGertz tell mode notice.

peterhuene:
@dotnet-bot test RHEL6 x64 Debug Build please

PR #8952, opened 2018-03-30
MSBuild 15.7.143.38429-preview

rainersigwald:
Bump master's MSBuild to the first preview3 build from MSBuild.

No urgent need for this--the changes are minimal.


AndyGerlicher:
Should this go into 3xx branch? This is in our release branch so it will get picked up by prodcon builds for 2.1 preview 2. And I think it's the only build that has the node reuse fix in it, right?

rainersigwald:
You're right: this build will go into preview2, so this should go to 3xx. I thought we'd already put it in 3xx but we had not.

I'm going to close this and submit a new PR with the right base and destination.

PR #9018, opened 2018-04-07
consume bring your own shim(byos)

wli3:
Bring your own shim

If there are shims packaged by convention in nupkg. Shim Repository will simply copy it to the right location.

The query interface ToolPackageInstance will be in charge of finding the shim folder and filter the right RID. Shim Repository will pick the right file after the folder is located since Shim Repository knows the shim name and it also book keep the files at uninstallation.

During development, due to the wrong adapter level. The mock duplicated too much logic. So, I corrected the abstraction level to lower (only create shim). And replaced the existing mock with a much smaller one without any atomic control and file move, copy logic. At the same time. The chmod, which is a IO action, causes problem during tests. So I added adapter layer to it and put it in Util


wli3:

wli3:
And tested end to end. I created a package with shim. Use this 
(very rough) https://github.com/dotnet/core/compare/master...wli3:sample-byos?expand=1 to get a package. And made an error during packaged shim creation to ensure that's the same file

src/dotnet/CommonLocalizableStrings.resx
```diff
@@ -640,4 +640,10 @@ setx PATH "%PATH%;{0}"
   <data name="FormatVersionIsMissing" xml:space="preserve">
     <value>Format version is missing. This tool may not be supported in this SDK version. Please contact the author of the tool.</value>
   </data>
-</root>
+  <data name="MoreThanOnePackagedShimAvailable" xml:space="preserve">
+    <value>More than one packaged shim available, they are {0}</value>
```
peterhuene: Perhaps just `More than one packaged shim is available: {0}.`

src/dotnet/CommonLocalizableStrings.resx
```diff
@@ -640,4 +640,10 @@ setx PATH "%PATH%;{0}"
...
+  </data>
+  <data name="FailedToReadNuGetLockFile" xml:space="preserve">
+    <value>Failed to read NuGet LockFile for tool package '{0}': {1}.</value>
```
peterhuene: Nit: generally I leave of periods from messages that contain an exception message, as that's likely to have it's own period.

src/dotnet/ShellShim/AppHostShimMaker.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            var appHostDestinationFilePath = shimPath.Value;
+            var appBinaryFilePath = PathUtility.GetRelativePath(appHostDestinationFilePath, entryPoint.Value);
+
+            EmbedAppNameInHost.EmbedAndReturnModifiedAppHostPath(
+                appHostSourceFilePath: appHostSourcePath,
+                appHostDestinationFilePath: appHostDestinationFilePath,
+                appBinaryFilePath: appBinaryFilePath);
+
+            new FilePermissionSetter().SetUserExecutionPermission(appHostDestinationFilePath);
```
peterhuene: Should this use a passed in `IFilePermissionSetter `?  Perhaps the same one that `ShellShimRepository` was passed?

src/dotnet/ToolPackage/ToolPackageInstance.cs
```diff
@@ -28,6 +32,58 @@ public IReadOnlyList<CommandSettings> Commands
             }
         }
 
+        public IReadOnlyList<FilePath> PackagedShims
```
peterhuene: I think this is too complicated (and expensive) for a property.  I think this should either be a `Lazy` initialized field or a method called `GetPackagedSims` to indicate it is performing work (I favor the first solution since this list shouldn't change).

test/Microsoft.DotNet.ToolPackage.Tests/LockFileMatcherTests.cs
```diff
@@ -1,28 +1,82 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
```
peterhuene: I'm confused by the changes to this file.  Was this file intended to move somewhere given it's no longer a test class?

src/dotnet/CommonLocalizableStrings.resx
wli3: and also @KathleenDollard for using facing string review.

src/dotnet/CommonLocalizableStrings.resx
wli3: and also @KathleenDollard for using facing string review.

src/dotnet/ToolPackage/ToolPackageInstance.cs
wli3: i genuinely confused with the line of Get vs property. In my mind most of the query methods in ToolPackageInstance do disk IO. For me that's the bar. However, do you think we should all convert it to method? Or because of most of them are Lazy so that is fine?

test/Microsoft.DotNet.ToolPackage.Tests/LockFileMatcherTests.cs
wli3: You are right, thanks for catching this. I was only look at the right side didn't realize this is a test file. Something is wrong

wli3: ah... bad bad merge....

wli3: that file is fixed

src/dotnet/ToolPackage/ToolPackageInstance.cs
peterhuene: It's not that it's inherently incorrect, I'm just of the opinion that developers tend to think of properties as "cheap" (return a field, for example) or "expensive at most-once" (lazy-init).  Having a property do work every time can be very surprising for users, especially for enumerable properties:

``` 
foreach (var shim in package.PackagedShims)
{
}

...

foreach (var shim in package.PackagedShims)
{
}
```

As someone looking at that code, I'd expect the two enumerables to be the same.  If I saw a method called `GetPackagedShims` or `EnumeratePackagedShims` instead, I'd expect each invocation to return a different enumeration.

nguerrera: "expensive-at-most-once" is the right bar for properties. Anything that does a bunch of computation or IO whenever invoked should be a method.

It looks like a field access, so it should behave like a field access. There's nothing immediately wrong looking with the following. (Other than "style nit: use foreach, but use your imagination to think of a case where indexing is actually needed.)

``` C#
for (int i = 0, i < package.PackagedShims.Count; i++)
{
     var x = package.PackagedShims[i];
}
```

But it calls package.get_PackageShims() twice per loop iteration!

src/dotnet/CommonLocalizableStrings.resx
wli3: Changed as suggested

src/dotnet/CommonLocalizableStrings.resx
wli3: Removed the last period

src/dotnet/ShellShim/AppHostShimMaker.cs
wli3: Changed as suggested

src/dotnet/ToolPackage/ToolPackageInstance.cs
wli3: Fixed, use Lazy

wli3:

wli3:
@peterhuene i addressed all the review. It is good?

wli3:

wli3:

Issue #9007, opened 2018-04-05
dotnet-install.sh installs linux-x64, not linux-arm, on an arm32v7 machine


natemcmaster:
## Steps to reproduce
Create a Dockerfile
```Dockerfile
FROM arm32v7/buildpack-deps:bionic-scm

# Install .NET CLI dependencies
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        libc6 \
        libcurl4 \
        libgcc1 \
        libgssapi-krb5-2 \
        libicu60 \
        liblttng-ust0 \
        libssl1.0.0 \
        libstdc++6 \
        libunwind8 \
        zlib1g \
    && rm -rf /var/lib/apt/lists/*

RUN curl https://dot.net/v1/dotnet-install.sh | bash -s - --version 2.1.300-preview2-008528  --install-dir ~/.dotnet
```
Execute `docker build . -t test`

## Expected  behavior
Downloads https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300-preview2-008528/dotnet-sdk-2.1.300-preview2-008528-linux-arm.tar.gz

## Actual behavior
```
dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300-preview2-008528/dotnet-sdk-2.1.300-preview2-008528-linux-x64.tar.gz
```

PR #9033, opened 2018-04-09
Dotnet install-script updates

ghost:
Introduce support for arm & arm64 architectures; when detecting a missing 'library' prerequisite - do not fail the installation.

Addresses: https://github.com/dotnet/cli/issues/9007

**Auto-detect tested successfully on:**
OS = raspbian.8
Arch = armv7l

### DIFF
--- a/scripts/obtain/dotnet-install.sh
+++ b/scripts/obtain/dotnet-install.sh
@@ -303,6 +303,28 @@ combine_paths() {
     return 0
 }
 
+get_machine_architecture() {
+    eval $invocation
+
+    if command -v uname > /dev/null; then
+        CPUName=$(uname -m)
+        case $CPUName in
+        armv7l)
+            echo "arm"
+            return 0
+            ;;
+        aarch64)
+            echo "arm64"
+            return 0
+            ;;
+        esac
+    fi
+
+    # Always default to 'x64'
+    echo "x64"
+    return 0
+}
+
 # args:
 # architecture - $1
 get_normalized_architecture_from_architecture() {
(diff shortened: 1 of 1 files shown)


natemcmaster:
Any reason we can't use `uname -m` to detect arch? Here are some sample values. I'm sure there are more. We could always err if we don't recognize the arch.

Platform | `uname -m` | Rid
---|---|---
macOS | `x86_64` | `osx-x64`
Ubuntu Bionic x64 | `x86_64` | `linux-x64`
Ubuntu Bionic arm32v7 | `armv7l` | `linux-arm`
Ubuntu Bionic arm64v8 | `aarch64` | `linux-arm64`

scripts/obtain/dotnet-install.sh
```diff
@@ -309,32 +303,24 @@ combine_paths() {
...
-            return 0
-            ;;
         amd64|x64)
             echo "x64"
             return 0
             ;;
-        x86)
-            say_err "Architecture \`x86\` currently not supported"
-            return 1
+         arm|arm32)
```
janvorli: Do we need to add the "arm32"? We don't use that name anywhere, I believe.

scripts/obtain/dotnet-install.sh
```diff
@@ -790,7 +776,7 @@ temporary_file_template="${TMPDIR:-/tmp}/dotnet.XXXXXXXXX"
 channel="LTS"
 version="Latest"
 install_dir="<auto>"
-architecture="<auto>"
+architecture="x64"
```
janvorli: A nit - should the default be the architecture of the current platform? the 'uname -m' returns a Linux specific name, so we can do:
```sh
CPUName=$(uname -m)
case $CPUName in
i686)
    architecture="x86"
    ;;
x86_64|amd64)
    architecture="x64"
    ;;
armv7l)
    architecture="arm"
    ;;
aarch64)
    architecture="arm64"
    ;;
esac
```

johnbeisner:
@natemcmaster 
@janvorli 
In the current version; the default architecture is 'x64' when "<auto>" is used.

I can see the value in trying to recognize the architecture; however, I feel strongly that we need to default to 'x64' when "<auto>" is used instead of an error-out. We would, at minimum, risk breaking fringe x64 Linux platforms if we now error-out on a peculiar return-value from 'uname -m'.

janvorli (approved):

natemcmaster (approved):

johnbeisner:

PR #9070, opened 2018-04-15
Give a different error to guide use to install via global tools

wli3:
Fix https://github.com/dotnet/cli/issues/8861

So, if several bundled DotnetTools cannot finish source build on time. The user can use global tools to get it.

The original plan that adding a different resolver is hard due to resolver can only find dll that will be used to spawn a process. However, the command constructor will give an error message when resolver find null. By adding a different error when the command name is part of the list, it can achieve the same goal.

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/Command.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/Command.cs
@@ -21,6 +21,8 @@ namespace Microsoft.DotNet.Cli.Utils
 
         private bool _running = false;
 
+        private static string[] _knownCommandsAvailableAsDotNetTool = new[] { "dotnet-dev-certs", "dotnet-ef", "dotnet-sql-cache", "dotnet-user-secrets", "dotnet-watch" };
+
         private Command(CommandSpec commandSpec)
         {
             var psi = new ProcessStartInfo
@@ -100,7 +102,11 @@ namespace Microsoft.DotNet.Cli.Utils
                 outputPath: outputPath,
                 applicationName: applicationName);
 
-            if (commandSpec == null)
+            if (_knownCommandsAvailableAsDotNetTool.Contains(commandName, StringComparer.InvariantCultureIgnoreCase) && commandSpec == null)
+            {
+                throw new CommandAvailableAsDotNetToolException(commandName);
+            }
+            else if (commandSpec == null)
             {
                 throw new CommandUnknownException(commandName);
             }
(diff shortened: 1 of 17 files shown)


src/Microsoft.DotNet.Cli.Utils/Command.cs
```diff
@@ -100,7 +102,11 @@ private Command(CommandSpec commandSpec)
                 outputPath: outputPath,
                 applicationName: applicationName);
 
-            if (commandSpec == null)
+            if (_knownCommandsAvailableAsDotNetTool.Contains(commandName, StringComparer.InvariantCultureIgnoreCase) && commandSpec == null)
```
nguerrera: Why InvariantCultureIngoreCase? Packages are OrdinalIgnoreCase as is Windows file system. I think it should be OridinalIgnoreCase.

nguerrera: Terms of `&&` should be re-ordered since there's no need to do N string comparison if command was already found.

Also consider hoisting the commandSpec == null to a single check.

nguerrera:
Looks good apart from the two comments.

src/Microsoft.DotNet.Cli.Utils/Command.cs
wli3: Changed to OrdinalIgnoreCase

wli3: hoisted. And added extra nesting. Still looks clean

src/Microsoft.DotNet.Cli.Utils/LocalizableStrings.resx
```diff
@@ -271,4 +271,9 @@
   <data name="EmbedAppNameInHostFileNameIsTooLong" xml:space="preserve">
     <value>Given file name '{0}' is longer than 1024 bytes</value>
   </data>
+  <data name="CannotFindCommandAvailableAsTool" xml:space="preserve">
+    <value>Cannot find command 'dotnet {0}', please run the following command to install
+
+dotnet tool install --global {1}</value>
```
nguerrera: Should we get a review of new user-facing text?

wli3: This text was given by Kathleen from original issue https://github.com/dotnet/cli/issues/8861

nguerrera: 👍

Issue #8563, opened 2018-02-07
Error message should have instruction to help investigate from NuGet error message


wli3:
NuGet is considered **not** an implementation detail of dotnet tools. However, when there is error of installation, there should be extra instruction to guide consumer to trouble shoot

KathleenDollard:
"The installation failed. If the NuGet message above indicates a near match that you want to install, use the --version switch as in 'dotnet install tool -g {0} --version <versionNumber>" 

@peterhuene

peterhuene:
What sort of guidance are we looking for?  Are we talking only about the prerelease issue?  I have a personal preference not to see generic guidance that isn't applicable on (potentially) a majority of failures.

KathleenDollard:
What are the possibility for failure

- The tool isn't found
- ?

KathleenDollard:
Do we still need text on this?

wli3:
the original idea i believe is to have a list of troubleshooting tips. when you see this do that... we currently have the "preview please add version" message, do we want to add more? like check your feed, you package name could conflict with a different package on other feed (often happen when you test your tool, and you call you tool something very common).

wli3:
i don't have strong opinion on this, on one side i am afraid this will become a rabbit hole and list all the place nuget error could happen. but sometimes it does help(while we don't have a reliable API to indicate different errors from nuget)

KathleenDollard:
@peterhuene Thoughts?

peterhuene:
I'm fine with expanding the guidance to a few things like @wli3 mentioned.

Something along the lines of:

```
The installation failed.  This failure may have been caused by:

* You are attempting to install a preview release and did not use the --version option to specify the version.
* A package of the same name was found from an unexpected NuGet source and it was not a .NET Core tool.
* ???
```

I wouldn't offer more than the top three, though.
*

KathleenDollard:
* Not being able to access a NuGet feed, such as not having an Internet connection
* Mistyping the name of the tool

Any others? Anything more interesting?

wli3:
@KathleenDollard I cannot think up more.

KathleenDollard:
How about:

```
The installation failed.  This failure may have been caused by:

* You are attempting to install a preview release and did not use the --version option to specify the version.
* A package by this name was found, but it was not a .NET Core tool.
* The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
* You mistyped the name of the tool.
```

[...]

PR #9079, opened 2018-04-16
Help text impoves

wli3:
fix #8869
fix https://github.com/dotnet/cli/issues/9023
fix https://github.com/dotnet/cli/issues/8563


src/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
```diff
@@ -161,12 +161,19 @@ Tool '{1}' (version '{2}') was successfully installed.</value>
...
+    <value>Tool '{0}' failed to install. This failure may have been caused by:
+
+* You are attempting to install a preview release and did not use the --version option to specify the version.
+* A package by this name was found, but it was not a .NET Core tool.
+* The required NuGet feed cannot be accessed, perhaps because of an Internet connection problem.
+* You mistyped the name of the tool.</value>
   </data>
   <data name="ToolInstallationFailedWithExplicitVersionGuide" xml:space="preserve">
-    <value>Tool '{0}' failed to install. You may need to specify the version using
+    <value>Tool '{0}' failed to install.
```
wli3: @KathleenDollard @peterhuene this is the message when nuget restore failed

src/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
```diff
@@ -161,12 +161,19 @@ Tool '{1}' (version '{2}') was successfully installed.</value>
     <value>The settings file in the tool's NuGet package is invalid: {0}</value>
   </data>
   <data name="ToolInstallationFailed" xml:space="preserve">
-    <value>Tool '{0}' failed to install.</value>
+    <value>Tool '{0}' failed to install. This failure may have been caused by:
```
wli3: @KathleenDollard @peterhuene this is the generic message no extra change other than the first sentence

peterhuene: Isn't this only used for `ShellShimException` handling?  Seems like none of these explanations can occur for that exception.

src/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
peterhuene: To me, the current `ToolInstallationFailed` message used above makes more sense for this use case, which covers restore failures (i.e. prerelease, package type mismatch, nuget feed down, couldn't find package, etc).

src/dotnet/commands/dotnet-tool/install/LocalizableStrings.resx
wli3: ShellShimException happens to be the only case. @peterhuene do you think we should revert to original? I am more worry about future usage.

wli3: @peterhuene or we should just keep them the same?

peterhuene: I think `ToolInstallationFailed` should stay the same and `ToolInstallationFailedWithExplicitVersionGuide` should use the text currently used here for `ToolInstallationFailed`.

wli3: Fixed, and renamed as suggested

PR #9122, opened 2018-04-23
do not set BaseIntermediateOutputPath in temp project of dotnet install

wli3:
Should use MsBuildProjectExtensionsPath instead.

- Change the property passin by project file instead of command line. It is more reliable passing path in xml and also the timing of MsBuildProjectExtensionsPath is controlled. (Before loading SDK)
- Change mock fake project to use “;” instead, since c:\path contains “:”.

### DIFF
--- a/src/dotnet/ToolPackage/IProjectRestorer.cs
+++ b/src/dotnet/ToolPackage/IProjectRestorer.cs
@@ -9,7 +9,6 @@ namespace Microsoft.DotNet.ToolPackage
     internal interface IProjectRestorer
     {
         void Restore(FilePath project,
-            DirectoryPath assetJsonOutput,
             FilePath? nugetConfig = null,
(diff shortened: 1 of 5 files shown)


wli3:
**Customer scenario**

After https://github.com/NuGet/NuGet.Client/pull/2131 and https://github.com/Microsoft/msbuild/pull/3059

Set BaseIntermediateOutputPath after common target will cause target not loading correctly. And it will trigger warning. Use MSBuildProjectExtensionsPath and set that property instead.

**Bugs this fixes**

https://github.com/dotnet/cli/issues/9124

**Workarounds, if any**

N/A

**Risk**

low

**Performance impact**

no

**Root cause analysis**

Dependency design change.

**How was the bug found?**

Issued by NuGet


@MattGertz ask mode for 15.8

PR #9133, opened 2018-04-24
Insert NuGet Build 4.7.0-rtm.5148 into cli

nugetlurker:
Insert NuGet Build 4.7.0-rtm.5148 into cli release/2.1.2xx branch

PR #9147, opened 2018-04-26
MSBuild 15.7.177 (2xx)

AndyGerlicher:
This is based on top of #9133 because that's the version of NuGet we have a compile-time reference to and which will ship with Preview 6. I think it caused an issue last time the assembly version changed.

Internal PR: https://devdiv.visualstudio.com/DevDiv/MSBuild/_git/VS/pullrequest/119072


AndyGerlicher:
Never mind the rebase comment. The assembly version didn't change between `5122` and `5148` (both are `4:7:0:5`) so I don't think there's any reason to gate this on NuGet since we include any of their assemblies.

AndyGerlicher:

Issue #7618, opened 2017-09-08
macOS installer is making every file exectuable


eerhardt:
## Steps to reproduce
Install the SDK .pkg on your macOS.

## Expected  behavior
Only the files that need to be executable are marked with the `x` permission.

## Actual behavior
Every file (even text files) are marked executable.

## Environment data
`dotnet --info` output:
N/A
Can be reproduced by installing the .pkg linked here https://www.microsoft.com/net/download/core

## Notes
See the code here https://github.com/dotnet/cli/blob/80c6635fdbaf4b8a04a33be53de9fbb1adbe9ff9/packaging/osx/clisdk/scripts/postinstall#L11-L12

PR #517, opened 2015-12-15
Fix the PKG permission issue and change the install location

blackdwarf:
A temporary fix for the PKG permission issue by using chmod -R 755 in the postinstall
script. Change the install location to remove the "cli" part of the path. Also
change the corehost to reflect this change.

Fixes #492 and #390 

/cc @piotrpMSFT @anurse @brthor

PR #9177, opened 2018-05-01
Remove mac mark all 755

wli3:
fix https://github.com/dotnet/cli/issues/7618

This temporary fix is added in https://github.com/dotnet/cli/pull/517 There was no core-setup at that time. And mark `dotnet` as executable should be enough. It is changing the whole folder as executable.

Today, it is covered by core-setup https://github.com/dotnet/core-setup/blob/master/src/pkg/packaging/osx/hostfxr/scripts/postinstall#L11


wli3:
cc @TheRealPiotrP

PR #9226, opened 2018-05-08
Install script updates

ghost:
**We should discuss this update:**
Adding the unsupported/unadvertised parameter: "RIDOSName"

**Other updates:**
Standardizing quotes in a conditional.
Removing superfluous method parameters.
Use lowercase with asset substrings as assets are lowercase.


kasper3:
Might be a good idea to integrate that shell checker linting in CI https://github.com/koalaman/shellcheck/wiki/Integration.
`shellcheck -f checkstyle` format is Jenkins compatible: https://github.com/koalaman/shellcheck/issues/20

PR #9279, opened 2018-05-15
MSBuild 15.8.61

AndyGerlicher:
This change removes our NuGet resolver in favor of an xml file manifest (in `Microsoft.Build.Runtime`) pointing to `Microsoft.Build.NuGetSdkResolver.dll` (next to `MSBuild.dll`) which is included with NuGet now. Removes the only direct compile-time dependency MSBuild had on NuGet.


AndyGerlicher:
@dotnet-bot test this please

AndyGerlicher:
@livarcocc think this should be good to go.

Issue #9289, opened 2018-05-16
global tools install transient failure


wli3:
This is a flaky error need to investigate.

wli3:
Because we deleted .stage folder when it is empty? But directory.create will create the subfolder regardless if there is a missing parent folder

PR #9313, opened 2018-05-18
Add retry when Directory.Move

wli3:
> Link to GitHub issue for open source changes  

https://github.com/dotnet/cli/issues/9289

> Link to GitHub dotnet:release/[version] pull request (see Git Prepping a PR from the Release Branch) 

https://github.com/dotnet/cli/pull/9313

> ProjectK TFS bug and changeset for closed source changes (security fix, package authoring, etc) 

N/A 

> Describe the issue and how it manifests to the customer. 

When running uninstall and then install on Windows. Due to global tools use Directory.Move underneath, there is a chance the indexer or the antivirus collided (had a handle still open on the prior name). And the installation/uninstalling will fail with access denied error.

> What is the impact on the customer?
 
This impact the producers most. During producer’s test, it is likely they will uninstall and install their tools several times. And the frequency of seeing this error varies machine by machine.

> Are there changes that cannot be made in GitHub? (security fixes, etc) 

No

> Characterize potential risks introduced by the fix. 

Low risk. Try action on failure. And Directory.Move, File.Move is atomic. Once it fails it can be retired.

> Does the change alter existing behaviors? (return values or types, error messages or exceptions, etc. In 
other words, is this a build-to-build breaking change?) 

No

 >> What changed?
>> What will teams need to do to respond?
>> Once check-in is approved send breaking change mail to netcorebreak

> Code reviewers:

> What testing has been completed? 

manual test

> Which platforms are impacted? (Linux | Mac | Windows) 

Windows

> Effected binaries and packages (exact name)

dotnet.dll, Microsoft.DotNet.Cli.Utils.dll

------------

Why check 0x80070005 on IOException
This error is the exact error thrown. It is just an IOException. The only way to detect that is by HResult.

Why don’t use the function above -- RetryOnFileAccessFailure?
RetryOnFileAccessFailure is async. And also in this case, it is awaiting a void. There is some gotcha around it. Since it is so close to release, I choose the safer way. And RetryOnFileAccessFailure ‘s exception is not specific. UnauthorizedAccessException is likely a real error for user to handle for –-tool-path.
 
Tested with 100 times tight loop. No error

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
@@ -50,13 +50,14 @@ namespace Microsoft.DotNet.Cli.Utils
         }
 
         /// <summary>
-        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with HResult 0x80070005. But this error is transient.
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with
+        /// HResult 0x80070005 due to Indexer. But this error is transient.
         /// </summary>
         internal static void RetryOnMoveAccessFailure(Action action)
         {
             const int ERROR_HRESULT_ACCESS_DENIED = unchecked((int)0x80070005);
-            int nextWaitTime = 1000;
-            int remainRetry = 3;
+            int nextWaitTime = 10;
+            int remainRetry = 10;
 
             while (true)
             {
@@ -68,7 +69,7 @@ namespace Microsoft.DotNet.Cli.Utils
                 catch (IOException e) when (e.HResult == ERROR_HRESULT_ACCESS_DENIED)
                 {
                     Thread.Sleep(nextWaitTime);
-                    nextWaitTime *= 10;
+                    nextWaitTime *= 2;
                     remainRetry--;
                     if (remainRetry == 0)
                     {
--- a/src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.DotNet.Cli.Utils
@@ -44,6 +45,35 @@ namespace Microsoft.DotNet.Cli.Utils
                     await Task.Delay(sleepDuration);
 
                     continue;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with HResult 0x80070005. But this error is transient.
+        /// </summary>
+        internal static void RetryOnMoveAccessFailure(Action action)
+        {
+            const int ERROR_HRESULT_ACCESS_DENIED = unchecked((int)0x80070005);
+            int nextWaitTime = 1000;
+            int remainRetry = 3;
+
+            while (true)
+            {
+                try
+                {
+                    action();
+                    break;
+                }
+                catch (IOException e) when (e.HResult == ERROR_HRESULT_ACCESS_DENIED)
+                {
+                    Thread.Sleep(nextWaitTime);
+                    nextWaitTime *= 10;
+                    remainRetry--;
+                    if (remainRetry == 0)
+                    {
+                        throw;
+                    }
                 }
             }
         }
--- a/src/dotnet/ShellShim/ShellShimRepository.cs
+++ b/src/dotnet/ShellShim/ShellShimRepository.cs
@@ -113,7 +113,7 @@ namespace Microsoft.DotNet.ShellShim
                         foreach (var file in GetShimFiles(commandName).Where(f => _fileSystem.File.Exists(f.Value)))
                         {
                             var tempPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
-                            _fileSystem.File.Move(file.Value, tempPath);
+                            FileAccessRetrier.RetryOnMoveAccessFailure(() => _fileSystem.File.Move(file.Value, tempPath));
                             files[file.Value] = tempPath;
                         }
                     }
@@ -137,7 +137,7 @@ namespace Microsoft.DotNet.ShellShim
                 rollback: () => {
                     foreach (var kvp in files)
                     {
-                        _fileSystem.File.Move(kvp.Value, kvp.Key);
+                        FileAccessRetrier.RetryOnMoveAccessFailure(() => _fileSystem.File.Move(kvp.Value, kvp.Key));
                     }
                 });
         }
--- a/src/dotnet/ToolPackage/ToolPackageInstaller.cs
+++ b/src/dotnet/ToolPackage/ToolPackageInstaller.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Xml.Linq;
 using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.Utils;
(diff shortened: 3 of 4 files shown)


src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,35 @@ public static class FileAccessRetrier
...
+            while (true)
+            {
+                try
+                {
+                    action();
+                    break;
+                }
+                catch (IOException e) when (e.HResult == ERROR_HRESULT_ACCESS_DENIED)
+                {
+                    Console.WriteLine(e.ToString() + "remainRetry: " + remainRetry);
```
wli3: oops let me remove

wli3:
fix https://github.com/dotnet/cli/issues/9289

wli3:
The following can repo the root cause 

```c#
using System;
using System.IO;
using System.Threading;
 
namespace console1
{
    class Program
    {
        static void Main(string[] args)
        {
            for (int i = 1; i < 1000000; i++)
            {
                Console.WriteLine(i);
                Directory.Move(@"C:\work\test1\folder", @"C:\work\test2\folder");
                Directory.Move(@"C:\work\test2\folder", @"C:\work\test1\folder");
            }
        }
    }
}
 
 ```
result
```
 …..
….
 
5873
5874
5875
5876
dotnet : 
At line:1 char:1
+ dotnet run >> result.txt 2>&1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Unhandled Exception:
 
System.IO.IOException: Access to the path 'C:\work\test1\folder' is denied.
   at System.IO.FileSystem.MoveDirectory(String sourceFullPath, String destFullPath)
   at System.IO.Directory.Move(String sourceDirName, String destDirName)
   at console1.Program.Main(String[] args) in C:\Users\wul\Documents\console1\Program.cs:line 14
 
```

File: src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs [hunk 2]
```diff
@@ -47,5 +48,34 @@ public static class FileAccessRetrier
                 }
             }
         }
+
+        /// <summary>
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with HResult 0x80070005. But this error is transient.
+        /// </summary>
+        internal static void RetryOnMoveAccessFailure(Action action)
+        {
+            const int ERROR_HRESULT_ACCESS_DENIED = unchecked((int)0x80070005);
```
wli3: Thank @tannergooding for dig this error code

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,34 @@ public static class FileAccessRetrier
                 }
             }
         }
+
+        /// <summary>
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with HResult 0x80070005. But this error is transient.
+        /// </summary>
+        internal static void RetryOnMoveAccessFailure(Action action)
```
peterhuene: Since this method no longer performs a move, could this be named `RetryOnAccessDenied`?

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,34 @@ public static class FileAccessRetrier
                 }
             }
         }
+
+        /// <summary>
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with HResult 0x80070005. But this error is transient.
```
peterhuene: Likewise the comment is out of date.

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,34 @@ public static class FileAccessRetrier
...
+            {
+                try
+                {
+                    action();
+                    break;
+                }
+                catch (IOException e) when (e.HResult == ERROR_HRESULT_ACCESS_DENIED)
+                {
+                    Thread.Sleep(nextWaitTime);
+                    nextWaitTime *= 10;
```
peterhuene: Maybe a factor of 5 (thus, 1 -> 5 -> 25)? 100 seconds for the final attempt seems like it will likely complete (I'd certainly ctrl-c long before then).

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
wli3: This is wrap FIle.Move and Directory.Move., _file.system.File.Move.  I don't know a easier way to enforce that, I can only put that in name. Is there a better way to convey this idea?

It could be dangerous If we use this on Directory.Copy. If the copy is partial, some file could be duplicated

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,34 @@ public static class FileAccessRetrier
...
             }
         }
+
+        /// <summary>
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with HResult 0x80070005. But this error is transient.
+        /// </summary>
+        internal static void RetryOnMoveAccessFailure(Action action)
+        {
+            const int ERROR_HRESULT_ACCESS_DENIED = unchecked((int)0x80070005);
+            int nextWaitTime = 1000;
```
tannergooding: Why not start with something like 10ms? That, combined with a 2x back-off factor will give us ~10s to give 10 tries.

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,35 @@ public static class FileAccessRetrier
...
+            {
+                try
+                {
+                    action();
+                    break;
+                }
+                catch (IOException e) when (e.HResult == ERROR_HRESULT_ACCESS_DENIED)
+                {
+                    Thread.Sleep(nextWaitTime);
+                    nextWaitTime *= 2;
```
wli3: this factor looks good?

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
```diff
@@ -47,5 +48,35 @@ public static class FileAccessRetrier
                 }
             }
         }
+
+        /// <summary>
+        /// Run Directory.Move and File.Move in Windows has a chance to get IOException with
```
wli3: @peterhuene an better idea to convey this only work for X.Move?

@nguerrera since we know this is the indexer now. I added indexer to the comment

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
tannergooding: LGTM. That gives ~10s to complete 10 retries.

peterhuene: 👍

src/Microsoft.DotNet.Cli.Utils/FileAccessRetryer.cs
peterhuene: I'm fine with the comment.

wli3:
We decide to not fix in 2.1.300 and set it to known issue. Re targeting to 2.1.4xx

wli3:
@livarcocc  need your approval for check in 2.1.4xx

PR #9315, opened 2018-05-18
LOC CHECKIN | dotnet/cli master | 20180518

v-zbsail:


peterhuene:
@livarcocc should this be targeting master?

nguerrera:
cc @tmeschter Generally, we should be merging these handbacks to whatever branch was used by loc team for the handoff. If we need another branch to be localized, we can coordinate that separately.

nguerrera:
Replaced by newer handback to master. A handback to 4xx will happen separately with handoff on July 9th and handback on July 16th.

PR #9330, opened 2018-05-22
Fix relative path tool path

wli3:
fix https://github.com/dotnet/cli/issues/9319

Replace PathUtility with new Path.GetRelativePath corefx api

Pass pull path to such api

--------------------

It would like to have a next step later that enforces DirectoryPath and FilePath always store full path. The problem I have is Path behaves differently in windows and mac. For example, /home means different thing in windows and unix. So if I store full path currently, a lot of test will fail. But I think it worth to do the proper fix to eliminate the whole class of bugs

### DIFF
--- a/src/Microsoft.DotNet.Cli.Utils/PathUtility.cs
+++ b/src/Microsoft.DotNet.Cli.Utils/PathUtility.cs
@@ -119,6 +119,11 @@ namespace Microsoft.DotNet.Tools.Common
         /// </summary>
         public static string GetRelativePath(string path1, string path2)
         {
+            if (!Path.IsPathRooted(path1) || !Path.IsPathRooted(path2))
+            {
+                throw new ArgumentException("both paths need to be rooted/full path");
+            }
+
             return GetRelativePath(path1, path2, Path.DirectorySeparatorChar, true);
         }
 
--- a/src/dotnet/ShellShim/AppHostShimMaker.cs
+++ b/src/dotnet/ShellShim/AppHostShimMaker.cs
@@ -39,8 +39,8 @@ namespace Microsoft.DotNet.ShellShim
                 appHostSourcePath = Path.Combine(_appHostSourceDirectory, ApphostNameWithoutExtension);
             }
 
-            var appHostDestinationFilePath = shimPath.Value;
-            var appBinaryFilePath = PathUtility.GetRelativePath(appHostDestinationFilePath, entryPoint.Value);
+            var appHostDestinationFilePath = Path.GetFullPath(shimPath.Value);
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.Cli.Utils/PathUtility.cs
```diff
@@ -119,6 +119,11 @@ public static string GetRelativePath(DirectoryInfo directory, FileSystemInfo chi
         /// </summary>
         public static string GetRelativePath(string path1, string path2)
         {
+            if (!Path.IsPathRooted(path1) || !Path.IsPathRooted(path2))
```
wli3: I still cannot remove this. It is used by testframework and it is netstandard. But at least I can add some guard

src/dotnet/ShellShim/AppHostShimMaker.cs
```diff
@@ -39,8 +39,8 @@ public void CreateApphostShellShim(FilePath entryPoint, FilePath shimPath)
                 appHostSourcePath = Path.Combine(_appHostSourceDirectory, ApphostNameWithoutExtension);
             }
 
-            var appHostDestinationFilePath = shimPath.Value;
-            var appBinaryFilePath = PathUtility.GetRelativePath(appHostDestinationFilePath, entryPoint.Value);
+            var appHostDestinationFilePath = Path.GetFullPath(shimPath.Value);
+            var appBinaryFilePath = Path.GetRelativePath(Path.GetDirectoryName(appHostDestinationFilePath), Path.GetFullPath(entryPoint.Value));
```
wli3: Path.GetRelativePath need to pass a directory. As suggested by its document

peterhuene:
Do we have a test to cover the execution of the shim so that we can ensure this regression doesn't happen again?

src/Microsoft.DotNet.Cli.Utils/PathUtility.cs
wli3: Another con I can think of is we cannot dogfood fix netstandard, netcore projects.

wli3:
let me add test

test/Microsoft.DotNet.ShellShim.Tests/ShellShimRepositoryTests.cs
```diff
@@ -45,6 +45,27 @@ public void GivenAnExecutablePathItCanGenerateShimFile()
             stdOut.Should().Contain("Hello World");
         }
 
+        // Reproduce https://github.com/dotnet/cli/issues/9319
```
wli3: @peterhuene @nguerrera added test to repro. I reverted 2 commits, see the repro and re apply the 2 commits back, and see it is fixed

wli3:

wli3:
🔔  @peterhuene @nguerrera  :)

PR #9333, opened 2018-05-22
Fixing the 'signing' step.

ghost:
Adding 'MicroBuild' props and targets so that the VSO build's signing steps can interact with the Signing PlugIn.

**Test:**
https://devdiv.visualstudio.com/DevDiv/Default/_build/index?buildId=1715666&_a=summary
Note: inspect the "Sign assemblies" step


build/signing/MicroBuild.Core.props
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tannergooding: Why are these checked in rather than being consumed from NuGet/MicroBuild like all the other repositores?

johnbeisner: rel/1.1.0 was not like other repositories - to wit: we were previously consuming them from a non-nuget location: "build_tools"
Import Project="$(RepoRoot)/build_tools/MicroBuild.Core.props"

We can take the extra step to consume from a nuget location; I have indeed looked into doing this.

tannergooding: We should do as such, we don't want to miss out on any critical fixes that were made to the MicroBuild props/targets

MattGal (approved):

PR #9339, opened 2018-05-23
Splitting the 'ItPublishesARunnableSelfContainedApp' test into 2 tests.

ghost:
Splitting the 'ItPublishesARunnableSelfContainedApp' test into: an 'implicit' and an 'explicit' version.

Addresses: https://github.com/dotnet/sdk/issues/2263

### DIFF
--- a/test/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
+++ b/test/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
@@ -75,7 +75,7 @@ namespace Microsoft.DotNet.Cli.Publish.Tests
         }
 
         [Theory]
-        [InlineData("1.1.9-servicing-001731-00")]
+        [InlineData("1.1.9-servicing-001739-00")]
         public void ItPublishesARunnableSelfContainedAppWithAnExplictRuntime(string runtimeFrameworkVersion)
         {
             var testAppName = "MSBuildTestApp";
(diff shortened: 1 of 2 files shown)


test/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -77,6 +74,46 @@ public void ItPublishesARunnableSelfContainedApp()
                      .And.HaveStdOutContaining("Hello World");
         }
 
+        [Theory]
+        [InlineData("1.1.9-servicing-001731-00")]
```
dsplaisted: I think we need to figure out a way not to hard-code this in the test.  Some options would be to look at what version is available in the Shared/Microsoft.NETCore.App folder in the .NET Core SDK that's being used, to read the version from Microsoft.DotNet.Cli.DependencyVersions.props, or use the MSBuild property to write a text file which would be included as an embedded resource in the test assembly.

test/dotnet-publish.Tests/GivenDotnetPublishPublishesProjects.cs
```diff
@@ -77,6 +74,46 @@ public void ItPublishesARunnableSelfContainedApp()
...
+                .WithRestoreFiles();
+
+            var testProjectDirectory = testInstance.Root;
+
+            var rid = DotnetLegacyRuntimeIdentifiers.InferLegacyRestoreRuntimeIdentifier();
+
+            new RestoreCommand()
+                .WithRuntime(rid)
+                .WithWorkingDirectory(testProjectDirectory)
+                .Execute($"/p:SkipInvalidConfigurations=true /p:RuntimeFrameworkVersion={runtimeFrameworkVersion}")
```
dsplaisted: What does `SkipInvalidConfigurations` do?  Why is it necessary here?

dsplaisted:
The two CI failures look like they might be real issues with either the product or the tests, although I don't know why they would only happen for two of the legs.

johnbeisner:
@dotnet-bot test this please

johnbeisner:
@dsplaisted 
Fedora 24 will always show as failed - it has been removed from the CI line-up.

johnbeisner:
closing in favor of: https://github.com/dotnet/cli/pull/9536

PR #9354, opened 2018-05-27
Change configuration completion suggestion to standard capitalization

dasMulli:
Changes completion suggestions for `-c` to standard capitalisation to prevent against issues with case sensitivity.

Tentatively against 2.1.4xx, but should be a clean merge into master as well.


dasMulli:
@livarcocc can this go in? the build failure requiring the test was a restore failure.

PR #9362, opened 2018-05-29
Fixing 'Channel' and 'BranchName': "release/2.1.4xx" to "master"

ghost:
Fixing 'Channel' and 'BranchName': "release/2.1.4xx" to "master" - also the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Looking for V1 of the .NET Core tooling?
 
 If you are looking for the v2.0 release of the .NET Core tools (CLI, MSBuild and the new csproj), head over to https://dot.net/core and download!
 
-> **Note:** the release/2.1.4xx branch of the CLI repo is based on an upcoming update of the SDK and is considered pre-release. For production-level usage, please use the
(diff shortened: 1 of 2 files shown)


peterhuene (approved):

dsplaisted (approved):
We should really be calculating the information in BranchInfo.props dynamically, otherwise we will continue to get this wrong.

We can't do that for the README.md file, but if we were calculating the `Channel` and `BranchName` dynamically, then we could have a test that ensured that README.md was up-to-date.

tannergooding:
We should probably have a validation job (possibly a MicroBuild job, to parity what Roslyn has).

That job can trivially validate things like that publish branch or created packages matching the branch that GitHub pulls down. It also avoids the problem of building from sources that don't have branch info or that weren't checked out via Git.

dsplaisted:
@tannergooding @johnbeisner Can one of you file an issue to improve things here?  IE either calculating these properties dynamically, or validating that they are correct in a validation job.

johnbeisner:
git rev-parse --abbrev-ref HEAD

PR #9363, opened 2018-05-30
Always store absolute full path in directory and file path

wli3:
There is no need to store relative path today. But some part of the system does not accept relative path and there is no indication if it is storing full path or not. This is the root cause of https://github.com/dotnet/cli/issues/9319

“someplace” means different full path for Path class on unix and Windows. And the mock file system uses real Path class. Change tests' setup to use essentially “TEMPATH/someplace” instead of  “someplace”

### DIFF
--- a/src/Microsoft.DotNet.InternalAbstractions/DirectoryPath.cs
+++ b/src/Microsoft.DotNet.InternalAbstractions/DirectoryPath.cs
@@ -10,8 +10,17 @@ namespace Microsoft.Extensions.EnvironmentAbstractions
     {
         public string Value { get; }
 
+        /// <summary>
+        /// Create DirectoryPath to repesent a absolute directory path. Note it may not exist.
+        /// </summary>
+        /// <param name="value">If the value is not rooted. Path.GetFullPath will be called during the consturctor.</param>
         public DirectoryPath(string value)
         {
+            if (!Path.IsPathRooted(value))
+            {
+                value = Path.GetFullPath(value);
+            }
+
             Value = value;
         }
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.DotNet.InternalAbstractions/DirectoryPath.cs
```diff
@@ -10,8 +10,17 @@ public struct DirectoryPath
     {
         public string Value { get; }
 
+        /// <summary>
+        /// Create DirectoryPath to repesent a absolute directory path. Note it may not exist.
+        /// </summary>
+        /// <param name="value">If the value is not rooted. Path.GetFullPath will be called during the consturctor.</param>
         public DirectoryPath(string value)
         {
+            if (!Path.IsPathRooted(value))
```
wli3: I am not sure I should automatic convert to fullpath or throw on not rooted.

Both of them has no type check (design time feed back). And if your tests always use full path, it will throw on runtime when it is not.

It is more pure to throw on not rooted. But GetParentPath are calling full path already.

wli3:
@dotnet-bot test opensuse.43.2 x64 Debug Build

wli3:

src/Microsoft.DotNet.InternalAbstractions/DirectoryPath.cs
```diff
@@ -46,7 +55,7 @@ public override string ToString()
 
         public DirectoryPath GetParentPath()
         {
-            return new DirectoryPath(Directory.GetParent(Path.GetFullPath(Value)).FullName);
+            return new DirectoryPath(Path.GetDirectoryName(Path.GetFullPath(Value)));
```
peterhuene: Does this still need the `Path.GetFullPath` call here?

wli3: Fixed

wli3:
@dotnet-bot test this place

wli3:
@dotnet-bot test this please

PR #9382, opened 2018-05-31
Update runtimeconfig.json doc for 2.1

steveharter:
Fixes https://github.com/dotnet/cli/issues/7861.

Update terminology, samples, and add new information for 2.1.0. There is still some detail regarding project.lock.json and other items that I didn't modify.

### DIFF
--- a/Documentation/specs/runtime-configuration-file.md
+++ b/Documentation/specs/runtime-configuration-file.md
@@ -2,43 +2,42 @@
 
 The runtime configuration files store the dependencies of an application (formerly stored in the `.deps` file). They also include runtime configuration options, such as the Garbage Collector mode. Optionally they can also include data for runtime compilation (compilation settings used to compile the original application, and reference assemblies used by the application).
 
-**Note:** This document doesn't provide full explanations as to why individual items are needed in this file. That is covered in the [`corehost` spec](corehost.md) and via the `Microsoft.Extensions.DependencyModel` assembly.
+**Note:** This document doesn't provide full explanations as to why individual items are needed in this file. That is covered in the [`apphost` spec](corehost.md) and via the `Microsoft.Extensions.DependencyModel` assembly.
 
 ## What produces the files and where are they?
 
 There are two runtime configuration files for a particular application. Given a project named `MyApp`, the compilation process produces the following files (on Windows, other platforms are similar):
 
 * `MyApp.dll` - The managed assembly for `MyApp`, including an ECMA-compliant entry point token.
-* `MyApp.exe` - A copy of the `corehost.exe` executable. Only present when the application is a self-contained application.
-* `MyApp.runtimeconfig.json` - An **optional** configuration file containing runtime configuration settings. This file is required for portable applications, but not for self-contained apps.
-* `MyApp.deps.json` - A list of dependencies, as well as compilation context data and compilation dependencies. Not technically required, but required to use the servicing or package cache/shared package install features.
+* `MyApp.exe` - A copy of the `apphost.exe` executable. This is present when the application is self-contained, and in newer functionality (2.1.0+) for framework-dependent applications that wish to support platform-specific (non-portable) executables.
+* `MyApp.runtimeconfig.json` - An **optional** configuration file containing runtime configuration settings. This file is required for framework-dependent applications, but not for self-contained apps.
+* `MyApp.runtimeconfig.dev.json` - An **optional** configuration file containing runtime configuration settings that typically only exists in a non-published output and thus is used for development-time scenarios. This file typically specifies additional probing paths. Depending on the semantics of each setting, the setting is either combined with or overridden by the values from `MyApp.runtimeconfig.json`.
+* `MyApp.deps.json` - A list of dependencies, as well as compilation context data, compilation dependencies and version information used to address assembly conflicts. Not technically required, but required to use the servicing or package cache/shared package install features. If the file is not present, all assemblies in the current folder are used instead.
 
 The `MyApp.runtimeconfig.json` is designed to be user-editable (in the case of an app consumer wanting to change various CLR runtime options for an app, much like the `MyApp.exe.config` XML file works in .NET 4.x today). However, the `MyApp.deps.json` file is designed to be processed by automated tools and should not be user-edited. Having the files as separate makes this clearer. We could use a different format for the deps file, but if we're already integrating a JSON parser into the host, it seems most appropriate to re-use that here. Also, there are diagnostic benefits to being able to read the `.deps.json` file in a simple text editor.
 
-**IMPORTANT**: Portable Applications, i.e. those published without a specific RID, have some adjustments to this spec which is covered at the end.
+**IMPORTANT**: Framework-dependent applications have some adjustments to this spec which are covered at the end.
 
 ## File format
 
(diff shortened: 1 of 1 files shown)


Documentation/specs/runtime-configuration-file.md
```diff
@@ -2,206 +2,155 @@
...
+* `MyApp.runtimeconfig.json` - An **optional** configuration file containing runtime configuration settings. This file is required for framework-dependent applications, but not for self-contained apps.
+* `MyApp.runtimeconfig.dev.json` - An **optional** configuration file containing runtime configuration settings that typically only exists in a non-published output and thus is used for development-time scenarios. This file typically specifies additional probing paths. Depending on the semantics of each setting, the setting is either combined with or overridden by the values from `MyApp.runtimeconfig.json`.
+* `MyApp.deps.json` - A list of dependencies, as well as compilation context data, compilation dependencies and version information used to address assembly conflicts. Not technically required, but required to use the servicing or package cache/shared package install features. If the file is not present, all assemblies in the current folder are used instead.
```
eerhardt: I think `compilation dependencies` is redundant with `compilation context data`, isn't it?

Documentation/specs/runtime-configuration-file.md
```diff
@@ -2,206 +2,155 @@
...
+* `applyPatches` - When `false`, the framework version is strictly obeyed by the host. When `applyPatches` is unspecified or `true`, the framework from either the same or a higher version that differs only by the patch field will be used.
   * For example, if `version=1.0.1` and `applyPatches` is `true`, the host would load the shared framework from `1.0.{n}`, where `n >= 1`, but will not load from `1.1.0`, even if present. When `applyPatches` is `false`, the shared framework will be loaded from `1.0.1` strictly.
-  * **Note:** This does not apply to `SemVer`'s `prerelease` versions, but only for `production` releases.
-  * **Note:** This section will not be used for standalone applications that do not rely upon a shared framework.
-
-* Others _TBD_
+  * **Note:** This does not apply to `pre-release` versions; it applies only to `production` releases.
+  * **Note:** This section should not exist self-contained applications because they do not rely upon a shared framework.
+* `rollForwardOnNoCandidateFx` - Determines roll-forward behavior of major and minor. Only applies to `production` releases. Values: 0(Off), 1 (roll forward on [minor]), Roll forward on [major] and [minor]
```
eerhardt: What's the value for `Roll forward on [major] and [minor]`?  We have `0`, `1` for the other two, but this one doesn't seem to have a value.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -2,206 +2,155 @@
...
+ See [roll-forward-on-no-candidate documentation](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md) for more information.
```
eerhardt: Can you use a relative link here?  They are easier to maintain across branches.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -2,206 +2,155 @@
...
 
-These settings are read by `corehost` to determine how to initialize the runtime. All versions of `corehost` **must ignore** settings in this section that they do not understand (thus allowing new settings to be added in later versions).
+These settings are read by `apphost` to determine how to initialize the runtime. All versions of `apphost` **must ignore** settings in this section that they do not understand (thus allowing new settings to be added in later versions).
```
eerhardt: IMO we shouldn't use "apphost" to mean both the `apphost.exe` and `dotnet.exe`.  Can we just say "host" when we mean them both?

Documentation/specs/runtime-configuration-file.md
```diff
@@ -2,206 +2,155 @@
...
-
-```json
-{
-    "compilationOptions": {
-        "allowUnsafe": true,
-        "defines": [ "CORE_CLR", "DEBUG_MODE" ]
-    }
-}
-```
+This section is created during build from the project's settings. The exact settings found here are specific to the compiler that produced the original application binary. Some example settings include: `defines`, `languageVersion` (C#/VB), `allowUnsafe` (C#), etc.
```
eerhardt: "`languageVersion` (C#/VB)" makes it sound like C# or VB is the possible values for these setting, when in reality it is a number like `7.3`

Documentation/specs/runtime-configuration-file.md
```diff
@@ -338,9 +236,9 @@ The shared framework's deps file will also contain a `runtimes` section defining
 
 The host will have a RID embedded in it during compilation (for example, `win10-x64` for Windows 64-bit). It will look up the corresponding entry in the `runtimes` section to identify what the fallback list is for `win10-x64`. The fallbacks are identified from most-specific to least-specific. In the case of `win10-x64` and the example above, the fallback list is: `"win10-x64", "win10", "win81-x64", "win81", "win8-x64", "win8", "win7-x64", "win7", "win-x64", "win", "any", "base"` (note that an exact match on the RID itself is the first preference, followed by the first item in the fallback list, then the next item, and so on).
```
eerhardt: `The host will have a RID embedded in it during compilation`
This line is incorrect. Can you fix it?  We do RID detection in the host now.

Documentation/specs/runtime-configuration-file.md
```diff
@@ -2,206 +2,155 @@
 
 The runtime configuration files store the dependencies of an application (formerly stored in the `.deps` file). They also include runtime configuration options, such as the Garbage Collector mode. Optionally they can also include data for runtime compilation (compilation settings used to compile the original application, and reference assemblies used by the application).
 
-**Note:** This document doesn't provide full explanations as to why individual items are needed in this file. That is covered in the [`corehost` spec](corehost.md) and via the `Microsoft.Extensions.DependencyModel` assembly.
+**Note:** This document doesn't provide full explanations as to why individual items are needed in this file. That is covered in the [`apphost` spec](corehost.md) and via the `Microsoft.Extensions.DependencyModel` assembly.
```
nguerrera: Aside: is this the best place for this document? Should it go in core-setup? Or docs.microsoft.com?

Documentation/specs/runtime-configuration-file.md
steveharter: I don't think relative across branches works. Also, if someone has a local branch the link probably won't work.

Documentation/specs/runtime-configuration-file.md
steveharter: Done

Documentation/specs/runtime-configuration-file.md
steveharter: Changing to "(e.g. the version of C# or VB)"

Documentation/specs/runtime-configuration-file.md
steveharter: Chaning to "The host will detect the RID at runtime"

Documentation/specs/runtime-configuration-file.md
steveharter: Yes I could see the "corehost" spec be moved to core-setup documentation to be consistent with other related documents also located there.

steveharter:
Any other comments? I need at least one "approval" to commit. Thanks

PR #9384, opened 2018-05-31
Update to RTM release of .NET Core 2.1 runtime (and ASP.NET)

dsplaisted:


peterhuene:
Shouldn't this be coming with a merge up from 2.1.3xx?

dsplaisted:
> Shouldn't this be coming with a merge up from 2.1.3xx?

@peterhuene Good point, perhaps we just need to merge #9350

PR #9386, opened 2018-05-31
Clear disabled package sources in generated NuGet.config

dsplaisted:
This will help prevent the repo being broken if you have the same feeds set up in your global NuGet.config but have them disabled

### DIFF
--- a/build/NugetConfigFile.targets
+++ b/build/NugetConfigFile.targets
@@ -32,6 +32,9 @@
       <NugetConfigSuffix>
         <![CDATA[
 </packageSources>
+<disabledPackageSources>
+   <clear/>
+</disabledPackageSources>
 </configuration>
         ]]>
       </NugetConfigSuffix>


nguerrera:
Should we put this infrastructure change in 3xx?

johnbeisner:
@dsplaisted 
Why do we need another "clear"? - we have one on line 15.

dsplaisted:
> Should we put this infrastructure change in 3xx?

I briefly considered this but I figured it was simpler to add it to 4xx for now and cherry-pick it to 3xx if we end up needing it.

> Why do we need another "clear"? - we have one on line 15.

The problem is the current clear applies to the `packageSources` section.  We need another clear for the `disabledPackageSources`.  That's not intuitive to me and I was very confused about why my builds were failing until I eventually figured it out (with the NuGet team's help), but it's how the NuGet.config works.

Issue #9420, opened 2018-06-05
Test failure in ProdCon: NewProjectRestoresCorrectPackageVersion


johnbeisner:
When moving the microsoft.netcore.app version forward to 2.1.1* from it's normal "2.1.0" - the NewProjectRestoresCorrectPackageVersion "console", "microsoft.netcore.app" test breaks:

Expected directory "2.1.1-servicing-26603-03" cannot be found inside directory "E:\A\_work\3761\s\bin\3\win-x86\test\dotnet-new.Tests\NewProjectRestoresCorrectPackageVersion_console\temp\packages\microsoft.netcore.app".

The "2.1.0" directory exists.

PR #9421, opened 2018-06-05
Pass "PB_AssetRootUrl" explictly on the MSBuild call and fix two tests

ghost:
Pass "PB_AssetRootUrl" explictly on the MSBuild call; we are looking for Microsoft.NETCore.App "2.1.0" or "2.1.1*"

Two basic ProdCon issues:
1) Some of the secondary or internal MSBuild calls need to know secret values
2) The version of Microsoft.NETCore.App could be something other than "2.1.0"

**Additional:** 
The AllLinux VSO build definition was altered to include 3 secrets as environmental variables - as adding the secret values explicitly to the secondary MSBuild call yielded this error: System.InvalidOperationException: An invalid request URI was provided.  /p:PB_PackageVersionPropsUrl=https:/dotnetfeed.blob.core.windows.net/orchestrated-release-2-1/20180603-02/cli-inputs/orchestration-metadata/PackageVersions.props

Please notice the escaped "//"

Related: https://github.com/dotnet/cli/issues/9420

### DIFF
--- a/build/Test.targets
+++ b/build/Test.targets
@@ -89,7 +89,7 @@
   </Target>
 
   <Target Name="TestsForBuildItself">
-    <DotNetMSBuild Arguments="/v:diag $(RepoRoot)/test/MsBuildScript.Tests/runtests.target"
+    <DotNetMSBuild Arguments="/v:diag $(RepoRoot)/test/MsBuildScript.Tests/runtests.target /p:PB_AssetRootUrl=$(PB_AssetRootUrl)"
                    ToolPath="$(PreviousStageDirectory)" />
   </Target>
 
--- a/test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectToolsCommandResolver.cs
+++ b/test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectToolsCommandResolver.cs
@@ -304,7 +304,7 @@ namespace Microsoft.DotNet.Tests
 
             result.Should().NotBeNull();
 
-            result.Args.Should().Contain("--fx-version 2.1.0");
+            result.Args.Should().Contain("--fx-version 2.1.");
         }
 
         [Fact]
--- a/test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectToolsCommandResolver.cs
+++ b/test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectToolsCommandResolver.cs
@@ -304,7 +304,7 @@ namespace Microsoft.DotNet.Tests
 
             result.Should().NotBeNull();
 
-            result.Args.Should().Contain("--fx-version 2.1.");
+            result.Args.Should().Contain("--fx-version 2.1.1");
         }
 
         [Fact]
--- a/test/dotnet-new.Tests/GivenThatIWantANewApp.cs
+++ b/test/dotnet-new.Tests/GivenThatIWantANewApp.cs
@@ -75,7 +75,7 @@ namespace Microsoft.DotNet.New.Tests
         }
 
         [Theory]
-        [InlineData("console", "microsoft.netcore.app")]
+        // [InlineData("console", "microsoft.netcore.app")] re-enable when this issue is resolved: "https://github.com/dotnet/cli/issues/9420"
         [InlineData("classlib", "netstandard.library")]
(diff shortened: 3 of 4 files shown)


build/Test.targets
```diff
@@ -89,7 +89,7 @@
   </Target>
 
   <Target Name="TestsForBuildItself">
-    <DotNetMSBuild Arguments="/v:diag $(RepoRoot)/test/MsBuildScript.Tests/runtests.target"
+    <DotNetMSBuild Arguments="/v:diag $(RepoRoot)/test/MsBuildScript.Tests/runtests.target /p:PB_AssetRootUrl=$(PB_AssetRootUrl)"
```
mmitche: Is it only the asset root url that needs to be passed here?

mmitche (approved):

test/Microsoft.DotNet.Cli.Utils.Tests/GivenAProjectToolsCommandResolver.cs
```diff
@@ -304,7 +304,7 @@ public void ItAddsFxVersionAsAParamWhenTheToolHasThePrefercliruntimeFile()
 
             result.Should().NotBeNull();
 
-            result.Args.Should().Contain("--fx-version 2.1.0");
+            result.Args.Should().Contain("--fx-version 2.1.");
```
livarcocc: We should update this to 2.1.1.

test/dotnet-new.Tests/GivenThatIWantANewApp.cs
```diff
@@ -75,8 +75,7 @@ public void RestoreDoesNotUseAnyCliProducedPackagesOnItsTemplates()
         }
 
         [Theory]
-        [InlineData("console", "microsoft.netcore.app")]
```
livarcocc: In order to re-enable this test, just update the MicrosoftNETCoreAppPackageVersion to 2.1.1 in DependencyVersions.props.

test/dotnet-new.Tests/GivenThatIWantANewApp.cs
```diff
@@ -107,7 +106,7 @@ string GetFrameworkPackageVersion()
                 var sharedFxDir = dotnetDir
                     .GetDirectory("shared", "Microsoft.NETCore.App")
                     .EnumerateDirectories()
-                    .Single(d => d.Name.StartsWith("2.1.0"));
+                    .Single(d => d.Name.StartsWith("2.1."));
```
livarcocc: With the change above, you can set this to 2.1.1.

livarcocc (changes requested):
See my comments.

build/Test.targets
MattGal: As long as this works, great, but it may be worth wrapping in escaped quotes in the command to prevent weird escaping.

MattGal (approved):
LGTM

test/dotnet-new.Tests/GivenThatIWantANewApp.cs
mmitche: I think it will need to be the latest pre-release version of 2.1.1 right?

livarcocc (approved):

johnbeisner:
@livarcocc 
Renabling the NewProjectRestoresCorrectPackageVersion test doesn't pass for ProdConn type builds:

Failed   Microsoft.DotNet.New.Tests.GivenThatIWantANewApp.NewProjectRestoresCorrectPackageVersion(type: "console", packageName: "microsoft.netcore.app")
Expected directory "2.1.1-servicing-26603-03" cannot be found inside directory "E:\A\_work\577\s\bin\3\win-x64\test\dotnet-new.Tests\NewProjectRestoresCorrectPackageVersion_console\temp\packages\microsoft.netcore.app".

build/Test.targets
johnbeisner: @mmitche 
PB_PACKAGEVERSIONPROPSURL and EXTERNALRESTORESOURCES have already generated the 'nuget.config' and the 'OrchestratedPackageVersionsProps.props' - however, 'PB_AssetRootUrl' is needed to re-calculate the download locations for the secondary MSBuild call.

mmitche:
@dotnet-bot test this please

mmitche:
@dotnet-bot test this please

PR #9430, opened 2018-06-06
Drop the LZMA archive for every build

ghost:
**build/compile/LzmaArchive.targets** 
copy the LZMA to the packages folder for publishing

**build/Publish.targets** 
Only one leg should do the publish

### DIFF
--- a/build/compile/LzmaArchive.targets
+++ b/build/compile/LzmaArchive.targets
@@ -1,21 +1,23 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <PropertyGroup>
-      <FinalArchive>$(SdkOutputDirectory)/nuGetPackagesArchive.lzma</FinalArchive>
-      <NugetPackagesArchiveName>nuGetPackagesArchive-$(AspNetCoreVersion).lzma</NugetPackagesArchiveName>
-      <IntermediateArchive>$(IntermediateDirectory)/$(NugetPackagesArchiveName)</IntermediateArchive>
-      <NugetPackagesArchiveBlobUrl>$(AspNetCoreSharedFxRootUrl)$(AspNetCoreVersion)/$(NugetPackagesArchiveName)</NugetPackagesArchiveBlobUrl>
-    </PropertyGroup>
+  <PropertyGroup>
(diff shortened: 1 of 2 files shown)


build/compile/LzmaArchive.targets
```diff
@@ -17,5 +18,6 @@
                     ContinueOnError="WarnAndContinue" />
 
       <Copy SourceFiles="$(IntermediateArchive)" DestinationFiles="$(FinalArchive)" />
+      <Copy SourceFiles="$(IntermediateArchive)" DestinationFiles="$(PackagesArchive)" />
```
nguerrera: This looks like it will only copy the lzma if it is included in the SDK zip/installer. Is that what '$(IncludeNuGetPackageArchive)' == 'true' does? If so, then I think this won't address the request, which was to have flavors (ARM) where the lzma is not included in the zip/installer, but was easily acquired separately.

johnbeisner: @nguerrera 
I feel this would result in a 'double' publish [Linux-arm + Linux-arm64] which would fail one of the publishes as the name collides: "nuGetPackagesArchive.lzma"

Since the LZMA is not platform specific we are picking a known VSO build to do the singular publish -> Windows x64

nguerrera: Oh, I see that we'd only publish it with x64. Is the expectation that users get it from the x64 directory. I guess there's nothing platform specific about it. It will just be confusing in docker files to see ti get downloaded from x64.

nguerrera: Oh it's all in one folder. I see. OK.

nguerrera: Nevermind this comment thread, then.

nguerrera: Why do we need to copy it to yet another directory. This file isn't small. Can we just publish it from where it's already written?

johnbeisner: @nguerrera 
The path to pick this up would be: "http://dotnetcli.blob.core.windows.net/dotnet/Sdk/2.2.100-preview1-008982/nuGetPackagesArchive.lzma"

nguerrera: Looks great.

nguerrera (approved):
Still curious if we can avoid an extra copy, but otherwise LGTM

MichaelSimons (approved):

johnbeisner:

PR #9465, opened 2018-06-12
Remove xml encoding

wli3:
fix https://github.com/dotnet/cli/issues/9413
Rely on CommandSpecFactory for the quoting

This fix can resolve the problem when the user has Non ASCII character in their $HOME directory. However, it will **not** resolve all the case when there is control character in folder name like “alexander, the great”. It will keep the “same level” with `dotnet build -o` which should be acceptable. Please see the detail in the issue. Since this goes to 2.1.3xx. I try to keep the change minimal. The existing escaping is far from ideal. I hope to keep investigating into this issue and find a general solution for 2.1.4xx.

### DIFF
--- a/src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
+++ b/src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
@@ -44,7 +44,7 @@ namespace Microsoft.DotNet.Tools.Tool.Install
             {
                 "--runtime",
                 AnyRid,
-                $"-property:BaseIntermediateOutputPath={assetJsonOutput.ToXmlEncodeString()}"
+                $"-property:BaseIntermediateOutputPath={assetJsonOutput.Value}"
             });
 
             argsToPassToRestore.Add($"-verbosity:{verbosity ?? "quiet"}");
--- a/test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageInstallerTests.cs
+++ b/test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageInstallerTests.cs
@@ -617,7 +617,9 @@ namespace Microsoft.DotNet.ToolPackage.Tests
         {
             var nugetConfigPath = WriteNugetConfigFileToPointToTheFeed();
 
-            string nonAscii = "ab Ṱ̺̺̕o 田中さん åä";
+            var surrogate = Char.ConvertFromUtf32(Int32.Parse("0301", NumberStyles.HexNumber));
+            string nonAscii = "ab Ṱ̺̺̕o 田中さん åä" + surrogate;
+
             var root = new DirectoryPath(Path.Combine(TempRoot.Root, nonAscii, Path.GetRandomFileName()));
             var reporter = new BufferedReporter();
             var fileSystem = new FileSystemWrapper();
@@ -635,6 +637,7 @@ namespace Microsoft.DotNet.ToolPackage.Tests
                 nugetConfig: nugetConfigPath);
 
             AssertPackageInstall(reporter, fileSystem, package, store);
+            Char.IsSurrogate(surrogate).Should().BeTrue();
 
             package.Uninstall();
         }
--- a/test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageInstallerTests.cs
+++ b/test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageInstallerTests.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and contributors. All rights reserved.
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -606,6 +606,33 @@ namespace Microsoft.DotNet.ToolPackage.Tests
             reporter.Lines.Should().Contain(l => l.Contains("warning"));
             reporter.Lines.Should().NotContain(l => l.Contains(tempProject.Value));
             reporter.Lines.Clear();
+
+            AssertPackageInstall(reporter, fileSystem, package, store);
+
+            package.Uninstall();
+        }
+
+        [Fact]
+        public void GivenARootWithNonAsciiCharactorInstallSucceeds()
+        {
+            var nugetConfigPath = WriteNugetConfigFileToPointToTheFeed();
+
+            string nonAscii = "ab Ṱ̺̺̕o 田中さん åä";
+            var root = new DirectoryPath(Path.Combine(TempRoot.Root, nonAscii, Path.GetRandomFileName()));
+            var reporter = new BufferedReporter();
+            var fileSystem = new FileSystemWrapper();
+            var store = new ToolPackageStore(root);
+            var installer = new ToolPackageInstaller(
+                store: store,
+                projectRestorer: new ProjectRestorer(reporter),
+                tempProject: GetUniqueTempProjectPathEachTest(),
+                offlineFeed: new DirectoryPath("does not exist"));
+
+            var package = installer.InstallPackage(
+                packageId: TestPackageId,
+                versionRange: VersionRange.Parse(TestPackageVersion),
+                targetFramework: _testTargetframework,
+                nugetConfig: nugetConfigPath);
 
             AssertPackageInstall(reporter, fileSystem, package, store);
 


test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageInstallerTests.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
wli3: this is encoding change

wli3:
@livarcocc do you have context on escaping quotes? Do you think this solution for 2.1.3xx is good enough?

src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
```diff
@@ -44,7 +44,7 @@ public ProjectRestorer(IReporter reporter = null)
             {
                 "--runtime",
                 AnyRid,
-                $"-property:BaseIntermediateOutputPath={assetJsonOutput.ToXmlEncodeString()}"
+                $"-property:BaseIntermediateOutputPath={assetJsonOutput.Value}"
```
peterhuene: Does this work for directories that contain special MSBuild characters, such as `;`?

wli3: no, same as `,` But so is `dotnet build -o `. I think to be on the same level of `dotnet build` should be enough for this 2.1.3xx fix

wli3: My conclusion is there is not standard way to do escape. Maybe escape every control char (> , ;) with url data encoding but keep non ascii should be the right way (but it is so odd). That's why i hesitate to put it in 2.1.3xx

peterhuene: At any rate, I think we should have double quotes around the property's value to ensure MSBuild interprets the contents as a literal, even if semicolon won't work in the `BaseIntermediateOutputPath` property.

peterhuene:
On a sidenote, are you able to run the generated shim when it contains the characters from the test?

I get a crash in `libcoreclr`:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000000000000
    frame #1: 0x000000010204b51e libcoreclr.dylib`ClrAllocInProcessHeapBootstrap(unsigned int, unsigned long) + 30
    frame #2: 0x000000010204b872 libcoreclr.dylib`operator new(unsigned long) + 18
    frame #3: 0x000000010201b144 libcoreclr.dylib`DecoderReplacementFallback::CreateFallbackBuffer() + 20
    frame #4: 0x000000010201b80e libcoreclr.dylib`UTF8Encoding::GetCharCount(unsigned char*, int) + 1102
    frame #5: 0x000000010201b2f3 libcoreclr.dylib`UTF8ToUnicode + 179
    frame #6: 0x0000000102018231 libcoreclr.dylib`Initialize(int, char const* const*, unsigned int) + 737
    frame #7: 0x0000000102018872 libcoreclr.dylib`PAL_InitializeCoreCLR + 34
    frame #8: 0x000000010204ad35 libcoreclr.dylib`coreclr_initialize + 37
    frame #9: 0x0000000100588cfc libhostpolicy.dylib`___lldb_unnamed_symbol946$$libhostpolicy.dylib + 172
    frame #10: 0x000000010057148c libhostpolicy.dylib`___lldb_unnamed_symbol750$$libhostpolicy.dylib + 32428
    frame #11: 0x0000000100578af9 libhostpolicy.dylib`___lldb_unnamed_symbol773$$libhostpolicy.dylib + 297
    frame #12: 0x000000010038b3a3 libhostfxr.dylib`___lldb_unnamed_symbol931$$libhostfxr.dylib + 451
    frame #13: 0x00000001003af16e libhostfxr.dylib`___lldb_unnamed_symbol1000$$libhostfxr.dylib + 18606
    frame #14: 0x00000001003b1d42 libhostfxr.dylib`___lldb_unnamed_symbol1009$$libhostfxr.dylib + 1602
    frame #15: 0x00000001003b01f8 libhostfxr.dylib`___lldb_unnamed_symbol1007$$libhostfxr.dylib + 1976
    frame #16: 0x000000010038c419 libhostfxr.dylib`___lldb_unnamed_symbol934$$libhostfxr.dylib + 297
    frame #17: 0x000000010000ddc4 fix`___lldb_unnamed_symbol60$$fix + 7892
    frame #18: 0x000000010000e395 fix`___lldb_unnamed_symbol61$$fix + 165
    frame #19: 0x00007fff75d4f015 libdyld.dylib`start + 1
```

src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
wli3: https://github.com/dotnet/cli/blob/cc80822c16b955a1f664f8a65ed2c76c3c52a824/src/Microsoft.DotNet.Cli.Utils/CommandResolution/GenericPlatformCommandSpecFactory.cs#L17

It is added by CommandSpecFactory layer

wli3: And -o is doing the same https://github.com/dotnet/cli/blob/6f6f28dac6936352049cad0a03a3a06b8f8ceb88/src/dotnet/commands/dotnet-build/BuildCommandParser.cs#L27

peterhuene: `ArgumentEscaper.EscapeAndConcatenateArgArrayForProcessStart` quotes the entire argument string as-needed, which is not the same thing as quoting the property value.  `ArgumentEscaper` will **escape** the double quotes so that MSBuild can parse them on its command line.  When doing so, it will interpret the value **literally** rather than seeing a comma or semicolon and treating it as a multi-property option (e.g. `/p:First=foo,Second=bar`). 

You can check for yourself the semantic difference if you pass `--tool-path foo,bar` to `dotnet tool install`.  Without the double quotes, it fails.  With the double quotes, it succeeds.

The `-o` option for `dotnet build` is a bug.  I shouldn't have to quote it on the command line like `-o '"foo,bar"'` (bash/zsh) or `-o \"foo,bar\"` (Windows) for it to respect the contents literally.

wli3: We already know dotnet build -o cannot handle ';'. There is a bug in how we handle msbuild params for sure. However my stand for 2.1.3xx is to keep the same as -o (won't be worse). Or we should all in to complete understand what is the right thing to do and do the change. Which could delay the fix for at least a week. At the same time, if you pass a quoted "c:\temp" in xml and pass it to BaseIntermediateOutputPath, it will try to interpret it as CURRENTDIRECTOY\"c:\temp" for some reason.

wli3:
image

I can get a shim with such name

src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
peterhuene: We shouldn't copy a bug just for parity.  Use double quotes.  Note how my examples above all used comma, which should work with both `-o` (it doesn't because it's a bug) and `--tool-path` (let's not copy that bug, ok?) without the user having to run through any hoops.

peterhuene:
Not get the shim, run the shim.

Doesn't appear to be a problem with the shim itself.  If I rename the folder to containing "foo", it runs without issue.

I think the CoreCLR is having issue starting when the exe's path contains UTF-8 characters.

wli3:
image

Look at the stack trace. It might be mac issue.

src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
wli3: As long as you are sure it will not introduce more bugs. After discover CURRENTDIRECTOY"c:\temp"  case at least I am not so sure

peterhuene: I'm not sure what you mean by "quoted in XML".  In the project file?  It should never be quoted there because they would be literal quotes.

Passed on the command line, the quotes will be respected and removed during MSBuild's command line parsing.

wli3: I talked to Peter offline. We agree on this version. Since there is no solution can solve all the cases.

And I talked to @livarcocc later. He think this solution is also good. It supports a upper set of characters. We should keep investigate a better solution for 2.1.4xx.

peterhuene:
Linking to the issue tracking the crash we found while discussing this PR: https://github.com/dotnet/coreclr/issues/18439

src/dotnet/commands/dotnet-tool/install/ProjectRestorer.cs
peterhuene: 👍

test/Microsoft.DotNet.ToolPackage.Tests/ToolPackageInstallerTests.cs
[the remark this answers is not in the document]
wli3: Thank you, I am adding surrogate characters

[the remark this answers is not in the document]
wli3: Added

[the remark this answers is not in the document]
wli3: Tests fail on Linux, I need to investigate more on it

wli3:
to support surrogate character which could show up in user name. I changed the approach to pass by temp project. The approach is different enough to have a separate PR. Continue here   https://github.com/dotnet/cli/pull/9509

PR #9473, opened 2018-06-12
Update VS insertion package name

dsplaisted:
This will allow 2.1.2xx and 2.1.4xx to be separate components in the VS installer.


dsplaisted:
@MattGertz FYI for infrastructure /packaging change to 2.1.2xx in order to support VS Setup having it as an optional component (and 2.1.4xx included by default)

PR #9491, opened 2018-06-14
Fix the 'Microsoft.AspNet.Core.Mvc' versioning

ghost:
The version of 'Microsoft.AspNetCore.Mvc' should be independent of 'Microsoft.AspNetCore.All'

### DIFF
--- a/build/DependencyVersions.props
+++ b/build/DependencyVersions.props
@@ -4,6 +4,7 @@
     <MicrosoftAspNetCoreAllPackageVersion>2.1.1-rtm-30818</MicrosoftAspNetCoreAllPackageVersion>
     <MicrosoftAspNetCoreAppPackageVersion>$(MicrosoftAspNetCoreAllPackageVersion)</MicrosoftAspNetCoreAppPackageVersion>
     <MicrosoftAspNetCoreDeveloperCertificatesXPlatPackageVersion>2.1.0</MicrosoftAspNetCoreDeveloperCertificatesXPlatPackageVersion>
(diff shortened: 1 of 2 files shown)


mmitche (approved):

build/DependencyVersions.props
```diff
@@ -74,6 +74,7 @@
     <BuildTasksFeedToolVersion>2.1.0-prerelease-02430-04</BuildTasksFeedToolVersion>
     <VersionToolsVersion>$(BuildTasksFeedToolVersion)</VersionToolsVersion>
     <DotnetDebToolVersion>2.0.0-preview2-25331-01</DotnetDebToolVersion>
+    <MicrosoftAspNetCoreMvcVersion>2.1.1</MicrosoftAspNetCoreMvcVersion>
```
natemcmaster: Rename this to `MicrosoftAspNetCoreMvcPackageVersion` so future patches will automatically get updated by ProdCon.

johnbeisner: @natemcmaster 
ProdCon does not have a "MicrosoftAspNetCoreMvcPackageVersion" per @mmitche

natemcmaster: It doesn't for the current patch release, but it will in future patches. ProdCon only generates this version variable if a package is produced as a part of aspnet outputs. When we produce new versions, you will want to ensure you align with the version from ProdCon.

johnbeisner: @peterhuene 
Can you weigh-in on this issue?

mmitche: @natemcmaster  is right for this.  You want it to be PackageVersion so that it can be overridden.

livarcocc:
Do we need to pick up this change on the internal branch?

PR #9530, opened 2018-06-21
Update coresetup to 2.2.0-preview1-26620-03

eerhardt:
Updating the runtime in master to the latest runtime version before the move to `netcoreapp3.0` in the runtime `master` branches.

#9453 is already past the switch to `netcoreapp3.0`, so that will require more work.


eerhardt:
test Ubuntu x64 Release Build

eerhardt:
@dotnet/cli-contrib - can someone give me a thumbs up here?

PR #9546, opened 2018-06-24
Use blob feed instead of myget

wli3:


wli3:

wli3:
cc @livarcocc this is a infrastructure change

PR #9566, opened 2018-06-26
Log a verbose message when DOTNET_CLI_HOME is being used.

### LINKED ISSUE #9510: dotnet should warn users if preferring DOTNET_CLI_HOME over HOME or USERPROFILE.
## Steps to reproduce

Using 2.1.4xx:

1. `DOTNET_CLI_HOME=/tmp/foo dotnet tool install -g dotnetsay`

## Expected  behavior

`dotnet` should print a message saying it is respecting `DOTNET_CLI_HOME` over `HOME` or `USERPROFILE`.

See https://github.com/NuGet/Home/issues/6989#issuecomment-393704026.

## Actual behavior

`dotnet` does not print any such message.

## Environment data

```
 Version:   2.1.400-preview-009009
 Commit:    d3787295e2

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.13-x64
 Base Path:   /Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/sdk/2.1.400-preview-009009/

Host (useful for support):
  Version: 2.1.1-servicing-26605-02
  Commit:  6985b9f684

.NET Core SDKs installed:
  2.1.400-preview-009009 [/Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.1-rtm-30818 [/Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/shared/Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.1-rtm-30818 [/Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.1-servicing-26605-02 [/Users/peterhuene/src/cli/bin/2/osx-x64/dotnet/shared/Microsoft.NETCore.App]

```

peterhuene:
This commit logs a diagnostic message when the `DOTNET_CLI_HOME` variable is
used.  This enables users to determine where first-run-experience and global
tool files are being written to.

Fixes #9510.

### DIFF
--- a/src/dotnet/Program.cs
+++ b/src/dotnet/Program.cs
@@ -149,6 +149,8 @@ namespace Microsoft.DotNet.Cli
                         bool skipFirstRunExperience =
                             environmentProvider.GetEnvironmentVariableAsBool("DOTNET_SKIP_FIRST_TIME_EXPERIENCE", false);
 
+                        ReportDotnetHomeUsage(environmentProvider);
+
                         topLevelCommandParserResult = new TopLevelCommandParserResult(command);
                         var hasSuperUserAccess = false;
                         if (IsDotnetBeingInvokedFromNativeInstaller(topLevelCommandParserResult))
@@ -227,6 +229,21 @@ namespace Microsoft.DotNet.Cli
                 exitCode = result.ExitCode;
             }
             return exitCode;
+        }
+
+        private static void ReportDotnetHomeUsage(IEnvironmentProvider provider)
+        {
+            var home = provider.GetEnvironmentVariable(CliFolderPathCalculator.DotnetHomeVariableName);
+            if (string.IsNullOrEmpty(home))
+            {
+                return;
+            }
+
+            Reporter.Verbose.WriteLine(
+                string.Format(
+                    LocalizableStrings.DotnetCliHomeUsed,
+                    home,
+                    CliFolderPathCalculator.DotnetHomeVariableName));
         }
 
         private static bool IsDotnetBeingInvokedFromNativeInstaller(TopLevelCommandParserResult parseResult)
--- a/src/Microsoft.DotNet.Cli.Utils/LocalizableStrings.resx
+++ b/src/Microsoft.DotNet.Cli.Utils/LocalizableStrings.resx
@@ -276,4 +276,7 @@
 
 dotnet tool install --global {1}</value>
   </data>
+  <data name="DotnetCliHomeUsed" xml:space="preserve">
+    <value>Using home directory '{0}' set by the '{1}' environment variable.</value>
+  </data>
 </root>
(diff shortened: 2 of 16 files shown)


peterhuene:
Ping @nkolev92.  I opted for a "verbose" message regarding the `DOTNET_CLI_HOME` variable being used because we have commands like `dotnet run` that should not display output by default.  With the `-d` option passed to `dotnet`, this will at least enable users to determine if their home directory is being overridden.  Let me know what you think since you are the one who requested this behavior.  Thanks!

nkolev92:
I think that's reasonable!

Thanks!

The biggest reason for this is the fact that the global packages folder and the plugins directory change.
If/when customers see some weird issues, we can ask them to do verbose logs and it'll indicate clearly what's going on.

wli3:
I assume you talked to Kathleen

src/dotnet/Program.cs
```diff
@@ -149,6 +149,8 @@ internal static int ProcessArgs(string[] args, ITelemetry telemetryClient = null
                         bool skipFirstRunExperience =
                             environmentProvider.GetEnvironmentVariableAsBool("DOTNET_SKIP_FIRST_TIME_EXPERIENCE", false);
 
+                        ReportDotnetHomeUsage(environmentProvider);
```
wli3: nit: why here? not somewhere close to Reporter.Output.WriteLine("Performance Summary:");? Or else?

It's better to avoid adding stuff to Progam.cs but I don't know a better place

peterhuene: I wanted the output to be before command execution, close to where we would report telemetry being used or not, simply because this variable can alter the behavior of the command that's about to run.  It also affects first-run experience, so I'd like it printed before that too.

peterhuene:
I actually haven't, so ping @KathleenDollard 😄 

In a previous change into 2.1.4xx, we fixed the crash / unexpected behavior of when `HOME` (macOS/Linux) or `USERPROFILE` (Windows) was not set for the dotnet process.  As part of that change, I added support for a new variable called `DOTNET_CLI_HOME` which dotnet would respect as those variables if set.  However, NuGet still required the home environment variables to be set to work at the time.  This has now been addressed such that NuGet will work with just `DOTNET_CLI_HOME` set too, enabling CI scenarios that want to instruct the .NET Core SDK (and, by extension, NuGet) to put its files for users without a home directory.

It was requested that the CLI log a message when this variable is being preferred so that users know where the files will be put.  This commit fulfills that request, although only at the "diagnostic" output level with `dotnet -d`.

peterhuene:
@Pilchie ping for approval.

Pilchie:
Approved for P5, but I wonder about the loc constraints here.  Will we be able to merge this in time to get a loc handback @nguerrera ?

nguerrera:
Hold merge until we decide we're open for P5.

> I wonder about the loc constraints here.

@Pilchie We need to have all string changes merged into release/2.1.4xx *before* July 9th.

If we're holding checkins for P5 until after that, then we have a problem. It is possible that we can negotiate something else (use a different branch, arrange a one off loc cycle, etc), but I'd really prefer if we can stick to the plan with everything else going on.

Can we set a time before July 9th after which we're considered done for P4 and then merge these?

peterhuene:
Marking WIP until final approval for P5.

PR #9653, opened 2018-07-12
LOC CHECKIN | dotnet/cli release/2.1.4xx | 20180712

v-zbsail:


tmeschter:
I'm working on fixing the build errors.

peterhuene:
Tom's handed fixing this off to me since it was my late-to-the-party change that caused the failure.  I'm working on addressing it now.

peterhuene:
Investigating test failure.

peterhuene:
So the test is failing because somehow the expected result contains smart quotes but the actual output doesn't?  This is perplexing.

nguerrera:
This is due to console "best fit mapping".

nguerrera:
Last time this happened, I changed the smart quotes to straight quotes because there were many straight quotes in the other translations anyway. And I notified @v-zbsail and @janbaltus to make the same change in "translation memory" so that it would not come back in a subsequent handback.

peterhuene:
Ping @Pilchie.  This is a LOC insertion.

Pilchie:
LGTM, but we'll need the VSO bug to submit to the QB.

Pilchie:
@tmeschter - can you get that ready?

tmeschter:
@peterhuene @Pilchie Done: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/647733

nguerrera:
@tmeschter I notice that we haven't gotten the dotnet/sdk handback to release/2.1.4xx yet. The listed ETA was 7/16 for both, so is it just not done yet? sdk inserts into cli, which inserts into VS. Should we wait until we have dotnet/sdk done and use a single VSO bug to track inserting both loc changes together?

Pilchie:
Great point @nguerrera - I think we should wait.

tmeschter:
@Pilchie @nguerrera It's fine to use the one VSO bug for both, but we need to get this change in ASAP. Until it is merged I won't be able to tell for sure if we will need another pass from the Loc team; if we do I need to schedule it today.

tmeschter:
The incoming branch may be fully localized, but that doesn't mean the release/2.1.4xx branch will be once it is merged in. Loc freeze was technically last Friday but that hasn't stopped people from making Loc changes.

nguerrera:
We can check the merge result without merging.

tmeschter:
Sure, but there's no reason to wait on this merge...

nguerrera:
The reason to wait would be if there's any chance this will be rejected by shiproom, then we'd have to back it out, which would just make a mes in the history. If we're sure this is just a formality, then I agree. @Pilchie please make the call here.

nguerrera:
Sorry, just spoke to @tmeschter in person. Will get the bug approved and merge this, then when sdk comes in we'll use the same bug still.

PR #9684, opened 2018-07-18
Pin stage0 used for update-dependencies to stable 2.1 build

nguerrera:
@dagood @eerhardt 

We're getting builds from master and then trying to run on 2.1 tooling on 3.0. The builds from master aren't in great shape right now and I don't like having 3.0 dependent projects in 2.1.x branches. It seems to me that we can just use a stable build and stable TFM for the update-dependencies, but maybe I've misunderstood something about how this works.

This should fix the issue identified here: https://github.com/dotnet/versions/pull/331#issuecomment-405972675


dagood:
`update-dependencies.sh` is for anyone who wants to update dependencies semi-automatically using this flow on linux, but I don't know if anyone does this. Seems worth making this same change there.

PR #9699, opened 2018-07-19
Fix archiving to .zip.

crummel:
From [here](https://github.com/dotnet/cli/blob/5632ca0714f18e188228104849e52eb011461ae0/build/FileExtensions.props#L3), we are now always using zip files instead of tarballs on Linux.  This change makes it so that we do not generate tarballs with a .zip extension, which confuses some downstream repos.


crummel:
Not seeing how my change could have affected these tests so I'm just going to kick them once.

PR #9700, opened 2018-07-19
Fix OrchestratedPackageVersionProps for source-build.

### LINKED ISSUE #9680: [source-build] DownloadPackageVersionsProps target is no longer run
I think this must have gotten lost in the refactoring - the `DownloadPackageVersionProps` target still [exists](https://github.com/dotnet/cli/blob/master/build/DownloadPackageVersions.targets#L2) but nothing like [InitRepo](https://github.com/dotnet/core-sdk/blob/master/build/InitRepo.targets) depends on it any more, so it is never run.  Additionally, the destination for the props, `OrchestratedPackageVersionsProps` is no longer defined and, in core-sdk, depends on `GeneratedPropsDir`, which also no longer exists.  I tried to prep a PR for this but ended up not know where you would want all of it - let me know if I can help at all.

cc @nguerrera @dsplaisted

crummel:
When InitRepo was removed, the call to DownloadPackageVersionProps did not get moved elsewhere.  This change restores it and the properties that the target depends on.

Fixes #9680.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -17,6 +17,9 @@ tools\TestAssetsDependencies\TestAssetsDependencies.csproj
 
   <PropertyGroup>
     <RepoRoot>$(MSBuildThisFileDirectory)</RepoRoot>
+    <RelativeGeneratedPropsDir>bin/obj</RelativeGeneratedPropsDir>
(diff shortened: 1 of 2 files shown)


crummel:
@dotnet-bot test OSX10.12 x64 Release Build please (java.nio.channels.ClosedChannelException)

Issue #9701, opened 2018-07-19
UseBundledNETCoreAppPackageVersionAsDefaultNetCorePatchVersion not being set as expected for 2.2.0-preview1-*


nguerrera:
We have this:
https://github.com/dotnet/cli/blob/74cdff42c735d8e5b89dedc8d0f0c5ebcb86a902/build/MSBuildExtensions.targets#L135-L143

But we are generating this:
```
    <BundledNETCoreAppPackageVersion>2.2.0-preview1-26609-02</BundledNETCoreAppPackageVersion>
    <UseBundledNETCoreAppPackageVersionAsDefaultNetCorePatchVersion>false</UseBundledNETCoreAppPackageVersionAsDefaultNetCorePatchVersion>
```

Even though patch is 0 and this is a preview build

cc @wli3 @dsplaisted

PR #9702, opened 2018-07-19
Passing the right BundledNETCoreAppPackageVersion

wli3:
…pPackageVersion version

this is a face palm. $(BundledNETCoreAppPackageVersion) is stage0's BundledNETCoreAppPackageVersion. We want to use stage2's. Which is $(_NETCoreAppPackageVersion) since 

`<BundledNETCoreAppPackageVersion>$(_NETCoreAppPackageVersion)</BundledNETCoreAppPackageVersion>`

fixes https://github.com/dotnet/cli/issues/9701

no test caught it since we don't have a good way to assert preview version(it keeps changing). But we would caught it if it is non preview


wli3:

PR #9710, opened 2018-07-20
update F# to 10.2.2-rtm-180720-0

brettfo:


nguerrera:
Ubuntu x64 is known issue. Investigating separately.

PR #9755, opened 2018-07-27
Create binlogs by default.

ghost:
Create binlogs by default.

Addresses: https://github.com/dotnet/cli/issues/9671

Starting with release/2.1.3xx the change will flow into release/2.1.4xx; release/2.1.401; et al

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,6 +127,9 @@ bin/
 *.svclog
 *.scc
 
+# BinLog artifacts
+msbuild.*.ProjectImports.zip
+
 # Chutzpah Test files
 _Chutzpah*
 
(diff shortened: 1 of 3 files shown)


dagood (approved):
Sweet, thanks.

.gitignore
```diff
@@ -127,6 +127,9 @@ bin/
 *.svclog
 *.scc
 
+# BinLog artifacts
+msbuild.*.ProjectImports.zip
```
nguerrera: Just curious, is an external project imports zip now the default?

nguerrera (approved):

.gitignore
dsplaisted: I think this is the file that is created while the build is being run, but doesn't persist afterwards.  But if you look at the unstaged changes while a binlogged build is running you would see this file.

PR #9798, opened 2018-08-05
Fix build warnings and treat msbuild warnings as errors

nguerrera:
We were getting the warning that we set BaseIntermediateOutputPath to late because in this special case, we did it after SDK props and not via Directory.Build.props.

Also, preventing more warnings like this from creeping in by treating MSBuild warnings as errors, same as compiler warnings.


nguerrera:
Crossgen just trolled me into fixing another annoyance

nguerrera:
Ping @livarcocc Clear to merge?

PR #9800, opened 2018-08-06
Retrive the SDK version from the '.version' file

ghost:
**src/Microsoft.DotNet.Cli.Utils/SdkProduct.cs** 
retrieve the SDK version from the '.version' file

**test/dotnet.Tests/dotnet.Tests.csproj** 
test the version from the LKG SDK not the version from the current CLI - [see: test/dotnet.Tests/VersionTest.cs]

**build/test/OverlaySdkOnLKG.targets 
build/Test.targets**
use the version from the LKG SDK not the version from the current CLI for the directory-name of the SDK

**build/DependencyVersions.props** 
update the LKG CLI to: 3.0.100-alpha1-009408 - this is the first published SDK version with the corrected '.version' and '.toolsetversion' files

Reference: https://github.com/dotnet/core-sdk/issues/41

Follow-up: https://github.com/dotnet/cli/issues/9832


livarcocc:
No tests?

johnbeisner:
@livarcocc 
The tests are already present and they are correctly failing as they have uncovered a missing piece: https://github.com/dotnet/core-sdk/pull/45

johnbeisner:
@dotnet help

src/dotnet/Program.cs
```diff
@@ -24,7 +24,7 @@ namespace Microsoft.DotNet.Cli
 {
     public class Program
     {
-        private static readonly string ToolPathSentinelFileName = $"{Product.Version}.toolpath.sentinel";
+        private static readonly string ToolPathSentinelFileName = $"{SdkProduct.Version}.toolpath.sentinel";
```
nguerrera: I don't get why we have both Product and SdkProduct, if we're using SdkProduct exclusively at this stage. Can't we just change the implementation of Product.Version to use the .version file. 

If/when we surface the CLI version, the names 'Product' and 'SdkProduct' aren't very suggestive of the actual difference.  I'd wait until we actually do that to have both and then we can pick appropriate names.

johnbeisner:
@nguerrera 
@livarcocc 
@wli3 
@dotnet/dotnet-cli 
After much discussion and experimentation, this PR is in it's final form - please review and comment.

I will ensure that the PR gets squashed before merging...

src/Microsoft.DotNet.Cli.Utils/Product.cs
```diff
@@ -1,22 +1,17 @@
...
 
         private static string GetProductVersion()
         {
-            var attr = typeof(Product)
-                .GetTypeInfo()
-                .Assembly
-                .GetCustomAttribute<AssemblyInformationalVersionAttribute>();
-            return attr?.InformationalVersion;
+            DotnetVersionFile versionFile = DotnetFiles.VersionFileObject;
+            return versionFile.BuildNumber ?? string.Empty;
```
wli3: Just confirm that we decide to simply return empty in this case

wli3:
@dotnet/cli-contrib  Some thought for separating logic in general. Resolving version number in CLI repo is a circular dependency problem. In most of the case, core-sdk depends on cli and combine it with tool set. But cli also has some logic depends on the version number that only core-sdk know. Luckily, we found almost all cases that need a version number are not a scenario for cli (as in the verbs) so return empty is fine. But I don't think problem is a simply which version to use.

Also, since the repos are separated, we need to enforce boundary of the systems. If CLI, as in a portable app with verb and targets, doesn't really need logic from a specific toolset version, we should try to avoid the dependency. And also consider move logic that only deal with versions to core-sdk like creating sentinel file. Or we could have more features that requires 2 checkin

src/Microsoft.DotNet.Cli.Utils/Product.cs
```diff
@@ -1,22 +1,17 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Reflection;
-
 namespace Microsoft.DotNet.Cli.Utils
 {
     public class Product
     {
-        public static string LongName => LocalizableStrings.DotNetCommandLineTools;
+        public static string LongName => LocalizableStrings.DotNetSdkInfo;
```
nguerrera: Why are we changing this? Was that decided? It seems unrelated to me.

src/Microsoft.DotNet.Cli.Utils/Product.cs
nguerrera: When John and I talked last, I thought we were going to make it a hard failure with the reasoning that the version file can be considered a required part of the installation.

test/dotnet.Tests/dotnet.Tests.csproj
```diff
@@ -30,7 +30,7 @@
     <PropertyGroup>
       <ExpectedVersionFileInIntermediateFolder>$(IntermediateOutputPath)ExpectedSdkVersion.txt</ExpectedVersionFileInIntermediateFolder>
       
-      <VersionFileContent>$(SdkVersion)</VersionFileContent>
+      <VersionFileContent>$(DotNetCoreSdkLKGVersion)</VersionFileContent>
```
nguerrera: I'm having a really hard time understanding what $(SdkVersion) is supposed to represent and where it should be used. It should probably be renamed because it's hard to see that a line like this would be wrong as it was before.

build/test/OverlaySdkOnLKG.targets
```diff
@@ -2,22 +2,13 @@
   <Target Name="OverlaySdkOnLKG">
     <ItemGroup>
       <OverlaySDK Include="$(DotNetCoreSdkLKGPublishDirectory)/**/*" />
-      <OverlayCLI Include="$(OutputDirectory)/**/*" />
+      <OverlayCLI Include="$(CliOutputDirectory)/**/*" />
```
nguerrera: I'm having trouble understanding why the whole overlay process is changed by this. Can you describe what was happening before vs. what happens after this change?

johnbeisner: @nguerrera 
The overlay process was using the CLI version instead of the SDK-LKG version - the SDK versioned folder had a CLI version.

test/dotnet.Tests/dotnet.Tests.csproj
johnbeisner: @nguerrera 
In the CLI builds $(SdkVersion) is used for the CLI version.

src/Microsoft.DotNet.Cli.Utils/Product.cs
nguerrera: I think the change from ".NET Command Line Tools to .NET Core SDK" is a good one. But it's a separate issue. When we do this change we should also delete DotNetCommandLineTools. I'd prefer a separate PR, but if you want to keep this, please also delete the old string so that it's easier to see what is actually changing in the text.

src/Microsoft.DotNet.Cli.Utils/Product.cs
nguerrera: I'm OK with empty and Will makes a good point about not requiring the core-sdk bundling to execute dotnet.dll. We have a lot of related problems though, so at this point in time, I could go either way.

src/Microsoft.DotNet.Cli.Utils/Product.cs
johnbeisner: @nguerrera 
Old string is now deleted.

johnbeisner:
@nguerrera 
This PR should be ready to merge

nguerrera (approved):

PR #9819, opened 2018-08-10
Add restore passthrough 

wli3:
Allow disable-parallel, no-cache, ignore-failed-sources to pass through dotnet tool install

And refactor –-configfile to use the new way handling pass through.
Only few restore option make sense to dotnet tool install. So we need to handle it case by case. And it requires much more code than just pass the rest as string. And option like -–source is not directly pass through, since tool install will make it “additional” and also pass it via xml. Only disable-parallel, no-cache, ignore-failed-sources, configfile have no difference between restore and tool install.
I test it by make sure PackageLocation can get correct option from parse result and when feed argument generator’s result back to restore parser. Verify it can be understand by restore parser. We could further abstract the idea of pass through (less code). But so far only these 4 options are qualified to this concept.

### DIFF
--- a/src/Microsoft.DotNet.InternalAbstractions/FilePath.cs
+++ b/src/Microsoft.DotNet.InternalAbstractions/FilePath.cs
@@ -23,6 +23,16 @@ namespace Microsoft.Extensions.EnvironmentAbstractions
             Value = value;
         }
 
+        public static FilePath? CreateOrReturnNullWhenValueIsNull(string value)
+        {
+            if (value != null)
+            {
+                return new FilePath(value);
+            }
+
+            return null;
+        }
+
         public string ToQuotedString()
         {
             return $"\"{Value}\"";
--- a/src/dotnet/ToolPackage/PackageLocation.cs
+++ b/src/dotnet/ToolPackage/PackageLocation.cs
@@ -1,5 +1,7 @@
 ﻿
 using System;
+using Microsoft.DotNet.Cli;
+using Microsoft.DotNet.Cli.CommandLine;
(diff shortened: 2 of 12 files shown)


wli3:

src/Microsoft.DotNet.InternalAbstractions/FilePath.cs
```diff
@@ -23,6 +23,16 @@ public FilePath(string value)
             Value = value;
         }
 
+        public static FilePath? CreateOrReturnNullWhenValueIsNull(string value)
```
peterhuene: This method's name is confusing to me.  Perhaps `ToNullableFilePath`?

src/dotnet/ToolPackage/PackageLocation.cs
```diff
@@ -8,16 +10,33 @@ internal class PackageLocation
...
+            bool ignoreFailedSources = false)
         {
             NugetConfig = nugetConfig;
             RootConfigDirectory = rootConfigDirectory;
             AdditionalFeeds = additionalFeeds ?? Array.Empty<string>();
+            DisableParallel = disableParallel;
+            IgnoreFailedSources = IgnoreFailedSources;
+        }
+
+        public PackageLocation(AppliedOption appliedOption)
```
peterhuene: I'm not sure I like the CLI options "leaking" into the tool package types.  This is now another place we need to keep in sync with the command line parser for the commands.

wli3: actually that's should the right thing. Just put it is in CreateToolPackageStoreAndInstaller Factory

PR #9831, opened 2018-08-14
Follow XDG spec

Rosalie241:
This change allows the dotnet cli to follow the XDG spec on linux
and changes the OSX path to `$HOME/Library/dotnet`

sources: 
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
https://stackoverflow.com/questions/3373948/equivalents-of-xdg-config-home-and-xdg-data-home-on-mac-os-x

### DIFF
--- a/src/dotnet/MulticoreJitProfilePathCalculator.cs
+++ b/src/dotnet/MulticoreJitProfilePathCalculator.cs
@@ -51,7 +51,12 @@ namespace Microsoft.DotNet.Cli
 
         private static string GetNonWindowsRuntimeDataRoot()
         {
-            return $"{(Environment.GetEnvironmentVariable("HOME"))}/.dotnet/";
+            var XDG_DATA_HOME = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
+
+            if(string.IsNullOrEmpty(XDG_DATA_HOME))
+                XDG_DATA_HOME = Path.Combine(Environment.GetEnvironmentVariable("HOME"), ".local/share");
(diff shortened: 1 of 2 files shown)


tim241:
> 1: a discussion of whether we need to fully adapt XDG spec.
I feel like following the XDG standard would be really nice, especially
since a lot of applications already make another directory in $HOME so
adding to that won't make $HOME any less cluttered

> 2: the existing change is not enough. Only MulticoreJitProfilePath location is changed. There are at least dotnet cli cache, nuget fall back folder, global tools default location need to change. That the same time, we need to separate them according to the catalog of XDG spec
Alright.

> 3: XDG Base Directory Specification require different flags for different files. The existing files are all under .dotnet folder. And it is likely components are assuming locations. If these folders can be independently set-able, we also need test coverage for that.
May I know what kinda flags you're talking about? I don't really
understand this.

> 4: migration story of existing folders on users disk
Alright, but first the discussion about implementing this(or not..)

> 5: ensure sdk and runtime adapt the same convention.
> 6: ensure all the existing env that is changing the directories can work with these new XDGs .
Alright

> 7: maybe we need refactor to an abstraction to deal with folder location concern, and put it in one place. It also need to have a very good perf.
An abstraction and putting it in 1 place would make it a lot easier.

marcpopMSFT:
CLI repo has been archived and merged with https://github.com/dotnet/sdk/tree/master/src/Cli.  Closing all old PRs in the cli repo to master.

PR #9846, opened 2018-08-15
Fix warnings

wli3:
Using prebuild docker image eliminate warning during apt-get install build packages like python. There are warnings due to redundant remove dotnet-sdk when there is no dotnet-sdk before the test. And add delete fallback folder to prevent error on remove dotnet-sdk package from apt after the test


wli3:
I tested locally, since jenkins will not run all linux task

PR #9884, opened 2018-08-22
Refactor uninstall to a class

wli3:
Instead of a method on PackageInstance. Immediate goal is to reuse PackageInstance
on local tools, however local tools should not have Uninstall method.

At the same time, PackageStore is too powerful. It can create PackageInstance,
and PackageInstance can do side effect. It doesn’t make sense when PackageInstaller
has access of creating PackageInstance and delete other packages. I created IToolPackageStoreQuery
that can only return file locations to avoid this coupling.

### DIFF
--- a/src/dotnet/ToolPackage/IToolPackageStore.cs
+++ b/src/dotnet/ToolPackage/IToolPackageStore.cs
@@ -8,12 +8,16 @@ using NuGet.Versioning;
 
 namespace Microsoft.DotNet.ToolPackage
 {
-    internal interface IToolPackageStore: IToolPackageStoreQuery
+    internal interface IToolPackageStore
     {
-        IEnumerable<IToolPackage> EnumeratePackages();
+        DirectoryPath Root { get; }
 
-        IEnumerable<IToolPackage> EnumeratePackageVersions(PackageId packageId);
+        DirectoryPath GetRandomStagingDirectory();
 
-        IToolPackage GetPackage(PackageId packageId, NuGetVersion version);
+        NuGetVersion GetStagedPackageVersion(DirectoryPath stagingDirectory, PackageId packageId);
+
+        DirectoryPath GetRootPackageDirectory(PackageId packageId);
+
+        DirectoryPath GetPackageDirectory(PackageId packageId, NuGetVersion version);
     }
 }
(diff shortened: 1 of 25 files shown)


src/dotnet/ToolPackage/ToolPackageFactory.cs
```diff
@@ -21,13 +21,40 @@ internal static class ToolPackageFactory
...
+            var toolPackageUninstaller = new ToolPackageUninstaller(
+                toolPackageStore);
+
+            return (toolPackageStore, toolPackageUninstaller);
+        }
+
+        public static (IToolPackageStore,
+            IToolPackageInstaller,
+            IToolPackageUninstaller)
+            CreateToolPackageStoreInstallerUninstaller(
```
peterhuene: Not a huge fan of this method name, but I can't suggest a better one.  Perhaps add an `And` for consistency with the one above?

src/dotnet/ToolPackage/IToolPackageStore.cs
```diff
@@ -8,18 +8,8 @@
 
 namespace Microsoft.DotNet.ToolPackage
 {
-    internal interface IToolPackageStore
+    internal interface IToolPackageStore: IToolPackageStoreQuery
```
peterhuene: I don't agree with this refactoring.  These are all methods relating to the management and inspection of the package store.  If anything, the users of `IToolPackageStore` after this refactoring are only interested in what should be the "query" part of the interface (i.e. they only care about what's in the store, but not the management methods used by the installer and uninstaller).  Thus this seems backwards to me because the `IToolPackageStoreQuery` interface is more about management of the store and the `IToolPackageStore` interface has both query and management.

Stepping back, is there really a need for this?  All of these methods now relate to the store, albeit some are used to query the store and some are used by the install/uninstaller for managing the store.  This refactoring seems only intended to further abstract what should be possible to do from the installer and uninstaller and I don't feel that's necessary.

wli3: summery, make them separate interface and swap the 2 names

wli3:
@peterhuene I split the interface renamed factor method in d19d375

I renamed the concrete class to `ToolPackageStoreAndQuery : IToolPackageStoreQuery, IToolPackageStore` does it look good?

